PR #11, opened 2023-07-13
Test of Contributing to GEECS-Plugins

cedoss-lbl:
Currently making some Python analysis of HiResMagSpec images.  Testing if this push works or if I need to get more permissions.


cedoss-lbl:
This is a month old now, just going to merge later

PR #13, opened 2023-08-22
Integrate live postprocessing into geecs api chris edits

cedoss-lbl:
Includes a unit test for the analyze_image function, fixed the output of the LabView wrapper function to return dictionary values in a set order, and removed personal scripts from the repository.


TAU-Reinier (approved):
@cedoss-lbl Looks great! I added one commit* to separate the value assertions into self.assertAlmostEqual clauses. 

* Note that you can commit on a branch that's in the middle of a pull request, and the pull request will be updated with the new commit.

PR #18, opened 2023-08-29
Recreated Sam's "Optimization Factor" metric as a scalar representing…

cedoss-lbl:
… the quality of the shot.  HiResMagSpec now returns another scalar that is the sum of the charge density array multiplied against a Gaussian weight function.  By default, the Gaussian weight function is centered on 100 MeV with a sigma of 2 MeV


TAU-Reinier (approved):
Looks good! Please do add documentation for all of the parameters in U_HiResMagSpecImageAnalyzer.__init__(). In particular, for optimization_central_energy and optimization_bandwidth_energy, specify the units! 
(Though in a future version, you can use Quantities, which are unit-aware floats or np.arrays)

PR #24, opened 2023-09-28
Online analysis refactor chris

cedoss-lbl:
Made a lot of edits to how my analysis scripts are organized and how they interface with outside calls to them.  Despite this, the LabView interface should be unchanged.  In addition to reorganizing the ImageAnalysis/image_analysis/analyzers/ folder to be better (I think?) I've made a lot of improvements to the image analyzers themselves:

MagCams AND UndulatorExitCam:
- The returns of the analyzer's analysis is now a dictionary that contains the processed image, dictionary of return scalars, dictionary of input parameters, and a 2D array of vstacked lineouts.  (The LabView wrappers still return a 2D array, list of floats, and another 2D array).
- Both the input and return dictionary keys are renamed to be more descriptive.
- The LabView wrapper functions in labview_adapters.py were adjusted to accommodate the new return structures, but the overall flow should be the same as before.

Only MagCams (ACaveMagCam3 and HiResMagCam):
- Combined into a single class (UC_GenericMagCam)
- Class is initialized either the old way using keyword args in __init__, OR by using functions that generate default analyzer settings using config files.  There is a config file for both acave3 and hires that the camera's respective LabView function call uses.  One could also provide a custom config file if it is written with the same structure.  The goal is that if needed one could edit the "default" config file to adjust the analyzer settings without needing to reinstall the entire package on that camera's computer.
- The analysis procedure was moved from mag_spec_analysis.py into the class itself.  Much of the code was cleaned up to use the class variables rather than the input parameter dictionary.  mag_spec_analysis.py still exists and has much of the basic analysis functions.

Open Questions:
- Is the organization of UC_GenericMagSpecCam acceptable?  Can this be a good template for further camera analysis scripts?
- Should I go ahead and make UC_UndulatorExitCam and other cameras into this format?
- In UC_GenericMagSpecCam, I also made the debugging timing print functionality a class function.  But, since I do the same thing in UC_UndulatorExitCam's analysis, should I consider putting this "print_time" function in the super class?
- In UC_GenericMagSpecCam, I purposefully did not use the utils.ROI class that was used by Reinier's sample analysis class.  This is because I didn't like the idea of having to import this class every time I wanted to make an analyzer with a specific roi in the __init__ call.  Also, on appserver I can't import the module imageio.v3 and appserver doesn't let me use pip. (The camera computers don't have this issue at least, but it does make it annoying for me to develope on appserver).  I just use a List[int] to do the same thing for now.  Is this fine?
- When run through LabView, will python be ok at using configparser?


TAU-Reinier:
Good stuff @cedoss-lbl ! Thank you very much for the detailed explanation.

Ideally the UC_GenericMagSpecCam.py stays "Generic", i.e. does not contain any device-selection functions and methods, and the client (labview_adapters in this case) does the selection. 

I see that there are essentially two device-specific parameters: the default_config filename; and the energy axis function. I can think of a few ways to implement this in labview_adapters.py: 
* Load the config file in hi_res_mag_spec_labview() or acave_cam3_mag_spec_labview(), and instantiate the UC_GenericMagSpecCam with parsed config values. In addition, derive the energy axis here and pass it as a new `__init__` parameter, instead of mag_spec_name
* Add a load_config(config_filename) method to UC_GenericMagSpecCam, then call that from labview_adapters. Derive the energy axis in labview_adapters and add that as a new `__init__` parameter instead of mag_spec_name (as above)
* Add a class variable default_config_filename, a method load_config(), _and a method derive_energy_axis()_ to UC_GenericMagSpecCam, _then subclass it_ with UC_HiResMagSpecCam and UC_ACaveMagCam3 that implement only those three items, either in labview_adapters or in a new UC_MagSpecCams.py file. In the UC_GenericMagSpecCam class, they can either return defaults, or raise NotImplementedError

TAU-Reinier:
your open Questions:

> Is the organization of UC_GenericMagSpecCam acceptable? Can this be a good template for further camera analysis scripts?

See [my previous comment](https://github.com/GEECS-BELLA/GEECS-Plugins/pull/24#issuecomment-1739823132)

> Should I go ahead and make UC_UndulatorExitCam and other cameras into this format?

Not sure how this Cam relates - let's talk. 

> In UC_GenericMagSpecCam, I also made the debugging timing print functionality a class function. But, since I do the same thing in UC_UndulatorExitCam's analysis, should I consider putting this "print_time" function in the super class?

I'm not sure the ImageAnalyzer base class is the right place for this; I think a Mixin class is the way to go here. 

> In UC_GenericMagSpecCam, I purposefully did not use the utils.ROI class that was used by Reinier's sample analysis class. This is because I didn't like the idea of having to import this class every time I wanted to make an analyzer with a specific roi in the init call. Also, on appserver I can't import the module imageio.v3 and appserver doesn't let me use pip. (The camera computers don't have this issue at least, but it does make it annoying for me to develope on appserver). I just use a List[int] to do the same thing for now. Is this fine?

I don't quite understand the objection - are you creating analyzers outside of ImageAnalysis modules? In any case, a List[int] is fine, as long as the expected order and location of origin is clear everywhere. 

> When run through LabView, will python be ok at using configparser?

Yes, configparser is a built-in library

TAU-Reinier:
Note that because the .ini files are inside the ImageAnalysis package, the package does need to be reinstalled if they are updated, at least if the package is installed as a copy rather than --editable.

cedoss-lbl:
Thanks for your comments @TAU-Reinier!  Just to follow up on your suggestions:

Device-selection functions in UC_GenericMagSpecCam:
- The original idea is that for post-analysis I do make analyzer instances, and it is nice to have a way to load a default config file.  I just thought it would be a little cleaner to have this functionality within UC_GenericMagSpecCam.py rather than in labview_adapters.py because I don't like the idea of adding too much stuff to labview_adapters.py.  But, perhaps a better way is to just have a module that lives in the same folder as all of the image analyzers whose job it is to organize creating classes from default config files.  That way, both labview_adapters.py and my post analysis scripts can call this module.
- I'm still not sold on making the energy axis an __init__ parameter.  It requires having the dimensions of the image before the image is given to the analyzer, and this just seems like it can be a bit roundabout to me.  The only positive I can think of is if we want the possibility to initialize an analyzer with an arbitrary energy axis.  

UC_UndulatorExitCam:
- I wasn't going to make this into the same form as GenericMagSpecCam, just was going to move the analyze procedure into the class and add a default config file.  Just haven't done this yet before we decide if GenericMagSpecCam is super good enough.

ROI:
- Two objections: first is just me being lazy and not wanting to use import statements.  Second is a bit more crucial, I can't pip install imageio.v3 on my appserver.  This isn't the end of the world, but makes my workflow a bit more annoying.

Config file locations:
- Should we instead have the config files located outside of ImageAnalysis?  Or elsewhere on the Z drive?  But this is a good point, if we want to be able to edit these files to begin with, should they even be on this repo?

Conflicts with master:
- There is one "conflict," but it looks like Guillaume just moved a "->" after a function definition to a new line.  So this isn't really a conflict, but I'm not quite sure how to resolve these merge conflicts.

TAU-Reinier:
Ok great. You're right - let's not put too much in labview_adapters; it really should only be calling the right image analyzer and converting between inputs and returns. 

With that, plus the fact that the energy axis depends on the image size, it seems like subclassing UC_GenericMagSpecCam into a UC_MagSpecCams.py file, where the subclasses implement default_config_filepath and calculate_energy_axis(pixel_axis) is the cleanest. But your current approach is good enough. 

Ideally yes, the config files live separately from this repository. Is there a location on a shared drive for device configurations? Or should we put these in the device database? 

For conflicts with master: the best way to handle this is to first merge master into your own branch, resolve any conflicts, then create a pull request for your branch back into master. (I see that's what GitHub's suggestion is as well). You can do that while this PR is open; it should update with the new merge commit, and you should be able to then merge the PR.

TAU-Reinier (approved):

cedoss-lbl:
Looks like I was able to follow along the command line instructions and delete that file Guillaume edited without destroying everything else.  I'll go ahead and merge here in a bit.

PR #25, opened 2023-10-09
Analyzer cleanup chris

cedoss-lbl:
- Refactored UndulatorExitCam's analyzer to be organized the same was as the MagSpecCam's analyzer.
- Moved the default config files' location to be above ImagAnalysis so that it is external to the package and still within the repo.
- Wrote additional unit tests for ACaveMagCam3 and UndulatorExitCam
- Cleaned up the rest of my scripts a bit.

I'll be done messing with this repo for a bit now, but there are still a few todo's I can try and get to at some point:
- Docstrings on everything
- Make sure everything uses relative imports (and figure out why python sometimes crashes when I try to use relative imports)
- Maybe update UC_UndulatorRad2.py's analyzer to the current standard?


TAU-Reinier (commented):
This is good! There are a few things I'd recommend:
*  In [UC_GenericMagSpecCam:202](https://github.com/GEECS-BELLA/GEECS-Plugins/pull/25/files#diff-70e6f1819697a88d458329b03729be70f5a525a12fe687b3b0129d8a291649f8L202), the original is better because it's more explicit
* default_analyzer_generators:
  * in Python generators means something different (iterable functions) so better to rename
  * there's a lot of repetition here, so it's likely that this can be written more concisely. 
* the linear fit in [single_scan_charge_calibration:148](https://github.com/GEECS-BELLA/GEECS-Plugins/pull/25/files#diff-7fd31f02782ba00fea1845614fdd9300d077f3d5a7a69224a49350c51c56c9aaR148) doesn't seem right. The best slope with intercept 0 (i.e. y=b*x fit) isn't the same as a linear curve fit and setting the intercept to 0. And why do both polyfit and curve_fit? 
* [shot_charge_reader:32](https://github.com/GEECS-BELLA/GEECS-Plugins/pull/25/files#diff-3abef849ca73fbd91815ac26f1dad69c66db6e6a5798fbdf239c26be7e789df7R32): the original is better - make explicit that this is picoCoulomb, otherwise lowercase "pc" is unclear.

I think the cleanest option - [as recommended before](https://github.com/GEECS-BELLA/GEECS-Plugins/pull/24#issuecomment-1739963750) - is to subclass UC_GenericMagSpecCam and put things like `default_input_config_filename` there.

cedoss-lbl:
I have made the edits you've suggested.  Just two comments:
- Still just have the default_analyzer python file, might change it over to be subclasses at some point.
- The code in single_scan_charge_calibration was just really messy.  It wasn't actually using the polyfit, but I just had it in there when testing different fits and never took it out.  Removed those extra calculations.

PR #33, opened 2023-11-22
Two quick bug fixes

cedoss-lbl:
Stumbled upon these while running my scripts on control room PC:
-In pngTools.py, the method I was previously using for loading an image (used in post-processing, not live analysis) was deprecated.  Moved the newer version of im.imread into a try block. 
-In UC_GenericMagSpecCam.py, found a very strange bug where I was analyzing an image that just barely had enough counts on the screen to register that the sum of the image was > 0.  But when I took the average of several vertical slices this residual camera count was averaged away and so numpy functions with weights crashed.  This should be a fringe case we don't care about, so now I explicitly look for if this and set subsequent calculations to default to zero if this is detected.


TAU-Reinier (approved):
Ok to merge, but a few suggestions:
* Best practice is to keep your commits atomic, so two unrelated bug fixes should have two commits. (too late for this PR but for future reference)
* When using a try-except block to switch flow, it's best to be explicit about which exception, so that errors not related to the one you're trying to catch still halt the program. My guess is that it's a TypeError, so:  `try: ...    except TypeError:  ... `
* For png's created by LabView you should be using [my ](https://github.com/GEECS-BELLA/GEECS-Plugins/blob/c7653db750254846408ce3fd31a4709f41fd6c4a/ImageAnalysis/image_analysis/utils.py#L11) or [Guillaume's](https://github.com/GEECS-BELLA/GEECS-Plugins/blob/c7653db750254846408ce3fd31a4709f41fd6c4a/GEECS-PythonAPI/geecs_python_api/tools/images/ni_vision.py#L8) png reader because of its unusual format

PR #41, opened 2024-03-12
Kyle Jensen - beam spot analyzer

kjensen11:
Generic beam spot analyzer and specific UC_Amp2Input analyzer.


ImageAnalysis/image_analysis/analyzers/default_analyzer_initialization.py
```diff
@@ -3,6 +3,7 @@
 
 from . import UC_GenericMagSpecCam as MagSpec
 from . import UC_UndulatorExitCam as PhotonSpec
+import UC_Amp2_IR_input as Amp2Input
```
TAU-Reinier, line 6: Please make this a relative import as well.

TAU-Reinier (commented):
Looks good, though there is a lot of duplicate code. 
The AnalyzerAmp2Input is almost identical to the BeamSpotAnalyzer. As far as I can tell, the only difference is that the centroid is split into centroidx and centroidy. 
Is this an important difference, i.e. will some analysis output require centroid as a 2-tuple and some as split scalars? If not, consider consolidating the two. 
If yes, consider calling `super().analyze_image(image)` from  AnalyzerAmp2Input.analyze_image() and altering the result.

TAU-Reinier (commented):
It looks like AnalyzerAmp2Input derives from BeamSpotAnalyzer but doesn't modify it (not even the centroid splitting from the [previous version](https://github.com/GEECS-BELLA/GEECS-Plugins/pull/41/files/8731a7a4a8950273bd387c1719732822fe42c565))

Can return_analyzer_from_config_file() just return a BeamSpotAnalyzer, removing AnalyzerAmp2Input altogether?

kjensen11:
> It looks like AnalyzerAmp2Input derives from BeamSpotAnalyzer but doesn't modify it (not even the centroid splitting from the [previous version](https://github.com/GEECS-BELLA/GEECS-Plugins/pull/41/files/8731a7a4a8950273bd387c1719732822fe42c565))
> 
> Can return_analyzer_from_config_file() just return a BeamSpotAnalyzer, removing AnalyzerAmp2Input altogether?

Thanks for the quick review.

This was done by design in that I imagined the BeamSpotAnalyzer could be inherited by future laser/electron cameras which observe a Gaussian-like beam spot. Then dedicated child class analyzers (i.e. AnalyzerAmp2Input) could then simply build off this analysis. Of course, for this specific case there is not yet any further analysis and it is effectively extra code.

I will remove AnalyzerAmp2Input and directly call BeamSpotAnalyzer, only creating dedicated child analyzers if they truly build off the parent class in the future.

TAU-Reinier (approved):

PR #46, opened 2024-03-27
Reworked how image analyzers talk with Labview, and added UC_ALineEBeam3 camera analyzer

cedoss-lbl:
Fairly big change to how everything is organized, but my goal is to remove all of the intermediate functions that needed to be created when making a new analyzer.  Now, using the config files, initializing the analyzer in a general way is done "out of sight" in labview_analyzers.py and base.py's LabviewImageAnalyzer class.

**Before**:
`labview_adapters.py`
 - dictionary item
 - calling function
 - function to parse the scalar returns

`default_analyzer_initialization.py`
 - function to point to the default .ini config

`Analyzer Class`
 - function to read from the config and parse into the __init__

**Now**:
`labview_adapters.py`
 - Only need a dictionary item; each element maps the device name to the analyzer class, the .ini config file, and the .json key list name

`labview_adapters_config.json`
 - Contains a list of the keys for the scalar returns.  Only need one list for each analyzer return.

`Analyzer Class`
 - The config reading is done by a intermediate parent class between the Analyzer and ImageAnalyzer.  So now the __init__ takes either a .ini config file or a list of keyword arguments and figures out how it is initializing itself, while the "configure" function assigns given keyword arguments to their respective class variables.  However, now all elements in the 'settings' of the .ini config file need to match up to the expected keyword arguments of configure()

**Notes**:
1. There might be a cleaner way to implement the __init__ of each child class.  One could even just not have an __init__ function and just let the parent class __init__ take care of it, but then my PyCharm complains that I shouldn't initialize class variables outside of __init__.  But as it stands now, __init__ handles class variable initialization and config() handles keyword argument assignment.
2. I hope the "base.py" was a good place to include the "LabviewImageAnalyzer" intermediate class.  I left the old "ImageAnalyzer" parent class the same so anything explicitly using that wont' be affected.
3. labview_adapters.py is the most cleaned up, and the three elements in the dictionary sort of force people to remember to implement the three things necessary now:  the .json key list name, the .ini settings config, and the analyzer itself.
4. Added the ALine3 analyzer.  Very basic analyzer for finding the centroid and fwhm using an roi of the camera and a circular crop of the screen.
5. Moved some repeated code between the analyzers to a new module called "image_processing_funcs"
6. Removed my old "math_tools" module that was outdated and barely used
7. Updated my unit test functions to use the new organization
8. Looks like I'll need to merge master into this branch to fix some merge issues.


kjensen11 (commented):
I think this this modified workflow generally works for me. Two notes/questions though:
1. I don't think your branch reflects the additions I recently made regarding the Amp 2 Input camera analysis and generic beam spot analyzer. I imagine this might be partially responsible for some of the expected merge conflicts. How should this be handled? Should I just reincorporate this after this pull request is accepted and branch merged? Or should I implement this now on the branch before acceptance and merge?
2. Roughly speaking, I think your UC_ALineEBeam3 analysis serves as a specific use case of the generic beam spot analyzer that I made, albeit with some new relevant functionality. I think it would be good if you add this functionality to the generic beam spot analyzer once things are properly merged and then you could just inherit the generic analyzer to reduce code redundancy.

TAU-Reinier:
Generally the source branch (ALineEBeam3-Analyzer-Dev in this case) should merge master into itself before doing a pull request, so that any merge conflicts are handled in the source branch. @cedoss-lbl , make sure your local master branch is up to date, then merge it into ALineEBeam3-Analyzer-Dev. The merge commit will be automatically included in this PR, and there will be no more merge conflicts for the PR.

cedoss-lbl:
> I think this this modified workflow generally works for me. Two notes/questions though:
> 
> 1. I don't think your branch reflects the additions I recently made regarding the Amp 2 Input camera analysis and generic beam spot analyzer. I imagine this might be partially responsible for some of the expected merge conflicts. How should this be handled? Should I just reincorporate this after this pull request is accepted and branch merged? Or should I implement this now on the branch before acceptance and merge?

As Reinier suggested, I can do the initial merge of master into this branch to fix the conflicts there.  The conflicts are generally to do with labview_adapters getting reorganized and the default_analyzer_initialization getting removed.  Next, I'll see if I can make some basic changes to your Analyzer to make it compatible with the new workflow (changing existing __init__ to configure(), checking the .ini file, adding the dictionary to .json).  After that, I can let you know and you can pull this branch to do your own testing.

> 2. Roughly speaking, I think your UC_ALineEBeam3 analysis serves as a specific use case of the generic beam spot analyzer that I made, albeit with some new relevant functionality. I think it would be good if you add this functionality to the generic beam spot analyzer once things are properly merged and then you could just inherit the generic analyzer to reduce code redundancy.

This a good idea!  We can keep them separate to complete the branch merging for now, but this will be a good change going forward.

kjensen11:
Regarding merge conflicts, those are the files I was suspecting. Otherwise it all sounds good to me. I am more or less ready to approve once you do the above and are ready to re-request.

cedoss-lbl:
Alright, merged master into here and made some quick changes.  I would think that you can now run the analyzer much the same way as before, but may need to initialize it differently in your tests.  Otherwise, `analyze_image()` is still the exact same.  @kjensen11 if you could run some tests to check that the functionality is the same as before that would be great!

Other notes on your anlayzer:
 - Noticed you were using the ROI class from the utils module.  I also tried using this in the past and I ran into some issue when running it on the old python version that Labview requires.  Something in utils just did not play nice with Python 3.9, so I ended up just making making "roi" a list of 4 ints.
 - The `self.background` class variable is handled by the LabviewImageAnalyzer now, so removed it from your __init__ function

TAU-Reinier (commented):
The changes you made to device configuration in labview_adapters.py make things clearer - good stuff. 

Adding common methods of image analyzers through LabviewImageAnalyzer, that you can use for several analyzers, is a good idea. 

I would prefer that you don't rewrite how `__init__` works for LabviewImageAnalyzer. 

The reason that the signature of `__init__` contains all of the configurable parameters, with type annotations and defaults, it that it's used by the LiveProcessing program. 

While the LabviewImageAnalyzers are meant to be invoked by Labview, not LiveProcessing, it shouldn't be an issue. But for the sake of consistency, please keep the form. 

The configure() method could be written to parse the settings and overwrite an ImageAnalyzer attribute.

ImageAnalysis/image_analysis/base.py
```diff
@@ -81,3 +83,122 @@ def analyze_image(self,
...
+            the roi bounds given by [top, bottom, left, right] pixel
+
+        """
+        roi_top = int(parser.get('roi', 'top')),
+        roi_bottom = int(parser.get('roi', 'bottom')),
+        roi_left = int(parser.get('roi', 'left')),
+        roi_right = int(parser.get('roi', 'right')),
+        return np.array([roi_top, roi_bottom, roi_left, roi_right]).reshape(-1)
+
+    def roi_image(self, image):
```
TAU-Reinier, line 157: Note that if self.roi is of [ROI](https://github.com/GEECS-BELLA/GEECS-Plugins/blob/8ca9adaa8dace69f78d0735386a17b93bdfb9d28/ImageAnalysis/image_analysis/utils.py#L48 )type, it has a .crop method.

ImageAnalysis/image_analysis/analyzers/online_analysis_modules/image_processing_funcs.py
```diff
@@ -0,0 +1,35 @@
...
+    return len(np.where(image > saturation_value)[0])
+
+
+def calculate_fwhm(projection_array, threshold=None):
+    maxloc = np.argmax(projection_array)
+    halfmax = projection_array[maxloc]/2
+    if threshold is not None:
+        halfmax = halfmax + 0.5*threshold
+
+    peak_region = np.where(projection_array >= halfmax)[0]
```
TAU-Reinier, line 34: Assumes that there are no peaks elsewhere in the array.

kjensen11:
> Alright, merged master into here and made some quick changes. I would think that you can now run the analyzer much the same way as before, but may need to initialize it differently in your tests. Otherwise, `analyze_image()` is still the exact same. @kjensen11 if you could run some tests to check that the functionality is the same as before that would be great!

Noted.

> Other notes on your anlayzer:
> * Noticed you were using the ROI class from the utils module.  I also tried using this in the past and I ran into some issue when running it on the old python version that Labview requires.  Something in utils just did not play nice with Python 3.9, so I ended up just making making "roi" a list of 4 ints.

I don't have a problem with this. We can either move away from the ROI class within the 'generic beam spot analysis' or modify the ROI class to play nicely.

ImageAnalysis/image_analysis/labview_adapters.py
```diff
@@ -1,82 +1,19 @@
...
+
+"""Dictionary to map "GEECs Device Name" --> [child of LabviewImageAnalyzer class,
+                                              'name of keys dictionary in .json file',
+                                              '.ini config file in "image_analysis_configs/"']"""
 DEVICE_FUNCTIONS = {
-    "UC_HiResMagCam": hi_res_mag_spec_labview,
-    "UC_ACaveMagCam3": acave_cam3_mag_spec_labview,
-    "UC_UndulatorExitCam": undulator_exit_cam_labview,
-    "UC_UndulatorRad2": undulator_rad2_labview,
+    "UC_HiResMagCam":       [MagSpec,   'MagSpecCam',   'default_hiresmagcam_settings.ini'],
```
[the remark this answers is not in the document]
cedoss-lbl: Ok, yeah I can make this change!  I was looking for a way to type-hint that the classes had to be derived from `LabviewImageAnalyzer` and this seems like a good solution.  It makes the `DEVICE_FUNCTIONS` dictionary a little more verbose, but probably still manageable to someone not familiar with the code.

cedoss-lbl:
> I would prefer that you don't rewrite how `__init__` works for LabviewImageAnalyzer.
> 
> The reason that the signature of `__init__` contains all of the configurable parameters, with type annotations and defaults, it that it's used by the LiveProcessing program.
> 
> While the LabviewImageAnalyzers are meant to be invoked by Labview, not LiveProcessing, it shouldn't be an issue. But for the sake of consistency, please keep the form.
> 
> The configure() method could be written to parse the settings and overwrite an ImageAnalyzer attribute.

Ah yeah I hear you.  This shouldn't be too difficult of a change to make.  What I'll do then is just have __init__ only create an analyzer using keyword arguments.  Would it be possible to also include a "configuration file" as a Path or str into __init__, and at the end of __init__ run configure() if it is given?  Or would you prefer that configure is not called from within __init__ at all?

ImageAnalysis/image_analysis/labview_adapters.py
```diff
@@ -121,10 +42,117 @@ def analyze_labview_image(device_type, image, background):
...
+    config_filename : str
+        String of the name of the config file
+
+    Returns:
+    --------
+    config filename: Path
+        String that points to the config file after determining its location
+
+    """
+    active_build_directory = r'Z:\software\control-all-loasis\HTU\Active Version\GEECS-Plugins\image_analysis_configs\\'
```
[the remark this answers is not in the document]
cedoss-lbl: Gotcha, noted

ImageAnalysis/image_analysis/base.py
cedoss-lbl: Right, we start using that if ROI works on the Labview python version.  I'll try to test and debug that in the near future.

ImageAnalysis/image_analysis/analyzers/online_analysis_modules/image_processing_funcs.py
cedoss-lbl: True.  I could fix this to account for multi-peaks, although this would be a kind of fringe case.  If our beam did have multiple spiky peaks, I would be hesitant to report that the FWHM was low for only a single peak.  Maybe there is a better term to use than FWHM, because at the end of the day if we are looking to minimize the FWHM then it is assumed we are also trying to minimize it in a single peak.

ImageAnalysis/image_analysis/labview_adapters.py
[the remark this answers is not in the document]
TAU-Reinier: Great! A more legible option could be
```python
"UC_HiResMagCam": DeviceConfig(
    labview_analyzer_class=MagSpec, 
    key_list_name='MagSpecCam', 
    default_settings_filename='default_hiresmagcam_settings.ini'
), 
"UC_xxx": DeviceConfig(
    labview_analyzer_class=MagSpec, 
etc.
```

TAU-Reinier:
> > I would prefer that you don't rewrite how `__init__` works for LabviewImageAnalyzer.
> > The reason that the signature of `__init__` contains all of the configurable parameters, with type annotations and defaults, it that it's used by the LiveProcessing program.
> > While the LabviewImageAnalyzers are meant to be invoked by Labview, not LiveProcessing, it shouldn't be an issue. But for the sake of consistency, please keep the form.
> > The configure() method could be written to parse the settings and overwrite an ImageAnalyzer attribute.
> 
> Ah yeah I hear you. This shouldn't be too difficult of a change to make. What I'll do then is just have **init** only create an analyzer using keyword arguments. Would it be possible to also include a "configuration file" as a Path or str into **init**, and at the end of **init** run configure() if it is given? Or would you prefer that configure is not called from within **init** at all?

I would recommend initializing the ImageAnalyzer and configure()'ing it separately, if only to keep a config_path keyword away from analyzer parameter keywords. 

That said, you could make a case for adding a configure() method to all `ImageAnalyzer`s that contain the annotated analyzer parameters instead of having `__init__()` hold them. If you want to go that way let me know - it'll require some changes to LivePostProcessing.

cedoss-lbl:
- Made some good progress I think!  __init__ is back to the way it used to be, where all class variables are defined and typed explicitly.  
 - Now the `configure()` function is fully defined in the super class to handle any list of keyword arguments, and so subclasses don't have to implement their own version of configure().  
 - Setting a config file to overwrite the default analyzer initialization is also quite easy, LabviewImageAnalyzer has an `apply_config()` function to do just that.
 - Made a one line change to `labview_adapters.py`, and all of my unit tests work without making any changes to them.

ImageAnalysis/image_analysis/base.py
```diff
@@ -191,14 +180,20 @@ def apply_background(self, background):
         self.background = background
 
     def configure(self, **kwargs):
-        """ Configures necessary parameters for the analyzer used by analyze_image
+        """ Given a dictionary of keyword arguments, updates environment variables for the anlayzer
 
-        To be implemented by children of this class.  This is called by the initialization function, where the keyword
-        arguments are given by either the "settings" block of the .ini config file or explicitly listed.
+        This function also requires that the class variables were previously initialized to their proper type.
+        Furthermore, passing in a None for a given keyword argument will skip over resetting the variable.
```
TAU-Reinier, line 186: Another way, which doesn't require that variables have the correct class*, is to get the expected types from the `__init__` method's annotations, using `typing.get_type_hints(image_analyzer_instance_or_class.__init__)`

\* This isn't guaranteed even with coercion in `__init__`, because users can still manually change the attributes from outside (`image_analyzer_instance.some_parameter = 1.23`).

ImageAnalysis/image_analysis/base.py
```diff
@@ -170,7 +159,7 @@ def roi_image(self, image):
             either the input image if there is no roi on this analyzer, or the cropped image if roi is defined
 
         """
-        if self.roi is None:
+        if (self.roi is None) | (isinstance(self.roi, list) and any(elem is None for elem in self.roi)):
```
TAU-Reinier, line 162: use `or` ; the pipe `|` is a bitwise OR.

cedoss-lbl:
Current status is that everything I wrote works with LabView.  I wrote a function in LabviewImageAnalyzer to build a return dictionary with the correct keys and typing, but Kyle's analyzer doesn't work still.  I think it has to do with the formatting of the return scalars.

I'll try to go through with the merge sometime next week unless there are any more outstanding issues.  We can try to merge Kyle's analyzer with the ALine analyzer then.

kjensen11 (approved):
Good on my side. Can adapt generic beam analyzer for compatibility post merge.

ImageAnalysis/image_analysis/base.py
```diff
@@ -122,6 +122,76 @@ def apply_config(self, config_file):
...
+                    for lineout in return_lineouts:
+                        shape = np.shape(lineout)
+                        if len(shape) != 1 or shape[0] < 2:
+                            print("return_lineouts must be passed as a list of 1d arrays!")
+                            return_lineouts = None
+                            break
+            if return_lineouts is None:
+                return_lineouts = np.zeros((1, 1), dtype=np.float64)
+            else:
+                max_length = max(len(lineout) for lineout in return_lineouts)
```
TAU-Reinier, line 180: Nice use of generator instead of list! If you want to get extra-fancy, you can write: `max(map(len, return_lineouts))`

ImageAnalysis/image_analysis/base.py
```diff
@@ -199,3 +269,17 @@ def configure(self, **kwargs):
...
+
+        Can be overwritten by implementing classes if you prefer more control over the return dictionary.
+        For example, adding units into the key names.
+
+        Returns
+        -------
+        dict
+            A compiled dictionary containing all class variables
+        """
+        self.__dict__.update(vars(self))
```
TAU-Reinier, line 284: What is this line supposed to do? I think `vars(self)` just returns `self.__dict__` so it just updates itself with its own values.

ImageAnalysis/image_analysis/base.py
cedoss-lbl: Ooh nice, I like that!  There's always a clever way to remove `for` usages it seems

PR #49, opened 2024-04-18
Fix geecs api scan info updating

skbarber:
the scan info .ini file created by MC when operating in 'remote scan' doesn't capture all of the correct metadata. We use the python api to edit it and add things like a description of the scan, scan parameter etc.

The previous method to edit this file did not use configparser, which is a cleaner way to do it. Additional string entries we're prepended to the file, but this breaks things like GEECSPlotter and the Automated scan logger. 

Note, the 'scan parameter' entry is hardcoded to be "Shotnumber". This is the way the original version worked. Seems like this would need to be changed if other kinds of scans are used via "remote scan".

### DIFF
--- a/GEECS-PythonAPI/geecs_python_api/controls/devices/geecs_device.py
+++ b/GEECS-PythonAPI/geecs_python_api/controls/devices/geecs_device.py
@@ -383,55 +383,6 @@ class GeecsDevice:
         if not comment:
             comment = f'{var_alias} scan'
         return GeecsDevice.file_scan(self, comment, timeout)
-        
-    def update_ini_file(ini_file_path, comment):
-        backup_file_path = str(ini_file_path) + '~'
-
-        # Extract number from the filename
-        match = re.search(r'ScanInfoScan(\d{3})\.ini$', str(ini_file_path))
-        if match:
-            scan_number = int(match.group(1))
-        else:
-            print("Filename does not match expected format. Aborting update.")
-            return
-        try:
-            # Backup the original INI file
-            shutil.copy2(ini_file_path, backup_file_path)
-
-            # Create a ConfigParser object and read the backup file
-            config = configparser.ConfigParser()
-            config.optionxform = lambda option: option  # Preserve case
-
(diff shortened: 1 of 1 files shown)


GEECS-PythonAPI/geecs_python_api/controls/devices/geecs_device.py
```diff
@@ -503,7 +459,56 @@ def _process_scan(cmd: str, comment: str = 'no scan', monitoring_device: Optiona
                     pass
 
         return next_folder, next_scan, accepted, timed_out
+        
+    def update_ini_file(ini_file_path, comment):
```
TAU-Reinier, line 463: Recommend marking this a `@staticmethod`. Good for clarity, but also that way geecs_device_instance.update_ini_file(path, comment) will work correctly.

TAU-Reinier (approved):
Good stuff. configparser is the way to go.

GEECS-PythonAPI/geecs_python_api/controls/devices/geecs_device.py
skbarber: Will do. Admittedly, I’m not too knowledgeable about static vs non static

guillaumeplateau: Nice improvement. Recommend adding a check that ini_file_path is indeed a text file and not something dramatic like "C:/Program Files". I believe the "else" case would erase it all.

skbarber: I think it's safe, this part should catch any issues i think:
        # Extract number from the filename
        match = re.search(r'ScanInfoScan(\d{3})\.ini$', str(ini_file_path))
        if match:
            scan_number = int(match.group(1))
        else:
            print("Filename does not match expected format. Aborting update.")
            return

also, os.remove wouldn't delete directories either. 

I'll go ahead and merge with master and delete the branch

guillaumeplateau (approved):
Nice improvement

PR #51, opened 2024-05-14
Incorporate spectral correction for camera based spectrometers

kjensen11:
Extend functionality of UC_GenericLightSpecCam to correct for spectral efficiencies (quantum efficiency, transmission efficiency, etc.).

UC_GenericLightSpecCam analysis method modified to pass an additional boolean (to toggle functionality on and off) and a list of strings (file names for optic efficiencies to incorporate into correction). These arguments are defined in the respective config file.

UC_GenericLightSpecCam now includes two additional methods: perform_spectral_correction() and construct_spectral_correction(). The latter function constructs a complete detection response curve by iterating through the list of optic efficiencies file names, importing the data (wavelength and efficiency), interpolating the efficiency over the wavelength array extracted from the camera spectrometer calibration, and multiplying to the total efficiency. Note that this function does not extrapolate efficiency outside of the data range (setting outside efficiencies to 1 or not corrected). The perform_spectral_correction function constructs the efficiency curve, determines if any efficiencies are below a default defined threshold (currently 2.5%, setting efficiency to 1 or not corrected if below), and imposes the correction on the lineout spectrum and camera image.

Data files are referenced via hard coded relative filepath to the 'data' directory in ./ImageAnalysis/image_analysis/


ImageAnalysis/image_analysis/analyzers/UC_GenericLightSpecCam.py
```diff
@@ -130,6 +141,54 @@ def analyze_image(self, input_image: Array2D, auxiliary_data: Optional[dict] = N
                                                          input_parameters=self.build_input_parameter_dictionary())
         return return_dictionary
 
+    def perform_spectral_correction(self, wavelength, spectrum, image,
+                                    eff_low_bound=0.025):
+
+        # load detection efficiency
+        spectral_correction = self.construct_spectral_correction(wavelength)
```
TAU-Reinier, line 148: Consider calling construct_spectral_correction() from __init__ instead of from perform_spectral_correction(). That way, if the same analyzer is used on many (hundreds) of images, the spectral correction isn't constructed every time.

ImageAnalysis/image_analysis/analyzers/UC_GenericLightSpecCam.py
```diff
@@ -30,7 +30,9 @@ def __init__(self,
                  calibration_0th_order_pixel: float = 1186.24,
                  minimum_wavelength_analysis: float = 150.0,
                  optimization_central_wavelength: float = 420.0,
-                 optimization_bandwidth_wavelength: float = 10.0
+                 optimization_bandwidth_wavelength: float = 10.0,
+                 spec_correction_bool: bool = True,
+                 spec_correction_files: List[str] = []
```
TAU-Reinier, line 35: Don't forget to add documentation in the Parameters section!

ImageAnalysis/image_analysis/analyzers/UC_GenericLightSpecCam.py
```diff
@@ -130,6 +141,54 @@ def analyze_image(self, input_image: Array2D, auxiliary_data: Optional[dict] = N
...
+
+        # correct image
+        cor_image = image / spectral_correction['efficiency']
+
+        return cor_spectrum, cor_image
+
+    def construct_spectral_correction(self, input_wavelength):
+
+        # define relative path
+        rel_path = os.path.join('..', 'data')
```
TAU-Reinier, line 164: I'm not sure this works correctly - the path will likely be relative to the _current working directory_ rather than this python file. 
To make it explicit, use the special variable `__file__`: 
`rel_path = os.path.join(os.path.dirname(__file__), '..', 'data')`

or even better:
`spectral_correction_files_folder: Path = Path(__file__).parents[1] / 'data'`
using `from pathlib import Path`.

ImageAnalysis/image_analysis/analyzers/UC_GenericLightSpecCam.py
TAU-Reinier, line 35: Giving a mutable (such as a list or a dict) as a default for a keyword argument [can give unexpected results](https://docs.python-guide.org/writing/gotchas/#mutable-default-arguments). 
It's better to set None as default, followed by `if spec_correction_files is None: spec_correction_files = []` in the body.

TAU-Reinier, line 35: Consider writing out the variable names: spectral_correction_files, corrected_image, efficiency_lower_bound, etc.  Descriptive variable names effectively self-document code at negligible cost.

TAU-Reinier (commented):
I had a few minor comments, of which the rel_path one is the only one that might prevent this code from working correctly.

ImageAnalysis/image_analysis/analyzers/UC_GenericLightSpecCam.py
cedoss-lbl: Agreed.  While in the live python analysis generates a new analyzer for every image because the python session only persists for that single image from Labview, being able to reuse the same analyzer in post-analysis scripts has been quite convenient from my experience.

cedoss-lbl:
Looks good to me as well.  In addition to implementing Reinier's comments, I'll provide an example of the framework I typically use to test these analyzers so you can check functionality and benchmark the execution time.

cedoss-lbl (approved):

ImageAnalysis/image_analysis/analyzers/UC_GenericLightSpecCam.py
```diff
@@ -76,19 +82,30 @@ def __init__(self,
         self.optimization_central_wavelength = float(optimization_central_wavelength)
         self.optimization_bandwidth_wavelength = float(optimization_bandwidth_wavelength)
 
+        self.pointing_correction_bool = bool(pointing_correction_bool)
```
TAU-Reinier, line 85: Be careful with this, because it can give unexpected results. For example: `bool("False")` gives True (as does any string that's not empty). 
In addition, in if-statements where you use this variable, Python implicitly evaluates bool() anyway. 
If you want to enforce the type of this variable, as far as I know it'll take something like: 
```
if not isinstance(pointing_correction_bool, bool):
    raise TypeError("pointing_correction_bool should be of bool type")
```

TAU-Reinier (approved):

ImageAnalysis/image_analysis/analyzers/UC_GenericLightSpecCam.py
cedoss-lbl: Ah good tip!  I ran into this issue when debugging my "configure" function for the Labview image analyzer class, but I think all the child analyzer classes I wrote previously don't explicitly check for the boolean type in their `__init__` functions.  Might be a good time to update those as well.

ImageAnalysis/image_analysis/analyzers/UC_GenericLightSpecCam.py
```diff
@@ -67,10 +82,30 @@ def __init__(self,
...
+
+        # if spectral correction information available, do now to avoid redundant calculations
+        # if not available, retrieval and calculation performed in analyze_image
+        self.spectral_correction_bool = bool(spectral_correction_bool)
+        if self.spectral_correction_bool and spectral_correction_files is not None:
+            self.spectral_correction_files = list(spectral_correction_files)
+            self.spectral_correction_data = self.get_spectral_correction_data()
+        else:
+            self.spectral_correction_files = list()
+            self.spectral_correction_data = None
```
cedoss-lbl, line 95: I think like we talked about, if `self.spectral_correction_data` is only calculated in the `__init__` function, then this will not get updated when the config files are loaded using the `configure()` method.

I would suggest moving this logic somehow to the `analyze_image()` function.

ImageAnalysis/image_analysis/analyzers/UC_GenericLightSpecCam.py
```diff
@@ -67,10 +82,30 @@ def __init__(self,
...
+        #     self.spectral_correction_files = []
+        # else:
+        #     self.spectral_correction_files = list(spectral_correction_files)
+
+        # # calculate spectral correction from data files
+        # self.spectral_correction_data = self.get_spectral_correction_data()
+
         # Set do_print to True for debugging information
-        self.do_print = False
+        self.do_print = True
```
cedoss-lbl, line 105: Don't forget to turn this off before final merge

ImageAnalysis/image_analysis/analyzers/UC_GenericLightSpecCam.py
```diff
@@ -15,10 +15,16 @@
...
     from ..types import Array2D
 
+from pathlib import Path
+from scipy.interpolate import interp1d
+
 from ..base import LabviewImageAnalyzer
 from .online_analysis_modules import image_processing_funcs as process
 from .online_analysis_modules import photon_spectrometer_analyzer as analyze
 
+from ..utils import read_imaq_image, ROI
```
cedoss-lbl, line 25: This reminds me that we should test again if this old Python version on all the camera servers still complains about utils.

image_analysis_configs/undulatorrad2cam_may2_settings.ini
```diff
@@ -7,6 +7,8 @@ calibration_0th_order_pixel = 289.4770
 minimum_wavelength_analysis = 150.0
 optimization_central_wavelength = 420.0
 optimization_bandwidth_wavelength = 10.0
+spec_correction_bool = False
+spec_correction_files = []
```
cedoss-lbl, line 11: While we aren't using the python analysis for Rad2 at the moment, it will complain if the .ini file variable names don't match what is in `__init__`

ImageAnalysis/image_analysis/analyzers/UC_GenericLightSpecCam.py
TAU-Reinier: In that case, consider specifying a configure() method in this class that runs the parent class's configure() plus get_spectral_correction_data() : 
```python
def configure(**kwargs):
    super().configure(**kwargs)
    self.spectral_correction_data = self.get_spectral_correction_data()
```

Again, we want to avoid doing any "one-time" configuration in analyze_image(), since the same instance of this class will be used to call analyze_image() many times.

cedoss-lbl: I like this solution!

PR #52, opened 2024-05-16
05152024 mag spec charge fix chris

cedoss-lbl:
Had a wrong calculation for charge density.  Now I am appropriately using the energy axis.  Should work for both HiResMagCam and ACaveMagCam3.


cedoss-lbl:
Going ahead and merging this.

PR #56, opened 2024-06-14
Add context manager for HtuExp

TAU-Reinier:
Makes it possible to use HtuExp() in a with clause, which closes the instance at the end.


TAU-Reinier:
One question remains: Should the singleton instance also be deleted at the end of the with block?

cedoss-lbl:
Closing because outdated

PR #57, opened 2024-06-21
QuadAnalysis.data_dict: only keep setpoints and indexes for which there's analysis

TAU-Reinier:
In QuadAnalysis.data_dict, reduces list of 'setpoints' and 'indexes' to those setpoints and indexes that correspond to an analysis in the 'analyses' list. Thus it can be shorter than the setpoints used during the scan.

@guillaumeplateau I'm checking with you to make sure the 'setpoints' and 'indexes' aren't used anywhere else where you might need the full list.


cedoss-lbl:
Closing because outdated

PR #58, opened 2024-10-07
Geecs data acquision module addition

skbarber:
Adding a standalone data acquisition module using geecs_python_api. Allows for lighter weight and more configurable data acquisition. Also, broadens the scope of potential developers for data acquisition (doesn't require labview knowledge, just python)

Made a couple changes to existing geecs_python_api module. First, made an edit to tcp_subscriber to be able to register event based actions. e.g., an existing subscriber can now be used to add entries to a log when the status is updated. Also, noticed some unexpected behavior from the dequeue method for geecs_device. Seemed to only dequeue 1 command at a time, and only after a synchronous UDP call. The queue would build up and "randomly" dequeue commands after a synchronous UDP call. Now, it will dequeue all commands, though, it does seem some of the dequeued commands can get rejected if a "wait" isn't included.

### DIFF
--- a/GEECS-PythonAPI/geecs_python_api/controls/data_acquisition/__init__.py
+++ b/GEECS-PythonAPI/geecs_python_api/controls/data_acquisition/__init__.py
@@ -1,2 +1,2 @@
 from .data_acquisition import DeviceManager, DataLogger, DataInterface, ActionManager
-from .data_acquisition import setup_logging
+from .data_acquisition import setup_console_logging
--- a/GEECS-PythonAPI/geecs_python_api/controls/data_acquisition/__init__.py
+++ b/GEECS-PythonAPI/geecs_python_api/controls/data_acquisition/__init__.py
@@ -1,2 +1,2 @@
-from .data_acquisition import DeviceManager, DataLogger, DataInterface
+from .data_acquisition import DeviceManager, DataLogger, DataInterface, ActionManager
 from .data_acquisition import setup_logging
--- a/GEECS-PythonAPI/geecs_python_api/controls/data_acquisition/__init__.py
+++ b/GEECS-PythonAPI/geecs_python_api/controls/data_acquisition/__init__.py
@@ -1,2 +1,2 @@
-from .data_acquisition import DeviceManager, DataLogger, DataInterface, ActionManager
+from .data_acquisition import DeviceManager, DataLogger, DataInterface, ActionManager, ConfigManager
 from .data_acquisition import setup_console_logging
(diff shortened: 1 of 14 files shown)


cedoss-lbl:
I ended up merging this branch into my "geecsscanner-dev-10042024" branch to work on the gui, and there were a couple of things I needed to fix to get this working.  (As is, the example in the README does not work)

1. When you initialize the DataLogger, the `__init__()` function creates a new, empty DataInterface and DeviceManager.  This should instead be passed as an argument to the initialization, which is what I believe you originally had because that is what the README example would suggest.  Without this, no devices are ever set to save.

2. When reading the config file, if there are no actions then the default value returned by the `config.get()` function should be `None` instead of `{}`.  Otherwise it tries to do an empty command and crashes.

3. Minor edit, but for my purposes I wanted to use a dictionary to set the configuration rather than reading a yaml.  To do this, I just broke apart the function that reads from the config file into two functions:  A function to get a dictionary from a file and then a function to use the dictionary to set the configuration.

In addition, I have a few questions:
a) Are noscans possible with this framework?
b) When I was testing I was not running a scanlog generator.  But does this generate an entry in the daily scanlog?
c) In your `config.yaml` example, what is the purpose of the `scan_parameters` portion of the yaml?

skbarber:
1. yeah, at some point I put a jupyter notebook in the repo that I was
using for testing things. It's here:
GEECS-Plugins/GEECS-PythonAPI/htu_scripts
/test_scripts/. I think that should work. I realized the datalogger should
just initialize the other classes rather than pass them as arguments. Or,
at least, it's more efficient that way.

2. makes sense, I didn't get to testing the "no actions" case.

3. I thought a yaml was more or less a dictionary? Or at least functioned
in the same way.

a) well, they were in an early iteration. In fact was how I started with
this. But, I haven't tested for some time. In principle it is supported,
but probably there some minor debugging things that need to be done

b) There's no reason it shouldn't generate an entry in the daily scanlog.
That code is just looking at the data directory and reading files from
there. So, as long as I got all the formatting correct to match the MC
formatting, ti should be fine.

c) probably nothing at this point. My original plan was to have the scan
parameters as part of the config, but then I kind of realized the configs
are really more about diagnostic setting rather than choice of scan
parameter. I didn't delete it yet cause it wasn't a problem yet and I
wasn't sure if there might be some use for it down the line. probably not
at this point.

TAU-Reinier:
I would only suggest to call DataLogger something different, such as DataCollector or DataAcquirer, or something like ScanAcquirer. Partly this is to make clear which log threads, variables, and functions belong to console logging, and which to measurements.

skbarber:
Thanks. My plan was to refactor things a bit and have a "scanmanager" that
handles all of the setting of device variables and things like that. And
then the datalogger (to be renamed) would only deal with logging the tcp
events and asynchronous polling.

PR #59, opened 2024-10-16
Data acquisition refactor v2

skbarber:
some refactoring and debugging.

Migrating the console logging and some other things into a utils.py module. The configmanager class was overkill and was replaced with a single function in the utils. 

trigger is turned back on after waiting for ~2 seconds after sending the save>>off commands to devices, whcih prevents extraneous files getting created at the end of the scan.

In principle, no functionality should have changed.

### DIFF
--- a/GEECS-PythonAPI/geecs_python_api/controls/data_acquisition/__init__.py
+++ b/GEECS-PythonAPI/geecs_python_api/controls/data_acquisition/__init__.py
@@ -1,4 +1,3 @@
-from .data_acquisition import DeviceManager, DataLogger, DataInterface, ActionManager, ConfigManager
+from .data_acquisition import DeviceManager, DataLogger, DataInterface, ActionManager
 from .scan_manager import ScanManager
-
-# from .data_acquisition import setup_console_logging
+from .utils import ConsoleLogger, get_full_config_path, visa_config_generator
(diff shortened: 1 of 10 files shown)


cedoss-lbl:
These changes look good, and yeah they shouldn't effect the GUI.  I only really use scan_manager's `__init__`, `reinitialize`, `is_scanning_active`, and `stop_scanning_thread` right now.  Refactoring these large files with multiple classes into several files organized by class and/or utility helps a lot of readability, too.

Support for relative variables is also something that can be added to the GUI with relative ease, just need a checkbox to set a flag to pass in the dictionary.

I am hoping that once I get the ability to (1) specify "options" to easily change between VISA position and Velmex mode (2) add support for actions (which should just be understanding/including them in my yaml files) and (3) adding composite variable support; then my GUI should be ready for regular usage.

cedoss-lbl (approved):
Approved!  See previous comment

skbarber:
awesome, sounds good.

For 3, I don't think there is actually anything to do. The list of
composite variables is loaded by default as part of the device_manager. So,
if you choose a defined composite variable, the scan manager will handle it
by finding it in the composite_variables.yaml and using that to setup the
scan steps. I suppose you would just need to add mappings in your
"scan_devices.yaml". Or, maybe, you could probably even just populate the
scan_devices.yaml using the composite_variables definition to avoid
duplicating things.

For 2, using the setup_actions I don't think there is anything to do
either. If the setup_action is defined in the save_device yaml, the device
manager will take care of it. We could probably add an additional button or
two for some other basic actions. Like, the turning save to off for cameras
etc. should be very easy to do.

For relative variables, again, there might not be much that needs to be
done. I might have to understand a little bit more about how the
scan_config is generated in the GUI. But, the primary use case I had in
mind was doing a relative scan of steering elements, which would be
inherently a composite variable. And I already added the "relative" flag in
the composite_variable definition, so it should be handled natively in the
background

Also, I started a fresh gpt session to figure out the sounds and it worked
much faster and easier. So, sounds are supported now across windows/mac
platforms. I'll commit it soon.

PR #61, opened 2024-10-18
Geecs scan manager refactor

skbarber:
refactor of scan_manager into a second class ScanDataManager that is used solely for preparing, organizing and saving the logged scalar data from the scan, e.g. the "sfiles", and the .ini files etc. 

tested: no scan, composite variable scan in both jupyter-notebook and with the GUI and everything seems to work


skbarber:
i'm just directly merging to master since this was mostly an organizational change.

PR #63, opened 2024-10-29
Geecsscanner dev 10172024

cedoss-lbl:
Primarily adding the window for editing the config files.  Although there is some extra stuff in here to actually use the repetition rate and timing device.  I suspect this is what the merge conflict is, so I'll fix that and then merge afterwards.


GEECS-PythonAPI/geecs_python_api/controls/data_acquisition/scan_manager.py
```diff
@@ -36,15 +37,15 @@ class ScanManager():
...
+    def __init__(self, experiment_dir=None, shot_control_device=""):
+        self.device_manager = DeviceManager(experiment_dir=experiment_dir)
+        self.data_interface = DataInterface()
         self.action_manager = ActionManager(experiment_dir=experiment_dir)
         self.data_logger = DataLogger(experiment_dir, self.device_manager)  # Initialize DataLogger
 
         self.save_data = True
 
-        self.shot_control = GeecsDevice('U_DG645_ShotControl')
+        self.shot_control = GeecsDevice(shot_control_device)
```
skbarber, line 48: I get the intention of generalizing this beyond HTU. I suspect the shot_control_device is being read from the database or something? The downstream implementation is still hardcoded to use a specific channel for our DG645...I guess my point is, this generalization is only half of the story...Probably a good step in the right direction though

skbarber: Just looked a bit more. Looks like the shot_control_device is purely user defined from the GUI? A shot control device is generally (though not "required") as part of setting up any GEECS device and can be read from the database. In any case, probably something to think about later

skbarber (approved):

GEECS-PythonAPI/geecs_python_api/controls/data_acquisition/scan_manager.py
cedoss-lbl: Right, although I also have this as something that is stored within the .config file for the GEECS Python API.  So in general this is something that *could* be set in the GUI but should instead just be set in the config file and you won't have to think about it.

Now that I am thinking about it, something that could be good eventually is something on this GUI that automatically generates a .config file with appropriate elements if it doesn't exist.  I'll put that on the wishlist, but for now I may just provide a .config file in this folder to use for HTU.

cedoss-lbl:
Brought your additions into my branch.  Going to test that the scanner still works, fix that one bug with the newline characters in the scan info text box, and merge to master.

cedoss-lbl:
As per feature requests:
1. Added refresh list button for the save devices
2. Fixed bug with newline characters.  Now get replaced with spaces
3. Tab now correctly goes through the 1d scan boxes.

PR #65, opened 2024-11-01
Scan analysis update

kjensen11:
In scan_analysis.py, generalized CameraImageAnalysis to a parent class and created VisaEBeamAnalysis as a child class. Currently no change to analysis functionality. Pointed visa spectrometer set up to use VisaEBeamAnalysis instead of CameraImageAnalysis.

In the process, updated Visa 8 fixed crosshair coordinates (YAG screen crosshairs).

I do not foresee any merge conflicts as the scope of this work was pretty limited.

### DIFF
--- a/GEECS-PythonAPI/geecs_python_api/controls/data_acquisition/configs/experiments/Undulator/aux_configs/camera_analysis_settings.yaml
+++ b/GEECS-PythonAPI/geecs_python_api/controls/data_acquisition/configs/experiments/Undulator/aux_configs/camera_analysis_settings.yaml
@@ -9,6 +9,7 @@ UC_Phosphor1:
   Threshold: 0
   Rotate: False
   Calibration: 0.0000165
+
 UC_ALineEbeam1:
   Background Level: 60
   Left ROI: 1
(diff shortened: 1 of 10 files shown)


skbarber:
I think this is a good start. I don't think that methods like create_cross_mask should be in both the parent and child class. In fact, I think it makes sense to move some of these basic utilities to the image_analysis module which has a tools module. 

I'm also doing some work on the scan_analysis module. Why don't we merge this to master, then let's create a 'scan_analysis' only branch that we can work on simultaneously

kjensen11:
> I think this is a good start. I don't think that methods like create_cross_mask should be in both the parent and child class. In fact, I think it makes sense to move some of these basic utilities to the image_analysis module which has a tools module.

Oops. Agree that they shouldn't be in both. I thought I removed them from the parent. I'll do that before merging. I will leave the functions in the Visa class for now but we can move them after if we want. Being that there aren't any other analyses that really require removing crosshairs, I don't see an issue with postponing moving them to a basic utilities folder since they are somewhat specialized. But ultimately I'm open to whatever.

> I'm also doing some work on the scan_analysis module. Why don't we merge this to master, then let's create a 'scan_analysis' only branch that we can work on simultaneously

I'll merge now.

kjensen11:
Updated scan analysis (see recent commit). Removed remaining visa functions from parent class. Changed preprocessing function to 'crop_image' as that is the only remaining functionality. This can be further organized moving forward. Will merge now.

PR #66, opened 2024-11-04
Blue diode calibration kj

kjensen11:
Created a calibrations script directory (can be modified in the future) and blue diode undulator alignment calibration script. The blue diode calibration script contains an analysis class and the script can be run either standalone or the class can be used for scripted scan post analysis.

I also added camera config attributes for blue diode centroid x and y for each visa camera.


kjensen11:
IRL confirmation. Going to merge.

PR #67, opened 2024-11-04
Implement blue beam reference on visa analysis

kjensen11:
Implemented blue beam diode reference on Visa screen analysis. In particular, this is currently only implemented on the image array function.

Current implementation through a nearly duplicated 'create_image_array' function in the child VisaEBeamAnalysis class. I took this route for speed and simplicity since I did not know how widely we might want a reference point on image arrays. If it a widespread desire, we can generalize the create_image_array function in the parent CameraImageAnalysis class.


skbarber (approved):
I think this is ok for now. I am noticing we have lots of duplicated code both in run_analysis for the camera things. I think it's all functional for now, so that's good for near term, but we should looks for ways to consolidate

PR #68, opened 2024-11-04
Geecsscanner dev 10302024

cedoss-lbl:
New additions to the GEECS Scanner GUI.  Most notably the multiscanner.  Can now stack complex lists of presets to independently set what is saved and what is scanned.  These lists can then be saved and loaded, all within the GUI framework


cedoss-lbl:
Oh yeah, this branch also had the button to generate/fix the .config file if it doesn't exist or doesn't have the correct entries.

skbarber (commented):
looks good. no real change to the data_acquisition, merge!

PR #70, opened 2024-11-06
Create MultiScanAnalysis and MultiScanAnalysis_Visa

kjensen11:
Created base MultiScanAnalysis class which has framework to parse and organize scan data for a list of input scans.

Created child MultiScanAnalysis_Visa class which builds upon the parent class. The child class can either perform the binning analysis of VisaEBeamAnalysis or load in processed images from a previous analysis. Then, a matrix of averaged and processed images is constructed and saved to the analysis folder for the first specified scan.

Note the processed image loading option currently does not work as the image saving function was not up to date with master. This should be troubleshot after merging.

We should probably find a proper home for this analysis. Likely in the post processing directory under construction on master.


GEECS-PythonAPI/geecs_python_api/controls/data_acquisition/multiscan_analysis.py
```diff
@@ -0,0 +1,294 @@
...
+                      else list(input_scans['month']))
+        scan_days = ([input_scans['day']] * len(scan_numbers)
+                      if isinstance(input_scans['day'], str)
+                      else list(input_scans['day']))
+
+        # check compatibility of scan dates
+        if (len(scan_years) != len(scan_numbers)
+            or len(scan_months) != len(scan_numbers)
+            or len(scan_days) != len(scan_numbers)):
+            raise Exception("Error: Incompatible dates specifications.")
```
cedoss-lbl, line 55: This formatting of `scan_years`, scan_months`, and `scan_days` seems a little complex at first glance, but looking at an example down on line 279 looks pretty reasonable.  I suppose here you have the option to define a list of scans spanning multiple days/months/years to fully customize your dataset rather than assuming just a single day.  So long as the input formatting is clear than should be fine I think

GEECS-PythonAPI/geecs_python_api/controls/data_acquisition/multiscan_analysis.py
```diff
@@ -0,0 +1,294 @@
...
+
+    @staticmethod
+    def load_sfile(filepath):
+        dataframe = pd.read_table(filepath, header=0)
+        return dataframe
+
+class MultiScanAnalysis_Visa(MultiScanAnalysis):
+
+    def __init__(self, input_scans):
+
```
cedoss-lbl, line 92: if you are making a child class, proper notation is to call the init of the super:
`super().__init__()`
then you can just pass in input scans to the super's initialization if they get handled there.  And, if there is nothing special about initializing the VISA multiscan you could just skip it alltogether

GEECS-PythonAPI/geecs_python_api/controls/data_acquisition/multiscan_analysis.py
```diff
@@ -0,0 +1,294 @@
...
+        analysis_directory = analysis_path / f"Scan{scan_info['num']:03d}"
+        sfile_path = analysis_path / f"s{scan_info['num']}.txt"
+
+        return scan_directory, analysis_directory, sfile_path
+
+    @staticmethod
+    def load_sfile(filepath):
+        dataframe = pd.read_table(filepath, header=0)
+        return dataframe
+
```
cedoss-lbl, line 88: More down-the-road when we generalize things, it would be good to define classes that every multiscan analysis should have implemented.   In that train of thought, a blank `analyze()` function that just returns an error when it is called could be good, and then the idea would be that all child classes would need to write their own `analyze()` function to avoid that error.

GEECS-PythonAPI/geecs_python_api/controls/data_acquisition/multiscan_analysis.py
```diff
@@ -0,0 +1,294 @@
...
+        self.scans = self.parse_scans(input_scans)
+
+    def parse_scans(self, input_scans):
+        """
+        Parse input scans and organize.
+        """
+
+        # use super parse function
+        scan_dict = super().parse_scans(input_scans)
+
```
cedoss-lbl, line 105: I think here as well I like the idea of moving `parse_scans()` into the `__init__`() function, since parsing the scans into proper lists seems to be something that has to be done at the initialization to begin with.

GEECS-PythonAPI/geecs_python_api/controls/data_acquisition/multiscan_analysis.py
```diff
@@ -0,0 +1,294 @@
...
+        # identify scan parameter bin values
+        bins = sfile[scan_parameter].unique()
+
+        # identify shots for each bin
+        binned_shots = {val: sfile.loc[sfile[scan_parameter] == val, 'Shotnumber'].tolist() for val in bins}
+
+        # for each bin, load images and average
+
+        pass
+
```
cedoss-lbl, line 140: If backup is not VISA-specific, could add this into the parent class

GEECS-PythonAPI/geecs_python_api/controls/data_acquisition/multiscan_analysis.py
cedoss-lbl: or `run_routine()` as you have at the bottom of this script

GEECS-PythonAPI/geecs_python_api/controls/data_acquisition/multiscan_analysis.py
```diff
@@ -0,0 +1,294 @@
...
+        # get bin numbers and values (assumes correlated bin number and value!!)
+        bin_num = np.unique(analyzer.bins)
+        bin_vals = analyzer.binned_param_values
+
+        # load already processed images
+        img = cv2.imread('Z:/data/Undulator/Y2024/10-Oct/24_1031/scans/Scan053/UC_VisaEBeam1_1_processed.png', cv2.IMREAD_UNCHANGED)
+        scan['binned_data'] = {val: {'value': bin_vals[ind],
+                                     'image': read_imaq_image(analyzer.scan_directory / f"{scan['device']}_{val}_processed.png")}
+                               for ind, val in enumerate(bin_num)}
+
```
cedoss-lbl, line 182: Down-the-road comment:  could be cool if there was a way to generalize this to all multiscan analysis routines and put in the parent.  I could see us rewriting this kind of function often

GEECS-PythonAPI/geecs_python_api/controls/data_acquisition/multiscan_analysis.py
```diff
@@ -0,0 +1,294 @@
...
+            if run_analysis:
+                self.scans[key] = self.scan_analysis(item)
+            else:
+                self.scans[key] = self.load_analysis_results(item)
+
+        # compile image matrix
+        self.compile_image_matrix(self.scans)
+
+        return
+
```
cedoss-lbl, line 269: A broad comment for the class as a whole:  you can remove a lot of arguments in class functions by instead assigning them as class variables during the initialization.  IE, you could have a `run_analysis=False` during the `__init__()` then you won't need to pass it in `run_routine()`.  This can potentially simplify a lot of re-reading down the line because all of the options are just in one place instead of scattered throughout the arguments of methods

skbarber:
Hey kyle,

I'm definitely not much of an authority on github practices, but it looks
like this pull request is to add your changes directly into master. I think
it's best to first merge Master into your branch and sort out merge
conflicts in that space, rather than in the master branch. I did just this.
Merged master into your branch and went through and sorted out the
conflicts in the scan_analysis class. I tested doing analysis on one of the
visa cameras and it seems to work. One important thing, the crosshair mask
needs to be applied before cropping.

cedoss-lbl (approved):
Looks good to me to start!  Like Kyle mentioned we may move this outside of GEECS-PythonAPI at some point and/or generalize it to any HTU (or BELLA) multiscan analysis, but we'll cross those roads when we get there.

skbarber:
also, it seems the perform_bulk_analysis method is a good general start but not really implemented to it's full potential as it seems to be more or less copy pasted to the visa analysis class. I think what would be good would be to define a fully generic analysis workflow that could be used by the generic class and exploited by the child class. For example: i) average images by bin and save, ii) apply a mask (could be crosshair or any other mask), iii) crop image, iv) apply filtering v) perform calculations, e.g. center of mass, fwhm etc.). I think it's good to always apply any mask to the full ROI before cropping so that there's never any question as to what it is reference to.

kjensen11:
> Hey kyle, I'm definitely not much of an authority on github practices, but it looks like this pull request is to add your changes directly into master. I think it's best to first merge Master into your branch and sort out merge conflicts in that space, rather than in the master branch. I did just this. Merged master into your branch and went through and sorted out the conflicts in the scan_analysis class. I tested doing analysis on one of the visa cameras and it seems to work. One important thing, the crosshair mask needs to be applied before cropping.

Ya I think we are all learning as we go unfortunately. When I made this branch, I had not yet actually merged my previous pull request into master (you were still working on some stuff and then went through and resolved merge conflicts). Then somehow, after the previous pull request was merged, I don't think this new branch properly caught up to master. Hence why I didn't have the correct image save functions.

Now, we both actually did simultaneous merge conflict resolutions and I have no idea how that will come together... ha

I think what I will do is hold off on pushing this merge through and look back to see what my current branch looks like and retest things. At the same time, I will try to apply changes regarding some of Chris's comments and better streamline the perform bulk analysis method. BTW, the initial goal of the bulk analysis was to separate all analysis from plot making so the analysis could be called in other scripts. Still this does not negate your comment about better streamlining.

Fingers crossed this sorts out...

kjensen11:
@skbarber Ok, so brief look over things and I think merging master into this branch removed a lot of the gaussian filter analysis stuff you added. Because I noticed the simultaneous conflict resolutions prior to submitting my merge, I still have my fully resolved script in a web browser. When there is time tomorrow, I will go back through line by line and sort out this branch and then revisit pull requesting/merging. If you make much for changes before then, maybe shoot me a short note so I can be sure to include that as well.

skbarber:
honestly, i wouldn't worry too much about it. A lot of stuff I had added
was slap dash to try and make it useful immediately during HP. It's as
functional in its current state as it needs to be for now. I did add the
filter method back into the merged version, but it's not being used for the
moment. Again, no big loss. I would rather focus on removing these types of
methods (like cropping, filtering and even crosshair masking) from the
scan_analysis module and move them into the tools/utils for the
image_analysis module. scan_analysis should be used to define a workflow of
image_analysis steps to apply to the data.

PR #71, opened 2024-11-07
Livescananalysis dev 11052024

cedoss-lbl:
After looking at some of Reinier's code, seemed to me like the new data_acquisition and analysis paradigm of binning images before analysis would benefit from a newer implementation of some of his ideas.  Here I made a new folder `livescananalysis\` where I just have two simple files for the time being.  

One of them is able to take a scan folder and check what analyses could be performed on it, based on the contents of that scan.  For instance, a Rad2 analysis would required the Rad2 camera and at least one of the Visa cameras.  These requirements can be expressed with AND/OR logic.

The other script sets up a "lookout" thread that periodically checks the daily data path for new scan folders.  I do this by seeing if there is a new `ScanXXX/ScanXXX.tdms` file, as that is generated at the end.  If there is a new one, then it checks what analyses can be performed and then adds this scan to a big yaml that contains all of the scans previously analyzed.  This thread continues until we stop in (which, if we made a simple gui this could be a button).

Currently living in LivePostProcessing with the intention of calling our scan analysis functions (wherever they may end up living), but this doesn't need to stay here.


skbarber (approved):
looks good. Only substantive comment is to switch to pathlib rather the os for file operations. 

And then, maybe a more down the line thing would be to think about centralizing some of these standard data file type operations. That was the initial idea behind the DataInterface class i started as part of the data_acquisiiton. But, it quickly got abused and got a bunch random methods added to it...

skbarber:
ahhhh, right. Kind of forgot about the stupid python 3.7 requirement for
image analysis...

PR #72, opened 2024-11-09
Geecsscanner dev 11072024

cedoss-lbl:
Lots of small features and bug fixes to the GUI.  Also started on an installation README to help out with installing the virtual environment.


skbarber (approved):
good catches on tolerance and the async listeners.

PR #78, opened 2024-11-20
Visa analysis spatial calibration

kjensen11:
Create a working spatial/angular calibration of electron beam transport into the undulator based on magnet currents.


kjensen11:
No merge conflicts so going ahead and merging so we can proceed with greater code migration and organization.

PR #80, opened 2024-11-21
Update scan manager mapping

kjensen11:
Updated scan manager mapping of analysis locations. Updated geecs_python_api poetry toml and lock. Updated the name in ScanAnalysis poetry toml to appease poetry (it didn't like that the name was scan_analysis while the folder was ScanAnalysis).


kjensen11:
I don't think this requires much oversight but will leave it here until end of day in case you want to check.

cedoss-lbl (approved):
Should be all good!  Probably just a conflict due to all of the whitespace changes, and in that case resolving them will be easy

kjensen11:
Merge conflict was the new optional argument "MC_ip". Conflicts resolved and going ahead with a merge to master.

PR #83, opened 2024-11-21
Small updates to set up default analysis for UC_UndulatorRad2 and modify save paths for cam analysis

kjensen11:
Set up default analysis for UC_UndulatorRad2 to be CameraImageAnalysis until a more specific analysis exists. Modified the save paths for CameraImageAnalysis and VisaEBeamAnalysis to point to the analysis folder rather than the scan folder.


cedoss-lbl (approved):
Looks good, once everything is neatly in ScanAnalysis we can try deleting the duplicates and check that it all works.

kjensen11:
Moving ahead with merge.

PR #86, opened 2024-11-21
Scandirectory dev 11212024

cedoss-lbl:
Moving the recently-created and stripped-down `scan_folder.py` into the old directory where `scan_data.py` currently lives.  Now, `ScanData` is a daughter class of this new `ScanFolder` class, and all of the basic file operations are contained in `ScanFolder`.

I removed all of the functions that were equivalent between ScanData and ScanFolder on the `scan_data.py` side, the one exception being `load_scalar_data.py`.  In the new version, I had the tdms read within a try-except block because some recent data we took had corrupted tdms files (an error we have since corrected).

The other issue is that `ScanData.__init__()` is built slightly differently from `ScanFolder.__init__()`, mostly from the ScanData version doing some additional checks with `GEECSDevice` to get the experiment name and see if the folder is "online" and in the Z drive.  I currently just have both initialization functions intact, (with a call to the super's initialization inside ScanData), but I suspect there is a better way to write this.
- My first thought would be to split up the initialization functions between so that ScanData does not need to go through all of ScanFolder's initialization.  If you agree, I can make this change and update `scan_data.py` further.


TAU-Reinier (approved):
Hey Chris, this is excellent.

TAU-Reinier:
Oh I see now that the __init__() isn't quite right: 

The duplicate code in the __init__() methods can be removed from ScanData.__init__(). Calling super().__init__ within ScanData.__init__() runs the init code from ScanFolder on the ScanData object, so it now runs twice.

cedoss-lbl:
Trimmed away most of the duplicate code, and I believe the functionality should all be preserved for scan_data.  Namely, one can either give an exact folder and have it ignore GEECSDevice and work offline, or it can use the information in GEECSDevice to build the data path.

Ran into an issue where the child class could not access protected class variables of the parent.  This is the case for `self.__folder` and 3 other variables.  I now have it so that these get passed now through a `self.folder` class variable, but this just seems bad.  Will probably just try to remove all of the protected class variables and refactor the rest of the code in these two files so that we don't have two different sets of protected class variables floating around.

skbarber:
hey chris, I like the progress. A couple of things: to me it makes sense to move all of this stuff up one level from analysis/images/scans to analysis/scans. Doesn't make a lot of sense to have it in 'images'. I think this was something that Reinier/Guillaume had been working on in an unmerged branch. 

Also, at this point, I'm not entirely sure what the purpose of scan_data is...to me, what is in scan_folder seems more like it should be scan_data and what's left in scan_data looks to be more related to doing specific types of magspec analysis for HTU...I guess my suggestion is to rename scan_folder to scan_data

cedoss-lbl:
I do like scan_folder/data being in geecs_python_api somewhere because it does still use a few things from the api, but I am not opposed to it living in a different folder.  Maybe `tools/interface/`, or top level `analysis/`?

Keeping the old scan_data called scan_data and just creating a new file called scan_folder would be ideal, as it preserves existing functionality.  But you are also right that the current version of scan_data is just a magspec analysis script.  I will write Guillaume an email to ask if he is fine with us renaming/moving scan_data.

TAU-Reinier (approved):
Looks good!

skbarber (approved):
i like it.

PR #88, opened 2024-11-25
Geecs python api dev 11252024

cedoss-lbl:
Here I have added some more functionality that would be useful to have in ScanData.  Most of which are included here as static methods within the ScanData class.
- Given the year, month, day, and scan number, return a ScanTag with the proper formatting.  For instance, month can be 11, '11', 'nov', or 'November'
- Return a ScanData class for the last available scan of the given day.  If no day is given, assume today
- Return the folder for what the next scan would be.  Defaults to today if no date given
- Returns the ScanData for the next scan, creating the folder in the process.  Defaults to today if no date given.

Changes:  (Mostly things to speed up the __init__ method)
- Moved a lot of functionality out of the __init__ method.  Now a lot of this is done within the class methods.  For instance, the `analysis` folder is only created when the `get_analysis_folder` method is called for the first time.
- Added a "read_mode" boolean flag to the __init__ method.  Defualts to `True`, and is used if the requested folder does not exist.  If in read_mode, then throw an error.  If not in read_mode, then make the folder.

Other thoughts now that I am writing this up:
- Would it be more useful to return the Scan Tag of the next scan rather than the Path?  That way, you can easily get the scan number from the Tuple rather than parsing it from the path, and then you can use a built-in method to make the folder from the tag.
- The generalized `get_scan_tag` method was the last thing I made, so I might go through and add this where applicable instead of making a ScanTag manually.
- Most of this ScanData class is now in static methods.  Does this mean we should refactor to a standalone module?


TAU-Reinier (commented):
This PR adds some more date flexibility but I would argue that less date flexibility is better, and I also wonder if this is a good time to bite the bullet and change all year, month, day references in ScanTag and elsewhere into datetime.date objects. * 

Parsing dates in various formats is important is good if it comes from unknown inputs such as text files or prompted keyboard input, but I think in all cases here dates come from Python scripts using GEECS-PythonAPI. So forcing a date format - or better yet a datetime.date object - reduces the amount of case handling you have to do and thus prevents edge cases and errors. 

\* This would be a breaking change, so either ensure that no one is using ScanData in its current form, or continue to support both old and new for a time with a deprecation warning.

cedoss-lbl:
While I'm not opposed to having a more rigid datetime.date object to standardize the inputs for the inner workings of geecs-python-api, I think having more flexibility is useful for the environment of the BELLA center.  While I'm not sure just how well used this will become, it would be good to have an easy on-ramp for people with little coding experience.  I've talked to people that think importing numpy is a huge hassle, so making it as easy as possible would be beneficial.  With the varying levels of coding experience and varying styles of writing dates, it is almost as if the dates are coming from unknown sources.

GEECS-PythonAPI/geecs_python_api/analysis/scans/scan_data.py
```diff
@@ -119,6 +108,56 @@ def build_device_shot_path(tag: ScanTag, device_name: str, shot_number: int, fil
...
+        """
+        if year is None or month is None or day is None:
+            today = datetime.today()
+            year = today.year
+            month = today.month
+            day = today.day
+
+        i = 1
+        new_scan_flag = True
+        while new_scan_flag:
```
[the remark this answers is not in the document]
cedoss-lbl: Good call, while it would be nice for a new day to start with the fresh scan 001 as the next, a previous day should say something if the latest scan for that day was to be scan000.  Probably return None is fine

GEECS-PythonAPI/geecs_python_api/analysis/scans/scan_data.py
```diff
@@ -119,6 +108,56 @@ def build_device_shot_path(tag: ScanTag, device_name: str, shot_number: int, fil
...
+            try:
+                ScanData(tag=tag, experiment=experiment, load_scalars=False, read_mode=True)
+            except ValueError:
+                break
+            i = i+1
+        return ScanTag(year, month, day, i-1)
+
+    @staticmethod
+    def get_latest_scan_data(experiment: str, year: Optional[int] = None,
+                             month: Optional[int] = None, day: Optional[int] = None) -> 'ScanData':
```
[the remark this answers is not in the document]
cedoss-lbl: Ahhh I see!  That's what this is used for

GEECS-PythonAPI/geecs_python_api/controls/api_defs.py
```diff
@@ -28,3 +28,33 @@ class ScanTag(NamedTuple):
...
+        'dec': 12, 'december': 12,
+    }
+    try:
+        month_int = int(month)
+        if 1 <= month_int <= 12:
+            return month_int
+    except ValueError:
+        pass
+
+    month_lower = month.lower()
```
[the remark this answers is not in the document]
cedoss-lbl: Gotcha, this is definitely more compact

GEECS-PythonAPI/geecs_python_api/analysis/scans/scan_data.py
```diff
@@ -38,68 +42,55 @@ def __init__(self, folder: Optional[SysPath] = None,
...
-                    raise ValueError("Folder path does not appear to follow convention")
-                elif not folder.exists():
-                    raise ValueError("Folder does not exist")
+            folder = Path(folder)
 
-                self.__tag_date = dtime.strptime(date_folder_name, "%y_%m%d").date()
-                self.__tag = \
-                    ScanTag(self.__tag_date.year, self.__tag_date.month, self.__tag_date.day, int(scan_folder_name[4:]))
+            (exp_name, year_folder_name, month_folder_name, date_folder_name,
+             scans_literal, scan_folder_name) = folder.parts[-6:]
```
skbarber, line 64: (exp_name, year_folder_name, month_folder_name, date_folder_name,
             scans_literal, scan_folder_name) = folder.parts[-6:]

I think this part should come after the checking if the folder follows expected formatting (lines 71+)

skbarber (approved):
I think there was the one small bug noted, but looks good otherwise.

GEECS-PythonAPI/geecs_python_api/analysis/scans/scan_data.py
cedoss-lbl: This is more of an issue with how Github decided to present the changes.  I removed a try-except block so everything got un-indented, and it got very confused as to how to present the changes.  But if you look at the line numbers, the parsing is on lines 62-63 and the if statement is on lines 71-76

GEECS-PythonAPI/geecs_python_api/controls/api_defs.py
```diff
@@ -32,29 +33,11 @@ class ScanTag(NamedTuple):
...
-        'aug': 8, 'august': 8,
-        'sep': 9, 'september': 9,
-        'oct': 10, 'october': 10,
-        'nov': 11, 'november': 11,
-        'dec': 12, 'december': 12,
-    }
     try:
         month_int = int(month)
         if 1 <= month_int <= 12:
             return month_int
```
TAU-Reinier, line 39: Do handle the case if month_int is outside 1..12, because otherwise dateparse(month) will return unexpected results. Consider checking isinstance(month, str) before the dateparse.

cedoss-lbl: Good call, added it in!

PR #89, opened 2024-11-26
Geecsscanner dev 11222024

cedoss-lbl:
For the v0.2 GEECS Scanner version, lots of bug fixes and other small code edits.

- Fixed multiscanner skipping steps (was related to threads issue)
- Changed all "os" usage to cleaner "pathlib" usage
- Added type hints to most returns and method arguments
- Renamed "Cancel" button in element editor to "Close"
- Fixed runaway threads from SoundPlayer
- Fixed ScanManager throwing crashes if base_monitoring_devices, actions.yaml, or composite variables don't exist.  This would be an issue if other experiments ever started trying to use this.
- Fixed bug with preset list not updating when experiment is changed
- Removed extraneous "Scan Parameters" component of input dictionary to Scan Manager
- Fixed issue in TDMS file creation where "Bin #" and "Elapsed Time" would add an extra whitespace to the end of their variable.  This would be annoying when trying to work with older analysis scripts that used previous TDMS variable formatting.


GEECS-PythonAPI/geecs_python_api/controls/data_acquisition/data_acquisition.py
skbarber, line 1: not entirely sure I follow the new logic for the sound threading, but if it's tested and clears threads better that's good. There's no need for a stop_queue method?

skbarber (approved):

GEECS-PythonAPI/geecs_python_api/controls/data_acquisition/data_acquisition.py
cedoss-lbl: There's a stop_logging method in data logger (line 1381) that calls Sound Player's "stop"

PR #90, opened 2024-12-03
No scan camera image analysis

kjensen11:
Create a no scan analysis for CameraImageAnalysis that applies image processing to each image, averages the images (and saves), and creates a gif of the images for visualization (and saves).

Refactored some code in CameraImageAnalysis and VisaEBeamAnalysis. Removed the bulk image analysis function as it began to feel like an unnecessary wrapper function. Bin and image looping is handled in the scan and no scan analyses while image processing is contained within the respective image processing functions using inheritance as needed.

Each case has been tested (no scan and scan for both classes).


kjensen11:
No merge conflicts exist and scope of work is quite limited to the above classes. Going ahead with merge. Will return to properly document... (i know, i know, I should have done it already...)

PR #91, opened 2024-12-03
Geecs scan manager refactor

skbarber:
This PR is to eliminate the DataInterface class that was initially used for handling processing data generated during scans. The functionality that was in this class has been migrated to ScanData and ScanDataManager. Note, ScanDataManager is a class used exclusively by ScanManager to handle data generated by ScanManager, which executes the scans. Probably it should be renamed down the line

One general change has been added to ScanData which allows ScanManager etc. to work more seamlessly across platforms. The root paths to the data can now be specified in the .config file, and if not present, default values are used (e.g. root paths for HTW CR windows computers). Console logging was also added to scan_data for improved debugging.

Tested single parameter scans and 1D parameter scans on both remote mac and internal windows computer and all seems functional

FUTURE WORK:
The ScanDataManager is used to organize and rename saved files to follow MC convention as well as update the shotnumber in the file name using the timestamp embedded in the file meta data and matching that to the time stamp in the sFiles. Right now, ScanDataManager is makes use of the device_manager for a couple things. This makes it inconvenient to use the renaming functionality after a scan has completed. In principle, it is possible to restructure a bit so that ScanDataManager only needs ScanData, in which case ScanDataManager could be completely standalone.


GEECS-PythonAPI/geecs_python_api/controls/data_acquisition/scan_manager.py
```diff
@@ -336,6 +365,152 @@ def fill_async_nans(self, log_df, async_observables, fill_value=0):
...
+        """
+        Rename master files based on scan number, device name, and matched row index.
+        """
+        for file_path, row_index in matched_rows:
+            row_number = str(row_index + 1).zfill(3)
+            new_file_name = f"{scan_number}_{device_name}_{row_number}{file_path.suffix}"
+            new_file_path = file_path.parent / new_file_name
+            if not new_file_path.exists():
+                logging.info(f"Renaming {file_path} to {new_file_path}")
+                os.rename(file_path, new_file_path)
```
cedoss-lbl, line 511: Was this `os` here because it is more compatible with the `ThreadPoolExecutor()` of line 402 than `pathlib.Path` for multithreaded file operations?  Otherwise, could consider changing to a `Path` to entirely remove `os`

skbarber: yeah, I tend to agree. Originally, I had some other code here to automatically determine relevant paths based off specific machine details (domain, computer name etc.). I switched to this config file based approach so as to not have specific machine details in the code. 

The load_config function in geecs_database was added by me some time ago when we first introduced this config file. But it's not really the right spot for it. Probably it makes sense to consolidate/move the stuff in geecs_database and here to geecs_python_api/tools/interface. But, I'll leave that as a future task.

GEECS-PythonAPI/geecs_python_api/analysis/scans/scan_data.py
```diff
@@ -7,23 +7,134 @@
...
 from geecs_python_api.tools.distributions.binning import unsupervised_binning, BinningResults
 
+# Create a module-level logger
+logger = logging.getLogger(__name__)
+
+# Set up default logging only if no handlers are present
+if not logging.getLogger().hasHandlers():
+    logging.basicConfig(level=logging.INFO, format="%(asctime)s - %(name)s - %(levelname)s - %(message)s")
+
+class GeecsPathsConfig:
```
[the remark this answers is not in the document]
skbarber: yeah, I tend to agree. Originally, I had some other code here to automatically determine relevant paths based off specific machine details (domain, computer name etc.). I switched to this config file based approach so as to not have specific machine details in the code.

The load_config function in geecs_database was added by me some time ago when we first introduced this config file. But it's not really the right spot for it. Probably it makes sense to consolidate/move the stuff in geecs_database and here to geecs_python_api/tools/interface. But, I'll leave that as a future task.

GEECS-PythonAPI/geecs_python_api/controls/data_acquisition/scan_manager.py
skbarber: switched to pathlib

ImageAnalysis/image_analysis/utils.py
```diff
@@ -118,13 +118,23 @@ def get_magspecstitcher_timestamp(fname):
...
+        file_path (str or Path): Path to the TDMS file.
+
+    Returns:
+        The timestamp if it exists, or None otherwise.
+    """
+    try:
+        # Open the TDMS file in lazy mode (if supported by the library)
+        with TdmsFile.open(file_path) as tdms_file:
+            return tdms_file.properties.get('timestamp')
+    except Exception as e:
```
TAU-Reinier, line 134: Be more specific with which Exception you're catching - i.e. only catch the Exception that you expect from file reading. You want any unexpected Exceptions to fail loudly.

TAU-Reinier (approved):

skbarber:
Made some minor debug/config file changes after using scan_manager during HP run. Will merge now

PR #92, opened 2024-12-04
Scan analysis dev 11252024

cedoss-lbl:
Not the final version of ScanAnalysis, but this is a working version and we thought it would be good to go ahead and make a pull request while making the finishing touches.  As of now, the live watch is performed by running the `__main__` function of `live_watch\scan_watch.py` with the specified date and experiment name.  It is standalone, so there is no immediate rush to merge to master as this won't impact the python scanner's functionality.

## Additions:
- In `live_watch\scan_watch.py` there is a watchdog that gets notified if a new scan was added.  This started as a copy of Reinier's original watchdog in `LivePostProcessing`, but it has been changed a fair bit.  It still looks for s-files, and additionally it will also do a sweep of the daily folder to see if there are previously untouched scans.
- The watchdog thread only activates on signals, but it also just adds the scantag of a new scan to a Queue.  The main thread is still in an infinite loop that checks the queue every 0.5 seconds and attempts an analysis if there is a ScanTag in the queue, and this analysis is performed on the main thread.
- `devices_to_analysis_mapping.py` contains logic for determining which combinations of saved devices correspond to which analyzers.
- `execute_scan_analysis.py` maps the name of analyzers to their respective analysis class instance and performs the `run_analysis()` function for a given `ScanTag`
- We added an optional `experiment` field to the `ScanTag` NamedTuple.  This was nice because now you can specify a full scan folder with just the information in the ScanTag and using ScanData.  However, we may now have too many avenues to specify the experiment name.  (ScanData argument, ScanTag field, the global config file, maybe more...)
- `scan_analysis\base.py` has a parent class of `ScanAnalysis` that sets the formatting all scan analyzers should follow, but the child classes need to implement their own `run_analysis()` method.

## Known things we want to change:
- A GUI to launch and control the main while loop of `scan_watch.py` (long-term goal)
- Implement Rad2 spectrometer analyzer, as well as asign more cameras to the `CameraImageAnalysis` analyzer
- Move the two separate dictionaries in `execute_scan_analysis.py` and `devices_to_analysis_mapping.py` to a combined dictionary that is separated by experiment name and maps both the required devices and analysis class simultaneously.
- Better handling of logging messages
- Fix old calibration scripts using the new ScanData notations
- Return to `multiscan_analysis` and implement functionality to use `scan_analysis` to prepare data for `multiscan_analysis`.
- Implement optional config files to customize ScanAnalysis routines.


kjensen11:
I approve a merge

kjensen11 (approved):

ScanAnalysis/scan_analysis/mapping/map_Undulator.py
```diff
@@ -0,0 +1,25 @@
...
+
+from scan_analysis.analyzers.Undulator.MagSpecStitcherAnalysis import MagSpecStitcherAnalysis
+from scan_analysis.analyzers.Undulator.CameraImageAnalysis import CameraImageAnalysis
+from scan_analysis.analyzers.Undulator.VisaEBeamAnalysis import VisaEBeamAnalysis
+from scan_analysis.analyzers.Undulator.Rad2SpecAnalysis import Rad2SpecAnalysis
+from scan_analysis.analyzers.Undulator.HiResMagCamAnalysis import HiResMagCamAnalysis
+
+undulator_analyzers = [
+    Info(analyzer_class=MagSpecStitcherAnalysis,
+         requirements={'AND': ['U_BCaveICT', 'U_BCaveMagSpec']},
```
skbarber, line 11: Not sure I understand why U_BaveICT is a requirement for the magspecstitcher. Same goes for the other classes

GEECS-PythonAPI/geecs_python_api/analysis/scans/scan_data.py
skbarber, line 1: looks like there are conflicts in scan_data now with master. I made some changes and updates to scan_data in the scan_manager_refactor branch. Mostly organizational and enabling scan_data to be more platform agnostic. Looks like the only change here in scan_data is the use of 'experiment' as part of ScanTag. My suggestion would be to resolve conflict using master and just adding the 'experiment' variable/arg where needed

skbarber (approved):
Given this is pretty standalone, no urgency to merge. I would try to do that little bit of clean up in the base.py analysis type to use the scan_data methods for loading scalar data etc. When that's done, merge into master

ScanAnalysis/scan_analysis/mapping/map_Undulator.py
cedoss-lbl: You are right that U_BCaveICT isn't really used for the magspecstitcher as it is, so I will remove it from the required for that analysis.  But, I like having this for the HiResMagCam analysis to get the scalars of charge and I *will* like having this for when we do UCRad2 plots of light vs charge

ScanAnalysis/scan_analysis/base.py
```diff
@@ -58,77 +92,73 @@ def __init__(self, scan_directory, use_gui = True, experiment_dir = "Undulator")
...
         Returns:
             str: The scan parameter with colons replaced by spaces.
         """
         config = configparser.ConfigParser()
-        config.read(ini_file_path)
+        config.read(self.ini_file_path)
         cleaned_scan_parameter = config['Scan Info']['Scan Parameter'].strip().replace(':', ' ').replace('"', '')
         return cleaned_scan_parameter
 
     def load_auxiliary_data(self):
```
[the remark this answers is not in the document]
cedoss-lbl: Was able to change `load_auxiliary_data` and `extract_scan_parameter_from_init` to use respective functions from ScanData, but did not see a way to refactor this functionality out completely.  While ScanData has methods to load scalars and read these files associated with the scan, it is up to ScanAnalysis to get the useful stuff out of it (ie, bins and extracting the scan parameter out of the ini file contents).

cedoss-lbl:
Thanks for the suggestions!  We cleaned up some sloppy loose ends in the calibrations folder, updated existing unit tests, and included some type-hinting improvements in various locations.  The main functionality of running scans and running the live scan analysis appears to be working, so if there are no objects we'll proceed with a merge by end of day.

PR #96, opened 2024-12-12
Correct a few imports, and one more reference to renamed function

TAU-Reinier:
One remaining large issue is many references to DataInterface, which no longer exists.


cedoss-lbl:
Thanks for pointing these out!  I checked again and found a few more "build" -> "get" changes is a couple of places, so I'll include those in the original scan_analysis branch.  Meanwhile, I am looking at getting the old calibration scripts functional again and Kyle is working on multiscan.

I'll also clean up some of thee __main__ functions in the scan analyzers.  They are useful to have for testing/running single instances, but some of them are outdated with the old DataInterface usage

PR #97, opened 2024-12-12
Multiscan analysis update

kjensen11:
Fixed multi scan analysis to handle upstream current data and class structures. Currently only has functionality to handle scan data (not no scan data).


cedoss-lbl (commented):
Looks good, had to do another conflict resolution in multiscan_analysis but it was just an outdated import of DataInterface still in here

cedoss-lbl (approved):

PR #99, opened 2024-12-16
Geecsscanner dev 12052024

cedoss-lbl:
Fixes to the GUI to reflect some of the refactoring done around ScanAnalysis.  Calling this a v0.3 update.

- Removed Post-Analysis Class for the save elements
- Added a new button to open a timing system configuration editor, but this is not yet finished
- Scan Element Editor has been visually reorganized
- Closeout actions are now passed to Scan Manager
- Closeout actions no longer buggy in the Scan Element Editor
- Added buttons to execute the actions found within an element
- Reordered how the cursor moves when tabbing through Scan Element Editor
- `experiment` can be used instead of `experiment_name` for `ScanData.get_scan_tag`, where the latter is still valid but prints a warning


GEECS-PythonAPI/geecs_python_api/controls/data_acquisition/scan_manager.py
```diff
@@ -1175,7 +1176,11 @@ def _execute_step(self, component_vars, wait_time, is_composite, max_retries=3,
 
 
             time.sleep(interval_time)
-            current_time += interval_time
+            current_time = time.time() - start_time
+
+            if current_time % 1 < interval_time:
+                log_df = self.scan_data_manager._process_results(self.results)
+                self.scan_data_manager.dataframe_to_tdms(log_df)
```
skbarber, line 1183: I don't think you need the 'self.scan_data_manager.dataframe_to_tdms(log_df)' bit of code. That's already in the _process_results method. 

I'm not sure I understand this logic: if current_time % 1 < interval_time...

I get the idea here (just write the file while doing the steps), but maybe not the best implementation. I think effectively this is just dumping the file at various time steps. So, just overwriting existing files. What I think tony was after was to write the .tdms file in a similar fashion to MC. In MC, the file gets opened and written to continuously. And it is also possible to read/write to this file simultaneously. 

It all depends a little bit on what is trying to be achieved with the live dumping of the data files. Again, i don't quite follow the logic, but if you tested it and it works, this is a pretty brute force way to get the data files to update and maybe that's enough...

cedoss-lbl: We saw the `_process_results` method, but since they only wanted the tdms file to be dumped every shot I figured it would be best to save only the tdms and not also the txt file.  Otherwise, like you said this current implementation is pretty brute force and so writing more than we have to might not be efficient.

The `current_time % 1 < interval_time:` statement is just so that this code only executes once per second.  Otherwise it will get called every 0.1 seconds and that is too much.

I tested it with 1 camera, so we'll have to see what happens when there's a lot more going on in a scan.  Might be a significant chunk of time.  Would be good to debug just how long it takes to write a tdms file for a scan with 10 devices and 1000+ shots.  Might even want to consider a flag to turn off this feature if we plan on doing long, overnight scans.

Perhaps the biggest issue with leaving the tdms file open during the full run is the possibility of a crash occurring and then it gets corrupted?  But if that's not a problem then yeah we can revisit this at some point.

skbarber: my point on _process_results is that this is already in that method:
            # Write TDMS files (data and index)
            self.dataframe_to_tdms(log_df)
so no need to call it again after calling _process_results() as you have it on 1183.

I think I take back my point about overwriting the files. The tdms_writer stays open. So each time you call it, it is writing the full current data. I'm not sure there is anything that prevents you from writing duplicate entries in a tdms, so you might want to look at if you are just appending the full available dataframe at each step. 

I guess I understand the idea behind the code below now. But, I feel like it's not very deterministic. With current_time updating each time through the loop, in principle you would expect it to only increment by the interval_time. But, with other threads and processes going on, isn't it possible execution of that loop might take more than the interval_time? In which case, the decimal part of the number is going to be kind of random. 
            current_time = time.time() - start_time

            if current_time % 1 < interval_time:

cedoss-lbl:
Ok, I see your point on what I had written.  I must have confused myself and accidentally wrote `_process_results` instead of `convert_to_dataframe`.  Now it only writes the tdms, so should be a bit quicker.

I also made a flag (that is currently always False) which we can later make settable from the GUI to control if this happens.  But right now it is disabled.  Also talked to Chetanya about how we would change this script so that it changes from overwriting the file every shot to appending only the new data.  Seems pretty easy to accomplish.

Yeah, thats why we have it so that the mod operator is less than the interval time (0.1s) rather than == 0 for if current_time is always a multiple of 0.1.  If no tdms writing is happening, this loop is basically just a quick checking of flags and waiting for 0.1s.  That way it will be impossible for this tdms writer to ever get called more than once per 1s interval.

*However*, this logic may be flawed anyways because if tdms writing does take more than 1s then we may skip shots because the rest of the loop is waiting on the tdms writing.  To make this foolproof, we probably would want to send a new dataframe row to a queue and have a separate thread where a tdms writer just polls the queue and writes what it finds at its own pace.  But hopefully once we ensure that the writer is only appending and not overwriting then we won't have to worry about this speed issue.

skbarber:
> Ok, I see your point on what I had written. I must have confused myself and accidentally wrote `_process_results` instead of `convert_to_dataframe`. Now it only writes the tdms, so should be a bit quicker.
> 
> I also made a flag (that is currently always False) which we can later make settable from the GUI to control if this happens. But right now it is disabled. Also talked to Chetanya about how we would change this script so that it changes from overwriting the file every shot to appending only the new data. Seems pretty easy to accomplish.
> 
> Yeah, thats why we have it so that the mod operator is less than the interval time (0.1s) rather than == 0 for if current_time is always a multiple of 0.1. If no tdms writing is happening, this loop is basically just a quick checking of flags and waiting for 0.1s. That way it will be impossible for this tdms writer to ever get called more than once per 1s interval.
> 
> _However_, this logic may be flawed anyways because if tdms writing does take more than 1s then we may skip shots because the rest of the loop is waiting on the tdms writing. To make this foolproof, we probably would want to send a new dataframe row to a queue and have a separate thread where a tdms writer just polls the queue and writes what it finds at its own pace. But hopefully once we ensure that the writer is only appending and not overwriting then we won't have to worry about this speed issue.

GEECS-Scanner-GUI/action_control.py
```diff
@@ -0,0 +1,18 @@
...
+class ActionControl:
+    def __init__(self, experiment_name: str):
+        self.action_manager = ActionManager(experiment_dir=experiment_name)
+
+    def perform_action(self, action_list: dict):
+        if 'steps' in action_list and len(action_list['steps']) > 0:
+            self.action_manager.add_action({'closeout_action': action_list})
+            self.action_manager.execute_action('closeout_action')
+
+        # TODO any way to close the TCP connection to the devices we changed?
```
skbarber, line 18: ActionManager instantiates it's own set of GeecsDevice objects to interact with. I think I was trying to be "efficient" and so I was having it track the devices it instantiates in an effort to not always instantiate new objects each time it gets interacted with. But, it's true that these objects never close out. I was incorrectly under the impression that TCP subrcribers don't get opened until they are explicitly called for through subscribe_var_vals, but a TCP connection does get opened at initialization of a GeecsDevice. Not sure that is strictly necessary. But, probably the easiest thing to do is to just close the device after each time it gets used in ActionManager. This probably adds a very small amount of overhead that is insignificant in the context of executing real actions.

skbarber (approved):

GEECS-PythonAPI/geecs_python_api/analysis/scans/scan_data.py
```diff
@@ -139,17 +139,20 @@ def get_scan_tag(year: Union[int, str], month: Union[int, str], day: Union[int,
...
         """
         year = int(year)
         if 0 <= year <= 99:
             year += 2000
         month = month_to_int(month)
 
-        experiment = experiment_name or CONFIG.experiment
+        exp = experiment or experiment_name or CONFIG.experiment
+        if experiment_name is not None:
+            print("WARNING: Recommended to use 'experiment' instead of 'experiment_name' for 'get_scan_tag'...")
```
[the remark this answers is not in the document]
cedoss-lbl: Right, yeah we should definitely standardize something for this.  Probably not too difficult, just got to get on the same page and get comfortable with it.  For both the scanner and analyzer we'll want some way to print information to the gui at varying levels of verbosity.  For now though I'll change it to a `logging.warning()` statement.

GEECS-Scanner-GUI/action_control.py
skbarber: Actually, this comment triggered me to look into a bit more detail of the geecs_device. So, there shouldn't actaully be any tcp_subscribers opened at all by action_manager. Action manager just does basic set/get things via udp, which is simpler communication protocol. So, maybe the question goes back to you a bit, what is it that suggested you wanted to close tcp connections in action manager?

cedoss-lbl: Ah ok, if it is UDP and there are no lasting connections and/or it is simpler and easier to maintain then maybe there is no problem at all.  I guess I was looking at the logging statements while testing the "Execute action" buttons and was wondering if these connections do need to be explicitly closed at all.  But from what I could tell there wasn't any active threads when closing the gui so might be all good.

GEECS-PythonAPI/geecs_python_api/analysis/scans/scan_data.py
```diff
@@ -25,8 +25,11 @@
 if not logging.getLogger().hasHandlers():
     logging.basicConfig(level=logging.INFO, format="%(asctime)s - %(name)s - %(levelname)s - %(message)s")
 
-CONFIG = GeecsPathsConfig()
-
+try:
+    CONFIG = GeecsPathsConfig()
+except ValueError:
+    print("'CONFIG' not set for ScanData")
```
skbarber, line 31: logging() instead of print here?

GEECS-PythonAPI/geecs_python_api/controls/interface/geecs_database.py
```diff
@@ -100,6 +105,10 @@ def _close_db(db, db_cursor):
     @staticmethod
     def collect_exp_info(exp_name: str = 'Undulator')\
             -> dict[str, Union[ExpDict, dict[str, Path], Path, int]]:
+        print(GeecsDatabase.name, "current name")
+        if GeecsDatabase.name is None:
+            raise AttributeError("Geecs Database not set properly")
+
         db = GeecsDatabase._get_db()
         db_cursor = db.cursor(dictionary=True)
```
skbarber, line 114: looks like print is widely used in geecs_database, so i guess ok to be consistent here for now.

skbarber (approved):
lets add an 'issue' in github to review/standardize our logging/verbosity operations in these packages.

cedoss-lbl:
> lets add an 'issue' in github to review/standardize our logging/verbosity operations in these packages.

Yeah, this should be needed.  I've been going through trying to add this to all the print statements I could find in our code.  But I am running into a lot of issues with setting the verbosity and can't seem to understand why.  Some scripts it seems like the logger (through `logging` itself or as `logger = logging.getLogger(__name__)`) has defaulted to an INFO level, others it seems like it defaults to a WARNING level, and in one case just by using `logging.info("text")` has the effect of removing the INFO level elsewhere alltogether.

In short, yes standardizing logging and figuring out the intricacies with logging across multiple files will need to be a separate issue.

cedoss-lbl:
Made some additional improvements with being able to launch the GUI on a disconnected PC, and otherwise made some notes for the future works of (1) improving logging and (2) improving on-shot tdms writing.

Assuming this still works on a HTU PC, we can merge.

cedoss-lbl:
Works on an HTU PC, will merge into master.  Can finalize on-shot tdms writing and logging standards in a future PR

PR #104, opened 2024-12-18
establish whether or not a geecs device is off or on upon instantiation

skbarber:
Small improvement. Realized there didn't seem to be a good way to establish whether a device was off/on after instantiation of a GeecsDevice. Just made use of dev_tcp.connect() to see if a connection is possible. If yes, flip a boolean in the state dict, if not, throw an error and close the device. Closing the device prevents inadvertently getting stuck in long timeout loops trying to set/get parameters from a device.

### DIFF
--- a/GEECS-PythonAPI/geecs_python_api/controls/devices/geecs_device.py
+++ b/GEECS-PythonAPI/geecs_python_api/controls/devices/geecs_device.py
@@ -121,12 +121,17 @@ class GeecsDevice:
                     if self.dev_tcp is None:
                         try:
                             self.dev_tcp = TcpSubscriber(owner=self)
-                            ###try to connect a TCP subscriber to see if device is on and accessible
-                            is_device_on = self.dev_tcp.connect()
                         except Exception:
                             api_error.error('Failed to connect a TCP subscriber', 'GeecsDevice class, method "__init__"')
+                        
                 else:
                     api_error.warning(f'Device "{self.__dev_name}" not found', 'GeecsDevice class, method "__init__"')
+                
+                if self.dev_tcp:
+                    ###try to connect a TCP subscriber to see if device is on and accessible
+                    is_device_on = self.dev_tcp.connect()
+                else: 
+                    is_device_on = False
                 
                 if not is_device_on:
                     api_error.error(f'Failed establish test connection, {self.__dev_name} likely not on', 'GeecsDevice class, method "init_resources"')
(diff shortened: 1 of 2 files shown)


GEECS-PythonAPI/geecs_python_api/controls/devices/geecs_device.py
```diff
@@ -58,7 +58,7 @@ def __init__(self, name: str, virtual: bool = False):
         self.use_alias_in_TCP_subscription = True
 
         self.setpoints: dict[VarAlias, Any] = {}
-        self.state: dict[VarAlias, Any] = {'fresh': True,"shot number":None,"GEECS device error":False}
+        self.state: dict[VarAlias, Any] = {'fresh': False,"shot number":None,"GEECS device error":False, "Device alive on instantiation":False}
```
TAU-Reinier, line 61: For a container with a fixed set of named keys, consider a NamedTuple.

GEECS-PythonAPI/geecs_python_api/controls/devices/geecs_device.py
```diff
@@ -121,10 +121,20 @@ def init_resources(self):
                     if self.dev_tcp is None:
                         try:
                             self.dev_tcp = TcpSubscriber(owner=self)
+                            ###try to connect a TCP subscriber to see if device is on and accessible
+                            is_device_on = self.dev_tcp.connect()
                         except Exception:
```
TAU-Reinier, line 126: When using a try-except block, especially one that is used as "if this one command raises an error" , be specific about the Exception you're catching. You want all other Exceptions to fail loudly.

GEECS-PythonAPI/geecs_python_api/controls/devices/geecs_device.py
```diff
@@ -121,10 +121,20 @@ def init_resources(self):
...
-                            api_error.error('Failed creating TCP subscriber', 'GeecsDevice class, method "__init__"')
+                            api_error.error('Failed to connect a TCP subscriber', 'GeecsDevice class, method "__init__"')
                 else:
                     api_error.warning(f'Device "{self.__dev_name}" not found', 'GeecsDevice class, method "__init__"')
+                
+                if not is_device_on:
```
TAU-Reinier, line 131: `is_device_on` may not be defined, if `is_device_on = self.dev_tcp.connect()` raises an Exception.

TAU-Reinier (changes requested):
Besides one bug (is_device_on may not be defined), I added a few suggestions.

skbarber:
I think cleaned up the logic and clarified the potential bug. As for changing 'state' to a named tuple, I'm hesitant to do that as it is being used in many places and getting updated by various processes. Variables can be dynamically added to it as needed. But, maybe it would make sense to create a different attribute not called 'state' to track those other variables.

TAU-Reinier (approved):

PR #106, opened 2025-01-06
Breaking Change: Shot Control Setup (Geecsscanner dev 12192024)

cedoss-lbl:
This PR includes the long-awaited change of moving the hard-coded shot control features that control the trigger on/off into a customizable input.  Now, the user has access to a window to create/edit "shot control configurations" which specify a shot control device and any number of variables associated with that device.  These variables have three states:  OFF, SCAN, and STANDBY.

The reason for this being a breaking change is that I had to do two things:
1. Scan Manager now accepts different inputs, and will not work for HTU unless the new, updated timing configuration file (or dict) is passed into it.  (See `GEECS-Scanner-GUI/shot_control_configurations/Undulator/HTU-Normal.yaml`)
2. When loading the PC config file (the .ini), the old `shot_control_device` is now deprecated and doesn't do anything while the new `timing_configuration` is read to populate the initial shot control device.

I tested this before I left on Friday and it worked (both with the laser on and laser off).  In this PR there are three configurations for HTU as examples.

Known Issues:
- The experiment database dict lives in scan manager, and this is used to make the drop-downs for device and variable names.  The issue is that if the timing device is not specified, scan manager does not get initialized.  Need to make a way to get the database with just the experiment name, so that these completer drop downs are visible when setting up the timing devices.
- Should increment the version number by 0.1


cedoss-lbl:
I added a few more commits to the PR:
- Version 0.4
- Now pulls the database dictionary using the experiment name as a backup if timing device not set
- Some other misc HTU files for the dispersion studies

Tested this branch on Finn's PC and it works, I'll stop posting commits to this until we merge it.

PR #108, opened 2025-01-17
GEECS Scanner V0.5

cedoss-lbl:
Lot's of small quality of life changes, but a few of these reached into GEECS-Python-API.

- Refactored `data_aquisition` into `GEECS-Scanner-GUI` sub-repo
- Reorganized the sub-repo to separate out the modules and to better organize all of the config files
- Added a `main.py` on the top level to easily launch the GUI
- Started using relative package imports defined in `__init__.py`
- Renamed modules to be in snake_case rather than CamelCase
- GUI now doesn't crash when a new experiment is selected that doesn't have everything built up.  
- Once everything is configured from within the application the scanner will be functional.
- Added menu bar actions:  `Open` has actions to open various folders and the Github page, `Options` has a list of optional parameters one can define that are sent into scan_manager.  These options are very easy to define in `geecs_scanner.py`, and just need to be specified as either a `str` or `bool` and given a name.  They are then written and loaded from the user `config.ini` file
- Lots of fixed encountered when getting the GUI up and running for PW (Github Desktop was under Chetanya's name for those)
- PW Request:  Added optional parameter to turn saving off for a period of time to instead save 1 shot every x seconds
- PW Request:  A timing device is not required to run scans through scan manager
- Lots of small changes within `GEECS-Python-API` mostly to reload static variables in a couple of modules.  This is so that the experiment name can change and the database will get re-read.

The next steps after this PR are going to be (a) tackling the new GUI's for Action Library and Scan Variables, and (b) going through and cleaning up some of the code in `data_acquisition`.


skbarber:
It looks like files were moved out of the original controls/data_acquisition location but those folders still exist. I would delete the data_acquistion folder and it's subcontents since it all looks empty now

GEECS-Scanner-GUI/geecs_scanner/data_acquisition/scan_manager.py
```diff
@@ -652,19 +100,24 @@ def resume_scan(self):
...
             config_path (str, optional): Path to the configuration file.
             config_dictionary (dict, optional): Dictionary containing configuration settings.
+            scan_data (ScanData, optional):  If given, scan_data_manager will use an alternative scan folder
         """
 
         self.initial_state = None
         self.device_manager.reinitialize(config_path=config_path, config_dictionary=config_dictionary)
+        self.scan_data_manager = ScanDataManager(self.device_manager, scan_data, database_dict)
+
+        self.options_dict = config_dictionary['options']
```
skbarber, line 117: I don't think config_dictionary is a required input for reinitialize since it should be able to use the config_path. So, this throws an error that config_dictionary = None if config path is used. Also, not sure what the config dict entry should look like

GEECS-Scanner-GUI/geecs_scanner/data_acquisition/scan_manager.py
```diff
@@ -652,19 +100,24 @@ def resume_scan(self):
...
         self.data_logger.reinitialize_sound_player()
         self.data_logger.last_log_time_sync = {}
+        self.data_logger.update_repetition_rate(self.options_dict['rep_rate_hz'])
```
skbarber, line 120: same here, options_dict is not defined if using config_path rather than config_dictionary

GEECS-Scanner-GUI/geecs_scanner/data_acquisition/scan_manager.py
```diff
@@ -605,13 +48,19 @@ def __init__(self, experiment_dir: str, shot_control_information: dict, device_m
...
-        self.shot_control = GeecsDevice(shot_control_information['device'])
-        self.shot_control_variables = shot_control_information['variables']
+        self.shot_control: Optional[GeecsDevice] = None
+        self.shot_control_variables = None
+        shot_control_device = shot_control_information.get('device', None)
+        if shot_control_device:
+            self.shot_control = GeecsDevice(shot_control_information['device'])
+            self.shot_control_variables = shot_control_information['variables']
+            self.enable_live_ECS_dump(client_ip=self.MC_ip)
+
```
skbarber, line 63: This is another area where I think some of these things should be initialized not as "none" but to some default values. The experiment name can and should be used to determine things like shot control device etc. Feels too cumbersome to have to specify the expeirment name and them a full dict to describe the shot control device. 

I suspect some of these issues arise due to you focusing on the GUI dev and I am working mostly on backend code

GEECS-Scanner-GUI/geecs_scanner/app/lib/__init__.py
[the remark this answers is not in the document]
cedoss-lbl: This was patched in the `Refactored standalone modules into utils package` commit

GEECS-Scanner-GUI/geecs_scanner/data_acquisition/scan_manager.py
cedoss-lbl: `Config dictionary` contains everything, so all the save devices, actions, and options for the acquisition.  This is used to update data_logger's repetition rate, load all of the scan `options_dict` into a class variable, and initialize the device manager.

I can work on a fix for this, but this is what I will do:
1. if config_dictionary is None, options_dict = {} and rep rate = 1 (some default)
2. if the options_dict is empty, return default values when it is checked (by using `dict.get("key", default)` instead of `dict["key"]`)

cedoss-lbl:
I believe I solved the issue of running scans without the GUI.  Was able to test and successfully ran a scan using both the GUI and the new utility module of `default_scan_manager.py`.  Try it out and see if it works for you.

Assuming now we are happy, I'd like to wrap up this pull request by tomorrow.  Going to do some quick tests to make sure that GEECS-Python-API is still functional for analyzing scans and doing Xopt stuff.

.gitignore
```diff
@@ -22,3 +22,5 @@ GEECS-PythonAPI/htu_scripts/test_scripts/credentials.json
 LogMaker4GoogleDocs/logmaker_4_googledocs/HTUcurrentvalues.in
 LogMaker 4 GoogleDocs/source/credentials.json
 LogMaker 4 GoogleDocs/source/token.pickle
+GEECS-Scanner-GUI/testing.ipynb
+GEECS-Scanner-GUI/poetry.lock
```
TAU-Reinier: poetry [recommends ](https://python-poetry.org/docs/basic-usage/#committing-your-poetrylock-file-to-version-control) committing poetry.lock to the repo so that users who use `poetry install` get the same versions of all packages as the developer who worked - and presumably tested - with those packages.

skbarber: got it. But, it's also possible that the packages are not cross platform compatible, no? And then users ostensibly need to run 'poetry lock' to solve for their machine?

GEECS-PythonAPI/geecs_python_api/analysis/scans/scan_data.py
```diff
@@ -25,15 +25,22 @@
...
+        ScanData.paths_config = GeecsPathsConfig()
+    except ValueError:
+        logger.error("'CONFIG' not set for ScanData")
+        ScanData.paths_config = None
+
 
 class ScanData:
     """ Represents a GEECS experiment scan """
 
+    paths_config = None  # TODO the "reload" function is called statically at the bottom of this file, seems kinda bad
```
TAU-Reinier, line 41: The reload_paths_config() function could be a classmethod (not to be confused with staticmethod).
Also consider changing the behavior of GeecsPathsConfig(). It's very strange for an `__init__` method to throw a ValueError.

GEECS-Scanner-GUI/geecs_scanner/data_acquisition/action_manager.py
```diff
@@ -0,0 +1,168 @@
...
+        Args:
+            action (dict): A dictionary containing the action name and steps.
+
+        Raises:
+            ValueError: If the action dictionary does not contain exactly one action name.
+        """
+
+        # Parse out the action name and steps
+        if len(action) != 1:
+            raise ValueError("Action must contain exactly one action name")
```
TAU-Reinier, line 63: Why not just make this function take action_name and steps arguments?

.gitignore
TAU-Reinier: Good question. Yes, there are some packages where you might want different versions to run on Windows vs Mac vs Linux, and poetry does provide [a mechanism for this](https://stackoverflow.com/a/70440630). I believe that the same poetry.lock file is generated and distributed when a developer `poetry lock`s on any platform, and that `poetry install` picks the right version based on what platform a user is on - i.e. poetry.lock looks the same to everyone.

GEECS-Scanner-GUI/geecs_scanner/data_acquisition/scan_data_manager.py
```diff
@@ -0,0 +1,583 @@
...
+
+        self.scan_data = ScanData.build_next_scan_data()
+
+        for device_name in self.device_manager.non_scalar_saving_devices:
+            data_path = self.scan_data.get_folder() / device_name
+            data_path.mkdir(parents=True, exist_ok=True)
+
+            device = self.device_manager.devices.get(device_name)
+            if device:
+                save_path = str(data_path).replace('/', "\\")
```
skbarber, line 82: There seems to be an issue here with this bit of code when running from a mac out of CR network. Problem seems to originate from reload_paths_config(). Recall, the root path for 'build_next_scan_data()' might be different for the root path when seeting the data path for devices.

GEECS-PythonAPI/geecs_python_api/analysis/scans/scan_data.py
cedoss-lbl: Oooh interesting!  I've never seen a classmethod before, but yeah seems like the perfect use case for it.

And yeah I'll change the ValueError to something better, I remember back when I was browsing the python errors this seemed like the best one at the time.  I think what I'll do is make a new exception class and use that.

GEECS-Scanner-GUI/geecs_scanner/data_acquisition/action_manager.py
cedoss-lbl: Normally I'd be shy about making organizational changes, but this one isn't too embedded in the code.  Everywhere `add_action` is called is used as `add_action({name:steps}), so this one would be easy to do.  But yeah, in general I do support breaking up large dicts for clarity

GEECS-Scanner-GUI/geecs_scanner/data_acquisition/scan_data_manager.py
cedoss-lbl: Hmm, difficult to debug this on my own machine.  But I have an idea for you to consider if you want to find a solution:

The paths are loaded into ScanData using `reload_paths_config()`, which does not take any arguments.  However, the class `GEECSPathsConfig` does allow for keyword arguments to specify the base path and experiment name to use.  So what I'll do is I'll allow for `reload_paths_config()` to take optional arguments to pass into `GEECSPathsConfig`, and then there might be a way in `data_acquisition` to override the default server address using this.

GEECS-PythonAPI/geecs_python_api/analysis/scans/scan_data.py
TAU-Reinier: Oh I meant I don't think an `__init__` function should throw an error at all, except maybe validation of constructor arguments. Especially in this case, where GeecsPathsConfig is dataclass-like. 

There are a few options: 
* Add a load() method, to be called by the caller, who can catch errors.
* Replace raising ValueError by setting things to None, and then check for that in the caller. 
* Add a staticmethod load() that returns a GeecsPathsConfig. 

And finally, after thinking about it a bit more and reading a bit, it's probably ok just to leave it as is - it's the simplest and cleanest, and it isn't bad form like I thought before.

GEECS-Scanner-GUI/geecs_scanner/data_acquisition/action_manager.py
```diff
@@ -46,27 +46,24 @@ def load_actions(self):
...
         """
-
-        # Parse out the action name and steps
-        if len(action) != 1:
-            raise ValueError("Action must contain exactly one action name")
-
-        action_name = list(action.keys())[0]
-        steps = action[action_name]['steps']
+        # Check that the steps are formatted correctly
+        if 'steps' not in action_steps or len(action_steps) != 1:
```
TAU-Reinier, line 62: Why require `{'steps': steps}` as an argument instead of just `steps`? If your argument is a dict of required length 1, you likely have better options, such as tuple, or splitting into two args/kwargs.

GEECS-PythonAPI/geecs_python_api/analysis/scans/scan_data.py
cedoss-lbl: Ok, I think I see what you're saying.  I suppose if `paths_config` were always a GEECSPathsConfig class instance then it moves all of the error handling when not properly initialized to the class functions of GEECSPathsConfig rather than in ScanData or anywhere else using it.  And this would probably be cleaner if more modules were using this class.

I'll keep this in mind, the refactoring would not be too expansive I think.  But I'll save this for another day.

GEECS-Scanner-GUI/geecs_scanner/data_acquisition/action_manager.py
cedoss-lbl: tuple would make the most sense, but this will take some refactoring of a good bit of code wherever actions are used (@skbarber)

I'll add this to the To-Do list, but will leave as-is for the current PR

PR #109, opened 2025-01-29
Scan device development

skbarber:
Adding a new feature to simplify some device operations.

support for composite devices is currently shared across a couple of packages with the core functionality implemented ScanManager. In ScanManger, a composite variable is used to define execution steps in a scan which can include varying multiple components.

This new class is aimed at abstracting the composite device/variable operations at an earlier stage. It is a subclass of GeecsDevice. It can be instantiated identically as a GeecsDevice and behaves the same. It can also take in a dict describing a composite variable and create an object that behaves very similarly to a standard geecs device. The set/get methods are overwritten. 

The idea is, a user should just use this new class, ScanDevice, for all standard device operations. This should make it easier and more efficient to set up scans of composite variables. In principle, the operation should be the same for composite or standard device variables. 

This PR just creates the new class, which has been tested for both standard GeecsDevice operations, as well as get-type and set-type composite ScanDevices (which are inherently different)

Next steps will involve refactoring other key packages like DeviceManager, ScanManager and others to make use of this new simplified class.

Note, I merged in the existing PR for GeecsScanner v0.5 to test with that organization.


GEECS-PythonAPI/geecs_python_api/controls/devices/scan_device.py
```diff
@@ -0,0 +1,316 @@
...
+            # Standard GeecsDevice initialization
+            super().__init__(name_or_dict, virtual)
+            if self.is_valid():
+                self.is_composite = False
+                self.components = []
+            else:
+                raise ValueError(f"Device '{name_or_dict}' is not a valid GeecsDevice.")
+        
+        elif isinstance(name_or_dict, dict):
+            if len(name_or_dict) != 1:
```
TAU-Reinier, line 88: If you're requiring a dict of length exactly 1, you may have better options. For example, `name_or_tuple: str | tuple[str, dict]`. 
Even better is to use a `name` arg and an optional `composite_device_spec` kwarg.

GEECS-PythonAPI/geecs_python_api/controls/devices/scan_device.py
```diff
@@ -0,0 +1,316 @@
...
+            value (Any): The input composite variable value.
+
+        Returns:
+            Any: Computed sub-device value.
+        """
+        reference_value = (
+            self.reference.get(comp['device'], {}).get(comp['variable'], 0)
+            if self.mode == "relative" else 0
+        )
+        return reference_value + eval(comp['relation'], {'composite_var': value})
```
TAU-Reinier, line 213: Consider some sanitation of comp['relation']. Presumably this is read from a file and could contain badly formatted or malicious code.

GEECS-PythonAPI/geecs_python_api/controls/devices/scan_device.py
```diff
@@ -0,0 +1,158 @@
...
+from typing import Optional, Any, Dict, Union
+
+import time
+from functools import wraps
+from geecs_python_api.controls.interface.geecs_database import GeecsDatabase, load_config
+from geecs_python_api.controls.devices.geecs_device import GeecsDevice
+
+# Load experiment info globally
+expt = load_config().get('Experiment', 'expt')
+GeecsDevice.exp_info = GeecsDatabase.collect_exp_info(expt)
```
cedoss-lbl, line 11: This should happen in GeecsDevice instead of here

GEECS-PythonAPI/geecs_python_api/controls/devices/scan_device.py
```diff
@@ -0,0 +1,158 @@
...
+            # Dummy name for composite initialization
+            super().__init__('virtual_composite_device', virtual=True)
+
+            self.name, composite_info = next(iter(name_or_dict.items()))
+            self.is_composite = True
+            self.components = composite_info['components']
+            self.mode = composite_info['mode']
+            self.relation = composite_info.get('relation', None)
+            self.reference = {} if self.mode == "relative" else None
+            self.sub_devices = {comp['device']: GeecsDevice(comp['device']) for comp in self.components}
```
cedoss-lbl: Since most of these class variables are only used for composite variables, I think it would make more sense to have a different class that is a subclass of ScanDevice.  This would get rid of the need to have `if is_composite` checks everywhere, as if the ScanDevice is initialized as a CompositeDevice you would have all of these functions overwritten

GEECS-PythonAPI/geecs_python_api/controls/devices/scan_device.py
cedoss-lbl: Would this be some regular expression to check if comp['relation'] looks like a math equation or only contains numbers, operators, numpy functions, and `composite_var`?

GEECS-PythonAPI/geecs_python_api/controls/devices/scan_device.py
```diff
@@ -61,98 +97,167 @@ def __init__(self, name_or_dict: Union[str, Dict[str, Any]], virtual: bool = Fal
...
+                        'instance': GeecsDevice(device_name),
+                        'variables': []
+                    }
+                self.sub_devices[device_name]['variables'].append(comp['variable'])
 
             # Subscribe to necessary variables for caching in `state`
             self._subscribe_to_variables()
 
             if self.mode == "relative":
+                time.sleep(1)
```
cedoss-lbl, line 116: (assuming this is the sleep you were talking about). Any way to avoid sleeping for a set amount of time?  Just in case either 1 second is too long and we're just waiting around or even the other way where 1 second is too quick and it crashes anyways?

Like can we get a loop that checks for completion before moving on?  Or is this impossible?

cedoss-lbl (commented):
My main comment is I am not a fan of how Scan Device can either be a regular or composite device, and there are large `if is_composite` blocks that completely change how functions operate.  Would be cleaner I think if there was a CompositeDevice subclass of ScanDevice that instead overwrites the necessary functions.

skbarber:
> My main comment is I am not a fan of how Scan Device can either be a regular or composite device, and there are large `if is_composite` blocks that completely change how functions operate. Would be cleaner I think if there was a CompositeDevice subclass of ScanDevice that instead overwrites the necessary functions.

My thinking was to have a single class to represent all "geecs" type devices to streamline its use in other areas. Otherwise I need to have a GeecsDevicea and CompositeDevice. But, it's true that you have to do something to distinguish how they operate. 

For a normal geecs device, the ScanDevice doesn't actually do anything (except change the close method to address a race condition in the core GeecsDevice code). For a composite type device, only the set/get methods are overwritten.

GEECS-PythonAPI/geecs_python_api/controls/devices/scan_device.py
skbarber: gpt gave the same recommendation...will do.

GEECS-PythonAPI/geecs_python_api/controls/devices/scan_device.py
skbarber: In general, I think there is probably an overall better way to achieve the result here. This pulls the entire database for geecs devices into a dict. It would probably make a bit more sense to have a class (or something) to just query the DB as needed rather than load the whole DB. But, this is something of a baked in requirement.

GEECS-PythonAPI/geecs_python_api/controls/devices/scan_device.py
skbarber: When you subscribe to a device, there are no entries in the variables dict until a TCP message is sent out. TCP updates across all BELLA experiments are typically ~5Hz. So you need to wait a minimum of 200 ms before values will actually be updated. But, 1 second should be sufficient to get all variables from any number of devices.

You could set up an event handler for the TCP connection to get notified when each variable is first updated...But that just felt like too much effort in this case...

You could also get the values through a UDP get command, which would be a bit more deterministic, e.g. you request the value, you get the value then go on to the next device. But, for anything more than ~2 devices, this process can easily start to take longer than 1 second. 

So, agree, not great practice, but a faster more deterministic approach will require substantially more effort

cedoss-lbl:
> My thinking was to have a single class to represent all "geecs" type devices to streamline its use in other areas. Otherwise I need to have a GeecsDevicea and CompositeDevice. But, it's true that you have to do something to distinguish how they operate.
> 
> For a normal geecs device, the ScanDevice doesn't actually do anything (except change the close method to address a race condition in the core GeecsDevice code). For a composite type device, only the set/get methods are overwritten.

It wouldn't take much more code, just would need to know if a given device is initialized as one or the other (which you already do in the __init__ of `ScanDevice`).  If set up correctly, `CompositeDevice` would always be considered as a `ScanDevice` and anything that requires a `ScanDevice` or any functionality of a `ScanDevice` can seamlessly have a `CompositeDevice` there instead.  Still would be as you suggest: `ScanDevice` represents all "Geecs" type devices.  Would just be that `CompositeDevice` is a special case of `ScanDevice`

GEECS-PythonAPI/geecs_python_api/controls/devices/scan_device.py
skbarber: yeah, makes sense. Adopted the latter suggestion

cedoss-lbl:
Ok, I could not make a pull request of a branch off of this branch, so I'll just bring it up here and you are welcome to merge that single commit into this branch if you like:

In branch `scan_device_dev_refactor_02052025` I edited just the `scan_device.py` file.  
- ScanDevice almost completely inherits GEECSDevice, but has those few important changes for closing the device and having the get method return in a format used for data_acquisition.
- CompositeDevice is a subclass of ScanDevice, so it will be considered as a ScanDevice for anything that uses ScanDevice previously.  The set, get, close, and init methods are just overridden in CompositeDevice
- Now, if you want to initialize a device, use the `initialize_scan_device` function and that will determine whether or not to make a ScanDevice or CompositeDevice
- To check if a ScanDevice is composite or not, you can use the function `is_composite()`.

Feel free to use this or not, and there could be a better implementation of this.  Hopefully you find it useful nonetheless

skbarber:
> Ok, I could not make a pull request of a branch off of this branch, so I'll just bring it up here and you are welcome to merge that single commit into this branch if you like:
> 
> In branch `scan_device_dev_refactor_02052025` I edited just the `scan_device.py` file.
> 
> * ScanDevice almost completely inherits GEECSDevice, but has those few important changes for closing the device and having the get method return in a format used for data_acquisition.
> * CompositeDevice is a subclass of ScanDevice, so it will be considered as a ScanDevice for anything that uses ScanDevice previously.  The set, get, close, and init methods are just overridden in CompositeDevice
> * Now, if you want to initialize a device, use the `initialize_scan_device` function and that will determine whether or not to make a ScanDevice or CompositeDevice
> * To check if a ScanDevice is composite or not, you can use the function `is_composite()`.
> 
> Feel free to use this or not, and there could be a better implementation of this. Hopefully you find it useful nonetheless

I don't have a strong feeling either way on this. I see your point about organizing it this way. For now, this isn't affecting any functionality. Given there is a finalized refactored branch using the original single ScanDevice approach I think I will go ahead and merge and we can consider refactoring again in the future

PR #111, opened 2025-02-05
Modify camera_image_analysis and map_Undulator

kjensen11:
Modity camera_image_analysis to handle the case of no image settings or configs with default settings. Also modified save path definition for image array figure so that the save path could conveniently be appended to list to import to google docs.

Modify map_Undulator to include camera analysis for UC_ModeImager.


cedoss-lbl (approved):
Looks all good to me!

PR #114, opened 2025-02-12
Htu hama spectro analysis

kjensen11:
Create analysis for fiber-based spectrometer. Bare-bones at the moment. Imports data from scan (assuming hamaspectro file format; could be generalized). Bifurcated no scan and scan analyses that currently just plot appropriately for both.

Generally tested on select scans. The only thing that may not work right now is compatibility with the Google Docs auto importer. Nevertheless, this is isolated so merging should not conflict with anything else.

Note: As more analyses are being created, refactoring to minimize duplicate code and increase standardization is becoming more relevant. Much of this would likely live in ScanAnalysis base class. This should be the focus of a near-future pull request.


ScanAnalysis/scan_analysis/mapping/map_Undulator.py
```diff
@@ -36,4 +36,7 @@
          device_name='UC_ModeImager'),
     Info(analyzer_class=HiResMagCamAnalysis,
          requirements={'AND': ['UC_HiResMagCam', 'U_BCaveICT']}),
+    Info(analyzer_class=FiberSpectrometerAnalysis,
```
TAU-Reinier, line 39: FiberSpectrometerAnalysis isn't imported. 

I would urge taking this opportunity to make the imports of all the analyzer classes explicit, that is: 
* Don't use wildcard imports (see https://peps.python.org/pep-0008/#:~:text=Wildcard%20imports)
* When importing from within the same package, import from the original location, not from the "shortcut" in 
  `Undulator.__init__` - leave those for end-users. 

So: 
`from scan_analysis.Undulator.camera_image_analysis import CameraImageAnalysis`
(not `import *`, not `from scan_analysis.Undulator`)

ScanAnalysis/scan_analysis/analyzers/Undulator/hamaspectro-analysis.py
```diff
@@ -0,0 +1,290 @@
...
+
+        # initialize storage
+        wavelength = None
+        spectra = [[]] * len(shot_numbers)
+
+        # iterate shot numbers
+        for ind, shot_num in enumerate(shot_numbers):
+            try:
+                # get file for shot number
+                file = next(self.path_dict['data'].glob(f"*_{shot_num:03d}.txt"), None)
```
TAU-Reinier, line 113: ScanData has a method for this: get_device_shot_path

ScanAnalysis/scan_analysis/mapping/map_Undulator.py
cedoss-lbl: Makes sense, I agree

kjensen11: Agreed, that makes sense. Fixed. Also was a good comment because I realized the Fiber analysis wasn't imported to map_Undulator.py. I suppose that a prime example of why not to do this.

ScanAnalysis/scan_analysis/analyzers/Undulator/hamaspectro-analysis.py
```diff
@@ -0,0 +1,290 @@
...
+
+        # plot binned waterfall plot
+        self.plot_waterfall_binned(self.wavelength, self.spectra_binned, save=self.flags['save_image'])
+
+    def run_noscan_analysis(self):
+
+        # plot waterfall
+        self.plot_waterfall_noscan(self.wavelength, self.spectra, save=self.flags['save_image'])
+
+    def load_spectrometer_data(self) -> [np.ndarray, np.ndarray]:
```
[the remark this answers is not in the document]
kjensen11: done

ScanAnalysis/scan_analysis/analyzers/Undulator/hamaspectro-analysis.py
kjensen11: Got it. Fixed.

ScanAnalysis/scan_analysis/analyzers/Undulator/hamaspectro-analysis.py
```diff
@@ -0,0 +1,290 @@
...
+                          'save': (self.scan_directory.parents[1] / 'analysis'
+                                   / self.scan_directory.name / f"{device_name}" / "FiberSpectrometerAnalysis")
+                          }
+
+        # initialize future variables
+        self.wavelength = None
+        self.spectra = None
+        self.spectra_binned = None
+
+    def run_analysis(self) -> None:
```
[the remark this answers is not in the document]
kjensen11: Fixed.

TAU-Reinier (approved):

PR #115, opened 2025-02-14
Moved 'processed_scans' yaml to analysis folder.

kjensen11:
Moved processed scans yaml to analysis folder. Simple as that.


kjensen11:
Just going to merge because its one line of code just changing a file path and it is tested.

cedoss-lbl (commented):
Nice!

PR #116, opened 2025-02-14
Improve device not on error handling in scan manager

skbarber:
Running lots of various data acquisitions scans revealed the vulnerability of geecs_scanner when devices are not behaving. In particular, if a device is off or not responding, the scanner goes into hard crash mode (need to force quit), with ambiguous error codes. 

This PR adds in some error tracking and catching during the process of device instantiation. The errors are propagated up and handled by device manager, scan_manager and run_control. Now, if either a save device or a scan variable device fails to instantiate properly, processes are terminated cleanly with errors clearly logging the issue. 

to do: could be nice to add a pop up that notifies the user which devices failed and ask if the data acquisition should continue without them. 

Tested with save and scan devices off and the code exists cleanly, and is read for another scan. Also tested cases of non existent devices, which also exited cleanly.


GEECS-Scanner-GUI/geecs_scanner/data_acquisition/scan_manager.py
```diff
@@ -131,6 +141,8 @@ def reinitialize(self, config_path=None, config_dictionary=None, scan_data=None)
         self.console_logger.stop_logging()
         self.console_logger.setup_logging()
 
+        return True
```
cedoss-lbl, line 144: Better yet, could assign the boolean True/False to a class variable, and then check this class variable in the `start_scan_thread` method.  This could move a lot of the `if success:` logic out of the outer parts of code and back to the scan_manager.

cedoss-lbl (approved):
Apart from making `reinitialize` assign a class variable instead of returning a bool, looks good to me!

cedoss-lbl:
Added these changes, along with a error pop-up on the main gui window.  Currently it is just a bool so the pop-up doesn't display problem devices, but the user could check the log.

TAU-Reinier (approved):

Issue #105, opened 2024-12-19
Standardize logging and print statements across packages


cedoss-lbl:
For the GEECS Scanner and Scan Analysis GUI's, having a standardize way to print information and set the verbosity level (ie, all messages, or show only warnings and errors) would be great for printing out information.  To do this we need to replace `print` statements with something more standardized (such as the `logging` package), but this is easier said than done.

PR #117, opened 2025-02-20
Api error refactor to use logging in geecs device rather than print statements

skbarber:
Only slight change to use python logging rather than print statements for error messaging in geecsdevice

Using logging, the error messages from geecs device will now propagate up to the scan execution logs etc. for improved debugging


TAU-Reinier (approved):
Good idea to use the logging facility rather than print statements. 

One thing to consider is that - because error handling is centralized - the logger name is always going to be controls.interface.geecs_errors. i.e. not really a "module level logger". (Normally, when using `getLogger(__name__)` each module gets its own logger). 

One way would be to call getLogger within the error_handler function, but with what name? The `src` unfortunately doesn't appear to be a good identifier. You might have to do something horrible like ascertain the module from where the .error() or .warning() call came. 

So just be aware that only one logger is used throughout geecs_python_api, which only means that you won't be able to configure loggers at the module level (not a big deal) and that you'll rely on the `src` field in api_error.error|warning calls to tell you where the error came from.

skbarber:
I think this discussion is a bit over my head here...

But, I am generally aware that our logging etc. is a bit of a mess...We created an issue here: https://github.com/GEECS-BELLA/GEECS-Plugins/issues/105. Not entirely sure the best way to handle this though

PR #118, opened 2025-02-20
HTU frog analysis

kjensen11:
Implemented base FROG analysis which calculates temporal second moment and peak value on a single shot basis. Then, the results are appended to the sfile (analysis folder). Mapped analysis.

Modified ScanAnalysis to have 'auxiliary_data' imported from the sfile in the analysis folder rather than the ScanData file in the scan folder. 

Future work:
The FROG analysis currently does the same thing for a scan and no scan for quick implementation but framework exists for both. Binning and unique plotting should be implemented for the two cases.

General analysis framework and common functions should be moved from analysis classes (frog analysis, camera image analysis, etc) to ScanAnalysis parent class and inherited. Clean up can be scope of next pull request when sufficient time is available.


TAU-Reinier (approved):
Looks good. Though some things to consider here, but also more generally for all the other ScanAnalyzers: 

* run_analysis is nearly identical in many ScanAnalyzers - it does a few checks and then dispatches to scan or no_scan. This could be implemented in the base ScanAnalyzer. The way flags are handled would need to be standardized. 
* Methods like load_images, average_images, and append_to_sfile, aren't specific to FrogAnalysis. They should probably be in the base class. 
* Job separation between ImageAnalysis and ScanAnalysis is getting muddied. ImageAnalysis contains analyzers for single images, and ScanAnalysis was meant for analysis that needs auxiliary data or analyzes a sequence of images. I don't think ImageAnalysis is obsolete, so consider keeping all single_shot_analysis in ImageAnalysis. 

One more general tip: please make the topline commit message a bit more specific.

kjensen11:
Good comments to keep in mind Reinier. Based on any comments from the other two, I will likely merge and then try to tackle some of those bullet points in a follow-up. Agree with them all.

cedoss-lbl:
Just adding my two cents:

> Looks good. Though some things to consider here, but also more generally for all the other ScanAnalyzers:
> 
> * run_analysis is nearly identical in many ScanAnalyzers - it does a few checks and then dispatches to scan or no_scan. This could be implemented in the base ScanAnalyzer. The way flags are handled would need to be standardized.

This is true, though I anticipate that some analyzers will have the same "scan" and "noscan" behavior.  (Such as laser cameras that are independent of what's going on in the experiment).  Then, in that case you would need to specify that running a "scan" analysis just runs the "nonscan" analysis.  However, we could also add into the `base` a default `run_scan` and `run_noscan` methods that could get overwritten by children classes.  Makes the abstract base class a little more complex, but gives more flexibility.

> * Methods like load_images, average_images, and append_to_sfile, aren't specific to FrogAnalysis. They should probably be in the base class.

I agree.  A standardized library of common functions would make this package more robust.  Would be great in either the base class or in a separate module accessible by all analyzers.

> * Job separation between ImageAnalysis and ScanAnalysis is getting muddied. ImageAnalysis contains analyzers for single images, and ScanAnalysis was meant for analysis that needs auxiliary data or analyzes a sequence of images. I don't think ImageAnalysis is obsolete, so consider keeping all single_shot_analysis in ImageAnalysis.

I definitely agree with this in principle, but in practice it is complicated for a few reasons: 
1. ImageAnalysis being compatible with LabView's basic Python 3.9 version
2. ImageAnalyzers having to return data in a specific format
3. Organization of ImageAnalyzers is not as robust as our current standard (separated by experiment)

May need to brainstorm what can be done to help make ImageAnalysis as useful to ScanAnalysis as it was to the initial LabView-based analysis.  My first thoughts on this would be to make a designated folder/package in ImageAnalysis for analyzers that are not for use with LabView, and organize this by Experiment name much as we've done in ScanAnalysis and GEECSScanner.

> One more general tip: please make the topline commit message a bit more specific.

skbarber:
This is a good conversation. I've been thinking about these things too. I think ScanAnalysis is mostly aimed at creating visualizations from existing data with minimal additional processing, e.g. making gifs, arrays of binned images, waterfall plots, xy plots of some scalars vs scan parameter etc. If there needs to be some true 'analysis' in which data from single device is getting reduced (e.g. integrating a scope trace, calculating moments of distribution) or processed (e.g. turn raw haso data into wavefront data, frog raw image to retrieved image) that should happen in ImageAnalysis and get called from ScanAnalysis if necessary. 

Chris is right about the restrictions on labview/python compatibility (which actually requires python 3.7). I see that the image analysis project .toml file actually allows up to 3.11, so I kind of forget how this all works. 

Anyways, as Reinier mentioned, things are getting a bit muddied and it's probably nearing time to review a few of these things more holistically and come up with a way to clean it up. At the same time, as long as we have the core functionality and can take good data, that supersedes spending too much effort on organization.

ScanAnalysis/scan_analysis/analyzers/Undulator/frog_analysis.py
```diff
@@ -0,0 +1,327 @@
...
+        try:
+            # copy auxiliary dataframe
+            df_copy = self.auxiliary_data.copy()
+    
+            # check column lengths match existing dataframe
+            lengths = {len(vals) for vals in dict_to_append.values() if isinstance(vals, (list, np.ndarray))}
+            if lengths and lengths.pop() != len(df_copy):
+                if self.flag['logging']:
+                    raise DataLengthError()
+
```
skbarber, line 260: I don't see why the column lengths need to match. In fact, there is a good chance they won't match most of the time if just a single image is missed. Since you are working with dataframes and you have the shotnumber associated with each image, it should be very easy to add the entries based off of that information. This would be much more robust.

PR #119, opened 2025-02-21
ScAnalyzer GUI execute and reorganization

kjensen11:
Reorganized the file structure of the ScAnalyzer GUI to more closely match that of the GEECS_Scanner GUI. Included a bash script executable to aid in operating from both C and Z drives.


ScanAnalysis/live_watch/scan_analysis_gui/ScAnalyzer.sh
```diff
@@ -0,0 +1,9 @@
+#!/bin/bash
+
+BASE_PATH='C:/GEECS/Developers Version/source/GEECS-Plugins/ScanAnalysis'
+#BASE_PATH='Z:/software/control-all-loasis/HTU/Active Version/GEECS-Plugins/ScanAnalysis'
+
+cd "$BASE_PATH"
+poetry install
```
cedoss-lbl, line 7: Just thinking aloud, we should see if there is a way to check if `poetry install` or `poetry lock --no-update` needs to be called every time.  Maybe there is a way to script a check to see if the current environment is up-to-date.

cedoss-lbl (approved):
At first I was going to suggest renaming `ScAnalyzer.py` to something not CamelCase, but the class within the module is `ScAnalyzer.ScAnalyzerWindow` which is maybe enough different so that the module name is not confused with the class name.  Also, `sc_analyzer.py`, `scanalyzer.py`, or `sc-analyzer.py` doesn't look as good.

ScanAnalysis/live_watch/scan_analysis_gui/ScAnalyzer.sh
kjensen11: Agree with this. Can look into it.

kjensen11: Also, I will be sad, but we can change the name of ScAnalyzer if we need to...

skbarber:
GUI stuff is a bit beyond me. Everything else looks fine

ScanAnalysis/live_watch/scan_analysis_gui/ScAnalyzer.sh
kjensen11: In commit [b317ad3], I modified executable script to check for an updated poetry lock file, only running poetry install if a lock file is found and poetry environment not up-to-date.

PR #120, opened 2025-02-21
V0.6 GUI update

cedoss-lbl:
Lots of changes here, including some to the backend `data_acquisition` package.  I'll list the main changes here, and in the next week I'm going to do a couple of minor cleanups and documentation edits.

Changes:
- Editor gui for scan variables and composite variables
- Action library gui to edit actions and have quick access to executing saved actions
- Added button to delete save elements on main window
- Added a line that displays what scan number is currently scanning/was previously scanned
- Added a top toolbar Open option to open the current daily scanlog on GDocs
- Moved `composite_variables.yaml` to the "scan_devices" folder
- Moved `actions.yaml` to a new "action_library" folder
- Added a `Background` scan mode that is functionally the same as a NoScan but writes a background flag to the scan config file
- Fixed the text that displays if a composite variable is relative or absolute on the main window
- Added an icon for the windows (but could not get the icon on the taskbar.  See note * below)
- Added preference option for Dark Mode
- Added preference option for randomized beeps during scan
- Added checkbox to set flag if all subscribed variables should be included for a device (**Note**, but not yet implemented in the backend)
- Added a jingle that plays upon successful completion of an action started from the gui (and another upon failure)
- Added error handling to actions.  Gets raised when an "execute" action has a wrong name or a "get" action does not return the expected value.  In the latter case, a pop-up is displayed to ask the user if it is safe to proceed or not.
- Improved dictionary combining before sending a scan.  Now it is possible for two different elements to contain the same device.  The variable list gets merged without any duplicates, and an error is thrown if the boolean flags have conflicting values.

Small stuff I'll do before merging this PR:
- Add documentation where missing
- I refactored some common functions to `lib` for the newer gui's, need to use them with the older gui's

Large stuff that I'll push to next version:
- Improved logging and displaying messages in terminal
- Unit tests ( 😅 )
- Implementing the option "subsribe to all variables"

*Note on icon:
Since we are always running GEECS Scanner from a python terminal, it will always display the default python icon (at least on Windows).  To get a custom icon visible in the taskbar, we'll need to properly build this in an exe.  I tried doing this with a package called `pyinstaller` and it was initially successful.  It at least ran and made an exe without breaking, and was compatible with the poetry toml.  However, the exe did not work.  Was giving an error about not being able to find a specific package.  It seems everything needs to be perfect for it to compile everything all at once, so this will need some work.


GEECS-PythonAPI/geecs_python_api/controls/interface/geecs_paths_config.py
```diff
@@ -94,12 +94,12 @@ def __init__(self,
...
 
-        self.base_path = base_path.resolve()
+        self.base_path = base_path  # .resolve()
         self.experiment = experiment
 
     def is_default_server_address(self) -> bool:
         """ Returns True if the base directory is equivalent to the default server address """
-        default_path = self.get_default_server_address(self.experiment).resolve()
+        default_path = self.get_default_server_address(self.experiment)  # .resolve()
         return self.base_path == default_path
```
skbarber, line 104: not sure the ramifications of removing the .resolve()...I'll have to test on Mac. But that's not something we need to wait on.

GEECS-Scanner-GUI/scanner_configs/experiments/Undulator/scan_devices/composite_variables.yaml
```diff
@@ -0,0 +1,64 @@
...
+    mode: relative
+  R56_at_100MeV:
+    components:
+    - device: U_ChicaneInner
+      relation: np.sqrt(100 ** 2 * composite_var / 560968.636)
+      variable: Current
+    - device: U_ChicaneOuter
+      relation: -np.sqrt(100 ** 2 * composite_var / 560968.636)
+      variable: Current
+    mode: absolute
```
skbarber, line 64: numpy is not supported anymore in the definitions of relations. Using numexpr which is evidently a safer alternative than straight eval()

skbarber (approved):
I mostly just looked at the core backend stuff, which doesn't seem like it changed all that much. Might be something with the geecs_path config that needs to be tested on mac, but that's not urgent.

GEECS-Scanner-GUI/geecs_scanner/app/lib/action_control.py
```diff
@@ -3,7 +3,12 @@
 
 -Chris
 """
+from typing import Union
 from geecs_scanner.data_acquisition import ActionManager
+import logging
+
+from ...utils.exceptions import ActionError
+from ...utils.sound_player import action_finish_jingle, action_faled_jingle
```
TAU-Reinier, line 13: Consider clarifying the action variable names. 
* in get_new_action(action), action is a string, naming an action
* in generate_action_description(action), action is a dict of action attributes (hint hint -> consider namedtuples or dataclasses?)
* in perform_action(action_list), action_list is in fact a dict, possibly the same dict as action in generate_action_description()?

GEECS-Scanner-GUI/geecs_scanner/app/action_library.py
```diff
@@ -0,0 +1,510 @@
...
+            self.update_action_list(index=index)
+
+    def get_selected_step_index(self) -> int:
+        """Finds the location of the currently-selected action in the visible list on the GUI.
+
+        :return: If no selection, returns -1.  Otherwise, return the index of the currently selected step
+        """
+        selected_action = self.ui.listActionSteps.selectedItems()
+        if not selected_action:
+            return -1
```
TAU-Reinier, line 322: Be careful with this. I think you mean to return an invalid index, but in fact steps[-1] will fetch something that I don't think you want.

GEECS-Scanner-GUI/geecs_scanner/app/action_library.py
```diff
@@ -0,0 +1,510 @@
...
+        for action in selected_action:
+            index = self.ui.listActionSteps.row(action)
+            return index
+
```
TAU-Reinier, line 326: index = self.ui.listActionSteps.row(selected_action[0])
return index
```suggestion
        for action in selected_action:
            index = self.ui.listActionSteps.row(action)
            return index

```

TAU-Reinier (approved):

GEECS-Scanner-GUI/geecs_scanner/app/lib/action_control.py
cedoss-lbl: Went about trying to tackle this today, and got pretty far in a separate branch (`actionmanager-dev-02252025`).  It's tricky to fully abandon the dict format because that is what gets saved in yaml files, but by the time any actions reach `ActionManager` I can have a list of these new `ActionStep` named tuples.  It feels much cleaner, but I'd like to fully implement them in the GUI's that edit actions before I merge to this PR or master.

GEECS-Scanner-GUI/geecs_scanner/app/action_library.py
cedoss-lbl: You're right, although I believe I do check that the index is >0 after every call to this function, I shouldn't rely on an external check of if the function did what it's supposed to.  Will change to returning `None`

cedoss-lbl:
Finished documentation and small refactoring efforts.  We also used this branch yesterday so we included a lot of new yaml files from experiment.  Couple of bug fixes related to refreshing ActionManager and allowing for 1D scans in the negative direction.

Began some stress testing (clicking on buttons when experiment name is `Thomson` or when it is undefined) and unfortunately things are breaking again at the GEECS-Python-API level.  This is all because there are a couple of files in GEECS-Python-API that attempt to load and use the .ini config file at the top of the file.  I previous worked around this by using `importlib` to import my RunControl that interfaced with anything GEECS-Python-API-related, and this was able to skip full initialization until experiment name is correctly defined.  However, now there are a few more `import` routes between GEECS Scanner and GEECS-Python-API (through ActionManager and through DatabaseDictLookup).

At this point, I think I want to try and tackle some GEECS-Python-API refactoring for the v0.7 version.  This would fit with the coming theme of v0.7 (quality control), and probably make unit testing easier if everything wasn't initialized at the import level.

As it stands right now, v0.6 works for Undulator.  It will crash if the .ini file isn't set up to a valid experiment before launching, but for us I think this is fine for the time being.  On the off-chance other experiments want to use v0.6 before we fix this, then there is more debugging to do initially.

I'll plan on merging this PR with master tomorrow.

GEECS-Scanner-GUI/geecs_scanner/app/geecs_scanner.py
```diff
@@ -815,15 +815,19 @@ def calculate_num_shots(self):
 
     def build_shot_array(self) -> list[int]:
```
TAU-Reinier, line 818: How about build_scan_parameter_value_array

GEECS-Scanner-GUI/geecs_scanner/app/geecs_scanner.py
```diff
@@ -815,15 +815,19 @@ def calculate_num_shots(self):
...
         """Given the parameters for a 1D scan, generate an array with the value of the scan variable for each shot."""
-        if (self.scan_stop - self.scan_start) / self.scan_step_size * self.scan_shot_per_step > MAXIMUM_SCAN_SIZE:
+        if abs((self.scan_stop - self.scan_start) / self.scan_step_size) * self.scan_shot_per_step > MAXIMUM_SCAN_SIZE:
             return []
```
TAU-Reinier, line 821: Maybe an error?

GEECS-Scanner-GUI/geecs_scanner/app/geecs_scanner.py
```diff
@@ -815,15 +815,19 @@ def calculate_num_shots(self):
...
         else:
             array = []
```
TAU-Reinier, line 823:
```python
from math import copysign
signed_scan_step_size = copysign(self.scan_step_size, self.scan_stop - self.scan_start)
array = sum([ [round(parameter_value, 10)] * self.scan_shot_per_step
    for parameter_value in range(self.scan_start, self.scan_stop + 1e-15, signed_scan_step_size)
    ], [])
```
The 1e-15 is there so that self.scan_stop is included in case (scan_stop - scan_start) is a multiple of step_size. It isn't pretty, so alternatively:

```python
from math import copysign, floor
signed_scan_step_size = copysign(self.scan_step_size, self.scan_stop - self.scan_start)
num_steps = floor((self.scan_stop - self.scan_start) / signed_scan_step_size) + 1
array = sum([ [round(self.scan_start + step * signed_scan_step_size, 10)] * self.scan_shot_per_step
    for step in range(num_steps)
    ], [])
```

GEECS-Scanner-GUI/geecs_scanner/data_acquisition/scan_manager.py
```diff
@@ -684,7 +689,7 @@ def estimate_acquisition_time(self, scan_config):
             wait_time = scan_config.get('wait_time', 1)# - 0.5  # Default wait time between steps is 1 second
 
             # Calculate the number of steps and the total time for this device
-            steps = ((end - start) / step) + 1
+            steps = abs((end - start) / step) + 1
```
TAU-Reinier, line 692: steps might be fractional.

PR #123, opened 2025-02-24
Add spectral second moment to FROG analysis.

kjensen11:
Previously only included a calculation for temporal second moment calculation. Now a spectral second moment is calculated too.


kjensen11:
YOLO merging because this is just an update to a single analysis and will not cause any merge issues.

PR #126, opened 2025-02-27
Scan analysis gui dev v021

kjensen11:
Implemented some degree of error handling and logging output to GUI text display box. 

Since scan watch and scan analysis record a large amount of logging to a dedicated log file, I decided to isolate the logging output for this GUI to only include its own dedicated logging messages.


ScanAnalysis/live_watch/scan_analysis_gui/app/ScAnalyzer.py
```diff
@@ -70,28 +71,35 @@ def start_analysis(self) -> None:
...
+    
+            # disable input fields and select buttons while analysis is running
+            to_disable = self.findChildren(QLineEdit) + [self.ui.buttonOverwrite, self.ui.buttonStart]
+            for field in to_disable:
+                field.setEnabled(False)
+
+            # initialize worker and thread
+            self.initialize_worker()
+
+        except (AnalysisRunningError, Exception) as e:
```
TAU-Reinier, line 95: AnalysisRunningError is a subclass of Exception, so Exception already catches AnalysisRunningError.

ScanAnalysis/live_watch/scan_analysis_gui/app/ScAnalyzer.py
```diff
@@ -70,28 +71,35 @@ def start_analysis(self) -> None:
...
+            self.log_error_message(str(e))
+            self.end_analysis()
+
+    def run_analysis(self,
+                     progress_callback: ProgressCallback,
+                     error_callback: ErrorCallback,
+                     is_running: RunningCheck,
```
TAU-Reinier, line 102: Good use of type aliases, but the types for ProgressCallback and ErrorCallback aren't quite right. They are `pyqtSignal` objects in at least one case, so most likely that should be the type hint (and an alias would be unnecessary). (Potentially you want to allow progress_callback and error_callback to be any object that has an emit() method, in which case you may want to look at a Protocol.)

ScanAnalysis/live_watch/scan_analysis_gui/app/ScAnalyzer.py
```diff
@@ -358,37 +326,73 @@ def set_ignore_list(self) -> None:
...
+        :return: No return.
+        :rtype: None
+
+        '''
+        # write to log display
+        self.ui.logDisplay.append(text)
+
+        # auto-scroll display to newest text
+        self.ui.logDisplay.verticalScrollBar().setValue(
+            self.ui.logDisplay.verticalScrollBar().maximum())
```
TAU-Reinier, line 396: Suggestion for a future PR: you can actually use the built-in logger to write to ui.logDisplay. You'd have to write a custom `Handler`. Then you could use the standard `logger = logging.getLogger(__name__)` and `logger.error(msg)`

TAU-Reinier (approved):

ScanAnalysis/live_watch/scan_analysis_gui/app/ScAnalyzer.py
cedoss-lbl: Ah that's interesting, I'll look into this as well

cedoss-lbl (approved):

PR #127, opened 2025-03-03
Improve data logger synchronicity

skbarber:
This PR is a solution to a recent discovery that some data can recorded can be 'off' by one shot between different devices. It seems the issue originates due to various pieces of hardware missing a hardware trigger. 

The way the synchronization was handled previously is a long wait was introduced after turning off the trigger. In this case, each device will timeout, but still send TCP events. On the backend in data_logger, the device timestamp (which refers to the timestamp for the last successful acquisition from the device) is logged and monitored. Though some code existed to check standby mode, it wasn't really used. 

Also, it was assumed that once triggering starting back up, all devices would receive that trigger and update their acquisition. however, it seems it is possible that a timed out device actually misses a hardware trigger event. This is likely an idiosyncrocy of specific hardware types and device driver logic. 

The solution here is to make use of and extend the standby status. And, critically, in scan_manager we execute a test shot and check that all devices successfully received the trigger. And, if not, try again.

### DIFF
--- a/GEECS-Scanner-GUI/geecs_scanner/data_acquisition/data_logger.py
+++ b/GEECS-Scanner-GUI/geecs_scanner/data_acquisition/data_logger.py
@@ -74,6 +74,7 @@ class DataLogger:
 
         self.last_timestamps = {}
         self.initial_timestamps = {}
+        self.synced_timestamps = {}
         self.standby_mode_device_status = {}
         self.log_entries = {}
 
@@ -97,7 +98,7 @@ class DataLogger:
 
         logging.info(
             'waiting for all devices to go to standby mode. Note, device standby status not checked, just waiting 4 seconds for all devices to timeout')
-        time.sleep(4)
+        time.sleep(1)
 
         logging.info("Logging has started for all event-driven devices.")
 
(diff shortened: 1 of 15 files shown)


GEECS-Scanner-GUI/geecs_scanner/data_acquisition/data_logger.py
```diff
@@ -72,56 +72,172 @@ def start_logging(self):
...
+        else:
+            return False
+
+    def check_device_standby_mode_status(self, device, timestamp: float):
+        # TODO: statuses are bit wonky and could be cleaned up. Right now, 'None'
+        #  really means the status is unknown. "True" indicates the device verifiably
+        #  went into standby mode. "False" indicates that device originally went into
+        #  standby mode, and has since received a verifiable hardware trigger. So,
+        #  once a "False" status has been flagged, it should stay that way until something
+        #  explicitly turns it back to "None".
```
cedoss-lbl, line 197: I suppose this is a decent way to make a 3-state boolean.  There are other ways to make something similar (like a `Union` of a number of different custom-types or something`) but this should suffice.

cedoss-lbl: Type hinting the returns also improves clarity:
` -> Union[bool, None]:`

cedoss-lbl (approved):
Looks good and seems to work.  I would say if things work good today we could go ahead and merge.  At some point adding in more doc-strings to the functions and type hinting args and returns would be nice.

skbarber:
I think all your comments are good. DataLogger was probably the first class I start developing for this whole project, and so, it is also probably the most in need of some clean up etc. The idea for this PR is to get us something quickly that allows real data synchronicity. If it works, then let's go through and clean up.

skbarber:
ran a bunch of scans in various configurations and the new logic appears sound and does appear to catch intermittent synchronization failures at the start of scans. Code is in need of some cleanup, but will deal with that in a future PR

PR #130, opened 2025-03-05
Scan analysis gui dev fix overwrite logic

kjensen11:
Branched from other active pull request. Only minimal modification. Changed overwrite button to a check box which sets a boolean state. The boolean state is passed to ScanWatch to either overwrite the existing yaml of processed scans or not overwrite.

Also added type hinting and documentation. This pull likely does not require a lot of oversight.


cedoss-lbl (approved):
All good here.  Since this also includes all of the stuff in PR129 I'd suggest waiting to merge either until you are happy with the comments of both Pull Requests

skbarber (approved):
don't have too much to contribute. Looks good, curious to play with it

PR #131, opened 2025-03-06
Version 0.6.1 (Geecsscanner dev 03052025)

cedoss-lbl:
Small update here to add in some new features.  Touched on a little bit of the backend code but nothing major.

Major:
- Added button in Action Library to import actions from save elements to the list of assigned actions.  This will enable us to use common actions (like the Visa Station Setups) without needing to open up the Save Element Editor every time.
- Added tool tip button to display the list of scan steps for 1D scans.  Tested and works with single variables, absolute composite variables, and relative composite variables.  The latter requires a GEECSDevice "get" command, and so this is only done when the button is clicked.

Minor:
- The logic behind creating this list of scan steps is used when starting a scan to check if everything is set up correctly.  Error handling will reveal if there are problems.
- Refactored static methods of `ActionControl` to a new module `action_api` that does not need an initialized GEECS-PythonAPI
- `ActionControl` now lives inside RunControl.  If RunControl is not set up then it will be impossible to perform actions.  (Also avoids complications when experiment name is not set up correctly)
- Error handling in `ScanDevice` to account for the case that the experiment name is not specified upon initialization.
- Fixed some crashes with the side GUI's when experiment name is not given
- Changed the buttons for when an action fails from "Yes/No" to "Abort/Ignore"


skbarber (approved):
Don't see anything concerning from my POV. I mean, probably not for the PR, but at some point we might want to reckon with the sprawling collection of config files in the repo. Maybe migrate them to a shared location outside of the repo?

PR #132, opened 2025-03-06
error catching added in the message_callback of async listener

skbarber:
Without this error catching, any error not handled by the callback method cause this listener to close. Now, if a message isn't handled as expected, it's logged and operations continue.

For a future PR, I will consider adding a listener that checks on the status of this async_listener and restarts it if it crashes.

### DIFF
--- a/GEECS-PythonAPI/geecs_python_api/controls/interface/tcp_subscriber.py
+++ b/GEECS-PythonAPI/geecs_python_api/controls/interface/tcp_subscriber.py
@@ -9,7 +9,7 @@ if TYPE_CHECKING:
     from geecs_python_api.controls.devices import GeecsDevice
 import geecs_python_api.controls.interface.message_handling as mh
 from geecs_python_api.controls.interface.geecs_errors import ErrorAPI, api_error
-
+import logging
 
 class TcpSubscriber:
     def __init__(self, owner: GeecsDevice):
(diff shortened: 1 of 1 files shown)


TAU-Reinier (commented):
This seems to me like the first jenga block in dismantling the accumulating-type error handling that is the basis for error handling in the geecs_python_api package. If that's the case, consider whether you want to a) overhaul the api_error based logging with the built-in logging, or b) have a mixed system, or c) consider another way to solve the problem that this commit solves.

cc @guillaumeplateau 

Note also that it's preferred to use a Logger instance, usually from `logging.getLogger(__name__)`, rather than the `logging` module itself to call `.error()` etc.

skbarber:
Yeah, this is kind of the motivation for the PR request: to try to get a
better handle on what should happen here...

Admittedly, I don't really understand the intended functionality of the
api_error vs logging. All I know is I need to add some error
logging/handling that propagates up in a sensible fashion to our execution
logs. I think I've defaulted to using the built in logging approach because
it is now marginally familiar. And it is what we are using throughout all
of the other packages at this point.

Also, the other packages are only using geecs_device and geecs_database.
Though, I would like to reconfigure geecs_database moving forward to be
more of a query based protocol rather than just pulling the whole DB at
runtime. Of course, that's not essential

skbarber:
I'm going to merge here with the understanding a more ground up standardized approach to logging across all the projects should be developed

PR #138, opened 2025-03-07
Scananalysis dev 02242025

cedoss-lbl:
Adding in my rad2 analysis for visa9 (Visa 1-8 can be added eventually, but they need to be calibrated from yesterday's scans).  Right now this is just an analysis of the camera counts, but room for also adding:
- Wavelength analysis
- Divergence analysis
- Multiscan analysis across all visa screens

Additionally, updated the Labview Analyzer calibration for the HiResMagSpec


kjensen11 (approved):
Looks good to me!

PR #139, opened 2025-03-12
Scananalysis dev 03102025

cedoss-lbl:
Includes work on the rad2 spec analysis.  Updated the cropping roi regions on Visa 1-8 and included a wavelength calibration to the Visa 9 saved data


cedoss-lbl:
Going to merge since I want to use it tonight and there's nothing too crazy here beyond rad2 analysis

PR #140, opened 2025-03-13
Camera image analysis refactor

skbarber:
Added a new base class for doing ScanAnalysis for data types that involve 2D array type data. In large part, this is a refactor of CameraImageAnalysis, however, I left that class alone for now. There are many child classes that would need to be refactored to use the new Array2DScanAnalysis class. 

All image processing is removed from the Array2DScanAnalysis. All image analysis is delegated to the ImageAnalysis class. Right now, Array2DScanAnalysis expects a return from the analyze_image method that is a dict of {'processed_image': image: NDArray, "analysis_results": dict[str, number]}. There should be sufficient error/exception handling in the case the return format is not of that type. 

Additional features added: Making use of multi threading/processing to speed up analysis. The approach was adopted from the existed LivePostProcessing project. Yields a speed up of ~5x. 

Still to do: refactor other ScanAnalysis classes to make use of image_analysis

### DIFF
--- a/ImageAnalysis/image_analysis/analyzers/ACaveMagCam3.py
+++ b/ImageAnalysis/image_analysis/analyzers/ACaveMagCam3.py
@@ -18,7 +18,7 @@ class ACaveMagCam3ImageAnalyzer(BasicImageAnalyzer):
         ----------
 
         """
-
+        self.run_analyze_image_asynchronously = True
         self.flag_logging = True
 
         super().__init__()
--- a/ImageAnalysis/image_analysis/analyzers/ACaveMagCam3.py
+++ b/ImageAnalysis/image_analysis/analyzers/ACaveMagCam3.py
@@ -1,0 +1,86 @@
+from __future__ import annotations
+
+from typing import TYPE_CHECKING, Union
+from pathlib import Path
+
+if TYPE_CHECKING:
+    from numpy.typing import NDArray
+    from ..types import Array2D
+
+import numpy as np
+from image_analysis.analyzers.basic_image_analysis import BasicImageAnalyzer
+
+class ACaveMagCam3ImageAnalyzer(BasicImageAnalyzer):
+
+    def __init__(self):
+        """
+        Parameters
+        ----------
+
+        """
+
+        self.flag_logging = True
+
+        super().__init__()
+
+    @staticmethod
+    def analyze_roi(image, roi):
+        x, y, w, h = roi
+        roi_image = image[y:y + h, x:x + w]
+        avg_count = np.mean(roi_image)
+        total_count = np.sum(roi_image)
+        max_count = np.max(roi_image)
+        return avg_count, total_count, max_count, roi_image
+
+    def analyze_image(
+            self,
+            image: Array2D = None,
+            file_path: Path = None
+    ) -> dict[str, Union[float | NDArray, dict[str, float]]]:
+        """
+        Analyze an image by simply loading it (if not already loaded) and returning it as the processed image.
+
+        Parameters
+        ----------
+        image : Array2D, optional
+            The image array to process. If not provided, it is loaded from file_path.
+        file_path : Path, optional
+            The path to the image file to load if image is None.
+
+        Returns
+        -------
+        dict
+            A dictionary with the processed image and placeholder for analysis results.
+        """
+
+        if image is None:
+            if file_path is None:
+                raise ValueError("Either an image or file_path must be provided.")
+            image = self.load_image(file_path)
+
+        # Define the ROIs (x, y, width, height) as a dictionary
+        rois = {
+            'ACaveMagCam3 High Energy ROI': (665, 130, 110, 110),
+            'ACaveMagCam3 Low Energy ROI': (815, 130, 110, 110),
+            'ACaveMagCam3 Total ROI': (660, 125, 270, 120)
+        }
+
+        # Start with a dictionary that includes the shot number
+        roi_analysis_result = {}
+
+        # Loop over each ROI and analyze it, then flatten the results into one dictionary
+        for roi_name, roi in rois.items():
+            avg, total, max_val, roi_image = self.analyze_roi(image, roi)
+            roi_analysis_result[f'{roi_name}:avg'] = avg.astype(np.float64)
+            roi_analysis_result[f'{roi_name}:total'] = total
+            roi_analysis_result[f'{roi_name}:max'] = max_val
+
+        # Append the flattened result dictionary to your list
+        return {'processed_image':roi_image, 'analysis_results': roi_analysis_result}
+
+if __name__ == "__main__":
+    image_analyzer  = ACaveMagCam3ImageAnalyzer()
+    file_path = Path('/Volumes/hdna2/data/Undulator/Y2025/03-Mar/25_0306/scans/Scan039/UC_ACaveMagCam3/Scan039_UC_ACaveMagCam3_001.png')
+    print(file_path.exists())
+    results = image_analyzer.analyze_image(file_path=file_path)
+    print(results['processed_image'])
--- a/ImageAnalysis/image_analysis/analyzers/basic_image_analysis.py
+++ b/ImageAnalysis/image_analysis/analyzers/basic_image_analysis.py
@@ -1,0 +1,56 @@
+from __future__ import annotations
+
+from typing import TYPE_CHECKING, Union
+from pathlib import Path
+
+if TYPE_CHECKING:
+    from numpy.typing import NDArray
+    from ..types import Array2D
+
+from ..base import ImageAnalyzer
+from ..utils import read_imaq_image
+
+class BasicImageAnalyzer(ImageAnalyzer):
+    def __init__(self):
+        """
+        BasicImageAnalyzer constructor that ignores ROI.
+        """
+        super().__init__()
+
+    @staticmethod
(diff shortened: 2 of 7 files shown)


ImageAnalysis/image_analysis/analyzers/ACaveMagCam3.py
```diff
@@ -0,0 +1,86 @@
...
+        Returns
+        -------
+        dict
+            A dictionary with the processed image and placeholder for analysis results.
+        """
+
+        if image is None:
+            if file_path is None:
+                raise ValueError("Either an image or file_path must be provided.")
+            image = self.load_image(file_path)
```
cedoss-lbl, line 59: Since this is copy/pasted from `basic_image_analysis.py`, you can instead use 
`image = super().analyze_image(self, image, file_path)['processed_image']`

ImageAnalysis/image_analysis/analyzers/basic_image_analysis.py
```diff
@@ -0,0 +1,56 @@
...
+        -------
+        dict
+            A dictionary with the processed image and placeholder for analysis results.
+        """
+        if image is None:
+            if file_path is None:
+                raise ValueError("Either an image or file_path must be provided.")
+            image = self.load_image(file_path)
+
+        return {'processed_image': image, 'analysis_results': {'test result':'hello'}}
```
cedoss-lbl, line 56: I feel like it would be a good idea to use the return structure that `LabviewImageAnalyzer` uses to build its dictionary.  That way any image analysis would return its results in the same format.

Could also use a class or NamedTuple to organize results instead of a big dict, but this would require more reworking and checking that LabviewAnalyzers still work.  But then we wouldn't need to remember the dictionary key names at all.

ScanAnalysis/scan_analysis/analyzers/Undulator/array2D_scan_analysis.py
```diff
@@ -0,0 +1,672 @@
...
+            if self.flag_logging:
+                logging.warning(f"Data directory '{self.path_dict['data_img']}' does not exist or is empty. Skipping")
+
+    def run_analysis(self, config_options: Optional[str] = None):
+        if self.path_dict['data_img'] is None or self.auxiliary_data is None:
+            if self.flag_logging:
+                logging.info("Skipping analysis due to missing data or auxiliary file.")
+            return
+
+        if config_options is not None:
```
cedoss-lbl, line 132: We could consider passing a camera's `settings.yaml` contents in here, rather than loading all camera settings for every camera in that one file.

This is a note more on me, but I also am reconsidering if config_options should instead be passed in the init and not here

ImageAnalysis/image_analysis/analyzers/basic_image_analysis.py
cedoss-lbl: Looking through `array2D_scan_analysis` has further convinced me this the best path forward.  Lots of type+format checking for something that should be compiled automatically at the image analysis level.

ScanAnalysis/scan_analysis/analyzers/Undulator/array2D_scan_analysis.py
```diff
@@ -0,0 +1,672 @@
...
+                    for key, value in analysis_dict.items():
+                        if not isinstance(value, numbers.Number):
+                            logging.warning(
+                                f"[{self.__class__.__name__} using {self.image_analyzer.__class__.__name__}] "
+                                f"analysis result for shot {shot_num} key '{key}' is not a number (got {type(value).__name__}). Skipping."
+                            )
+                        else:
+                            self.auxiliary_data.loc[self.auxiliary_data['Shotnumber'] == shot_num, key] = value
+
+            return image
```
cedoss-lbl, line 563: Why not just return the whole `results_dict` here?

ScanAnalysis/scan_analysis/execute_scan_analysis.py
```diff
@@ -13,19 +13,38 @@
...
     all_dispay_files = []
     for analyzer_info in analyzer_list:
         device = analyzer_info.device_name if analyzer_info.device_name else ''
         print(tag, ":", analyzer_info.analyzer_class.__name__, device)
         if not debug_mode:
             try:
                 analyzer_class = analyzer_info.analyzer_class
-                analyzer = analyzer_class(scan_tag=tag, device_name=analyzer_info.device_name, skip_plt_show=True)
+                # Instantiate the image analyzer if specified
+                if hasattr(analyzer_info, 'image_analyzer_class') and analyzer_info.image_analyzer_class:
```
cedoss-lbl, line 34: Could also pass the whole `analyzer_info` down into the `ScanAnalyzer` instead of extracting it here.

cedoss-lbl: Instead of having a full `if-else` block:
```
image_analyzer_class = analyzer_info.image_analyzer_class
image_analyzer = image_analyzer_class () if image_analyzer_class else None
analyzer = analyzer_class(scan_tag=tag, ... , image_analyzer=image_analyzer)
```

cedoss-lbl (commented):
Some things I'd like to see changed, but can work on them in the next iteration

TAU-Reinier (commented):
Please consider reverting this merge and taking another good look at this PR.

One big issue is that your derived class uses an incompatible signature of analyze_image() than its base class, and of many other derived classes. This will lead to real issues in our application because there are functions that accept any instance of any ImageAnalyzer subclass, and the instance.analyze_image() call would then fail for some and not for others. 

And to amplify @cedoss-lbl 's [comment](https://github.com/GEECS-BELLA/GEECS-Plugins/pull/140/files#r1994180647), it's not only useful to have analyze_image() return the same type for all ImageAnalyzers, it's essential for the same reason as above.

It's unclear to me what problem BasicImageAnalyzer solves. The return value of ImageAnalyzer.analyze_image() was already typed to be dict[str, float|np.ndarray], so you'd just return {'processed_image': image, 'scalar_result': some_float}. 

And finally, please check over everything that ChatGPT generates, including documentation. Something like [this](https://github.com/GEECS-BELLA/GEECS-Plugins/pull/140/files#diff-87d9cb3cc06e9b35fcf492e30e7716fc257f80d9f374f65573b89dc01b17c4a6R16) is nonsense.

skbarber:
"One big issue is that your derived class uses an incompatible signature of
analyze_image() than its base class, and of many other derived classes.
This will lead to real issues in our application because there are
functions that accept any instance of any ImageAnalyzer subclass, and the
instance.analyze_image() call would then fail for some and not for others."
Yeah, I knew this approach is not good. Though, i'm not sure how to address
the fundamental issue we are trying to solve...I'd really like to be able
to use a path to specify the 'image' to be analyzed. The .himg files (and
other file types for other devices) require the sdk to load and process
them, and those are the tools I need to do the image analysis. So, I don't
see a way that the ImageAnalyzer can be used for that kind of data unless
it can be aware of a path...Do we need to add an ImageLoader class or
something? Even then, the loader is going to generate a proprietary data
type, e.g. HasoSlopes, which is not an NDArray...

So i can revert, but then I don't know the path towards getting the
analysis done...

TAU-Reinier:
Ah ok I see. I can see a few ways to approach passing a himg file to ImageAnalyzer.analyze_image()
* The approach most in line with the current type hints is to pass the file path in auxiliary_data, i.e. `{'haso_filename': Path("/path/to/file.himg")}`. 
* You could stuff the file contents (a `bytes` object) in the `image` argument. Or even the file path. You could then consider updating the type hint on the base class to be `ImageLike`, which you could define somewhere as `NewType('ImageLike', (np.ndarray, bytes))`. Since the base class doesn't do anything with the `image` argument (implementation is left to derived classes), I'm quite sure this won't break anything.
* You could add `filepath` as a keyword argument at the end of the signature in ACaveMagCam3ImageAnalyzer.analyze_image() _after the argument list from the base ImageAnalyzer_. I'm also quite sure here that this wouldn't cause any incompatibilities.

skbarber:
That was my original thinking, to use the aux_data to pass the file name. I
guess, the image_analysis will still always expect something from the image
argument.

Maybe this is fundamentally an issue: we wanted the image_analysis to be
able to be used live with labview, which will pass an NDArray of uint16.
However, for post analysis, it feels far more natural and intuitive to pass
a file path and let the analyzer deal with it, rather than load the image
then send it to the analyzer. So, I kind of struggled with how to
accommodate this. Maybe a parent class above that switches whether to use
an NDArray arg vs file path? I think that was part of what was going on
with the addition of the BasicImageAnalyzer.

TAU-Reinier:
I suggest maybe adding a `analyze_image_file(image_filepath, auxiliary_data)` method to the ImageAnalyzer, which can be implemented in the base ImageAnalyzer as 
```python
image = self.load_image(image_filepath)
return self.analyze_image(image, auxiliary_data)
```
and the `load_image()` method as NotImplemented.

The Labview stuff can still call analyze_image directly, and ScanAnalyzer can use the analyze_image_file alternative.

skbarber:
I like this. Actually, I originally started down this path, but decided I
probably shouldn't mess with the base class...

Anyways, this seems to neatly handle using image paths and raw image data.
One consequence, which isn't a huge concern, for file types like .himg, the
load_image method is going to be doing essentially all of the work. Maybe
that makes sense though, because most of the work is in converting the
binary file to something interpretable by others.

PR #142, opened 2025-03-13
Scan analysis dev 03132025

cedoss-lbl:
Two things:
1. Checkbox on GUI to enable/disable uploading to scanlog gdoc
2. Shifted the `build_return_dictionary` function up to the base ImageAnalyzer class so that it available to the analyzers being called from ScanAnalysis.  In doing so, I updated the key names for the dict that gets returned so that they are consistent between all the types of ImageAnalyzers.

(At some point, could consider changing this complex return dict to a proper dataclass)


skbarber (approved):
looks good to me. I'll merge into my other branch and resolve conflicts there as needed, but it doesn't really look like htere will be any

PR #143, opened 2025-03-18
Haso4 3 dev with array2d scan analysis

skbarber:
adding haso lift himg ImageAnalyzers, improved Array2DScanAnalyzer, add density_from_phase_analysis.

This PR primarily adds functionality for handling and processing HasoLift recorded data. 

The Haso_himg_has_processor is a child of ImageAnalysis class. It uses the wavekit sdk to produce slopes, phase and intensity data from a .himg or .has file. It saves this data to the parent directory of the .himg or has file analyzed. To perform the analysis, the sdk has been added, but an appropriate config file is needed to make full use of it. Currently, the path to the HTU haso lift config is hardcoded. This will need to be dealt with for more broad use. 

The HIMG_with_average_saving.py is a child of Array2DScanAnalysis. The Haso_himg_has_processor is not meant to produce data that is immediately visualized. Mostly just to convert the proprietary binary files into useable data. HIMG_with_average_saving. However, it is still useful to create an average phase file for a given scan to make its use as a background much easier in downstream applications. This child class does just that.

the density_from_phase_analysis.py defines some classes and workflows to make use of phase files to extract useful data. Currently, the analysis is restricted to just do statistics and analysis on the phase and saving metrics related to the phase.


ImageAnalysis/image_analysis/base.py
```diff
@@ -55,7 +55,19 @@ def __init__(self,
...
+                that can be used by derived classes to initialize additional parameters. If not provided,
+                defaults will be used.
+
         """
+
+        # Default implementation does nothing with config.
+        # Subclasses can process config as needed.
+        self.config = config
+
         pass
```
TAU-Reinier, line 71:
```suggestion
```

TAU-Reinier (commented):
* Typically you wouldn't include a third party SDK in your repository, unless it's not publicly available or you've modified it. Consider leaving install instructions in the README, and putting the path to it in a config
* I see you set an ROI object in `HASOHimgHasProcessor.__init__`, but then you use a hard-coded ROI in apply_mask()
* 

Because this PR depends on #140 , maybe hold off on merging.

ImageAnalysis/image_analysis/offline_analyzers/HASO_himg_has_processor.py
```diff
@@ -0,0 +1,379 @@
...
+
+if TYPE_CHECKING:
+    from numpy.typing import NDArray
+    from image_analysis.types import Array2D
+
+try:
+    import image_analysis.third_party_sdks.wavekit_43.wavekit_py as wkpy
+except ModuleNotFoundError as e:
+    errmsg = "could not import wkpy, e.g. might be running on non windows machine"
+    e.add_note(errmsg)
```
TAU-Reinier, line 14: `add_note` was added in Python 3.11, whereas ImageAnalysis should support 3.7

```suggestion
    e.args += (errmsg,)
```

ImageAnalysis/image_analysis/offline_analyzers/density_from_phase_analysis.py
```diff
@@ -0,0 +1,1083 @@
...
+        gaussian_mask = np.exp(
+            -(((x - center_x) ** 2) / (2 * sigma_x ** 2) + ((y - center_y) ** 2) / (2 * sigma_y ** 2)))
+
+        # Apply the mask
+        masked_image = image * gaussian_mask
+
+        return masked_image
+
+    @staticmethod
+    def apply_gaussian_binary_mask(image, center_x=None, center_y=None, sigma_x=100, sigma_y=20, threshold_factor=0.01):
```
TAU-Reinier, line 421: This mostly duplicates apply_gaussian_mask()

Note also that ChatGPT's comments are not at all useful.

ScanAnalysis/scan_analysis/analyzers/Undulator/HIMG_with_average_saving.py
```diff
@@ -0,0 +1,107 @@
...
+
+
+class HIMGWithAveraging(Array2DScanAnalysis):
+    """
+    A child class of Array2DScanAnalysis that overrides the noscan postprocessing.
+    This version may, for example, save a custom average image and skip creating a GIF.
+    """
+
+    @staticmethod
+    def process_shot_parallel(
```
TAU-Reinier, line 33: The superclass's process_shot_parallel signature has changed, making this incompatible.

Remember that changing a function or method's arguments or return type is a breaking change so check all downstream uses of it in all projects that might use it.

ImageAnalysis/image_analysis/offline_analyzers/HASO_himg_has_processor.py
```diff
@@ -0,0 +1,379 @@
...
+          path to a slopes file for bkg subtraction.
+        laser_wavelength: float
+          Probe laser wavelength in nanometers.
+    """
+    mask: SlopesMask = SlopesMask(top=1, bottom=-1, left=1, right=-1)
+    background_path: Path = None
+    laser_wavelength: float = 800  # in nanometer
+
+    # global path to the wavekit config file for the specific serial number of HASO
+    wakekit_config_file_path: Path = Path('C:/GEECS/Developers Version/source/GEECS-Plugins/ImageAnalysis/image_analysis/third_party_sdks/wavekit_43/WFS_HASO4_LIFT_680_8244_gain_enabled.dat')
```
cedoss-lbl, line 56: Could potentially just put the path to the sdk in the config.ini file, and then could take the sdk out of the repo and place it anywhere on local PCs

ImageAnalysis/image_analysis/offline_analyzers/HASO_himg_has_processor.py
```diff
@@ -0,0 +1,379 @@
...
+                requires loading of the image and processing with a third party SDK
+
+        Returns:
+            A dictionary containing results (e.g., phase map and/or related parameters).
+
+        Raises:
+            ValueError: If the file type is not supported.
+        """
+
+        return  self.build_return_dictionary(return_lineouts=image)
```
cedoss-lbl, line 207: Should be `return_image=image`

ScanAnalysis/scan_analysis/base.py
```diff
@@ -40,6 +40,8 @@ class AnalyzerInfo(NamedTuple):
     is_active: bool = True
     # New optional field for explicit image analyzer import
     image_analyzer_class: Optional[Type[ImageAnalyzer]] = None
+    file_pattern: Optional[str] = "*_{shot_num:03d}.png"
+    image_analysis_config: Optional[any] = None
```
cedoss-lbl, line 44: This can probably be combined with the existing `config_file` field.  I don't believe `config_file` was getting any use, and I created it mostly because I thought we would get around to using config files eventually.

cedoss-lbl: I do like the addition of the `file_pattern` though.  My one worry is that if we ever are dealing with an analyzer with multiple types of non-ordinary file extensions, then just specifying one here wouldn't be enough.  That being said, `ScanData` still has tools getting filepaths for non-`.png` files so we could work with that if this happens.

ScanAnalysis/scan_analysis/mapping/map_Undulator.py
```diff
@@ -1,16 +1,41 @@
...
+
+
+bkg_file_path = get_path_to_bkg_file()
+phase_analysis_config: PhaseAnalysisConfig = PhaseAnalysisConfig(
+    pixel_scale=10.1,  # um per pixel (vertical)
+    wavelength_nm=800,  # Probe laser wavelength in nm
+    threshold_fraction=0.05,  # Threshold fraction for pre-processing
+    roi=(10, -10, 75, -250),  # Example ROI: (x_min, x_max, y_min, y_max)
+    background=bkg_file_path  # Background is now a Path
+)
```
cedoss-lbl, line 38: Hmmm, I see the dilemma.  This feels like it should belong in `density_from_phase_analysis`, but then you don't have access to `ScanData` when down in `ImageAnalysis.

While this does make me think that someday we should move ScanData to a new "Base Utilities" repo that can be accessed by any sub-repo, I think my suggestion for the time being would be just to hardcode the file path to the `average_phase.tsv` in the `density_from_phase_analysis`.  Feels clunky leaving it up here in the Undulator mapping dictionary.

cedoss-lbl (commented):
Couple of changes on the files I am more familiar with.  I think once we get things to a good working order it would be useful to keep `Array2DScanAnalysis` constant while we refactor existing scan analyzers and write unit tests

ImageAnalysis/image_analysis/offline_analyzers/HASO_himg_has_processor.py
skbarber: actually...this is a slight complication. The return_image is expected to be uint16, but, for some cases, I want to pass an array of floats. In this case, it's the wavefront in units of microns. Don't want to deal with converting to unit16 and tracking scaling. putting this into the 'lineouts' is a way around that.

ImageAnalysis/image_analysis/offline_analyzers/HASO_himg_has_processor.py
```diff
@@ -0,0 +1,350 @@
...
+        applied to the slopes using the sdk method apply_pupil
+        """
+
+        self.pupil = wkpy.Pupil(hasoslopes=self.processed_slopes)
+        pupil_buffer = self.pupil.get_data()
+        new_mask = self.apply_roi(pupil_buffer, 10,-10,75,-250)
+        self.pupil.set_data(datas=new_mask)
+        self.processed_slopes = self.post_processor.apply_pupil(self.processed_slopes, self.pupil)
+
+    def apply_roi(self, bool_array, x_start, x_end, y_start, y_end):
```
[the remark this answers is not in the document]
skbarber: done

ImageAnalysis/image_analysis/offline_analyzers/HASO_himg_has_processor.py
```diff
@@ -0,0 +1,350 @@
...
+            self.compute_phase_set.set_zonal_prefs(100, 500, 1e-6)
+
+            self.post_processor = wkpy.SlopesPostProcessor()
+
+        except Exception:
+            self._log_warning(
+                f"Not able to create necessary Wavekit objects, likely a result of Wavekit not being installed or missing/incorrect license file")
+            raise
+
+    def analyze_image(self, image: NDArray = None, file_path: Path = None) -> Tuple[wkpy.HasoSlopes, wkpy.HasoSlopes, NDArray, NDArray, NDArray]:
```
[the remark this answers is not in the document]
skbarber: fixed

ScanAnalysis/scan_analysis/analyzers/Undulator/HIMG_with_average_saving.py
skbarber: I had fixed this and comitted but forgot to push. no need to overwrite the process_shot_parallel method any more. Much cleaner now

ScanAnalysis/scan_analysis/base.py
skbarber: that's a good point. I kind of lost sight of the fact that ScanAnalysis can make use of data from multiple devices...I guess, the framework of image_analysis is still available without using these new attributes. But, it does feel like there are could be two 'competing' approaches to the analysis. one thought: we could consider bundling the image_analyzer, file_pattern, and image_analysis_config into a dict, like {'device1':image_analyzer, 'device2':image_analyzer}.

ScanAnalysis/scan_analysis/mapping/map_Undulator.py
skbarber: yeah, I agree with this. For other analyzers I was hardcoding paths, but found that annoying so I tried this. And I find this more annoying...ScanData as a util available to all projects would be great.

ScanAnalysis/scan_analysis/base.py
cedoss-lbl: Not sure we would want to have another dict floating around.  For a pure LabviewImageAnalysis, there is a dict in `labview_adapters.py` to map devices to analyzers and config files.  But I'm not too much of a fan of making a similar dict for the ScanAnalysis pipeline.  I think that all the information should either be in the `map_Undulator.py` dict, and if that is not possible then a custom ScanAnalyzer class should be made to correctly delegate data to their respective analyzers.

PR #144, opened 2025-03-19
Scan analysis dev 03182025

cedoss-lbl:
Changes:
- Updated ICT analysis scripts and added unit tests.  New version employs a sinusoidal fit
- `ScanData.get_device_shot_path` has more flexibility in specifying `file_extension` (now, for instance, `png`, `.png`, and `other_string.png` are all valid)
- Added more error handling and traceback printing for ScAnalyzer
- Updated Rad2 analysis with new image processing for Visa9, functionality to use UndulatorExitICT for Visa9, and appending a 'corrected' UndulatorExitICT charge value to the sfile for pre-March 18th scans.
- Updated `map_Undulator.py` with new requirements for Rad2, bugfixed an incorrect import statement that might have been responsible for crashing ScAnalyzer in the master branch.

This has been working on my branch from the C drive, but would be great to test the following before merging:
- Run ScAnalyzer on the Z drive and check that it appends
- Make sure the new ICT analysis works on the device computers.

TODO for future work in ScanAnalysis
- Migrate all existing scan analyzers over to the new paradigm:  Leave image analysis to ImageAnalysis and pipe basic analysis through 2DArrayAnalysis class
- Include specific analyzers to the list of "analyzed scans" for the day
- Add a "export priority" to help determine which plots should be uploaded to the Google Doc

### DIFF
--- a/GEECS-PythonAPI/geecs_python_api/analysis/scans/scan_data.py
+++ b/GEECS-PythonAPI/geecs_python_api/analysis/scans/scan_data.py
@@ -384,6 +384,34 @@ class ScanData:
         next_tag = ScanData.get_next_scan_tag(experiment, year, month, day, base_directory=base_directory)
         return ScanData(tag=next_tag, load_scalars=False, read_mode=False, base_directory=base_directory)
 
+    @staticmethod
+    def is_background_scan(tag: ScanTag, base_directory: Optional[Union[Path, str]] = None) -> bool:
+        """
+        Checks if the given scan tag references a scan that was designated as a background
+
+        Parameters:
+        ----------
+        tag : ScanTag
+            The scan tag containing year, month, day, and scan number.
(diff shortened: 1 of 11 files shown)


cedoss-lbl:
Couple more features:
- Rad2 analysis now saves the "raw camera counts" (includes roi and background processing, does not include accounting for detection limit)
- Rad2 analysis plots have information of the linear fit when in background mode (for spoiler in)
- ScAnalyzer and scan_watch now have a bit more error handling to deal with the FileNotFoundError that occurs randomly (at this point, I'm convinced it is either a strange race condition or a network error)
- Added a ScanData static method to determine if a scan was flagged as a "Background" or not
- Added a ScanData class function to make a fresh copy of the analysis sfile using the scans sfile

This branch was working quite well yesterday and overnight, so I think we can merge this into master soon

PR #147, opened 2025-03-25
Improve file moving branch cleanup

skbarber:
Top level change is the addition of the FileMover class in data_logger.py. This class handles all of the migration and renaming of files saved during a geecs_scanner scan. It uses a queue, workers and threads, which makes it relatively efficient. Note, file saving can now happen in one of two modes: i) local saving or ii) direct saving to netapp. Both have been tested. The user can toggle which mode to use from the GUI. It's been found that when saving large amounts of data (~50 MBs), saving to the netApp directly was less reliable (almost 50% data loss for haso .himg files). 

In order to use the local save approach, each device computer needs a folder called 'SharedData' in the C drive that is shared to the loasis.loasis user. A .bat script exists that can be copy/pasted to the client computers and double clicked to achieve this. Meanwhile, work with Progent (IT provider) is exploring ways to make this implemented through a domain rule. 

Some changes at the DataLogger level were needed to add FileMoveTasks to the queue each time a 'save' device registers a TCP event and adds data to the log. 

This change allowed the removal of a lot of code from ScanDataManager that was doing the file moving.

An additional auxiliary change has improved the overall reliability of generating ECS Live dumps, which had been very sporadic. 

Additional config files were added during HP operations and live testing


GEECS-Scanner-GUI/scanner_configs/experiments/Undulator/shot_control_configurations/HTU-LaserOFF.yaml
```diff
@@ -5,6 +5,6 @@ variables:
     SCAN: '4.0'
     STANDBY: '0.5'
   Trigger.Source:
-    'OFF': Single shot external rising edges
+    'OFF': Single shot
```
cedoss-lbl, line 8: Ah I see, is this the better way to do laser off?  In that case, might want to update the other configs?

GEECS-Scanner-GUI/scanner_configs/experiments/Undulator/save_devices/Base_Monitoring_Devices.yaml
```diff
@@ -1,6 +1,7 @@
 Devices:
   UC_Amp2_IR_input:
-    save_nonscalar_data: true
+    add_all_variables: false
```
cedoss-lbl, line 3: More of a note for the future, but I did end up adding this `add_all_variables` flag to each Scan Element.  Right now it does nothing and no devices have it marked `True`, but we should consider implementing this in device manager at some point.

GEECS-Scanner-GUI/geecs_scanner/data_acquisition/data_logger.py
```diff
@@ -1,18 +1,312 @@
...
+            logging.info(f"Processing variant '{variant.name}' with target '{adjusted_target_dir}'")
+
+            found_files_count = 0
+            for file in variant.glob("*"):
+                if not file.is_file():
+                    continue
+                if file in self.orphaned_files:
+                    continue
+
+                file_ts = extract_timestamp_from_file(file, device_type)
```
cedoss-lbl, line 122: Here and elsewhere in this class, we should really think of a way to more dynamically determine if there is a special suffix for a given device.  Definitely want to avoid hard-coding in HTU-specific or even device-specific things here, everything in data_acquisition should ideally be independent of experiment and devices.

Three options I can think of (but feel free to suggest others)
1. Include another field for devices in the Save Element Editor
2. Poll the experiment database and have that information set up there
3. Assume that all associated files to a given device will have the same root, and copy everything.

GEECS-Scanner-GUI/geecs_scanner/data_acquisition/scan_manager.py
```diff
@@ -314,7 +318,7 @@ def synchronize_devices(self) -> None:
         satisified, we can fire another test shot to see if all devices exit standby mode.
         This process repeats until successful or a timeout is reached.
         """
-        timeout = 17.5  # seconds
+        timeout = 25.5  # seconds
```
cedoss-lbl, line 321: Just curious, will this timeout be necessary once we have timestamp saving in the filename?

GEECS-Scanner-GUI/geecs_scanner/data_acquisition/scan_data_manager.py
```diff
@@ -385,226 +431,3 @@ def fill_async_nans(self, log_df, async_observables, fill_value=0):
         log_df = log_df.infer_objects(copy=False)
         logging.info(f"Filled remaining NaN and empty values with {fill_value}.")
         return log_df
-
-    def process_and_rename(self, scan_number: Optional[int] = None) -> None:
```
cedoss-lbl, line 389: Lots of removal here, but I suppose this is all outdated now with the filemover queue?

cedoss-lbl (commented):
I know we are kinda on schedule of "build for HTU first, generalize later" but I think it would be useful to have a path to remove the hardcoded device names out of `data_logger.py` before we get to comfortable with how it is.

Probably fine to merge once we know a path forward, then I can work on a branch implementing these future fixes while we work with these additions for HTU.

GEECS-Scanner-GUI/scanner_configs/experiments/Undulator/shot_control_configurations/HTU-LaserOFF.yaml
skbarber: yeah. well, previously, before the improved synchronicity, the 'off' mode wasn't actually used other than to turn off. But, now, I take advantage of the 'single shot' to fire off a synchronization shot. "Single shot" awaits a software trigger rather than hardware trigger like 'single shot external rising edges'.

GEECS-Scanner-GUI/geecs_scanner/data_acquisition/data_logger.py
skbarber: Yeah, probably something better could be done. I will say nothing here is 'htu' specific. This is just using device_type specifications, so it will work for all experiments. Part of the issue originates from the various GEECS devices. Some create extra save folders, some multiple files per shot within same folder. Also (and this part is on me for the current approach) some 'dependent' files don't have a timestamp that can be extracted. So, right now, there needs to be some level of special case handling. The 'expected file count' is important to keep the queue moving efficiently if a file is 'missed' during the live transfer. 

1. don't like it because it becomes even more granular, e.g. you have to specify for every individual device rather than by device type.
2. This info isn't in any database. 
3. This is kind of how the magspec is currently handled. 

I think it could be made cleaner by just parsing out this devive_type switching maybe into its own class/configuration somewhere.

GEECS-Scanner-GUI/geecs_scanner/data_acquisition/scan_data_manager.py
skbarber: yep

GEECS-Scanner-GUI/geecs_scanner/data_acquisition/scan_manager.py
cedoss-lbl: Gotcha.  We could always add another boolean option to the Scanner GUI to turn this on/off, too.

PR #155, opened 2025-04-25
v0.6.3:  Implementing timestamp in saved images

cedoss-lbl:
The major feature here was the change to the GEECS base class affecting how saved images are named.  Previously, images were saved with a shot number and this would be random unless you were running from `scan mode`.  Since our python scanner is running outside of scan mode, this resulted in us having to come up with methods to read the metadata to recover true timestamps for synchronizing shot number between devices.  This resulted in a myriad of functions to read metadata from different file types, and future devices would need labview coding to somehow save the timestamp.

Now, all GEECS devices save images with a timestamp in the filename.  This timestamp is formatted as the number of seconds since the last epoch, and includes 3 decimal places.  This allows us to load timestamps for each image and synchronize shots by only reading the filename.  We still need to get a t0 for each device in case the system timestamps are not synchronized before the scan, but this speeds up the renaming process significantly and future-proofs additional devices and file types getting added in the future.

Changes:
- Moved the timestamp retrieval functions to a new module in `GEECS-PythonAPI/geecs_python_api/tools/files/timestamping.py`.  If there ever is a device that doesn't play nice with the new filename system, we can add exception methods here.
- Added function to `ScanData` to return dataframe from the analysis folder sfile as a pandas dataframe.
- Added `scan` column to saved scalars for scans running through GEECS Scanner.  This column is necessary for GEECSPlotter's add columns functionality.
- `acq_timestamp` is now the preferred variable for tracking the timestamp of each device's last previous successful acquisition
- Removed `timestamp` from all element config yamls.  Now, `acq_timestamp` is automatically added to the list of variables if that device need to either save nonscalar data or is a synchronous device.
- Temporarily disabled the `Add All Variables` checkbox in the Element Editor GUI until we finalize how this should be implemented
- In ActionManager, before an action list is executed we now ping all relevant devices to check that they are all on and connected.  This is implemented by trying to `get` the `SysTimestamp` for a device, which is from the base class and should be subscribe-able for every device.


GEECS-Scanner-GUI/geecs_scanner/data_acquisition/action_manager.py
```diff
@@ -117,6 +119,19 @@ def execute_action(self, action_name):
...
+        GeecsDeviceInstantiationError if a device is not turned on
+        """
+        devices = []
+        for step in action_steps:
+            if device_name := step.get('device', None):
+                if device_name not in devices:
+                    devices.append(device_name)
+        for device_name in devices:
+            self.return_value(device_name, 'SysTimestamp')
+
```
skbarber, line 134: there's probably a few ways to do this faster. Right now, this waits for the response for the time for each device, which can take seconds. But, there is a 'fast' response to each UDP command which is read immediately and lets the user know if the command was accepted without having to wait for the response. I'll look into some faster alternatives to check if a device is 'alive'

GEECS-Scanner-GUI/geecs_scanner/data_acquisition/data_logger.py
```diff
@@ -287,12 +287,12 @@ def post_process_orphaned_files(self, log_df: pd.DataFrame, device_save_paths_ma
             target_dir = Path(device_info['target_dir'])
             device_type = device_info['device_type']
 
-            # Create a list of (shotnumber, timestamp) pairs from the df.
-            # Ensure the df columns are named appropriately.
+            # Create a list of (shotnumber, timestamp) pairs from the df. Ensure the df columns are named appropriately.
+            # *NOTE* Using `SysTimestamp` for data that was logged
```
skbarber, line 291: I think you mean acq_timestamp here, not SysTimestamp

GEECS-Scanner-GUI/geecs_scanner/data_acquisition/data_logger.py
```diff
@@ -779,7 +783,7 @@ def _log_device_data(self, device:GeecsDevice, elapsed_time:float) -> None:
                     target_dir=cfg['target_dir'],
                     device_name=device_name,
                     device_type=cfg['device_type'],
-                    expected_timestamp=observables_data['timestamp'],
+                    expected_timestamp=observables_data['acq_timestamp'],  # *NOTE* `SysTimestamp` for data logging
```
skbarber, line 786: SysTimestamp? think this is wrong

GEECS-Scanner-GUI/geecs_scanner/data_acquisition/device_manager.py
```diff
@@ -137,17 +136,19 @@ def _load_devices_from_config(self, config):
...
             synchronous = device_config.get('synchronous', False)
             save_non_scalar = device_config.get('save_nonscalar_data', False)
-            post_analysis_class_name = device_config.get('post_analysis_class', None)
             scan_setup = device_config.get('scan_setup', None)
-            logging.info(f"{device_name}: Post Analysis = {post_analysis_class_name}")
             logging.info(f"{device_name}: Synchronous = {synchronous}, Save_Non_Scalar = {save_non_scalar}")
 
             # Add to non-scalar saving devices if applicable
-            if save_non_scalar:
+            if save_non_scalar:  # *NOTE* `SysTimestamp allows for file renaming of nonscalar data
```
skbarber, line 143: SysTimestamp in the comments...should be acq_timestamp, no?

GEECS-Scanner-GUI/geecs_scanner/data_acquisition/action_manager.py
cedoss-lbl: In that case, yeah would just need to change that one line (line 133) with this UDP ping command.

cedoss-lbl:
You were right that I missed some uses of `SysTimestamp` in the comments.  I fixed those and I made a new function to house the old metadata timestamp extraction methods.  This method is not used currently in GEECS Scanner but the functionality is there.

I will go ahead and merge to master

PR #156, opened 2025-04-25
Debug some image and scan analysis issues

skbarber:
small PR for some debugging of Scan/Image analysis. Remvoed spyder kernels from .toml, added nptdms. Reorganized some tests


ScanAnalysis/scan_analysis/analyzers/common/array2D_scan_analysis.py
```diff
@@ -252,7 +252,7 @@ def process_error(shot_number: int, exception: Exception) -> None:
         # Gather tasks: each shot number paired with its file path.
         tasks = []
         for shot_num in self.auxiliary_data['Shotnumber'].values:
-            pattern = self.file_pattern.format(shot_num=shot_num)
+            pattern = self.file_pattern.format(shot_num=int(shot_num))
```
cedoss-lbl, line 255: I like this approach to dealing with the data that comes out of the sfile.  Seems like columns like shotnumber can have either a decimal (`3.0`) or not (`3`) and it is difficult to enforce one way over the other across all of Python and GEECS.  Makes sense it would be up to the user during ScanAnalysis to read the data in the right format they need.

ScanAnalysis/scan_analysis/execute_scan_analysis.py
```diff
@@ -21,8 +21,13 @@
...
 from scan_analysis.base import ScanAnalysis
-from image_analysis.offline_analyzers.density_from_phase_analysis import PhaseAnalysisConfig  # import your config class
-from image_analysis.offline_analyzers.HASO_himg_has_processor import HasoHimgHasConfig  # import your config class
+
+try:
+    from image_analysis.offline_analyzers.density_from_phase_analysis import PhaseAnalysisConfig  # import your config class
+    from image_analysis.offline_analyzers.HASO_himg_has_processor import HasoHimgHasConfig  # import your config class
+except:
+    logging.warning(f'from execute_analysis: could not load haso components')
+
```
cedoss-lbl, line 30: Kyle and I want to eventually refactor these HASO image analyzer things down to children of Array2DScanAnalysis.  But for now this is fine.

cedoss-lbl (approved):
Looks good!  Feel free to merge whenever you're ready

PR #160, opened 2025-05-14
Add batch image analysis workflow

skbarber:
this PR is mostly to modify the basic workflow of array2danalysis. The idea is, some analysis tasks require analyzing the complete set of images before performing individual image analysis. The initial use case is determining the background from a set of images, then subtract that background from the individual images before performing the analyze_image method.

To accomodate this, a analyze_image_batch method was added to the base class ImageAnalysis. Then, the work flow in array2danalysis is to load the images, perform the batch analysis, then perform the individual image analysis. Note, anything intended to be used as a result of the batch analysis should be stored as an attribute of the sepcific image analysis class, e.g. self.background. 

Some additional work was added to speed up the basic execution, including a (slightly) improved way of organizing and finding the relevant files.


ImageAnalysis/image_analysis/offline_analyzers/Undulator/ALine3.py
skbarber, line 1: I would say it isn't super necessary to read too deeply into this file. It's mostly just an initial example to test the new array2danalysis workflow. The plan is to implement a more generalized approach to doing the background subtraction and the beam_stats stuff should be refined and moved into image_analysis utils.

ScanAnalysis/scan_analysis/analyzers/common/array2D_scan_analysis.py
```diff
@@ -155,142 +194,199 @@ def run_analysis(self, config_options: Optional[str] = None):
...
             return
 
-    @staticmethod
-    def process_shot_parallel(
-            shot_num: int, file_path: Path, image_analyzer: ImageAnalyzer
-    ) -> tuple[int, Optional[np.ndarray], dict]:
+    def _process_all_shots_parallel(self):
+        self._load_all_images_parallel()
+        self._run_batch_analysis()
+        self._run_image_analysis_parallel()
```
cedoss-lbl, line 200: I like this function.  If in the future someone wanted to change the way that Array2DScanAnalysis works, they could easily make a subclass and overwrite this method.

cedoss-lbl (approved):
In general I like these changes.  The only open discussion I see is if we should make a specific `OfflineImageAnalyzer` subclass of the base `ImageAnalyzer` to house all of these new features.  As it stands now, `LabviewImageAnalyzer` will have access to these new methods and they will just cause errors or do unexpected things if one of the older Labview analyzers tries to run these new methods.  Making a fresh subclass to formally separate the two analyzer types could be the cleanest way to prevent this

ImageAnalysis/image_analysis/offline_analyzers/Undulator/ALine3.py
```diff
@@ -0,0 +1,182 @@
...
+        return_dictionary = self.build_return_dictionary(return_image=uint16_image, return_scalars=beam_stats)
+
+        return return_dictionary
+
+if __name__ == "__main__":
+    image_analyzer  = Aline3Analyzer()
+    file_path = Path('Z:\\data\\Undulator\\Y2025\\05-May\\25_0507\\scans\\Scan029\\UC_ALineEBeam3\\Scan029_UC_ALineEBeam3_001.png')
+    print(file_path.exists())
+    results = image_analyzer.analyze_image_file(image_filepath=file_path)
+    print(results)
```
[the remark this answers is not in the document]
skbarber: yep, I need to get into a better habit of doing that. Though, because the path is hardcoded, and that path is not available to everyone, maybe we should make a 'data' folder to hold test data.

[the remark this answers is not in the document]
cedoss-lbl: One workaround to this is to generate some fake data to test the script.  That way you could also make the fake data be a tiny array so that the test completes instantly.

skbarber:
> In general I like these changes. The only open discussion I see is if we should make a specific `OfflineImageAnalyzer` subclass of the base `ImageAnalyzer` to house all of these new features. As it stands now, `LabviewImageAnalyzer` will have access to these new methods and they will just cause errors or do unexpected things if one of the older Labview analyzers tries to run these new methods. Making a fresh subclass to formally separate the two analyzer types could be the cleanest way to prevent this

good point. Maybe you could just overwrite any unusable methods such that they just log a warning or something (e.g. 'analyze_image_batch method is not supported by LabViewImageAnalyzer types'

skbarber:
I'm going to go ahead and merge this branch. I don't see any backwards compatibility issues at this point

ScanAnalysis/scan_analysis/analyzers/common/array2D_scan_analysis.py
```diff
@@ -155,149 +195,206 @@ def run_analysis(self, config_options: Optional[str] = None):
...
 
-        if not isinstance(results_dict, dict):
-            logging.warning(f"Analyzer returned non-dict result for shot {shot_num}.")
-            return shot_num, None, {}
+            m = image_filename_regex.match(file.name)
+            if m:
+                suffix = m.group('suffix') or ''
+                file_format = m.group('format')
+                tail = f"{suffix}.{file_format}"
+                if tail != self.file_tail:
```
TAU-Reinier, line 228: Better than using a regex and then enforcing part of the captured groups to be something specific is just to add it into the regex: 

```python
        image_filename_regex = re.compile(
            r"Scan(?P<scan_number>\d{3,})_"  # capture scan number
            r"(?P<device_subject>.*?)_"  # non-greedy device/subject
            r"(?P<shot_number>\d{3,})"  # capture shot number
            + self.file_tail
        )

TAU-Reinier, line 228: Better than using a regex and then enforcing part of the captured groups to be something specific is just to add it into the regex: 

```python
        image_filename_regex = re.compile(
            r"Scan(?P<scan_number>\d{3,})_"  # capture scan number
            r"(?P<device_subject>.*?)_"  # non-greedy device/subject
            r"(?P<shot_number>\d{3,})"  # capture shot number
            + self.file_tail
        )
  ```

skbarber: implemented this and merged into master

Issue #159, opened 2025-05-13
change where uint16 is enforced in ImageAnalysis workflow


skbarber:
the build_return_dictionary is a method of ImageAnalysis that is used to structure returns from ImageAnalysis.analyze_image() in a semi-standard way. It's not required to be used by ImageAnalysis, but it is a useful shortcut. It was originally designed for use with ImageAnalyzers used for Labview, which is very strict with the return type, but it was added to the base class of ImageAnalysis. ImageAnalysis itself is more flexible with return types. In any case, I would like to change this part of the method which currently coerces the returned image to the uint16 type. 
https://github.com/GEECS-BELLA/GEECS-Plugins/blob/d310f360b7561e3105a289a6cd666e03e4bbc510/ImageAnalysis/image_analysis/base.py#L178

While this is necessary for the labview workflow, it can be enforced further downstream, for example, in labview_adapters.py:
(https://github.com/GEECS-BELLA/GEECS-Plugins/blob/d310f360b7561e3105a289a6cd666e03e4bbc510/ImageAnalysis/image_analysis/labview_adapters.py#L113C1-L114C1)

Reorganizing in this way allows ImageAnalysis to return non uint16 results, which is useful for things like returning background subtracted images that have potentially negative values. In principle, the negative values could be dealt with by an offset, but that would require tracking of additional information. Overall, allowing a bit more flexibility in the return type is useful and I think the uint16 requirement for labview can still be preserved. In fact, the conversion to uint16 should definitely be done in the ImageAnalyzer before sending as a return anyways. Probably, it makes sense to log a warning or error in the case that a non-uint16 image is being returned for use in labview rather than coercing it to 16 bit, which could lead to hard to track down errors.

cedoss-lbl:
I can work on a fix for this, but should I do this change on a fresh branch off of master or use the new pull request #160 and make this change on that branch?

skbarber:
I would say it makes sense to make a fresh branch off master and work on
this there. I don't think there will be any compatibility issues

skbarber:
I would think there isn't a huge amount of code that needs to be modified
(though, we may want to rename 'processed_image_uint16' key, which would
just need to be traced carefully)

cedoss-lbl:
[...]

PR #161, opened 2025-05-14
Image analysis return type 05142025

cedoss-lbl:
Here's the solution to Issue #159 that we discussed.

- `np.uint16` is no longer enforced at the image analyzer level
- It is instead enforced just before returning back to Labview
- A TypeError will be raised if the image contains pixels that are less than 0 or greater than 2e16-1
- The key name has been changed from `processed_image_uint16` to `processed_image`.  Most notably Array2DScanAnalysis had a lot of these instances but there could be more outside of ImageAnalysis and ScanAnalysis packages.

### DIFF
--- a/ImageAnalysis/image_analysis/labview_adapters.py
+++ b/ImageAnalysis/image_analysis/labview_adapters.py
@@ -110,7 +110,10 @@ def parse_results_to_labview(return_dictionary, key_list_name):
         - 2D double array
 
     """
-    return_image = return_dictionary['processed_image_uint16']
+    return_image = return_dictionary['processed_image']
+    if return_image is not None:
+        return_image = return_image.astype(np.uint16)
+
     scalar_dict = return_dictionary['analyzer_return_dictionary']
     return_lineouts = return_dictionary['analyzer_return_lineouts']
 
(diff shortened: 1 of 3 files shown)


ImageAnalysis/image_analysis/labview_adapters.py
```diff
@@ -109,8 +109,20 @@ def parse_results_to_labview(return_dictionary, key_list_name):
...
-    return_image = return_dictionary['processed_image_uint16']
+    return_image = return_dictionary['processed_image']
+    if return_image is not None:
+        if np.min(return_image) < 0:
+            raise TypeError("Return image cannot be cast as np.uint16.  Some pixels are < 0")
+        if np.max(return_image) > 2e16-1:
+            raise TypeError("Return image cannot be cast as np.uint16.  Some pixels are > 2e16-1")
+
+        return_image = return_image.astype(np.uint16)
+
```
skbarber, line 125: what about something more general that handles other issues that might make the image not properly castable:
def validate_uint16_castable(array: np.ndarray) -> None:
    if np.iscomplexobj(array):
        raise TypeError("Array contains complex values.")
    if not np.isfinite(array).all():
        raise TypeError("Array contains NaNs or infinite values.")
    if np.any((array < 0) | (array > np.iinfo(np.uint16).max)):
        raise TypeError("Array values out of uint16 range.")

skbarber (approved):
small suggestion about more general check for uint16 casting. Maybe unnecessary

ImageAnalysis/image_analysis/labview_adapters.py
cedoss-lbl: Yeah this is probably a good call.

PR #163, opened 2025-05-16
Add background object to image analysis

skbarber:
adding a background class in ImageAnalysis/tools/background.py.

This can help standardize and streamline background subtraction for ImageAnalyzers. In principle, subtracting a background is a very simple operation that maybe doens't require a dedicated class. However, we've recently found an extremely effective method for identifying and subtracting backgrounds from sets of images with transient signals, which is a very common type of dataset. The methods to do this subtraction are now all available in the background class, making it straightforward to implement in an ImageAnalyzer child class. 

the Aline3.py is a functional example showing how to use it.


ImageAnalysis/image_analysis/tools/background.py
```diff
@@ -0,0 +1,189 @@
...
+        self._min_subtracted_value: float = 0.0
+
+    def set_background_from_array(self, background: np.ndarray) -> None:
+        self._background = background.astype(np.float64)
+
+    def load_background_from_file(self, file_path: Union[str, Path]) -> None:
+        path = Path(file_path)
+        if not path.exists():
+            raise FileNotFoundError(f"Background file not found: {path}")
+        self._background = np.load(path).astype(np.float64)
```
TAU-Reinier, line 31: This only works for numpy array (.npy) files. You probably want to use `utils.read_imaq_image` for this.

ImageAnalysis/image_analysis/base.py
```diff
@@ -24,7 +25,7 @@ class ImageAnalyzer:
     # asynchronously, for example if it waits for an external process
     run_analyze_image_asynchronously = False
 
-    def __init__(self, config: Optional[Any] = None):
+    def __init__(self, config: Optional[Any] = None, background_obj: Optional[Background]=None):
```
TAU-Reinier, line 28: Is it presumptuous that all ImageAnalyzer subclasses would want a self._background? It may not be appropriate here. 

Also, typically you'd pass config arguments to implemented by derived classes as `**kwargs`. That way subclasses aren't required to have `config` as part of their signature in the strict typing sense.

TAU-Reinier (approved):

ImageAnalysis/image_analysis/tools/background.py
skbarber: yep, good point. will do

ImageAnalysis/image_analysis/base.py
skbarber: it's a fair point. I would say that probably at least 95% of 2d type data, a background makes sense. That might actually be 100%. I can't really think of an example in which a background would not make sense. It might not always be _necessary_ to use a background because a given diagnostic might be nicely set up and there is literally no stray signal, but it's really hard to think of a diagnostic for which background is inappropriate...

If I have this correct, are you suggesting doing something like this:
class ImageAnalyzer:
    run_analyze_image_asynchronously = False
    def __init__(self, **kwargs):
        self.config = kwargs.get("config", None)
        self.background_obj = kwargs.get("background_obj", Background())
        
There are a number of other bad practices around the implementation of ImageAnalyzers in conjunction with Array2DAnalysis that need to be addressed. I think this actually ties in to that work a bit (i.e. how the configs get passed, where the ImageAnalyzer is instantiated...)

TAU-Reinier: Ok, I would then arranging the arguments as this:
```python
class ImageAnalyzer:
    def __init__(self, background: Background = None, **config):
        self.background = background
```
where the config represents a series of keyword arguments (as current subclasses pass parameters), not some config dict or object.

This indicates that derived classes' `__init__` method should follow the signature `__init__(self, background: Background = None, analyzer_specific_kwarg1: float = 1.23, analyzer_specific_kwarg2: bool = False)`

skbarber: why not use a dict or object for config? That seems potentially cleaner to me if I have like 5-10 kwargs

skbarber: ok, i'll defer to you guys on this topic. This will require some basic changes to how Array2DAnalysis is used and the current definition of the AnalyzerInfo class. I created an 'issue' for this topic that can be worked on separately. I will merge this branch though.

PR #167, opened 2025-05-20
Geecs scanner unittests 04282025

cedoss-lbl:
Pull request with current progress for GEECS Scanner unit tests.  Added some tests for the basic functionality of DataLogger, DeviceManager, and ScanManager (running a noscan)


GEECS-Scanner-GUI/geecs_scanner/data_acquisition/device_manager.py
```diff
@@ -362,6 +374,8 @@ def add_scan_device(self, device_name, variable_list=None):
         Add a new device or append variables to an existing device for scan operations and
         recording their data.
 
+        # TODO is there ever an instance where variable_list is a list?  Can we make it just a single string?
```
skbarber, line 377: the method from geecs_device requires a list. I would say, it's almost always a list anyways. Once you want to record more than one variable from a device...

GEECS-Scanner-GUI/geecs_scanner/data_acquisition/device_manager.py
```diff
@@ -371,6 +385,7 @@ def add_scan_device(self, device_name, variable_list=None):
             logging.info(f"Adding new scan device: {device_name} with default settings.")
             self._subscribe_device(device_name, var_list=variable_list)
 
+            # TODO can we delete these lines of code for `self.nonscalar_saving_devices`?
```
skbarber, line 388: I don't think so...At least, not without looking carefully. I'm pretty sure the non_scalar_saving flag is used in other places. Now, can it be cleaned up: probably.

tests/ScannerTools/test_data_acquisition/test_scan_manager.py
```diff
@@ -0,0 +1,75 @@
...
+        if data_folder.exists() and data_folder.is_dir():
+            shutil.rmtree(data_folder)
+        if analysis_folder.exists() and analysis_folder.is_dir():
+            shutil.rmtree(analysis_folder)
+        if sfile.exists() and sfile.is_file():
+            sfile.unlink()
+
+
+if __name__ == "__main__":
+    pytest.main()
```
skbarber, line 75: This is maybe an inherent flaw to scan_manager, but I think it does require at least one syunchronous device to be added to acquire data (though, maybe that isn't a flaw...). I see this is using TC_Phosphor to serve that purpose. The GEECS devices do have a 'simulation' mode which could be useful here. The minor issue being that this test kind of only works for HTU. We could have a camera device called 'test_geecs_scanner' that is added to PW, HTU, HTT etc. that just runs in simulation mode.

skbarber (approved):

GEECS-Scanner-GUI/geecs_scanner/data_acquisition/device_manager.py
cedoss-lbl: Generally speaking, yes.  But this note was made within `add_scan_device()` where you will only ever be scanning a single variable (unless we start talking about 2D scans...).  The only two places this function are called have either (a) `variable_list=None` or (b) `variable_list=[var_name]` (see `handle_scan_variables()` at the bottom of `device_manager.py`)

That being said, all the general subscription functions use a list of variables so it is probably more of a pain to make `variable_list` not a list in this function.  We can leave it as is for now, it only marginally bothers me.

GEECS-Scanner-GUI/geecs_scanner/data_acquisition/device_manager.py
cedoss-lbl: I agree the flag is used in other places.  But this note was for the local `default_device_config` dict that is defined, used in only one spot, but then is False so the line `self.non_scalar_saving_devices.append(device_name) if default_device_config['save_non_scalar_data'] else None` always does nothing.  Maybe these lines did more in the past, but I don't see their purpose anymore

tests/ScannerTools/test_data_acquisition/test_scan_manager.py
cedoss-lbl: True, I don't have a great solution for this yet.  Best I can think of is if there is a "Test" experiment that all experiments would have access to, and in that experiment there is a test camera that can run in simulation mode.  I can ask Chetanya if this is realistic.

GEECS-Scanner-GUI/geecs_scanner/data_acquisition/device_manager.py
skbarber: ah, I see...maybe that was the thought: could be used for 2D scans or composite variables...I'm not opposed to updating it then

cedoss-lbl:
I'll go ahead and merge this to master.  I'll leave these TODO's in as food for thought in a later PR where we tackle more data_acquisition cleanup and we can decide then

GEECS-Scanner-GUI/geecs_scanner/data_acquisition/device_manager.py
skbarber: good point. delete!

PR #169, opened 2025-05-29
Implement image rendering utilities in image analysis

skbarber:
Makes several substantive changes to Scan/Image analysis. The initial intention centered on enabling more modular rendering options within Scan/Image analysis. As it stood, the rendering was generally hardcoded in the ScanAnalysis classes. Thus, customizing rendering required a dedicated child class. Now, a static rendering method can be defined in ImageAnalysis child classes that can be used by array2Danalysis. In setting this up, the general framework for creating creating image grids and gifs was improved.

As part of the above, a number of things were cleaned up/improved with the basic functionality of Array2DAnalysis. 

There are still more improvements to be made, but everything has been tested and is functional.


ImageAnalysis/image_analysis/offline_analyzers/Undulator/VisaEBeam.py
```diff
@@ -0,0 +1,469 @@
...
+
+from dataclasses import dataclass
+from dataclasses import asdict
+
+from image_analysis.base import ImageAnalyzer
+from image_analysis.tools.rendering import base_render_image
+
+import logging
+
+VISA_EBEAM_PRESETS = {
```
cedoss-lbl, line 21: My preference would be for this large dict to instead live in a yaml file somewhere.  Whether this would live in the Z drive under user_data or in this github repo is another question though...

ImageAnalysis/pyproject.toml
cedoss-lbl, line 1: Might make sense to upgrade the ImageAnalysis version number the same way we've been doing the other projects.

ImageAnalysis/tests/analyzers/test_visaebeam.py
cedoss-lbl, line 1: These test images are probably fine for now.  But in there future we could consider generating some quick, low-resolution test images in python to speed up these unit tests.

cedoss-lbl (approved):
I left some minor comments, but nothing outstanding from me

ImageAnalysis/image_analysis/tools/rendering.py
```diff
@@ -0,0 +1,46 @@
+# image_analysis/tools/rendering.py
+
+from typing import Optional, Union, Tuple
+import numpy as np
+import matplotlib.pyplot as plt
+
```
TAU-Reinier, line 6: Figure and Axes aren't exported by pyplot.

```suggestion
from typing import Optional, Union, Tuple, TYPE_CHECKING
import numpy as np
import matplotlib.pyplot as plt

if TYPE_CHECKING:
    from matplotlib.figure import Figure
    from matplotlib.axes import Axes
```

ScanAnalysis/scan_analysis/analyzers/common/array2D_scan_analysis.py
```diff
@@ -92,34 +104,45 @@ def load_image_wrapper_static(shot_num: int, path: Path, analyzer_class: type) -
...
-        return shot_num, None, {}
+        return {}
 
 # %% classes
 class Array2DScanAnalysis(ScanAnalysis):
 
     def __init__(self, scan_tag: ScanTag,
                  device_name: str,
                  image_analyzer: Optional[ImageAnalyzer] = None,
+                 image_analyzer_config: Optional[dict[Any]] = None,
```
TAU-Reinier, line 145:
```suggestion
                 image_analyzer_config: Optional[dict[str, Any]] = None,
```

TAU-Reinier (commented):
Aline3Analyzer.analyze_image_batch() needs to be updated to the output defined in base.py, and its references to background_obj need to be updated as well.

ImageAnalysis/image_analysis/tools/rendering.py
TAU-Reinier: Also in ALine3.py

skbarber:
> Aline3Analyzer.analyze_image_batch() needs to be updated to the output defined in base.py, and its references to background_obj need to be updated as well.

got it. I think I will ultimately deprecate and delete this class in favor of adopting the construction of the VisaEBeam analyzer (which I think might be turned into a more generic 'HTUEBeam' analyzer class to handle our yag/phosphor screens).

ScanAnalysis/scan_analysis/execute_scan_analysis.py
```diff
@@ -47,34 +47,15 @@ def instantiate_analyzer(tag: ScanTag, analyzer_info: AnalyzerInfo) -> ScanAnaly
...
         else:
             image_analyzer = analyzer_info.image_analyzer_class()
-        logging.info(f'the image analyzer config is {image_analyzer.config}')
     analyzer_instance = analyzer_info.analyzer_class(
         scan_tag=tag,
         device_name=analyzer_info.device_name,
         skip_plt_show=True,
-        image_analyzer=image_analyzer
+        image_analyzer=image_analyzer,
+        image_analyzer_config=config
```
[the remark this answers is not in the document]
skbarber: how about changing the ScanAnalyzer instantiation to work more like this:
```python
def instantiate_analyzer(tag: ScanTag, analyzer_info: AnalyzerInfo) -> ScanAnalysis:
    """
    Instantiate a ScanAnalysis subclass using the information from AnalyzerInfo.

    All kwargs for the subclass are defined in analyzer_info.analyzer_kwargs.
    """
    analyzer_cls = analyzer_info.analyzer_class
    init_kwargs = {
        "scan_tag": tag,
        "device_name": analyzer_info.device_name,
        "skip_plt_show": True,
        **(analyzer_info.analyzer_kwargs or {})
    }

    return analyzer_cls(**init_kwargs)
    ```
    
and adopting a signature for the base ScanAnalysis like this:

```python
class ScanAnalysis:
    def __init__(
        self,
        scan_tag: ScanTag,
        device_name: Optional[str] = None,
        skip_plt_show: bool = True,
        **kwargs
    ):
```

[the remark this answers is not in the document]
TAU-Reinier: The problem with this is that the `ScanAnalysis.__init__` kwargs are for the ScanAnalysis instance, whereas the argument in question is the dict of kwargs for the ImageAnalyzer instance. So it wouldn't be quite right to add **kwargs to the `ScanAnalysis.__init__` that aren't meant for the ScanAnalysis instance. You'll want to reserve this - down the line you may find that you'll want to pass kwargs to ScanAnalysis subclasses this way. 

One way to eliminate image_analyzer_config is to expect a configured ImageAnalyzer instance for the image_analyzer kwarg of ScanAnalysis.

TAU-Reinier:
Regarding [the discussion above](https://github.com/GEECS-BELLA/GEECS-Plugins/pull/169#discussion_r2122192360) with scan analyzer/analysis instances and image analyzer instances, please use a bit more explicit variable names `scan_analysis` and `image_analyzer`, `scan_analysis_class`, `image_analyzer_class`, etc. because currently `analyzer` is being used for both and it's really hard to keep track.

ScanAnalysis/scan_analysis/execute_scan_analysis.py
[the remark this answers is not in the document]
skbarber: ok, last several commits resolve this issue as mentioned above. ScanAnalysis now has as a simplified constructor that doesn't require ImageAnalysis. There is now a **kwargs argument. Updated derived classes to adopt the new signature. 

also, array2danalysis now only uses a fully instantiated ImageAnalyzer. To prevent issues, a check is performed to see if the instance is pickleable, and if not, switch to using threading while logging a warning.

TAU-Reinier:
> Regarding [the discussion above](https://github.com/GEECS-BELLA/GEECS-Plugins/pull/169#discussion_r2122192360) with scan analyzer/analysis instances and image analyzer instances, please use a bit more explicit variable names `scan_analysis` and `image_analyzer`, `scan_analysis_class`, `image_analyzer_class`, etc. because currently `analyzer` is being used for both and it's really hard to keep track.

It's probably best to rename the class ScanAnalyzer, because all variables referencing it are also called analyzer. 

Please also document AnalyzerInfo, and perhaps rename the attributes to `scan_analysis_class` (or `scan_analyzer_class`) and `scan_analysis_kwargs`/`scan_analyzer_kwargs`.

ImageAnalysis/pyproject.toml
skbarber: done

PR #171, opened 2025-05-30
Data acquisition    scan config refactor

skbarber:
ScanManager makes use of a 'scan_config' to execute scans. Previously, this was a dict that was passed around as an argument to many methods that are mostly used internally. In reality, it's really only the start_scan_thread method that should take the scan_config as an argument. 

This PR defines a ScanConfig dataclass in a types.py file and this type is used throughout ScanManager and other modules that require knowledge of the scan_config.


skbarber:
@SchroederSa

Issue #174, opened 2025-06-04
Suggestion: Make ScanAnalysis reusable by changing scan_tag from an attribute of ScanAnalysis to an argument of run_analysis


TAU-Reinier:
This is a suggestion to set up ScanAnalysis [^1] to be instantiated once, to be used on multiple Scans, rather than once for each Scan. Currently, scan_tag is an attribute of ScanAnalysis, so a ScanAnalysis with the same config each time needs to be instantiated for each new Scan. 

The suggestion is to instantiate a ScanAnalysis subclass only once as an engine that can be kept around and used on several scans. This is also how ImageAnalyzer is set up. 

For implementation, this means removing the scan_tag argument from the ScanAnalysis constructor and adding it as an argument to run_analysis.

[^1] Also [a suggestion to rename this](https://github.com/GEECS-BELLA/GEECS-Plugins/pull/169#issuecomment-2941046094) to ScanAnalyzer.

PR #181, opened 2025-07-24
[WIP] improve support for optimization workflows, resolve some issues

skbarber:
This pull request is mostly to add improved support for optimization workflows. it also resolves: #174


ScanAnalysis/scan_analysis/base.py
```diff
@@ -101,37 +101,66 @@ class ScanAnalyzer:
...
 
-        self.device_name = device_name
+        self.scan_tag: Optional[ScanTag] = None
+        self.scan_data: Optional[ScanData] = None
+        self.scan_directory: Optional[Path] = None
+        self.experiment_dir: Optional[str] = None
+        self.ini_file_path: Optional[Path] = None
+        self.scan_path: Optional[Path] = None
+        self.auxiliary_file_path: Optional[Path] = None
+        self.scan_parameter: Optional[str] = None  # the one you’ll *use*
```
TAU-Reinier, line 124: A possible way to make this neater is to create a small ScanParameter struct that can output in both styles: 

```python
class ScanParameter(NamedTuple):
    device: str
    metric: str

    def with_colon(self):
        return f"{self.device}:{self.metric}"
    def with_space(self):
        return f"{self.device} {self.metric}"
    def __str__(self):
        # default, used for example in f"{scan_parameter}"
        return self.with_colon()
```

ImageAnalysis/pyproject.toml
```diff
@@ -15,6 +15,7 @@ scipy = "^1.7"
 Pint = ">=0.18,<=0.22"
 pyabel = ">=0.9,<1.0"
 nptdms = "^1.7.0"
+h5py = "*"
```
TAU-Reinier, line 18: Better to only allow minor version upgrades:
h5py = "^3.8.0"

ImageAnalysis/image_analysis/tools/lcls_tools_gauss_fit.py
```diff
@@ -0,0 +1,27 @@
...
+
+import numpy as np
+from lcls_tools.common.image.fit import ImageProjectionFit
+
+def gauss_fit(img: np.ndarray, prefix: str = "") -> dict[str, float]:
+
+    result = ImageProjectionFit().fit_image(img)
+
+    # Add prefixes
+    prefix_x = f'{prefix}_gaussian_fit_x_'
```
TAU-Reinier, line 18: Same as in basic_beam_stats, organizing results of stats calculation should be left to the caller, so recommend just outputting {'x': {'param1': 123, ...}, 'y': {...}}

TAU-Reinier (approved):

PR #184, opened 2025-08-15
081125 htu dispersion analyzers

cedoss-lbl:
A PR of some recent work on new analyzers.  Still some future work to be done, but this can be lumped in with a future PR:

features:
- New actions in the action library geared towards closeout.  Tested during high power and it works
- Renamed some actions in the action library to be more clear (no more "tests")
- Created a "common" analyzer for simple line plots.  It is generalizable to any number of given parameters to plot, although after 3 or 4 the plot will likely be too cluttered.
- For this new analyzer, a HTU analyzer was created to automatically plot ICT charge
- Created a new analyzer to automatically analyze and plot ALine3 scans for dispersion signatures.  (*NOTE*: this is implemented with the old "camera image analysis" class, which is deprecated.  A future PR will attempt to migrate all deprecated analyzers to the new Array2DScanAnalysis approach)

### DIFF
--- a/GEECS-Scanner-GUI/scanner_configs/experiments/Undulator/action_library/actions.yaml
+++ b/GEECS-Scanner-GUI/scanner_configs/experiments/Undulator/action_library/actions.yaml
@@ -1,4 +1,36 @@
 actions:
+  Amp4_DUMP_HP:
+    steps:
+    - action: execute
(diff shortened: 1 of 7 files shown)


cedoss-lbl:
I'll add documentation and test with the new PR's soon

PR #254, opened 2026-02-07
Normalization and 1 dplus2 d scanalysis

rjacobBella:
Added Functionality for normalizing 2D images with 4 different normalization options. Added an ICT analyzer subclass to the 1D analyzer class. This subclass ignores the other default filters and pipeline, only allowing for the execution of desired high frequency filtering, sinusoidal BG subtraction, and signal region ID.


skbarber (approved):
Looks good. There are some opportunities for making code more concise, but that is not unique to this PR...

I had to some housekeeping (I had directly committed some changes to master that had formatting issues, which is why this PR was showing as 'checks failed')

Issue #273, opened 2026-03-05
Multi-device analyzer support + DNN Magspec implementation


skbarber:
## Problem

Some diagnostics require images from multiple devices (cameras) per shot to produce a single analysis result. The primary motivating use case is the __HTT magnetic spectrometer__, which uses 4 cameras to image different portions of the dispersed electron beam. These 4 images must be stitched together into a single composite image before energy spectrum extraction.

The current architecture assumes __1 device = 1 image per shot__ at every level:

- `ImageAnalyzer.load_image(file_path: Path)` accepts a single file path
- `ImageAnalyzer.analyze_image(image: Array2D)` accepts a single array
- `SingleDeviceScanAnalyzer` discovers files from one device folder and maps `{shot_num: Path}`
- Config models specify a single `device_name: str`

## Proposed Approach

Generalize the existing architecture to natively support lists of devices/paths, rather than creating a parallel class hierarchy. This keeps the codebase unified and allows any future multi-device diagnostic to use the same infrastructure.

Key ideas:

- __Widen `ImageAnalyzer` base signatures__ to accept `Union[Path, list[Path]]` and `Union[Array, list[Array]]`. The base `load_image` handles lists by iterating over the single-path logic, so all existing subclasses get list support for free.
- __Generalize `SingleDeviceScanAnalyzer`__ to accept `device_names: list[str]` alongside existing `device_name: str`. File discovery scans N folders and groups by shot number. All downstream pipeline methods work unchanged.
- __No renaming__ — keep existing class names. After live testing, a future PR can rename and add deprecation stubs.
- __Fully backward compatible__ — all existing configs, tests, and client code work unchanged.

### PR1: Widen ImageAnalyzer base signatures

Modify `base.py` to accept `Union[Path, list[Path]]` in `load_image` / `analyze_image_file` and `Union[Array, list[Array]]` in `analyze_image`. Add list-handling behavior to `load_image`. Verify existing tests still pass.

### PR2: Generalize SingleDeviceScanAnalyzer for multi-device

Extend file discovery and data loading to support multiple device folders per shot. Update downstream scan analyzer classes and config models to accept `device_names`. Backward compatible with existing `device_name` configs.

### PR3: HTT Magspec image stitching algorithm

Port the MATLAB stitching logic (background subtraction, vignette compensation, rotation, angle remapping, energy binning, concatenation).

### PR4: HTT Magspec analyzer

[...]

PR #276, opened 2026-03-05
Widen image analyzer base signatures

skbarber:
This is the first step in addressing #273. This does not affect any existing code, it only expands the data types that can be passed to the base ImageAnalyzer class. Previously, Array1D and Array2D were the only accepted types. This now permits the use of Lists of these types which paves the way to have ImageAnalyzers handle images from different data sources for a given shot. The first obvious use case is stitching magspec images.


ImageAnalysis/image_analysis/base.py
```diff
@@ -106,23 +121,34 @@ def analyze_image_file(
 
         return self.analyze_image(image, auxiliary_data)
 
-    def load_image(self, file_path: Path) -> Union[Array1D, Array2D]:
+    def load_image(
+        self, file_path: Union[Path, list[Path]]
+    ) -> Union[Array1D, Array2D, list[Union[Array1D, Array2D]]]:
```
SchroederSa: Is there a use-case for mixing 1D and 2D arrays in a list?
If not, I'd simplify the typing:
Union[Array1D, Array2D, list[Array1D], list[Array2D]]

ImageAnalysis/image_analysis/base.py
```diff
@@ -106,23 +121,34 @@ def analyze_image_file(
...
         Parameters
         ----------
-         file_path : Path
+        file_path : Union[Path, list[Path]]
+            A single file path or a list of file paths for multi-device loading.
 
         Returns
         -------
-         image : Union[Array1D,Array2D]
+        Union[Array1D, Array2D, list[Union[Array1D, Array2D]]]
```
SchroederSa: Same as above.

SchroederSa (commented):
Hi Sam,

Thanks for adding this feature!
First, my overview isn't good enough to determine whether any additional modifications are needed to make this work. But since it is mainly a change in typing and documentation, it shouldn't interfere. 
Another thought: the wording image is a little overused. Maybe diagnostic, device, data would be more appropriate at some points?

Thanks,
Sarah

PR #302, opened 2026-04-03
LiveWatch GUI Working Version

### LINKED ISSUE #300: Create a LiveWatch GUI for Configuring+Launching LiveWatch Analysis
The goal is to create a GUI for the configuration and launching of LiveWatch Analyses, providing the user the ability to view/modify and launch LiveWatch analyzers based on the settings offered in the live_watch.ipynb example notebook. These include date, experiment, scan number, displayed logging detail, dry_run, and rerun options. It should be flexible enough to add more functionality as the LiveWatch module grows. If feasible, a packaged app version will be developed as well.

rjacobBella:
Functional version of the LiveWatch GUI


skbarber:
FYI, take a look at the readme for geecs-plugins for how to use pre-commit. I have some code styles and formatting defined and pre-commit help enforce this consistent coding style across the whole project. If those styles are violated, github flags it with the red 'x'. If you install pre-commit and use claude for making commits, it will very quickly learn the style and adhere to it.

ScanAnalysis/LiveWatchGUI/doc_id_lookup.py
skbarber, line 1: My only question here is: is `DocIdLookup` specific to this GUI? This feels a little bit more like a library kind of functionality. It might make more sense to be a part of LogMaker. Or, if I were to be bold, I might argue that `DocIdLookup` makes more sense in geecs-data-utils. 

I hate to stymie progress though. I would say it is good to leave as is, but, create an issue that documents we should move this as a core functionality to geecs-data-utils and then import it from there.

skbarber (commented):
ultimately, everything looks good and if we merge it will be immediately useful. The only thing that stands out is the doc_id_lookup.py. It's the only part of this PR that is implementing core interfacing with data/etc. that doesn't exist natively in any project. It would be better if this was implemented in geecs-data-utils or logmaker. Given logmaker is 'out dated', I would vote for data utils. Could give claude/roo a crack at this comments and see the response

PR #306, opened 2026-04-08
Feat/software interlock

jram860:
# Summary 

* Created a software interlock system using TCP server-client.  
* Added `InterlockServer` class in `utils/geecs_interlock_server.py` as thread-safe TCP server for broadcasting interlock status flags.  
* `register_monitor()` supports registration of custom monitoring function using user-defined logic and real-time flag updates.

* Added `example_camera_interlock_test.ipynb` notebook that demonstrates how to define interlock logic and use the `InterlockServer` to monitor camera thresholds and broadcast status. The notebook includes step-by-step explanations and code cells for practical usage.

* Added `interlock_client.py` as a TCP client that connects to the interlock server, receives status updates, and displays real-time interlock states for debugging and monitoring purposes.


Interlocks/utils/geecs_interlock_server.py
```diff
@@ -0,0 +1,175 @@
+'''
+GEECS Interlock Server
+author: J. Ramirez
+'''
+
+import socket
+import threading
+import time
+from typing import Dict, Callable, Any
```
skbarber: `Any` isn't used. Pre-commit would catch this

Interlocks/utils/geecs_interlock_server.py
```diff
@@ -0,0 +1,175 @@
...
+    Usage:
+        server = InterlockServer(host="IP ADDRESS", port="PORT")
+        server.set_interlock("device1", True)
+        server.set_interlock("device2", False)
+
+        # or register some custom device monitoring functions
+        server.register_monitor("device1", my_custom_check, interval=0.5)
+    """
+
+    def __init__(self, host="192.168.14.14", port=5001):
```
skbarber: probably better to hardcoded the IP address as 0.0.0.0 as a safer default. It's not tragic, but maybe worth reconsidering putting real IP addresses in a public repo.

Interlocks/utils/geecs_interlock_server.py
```diff
@@ -0,0 +1,175 @@
...
+        """
+
+        with self.flags_lock:
+            # Get old state, default to False since we need to initialize.
+            old_state = self.interlock_flags.get(name, False)
+            self.interlock_flags[name] = is_active  # Set new state
+            if is_active != old_state:
+                # Use more descriptive status for logging
+                status = "ACTIVE" if is_active else "NOT ACTIVE"
+                print(f"[{name}] Interlock {status}")
```
skbarber: print statements are gross. It's generally a much better approach to use python's logging module:
```python
import logging
logger = logging.getLogger(__name__)
```

Then you have a whole variety of logging types that can be toggled off/on etc. For example,

```python
logger.INFO(f"[{name}] Interlock {status}")
```
or
```python
logger.DEBUG(f"[{name}] Interlock {status}")
```

Using logger like this also allows very clean, module specific tracebacks which is super helpful when integrated into larger projects.

At a minimum, replace print statements with logger.INFO. but, could also use other flags like logger.WARNING or logger.ERROR

Interlocks/utils/interlock_client.py
```diff
@@ -0,0 +1,75 @@
+"""
+TCP Client for monitoring interlock flag status
+Connects to the interlock server and displays real-time status (for debugging, mostly)
+"""
+import socket
+import time
+import struct
+
+
+def run_interlock_client(host="192.168.14.14", port=5001):
```
skbarber: host="0.0.0.0"

GEECS-PythonAPI/geecs_python_api/controls/interlocks/interlock_client.py
```diff
@@ -0,0 +1,75 @@
...
+                    length_data += chunk
+
+                # unpacl the length
+                message_length = struct.unpack('>I', length_data)[0]
+                # print(f"DEBUG: Expecting {message_length} bytes")
+
+                # Safety check for unreasonable message sizes
+                if message_length > 10000:  # 10KB should be plenty for status messages
+                    print(
+                        f"ERROR: Message length {message_length} is unreasonably large!")
```
skbarber, line 36: use logger.ERROR as noted above

GEECS-PythonAPI/geecs_python_api/controls/interlocks/interlock_client.py
```diff
@@ -0,0 +1,75 @@
...
+                # Now read exactly message_length bytes
+                message_data = b''
+                while len(message_data) < message_length:
+                    chunk = client.recv(message_length - len(message_data))
+                    if not chunk:
+                        print("Connection closed by server")
+                        return
+                    message_data += chunk
+
+                # print(f"DEBUG: Received {len(message_data)} bytes: {message_data}")
```
skbarber, line 50: see, no need to comment out debug print statements if you use logger!

Interlocks/utils/geecs_interlock_server.py
```diff
@@ -0,0 +1,175 @@
...
+
+                time.sleep(0.5)  # Adjust the sending interval as needed
+        except (BrokenPipeError, ConnectionResetError, OSError):
+            pass
+        finally:
+            try:
+                conn.close()
+            except:
+                pass
+            print(f"Client disconnected: {addr}")
```
skbarber: I think you could get away with a simpler:
```python
finally:
    conn.close()
    logger.debug("Client disconnected: %s", addr)
```
but, that doesn't really change anything

skbarber (commented):
I only reviewed the core code at a superficial level. 

Organizationally, we need to discuss some things. As it stands, this would land "Interlocks" as bare code inside of the repo, which is not super useful, or at least not easy to use. That means, Interlocks/ is just a folder of scripts. It has no __init__.py, so Python doesn't know it exists as a module. It has no pyproject.toml, so there's no way to install it as a dependency. This means the only way to use it is to manually add its path to sys.path, which is exactly what the example notebook does:

sys.path.append("c:\\GEECS\\Developers Version\\source\\GEECS-Plugins\\GEECS-PythonAPI")

This is quite brittle. It would make a lot more sense to add interlocks to an existing project with defined dependencies like geecs-python-api. Given you need to import geecs-python-api to use interlocks (as in the example), might as well land it there. Of course, right now, geecs-python-api is severely bloated with a lot of heavy dependencies and deprecated code. It's been my plan to pare geecs-python-api back to the bare essentials (which essentially means GeecsDevice). 

so my suggestion would be to put interlocks in geecs-python-api/controls/interlocks. I don't really see the need for the utils/ folder. I would also add an __init__.py to make it a proper package.

skbarber:
Checking in. What's the status of this PR? I left my review. I saw some changes. Would be nice to either see some responses to the comments before merging. Things move fast in geecs plugins and many Pr merges have happened, so probably need to merge in master before finishing this pr

PR #315, opened 2026-04-11
Add background_scan_number to BackgroundConfig

skbarber:
## Human summary

long overdue PR to allow specifying simply a scan number to use for background. Note, the background is generated by loading images using the `image_analyzer.load_image()` method, so it is generic for all file types. 

The whole implementation of background subtraction is a little bit of a mess with functionality split across ScanAnalysis and ImageAnalysis. I think this could be one of the last new background types to be added so i didn't embark on a larger rearrangement. It's simple, functional implementation for now.

## AI Summary

- Adds `background_scan_number: Optional[int]` to `BackgroundConfig` in `ImageAnalysis`. This field is inert at the image analysis layer — `BackgroundManager` ignores it entirely.
- Adds `_generate_scan_background()` to `SingleDeviceScanAnalyzer` in `ScanAnalysis`. When `background_scan_number` is set, `_resolve_background_paths()` calls this method which:
  - Resolves the background scan's device folder (same date/experiment, different scan number)
  - Checks for an existing cached average at `<bg_scan_analysis_dir>/<device>/<device>_background_avg.npy`
  - If absent: loads all device images from that scan, averages them, and saves to the cache
  - Injects the cache path into `BackgroundConfig` as `FROM_FILE` so `BackgroundManager` takes over normally

Caching in the background scan's own analysis folder means any scan referencing the same background scan number reuses the result on subsequent runs.

## Usage

```yaml
background:
  background_scan_number: 3
```

## Test plan

- [x] Configure a camera config YAML with `background_scan_number` pointing to a valid background scan
- [x] Run analysis — verify background is averaged and cached on first run
- [x] Run again — verify cached `.npy` is used (no re-averaging)
- [x] Confirm `BackgroundConfig` without `background_scan_number` behaves identically to before

### DIFF
--- a/ImageAnalysis/image_analysis/processing/array2d/config_models.py
+++ b/ImageAnalysis/image_analysis/processing/array2d/config_models.py
@@ -83,6 +83,11 @@ class BackgroundConfig(BaseModel):
         Additional constant to subtract AFTER primary background (default 0).
     dynamic_computation : Optional[DynamicComputationConfig]
         Configuration for dynamic background computation (batch processing only).
+    background_scan_number : int, optional
+        Scan number to use as background source. The scan analyzer will load
+        and average all images from that scan's device folder, cache the result
+        as a .npy file, and use it as a FROM_FILE background. Takes precedence
+        over file_path when set.
     """
 
     enabled: bool = Field(True, description="Enable background processing")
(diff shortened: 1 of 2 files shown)


antanthonylu (approved):

skbarber:
merging after testing on HTT for live data analysis for a day

PR #316, opened 2026-04-11
Add LineStitcher for composite multi-segment spectrometers

### LINKED ISSUE #301: HTT magspec stitcher
PR #277 resolved much of the core additions needed to enable HTT magspec analysis via the python analyzer framework.

The original matlab code which was used to implement camera specific charge and energy calibrations is now ported to the manual_magspec_calib_analyzer.py. Analysis of HTT magspec cams can now flow through the usual configuration pipeline. When wrapped into the `Array2DAnalyzer` scan analyzer, the raw camera images are analyzed and results are saved to 'interp' and 'interSpec' folders. The later folder will contain charge per MeV vs energy in tsv format. This is the usual format for magspec data. 

These individual analyzers for each magspec cam can be run individually to generate 4 tsv files per shot. What is needed now is an additional tool to stitch together these 4 tsv files into a single tsv file that is conducive to global energy spectrum analysis (like what is used for htu b-cave magspec).

There are at least two potential approaches: 
 - a new ImageAnalyzer class derived directly from the base class, which does not implement any config driven analysis like which is used for the StandardAnalyzer. This might be the simplest path forward, but, would essentially sit as a standalone 'stitcher'. And then another 1D analyzer should be used to run on the new stitched file. 
 - An explicitly path aware ImageAnalzyer that requires the home directory of one of the HTT-interpSpec folders and can resolve the path to the other three. This is not trivial since the analyze_image method expects a NDArray, not a file path. But, the analyze_image_file() method might be used. Just need to be aware this might be tricky to implement with the existing Array2D and Array1D ScanAnalyzers which leverage the analyze_image method...
 - There may be other ways to handle this...

GOAL: Using the 4 interpSpec tsv files generated each shot, create a new directory with that stiches the four files into a single file that can be used for downstream processing.

skbarber:
## Human Summary

New derived 'LineStitcher' class that is useful for stitching post process mag spec data acquired from multiple cameras. It's a fairly simple derived class that just overwrites the `load_image` method to load data from `sibling_devices`. It concatenates, sorts and interpolates the data. 

closes #301 

## AI Summary

- **`LineStitcher`** (`ImageAnalysis`): new analyzer that stitches TSV data files from multiple devices covering different portions of a shared physical axis (e.g. 4 MagSpec cameras each covering a different energy range). Extends `LineAnalyzer` — the only override is `load_image`, which resolves sibling device paths from the master file path, loads each file, concatenates, and sorts by x. The rest of the pipeline (ROI, interpolation, statistics) runs unchanged.
- **1D interpolation zero-padding** (`ImageAnalysis`): `apply_interpolation` now passes `left=0.0, right=0.0` to `np.interp`, so regions of the output grid outside the data range are zero-filled rather than held at the boundary value. Correct behaviour for density-type spectra (e.g. charge/MeV) when a fixed ROI is wider than the shot-to-shot data coverage.

> **Depends on #315** (`feat/background-scan-number`), which is merged into this branch. The diff will narrow to just this work once #315 lands on master.

## Usage

```python
from image_analysis.offline_analyzers.line_stitcher import LineStitcher

analyzer = LineStitcher(
    line_config_name="HTT-InterpSpec",   # configured for master device
    sibling_devices=[
        "HTT-C23_2_MagSpec2-interpSpec",
        "HTT-C23_3_MagSpec3-interpSpec",
        "HTT-C23_4_MagSpec4-interpSpec",
    ],
)
```

Sibling paths are resolved from the master file path using the GEECS naming convention. Directory suffixes not present in filenames (e.g. `-interpSpec`) are stripped automatically. A missing sibling file raises `FileNotFoundError`, causing the scan analyzer to skip that shot cleanly.

Enable uniform-grid interpolation in the line config YAML for waterfall plot consistency:

```yaml
interpolation:
  enabled: true
  x_min: 50.0
  x_max: 300.0
  num_points: 2000
```

## Test plan

- [ ] Run `LineStitcher` on a single file via `analyze_image_file` — verify all 4 device files are found and concatenated
- [ ] Verify shot is skipped cleanly when a sibling file is missing
- [ ] Run via `Array1DScanAnalyzer` across a full scan
- [ ] Confirm zero-padding behaviour at edges of interpolation range


antanthonylu (approved):

skbarber:
merging after testing on HTT for live data analysis for a day

PR #326, opened 2026-04-17
Config GUI creation

skbarber:
Making a draft PR as it easier to add some general discussion.

### DIFF
--- a/ScanAnalysis/ConfigFileGUI/README.md
+++ b/ScanAnalysis/ConfigFileGUI/README.md
@@ -1,0 +1,44 @@
+# Config File Editor GUI
+
+A PyQt5 GUI for viewing, modifying, and creating device configuration YAML files
+used by the GEECS image analysis system.
+
+> **Note:** This tool is under active development. Features are being added
+> incrementally — see [the plan](../../../plans/Config_File_GUI_Plan.md) for the
+> full roadmap.
+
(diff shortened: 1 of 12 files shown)


skbarber:
I just ran the GUI really quick, in general, it looks great. A couple of general comments:
 - If it's not too much trouble, I would maybe have each subsection be collapse-able so you don't have to scroll through the big long file. 
 - Organizationally, this so far is just an image analysis config GUI, but we have it in the ScanAnalysis project...It's really not the end of the world, but I would like to think bigger picture as to how we want to organize these 'apps'. Separate app directory in each project? One app directory at the root level? I don't really know what's best yet
 - Seeing all of the config models laid out cleanly in a GUI like this really lays bare some of the organizational deficiencies, inconsistencies etc. of the models themselves...I think this will help organize some of those things
 - This is ~4000 lines of code, so, going to be very difficult to actually review. It would be good to make sure it's not too difficult to change/add things as we add new features to the models.

skbarber:
Are you ready to merge this? Might be nice to just start working with it. 

BTW, i thought of an obvious next feature for the GUI: for a given config, there should be an option to analyze/visualize a single file. That way you can iterate the config settings live.

rjacobBella:
Merged! Feedback wanted

PR #335, opened 2026-04-23
Fix stitched MagSpec TSV output location and filename

antanthonylu:
## Summary
- Saves stitched lineout TSV to a new `HTT-C23_MagSpecAnalysis/` folder under the scan directory, instead of writing into the master camera's `interpSpec` folder
- Filename now follows the pattern `Scan{N}_HTT-C23_MagSpecStitched_{shot}.tsv` for clarity and consistency

## Motivation
Previously the stitched output was landing inside `HTT-C23_1_MagSpec1-interpSpec/`, making it hard to distinguish from per-camera data. The new folder keeps the stitched result clearly separate as a scan-level product.

## Test plan
- [x] Run scan analyzer on a local scan with magspec camera folders
- [x] Confirmed `HTT-C23_MagSpecAnalysis/` is created as a sibling to the camera folders
- [x] Confirmed TSV filename matches `Scan{N}_HTT-C23_MagSpecStitched_{shot}.tsv`
- [x] Confirmed per-camera `interpSpec` folders are untouched


ImageAnalysis/image_analysis/offline_analyzers/line_stitcher.py
```diff
@@ -126,3 +130,55 @@ def load_image(self, file_path: Path) -> Array1D:
...
+                self._save_stitched_output(result, Path(file_path))
+            except Exception as e:
+                logger.warning(
+                    "Failed to save stitched output for %s: %s", file_path, e
+                )
+
+        return result
+
+    _OUTPUT_FOLDER = "HTT-C23_MagSpecAnalysis"
+    _OUTPUT_LABEL = "HTT-C23_MagSpecStitched"
```
skbarber: recall, linestitcher is itself not specific to the HTT magspec. It can be recycled for other purposes. So, it would be much better for these `_OUTPUT_FOLDER` etc. 'global' variables to be part of the object constructor (e.g. arguments of the `__init__` method. They could be optionally defined with some place holder names like 'stitched files'.

skbarber (changes requested):
make that one small change then good to go I think. I assume you tested that the proper directory is generated an all?

antanthonylu:
Addressed Sam's review feedback: `_OUTPUT_FOLDER` and `_OUTPUT_LABEL` have been moved from hardcoded class-level constants into `__init__` constructor parameters (`output_folder` and `output_label`), with generic defaults (`"stitched_files"` / `"stitched_data"`) so `LineStitcher` remains reusable for non-HTT use cases.

The HTT-specific names are now passed explicitly at each instantiation point:
- `HTT-MagSpectStitcher.yaml` (livewatch path) via `image_analyzer.kwargs`
- `HTT-magspec.ipynb` and `HTT_magspec_analysis_tests.ipynb` in the constructor call

antanthonylu:
Also confirming that we tested this end to end — the correct directory structure is created as expected, with the stitched TSV landing in HTT-C23_MagSpecAnalysis/ as a sibling to the per-camera interpSpec folders.

skbarber (approved):
next time, don't commit changes to examples unless you really mean to!

PR #336, opened 2026-04-24
Config GUI for Devices and Scans, with Preview Window. Also small tweak to Docgen.py in the google logmaker

rjacobBella:
Added functionality to configure scan/experiments config files. Added a preview analyzer window (reads from memory, doesn't save output). Tweaked docgen.py to reduce duplication issues (probably). More details in push descriptions


skbarber (approved):
looks good enough to me. Given we've tested the GUI a bit, we should probably just deploy it and start to gather feedback.

PR #343, opened 2026-05-05
fix: default Standard1DAnalyzer to ThreadPoolExecutor to avoid BrokenProcessPool

### LINKED ISSUE #341: LineStitcher crashes with BrokenProcessPool when analysis runs for the first time
## Observed behavior

When the HTT MagSpec LineStitcher analyzer runs inside livewatch (after previously being blocked by a `no_data` state due to a directory name mismatch — now fixed), the following error appears in the livewatch log:

```
scan_analysis.analyzers.common.single_device_scan_analyzer - ERROR - Analysis failed for unit ##: A process in the process pool was terminated abruptly while the future was running or pending.
```

The stitcher task is correctly queued and sees the input data, but the analysis fails immediately. No stitched TSV files are created.

## Root cause

`LineStitcher` inherits `run_analyze_image_asynchronously = False` (set as an instance attribute in `Standard1DAnalyzer.__init__`, line 67). This causes `_analyze_units` in `SingleDeviceScanAnalyzer` to use `ProcessPoolExecutor`.

On Windows (where livewatch runs), `ProcessPoolExecutor` uses the `spawn` start method, which creates a **fresh Python subprocess** for each worker. That subprocess must re-import all modules, including `matplotlib.pyplot` (imported at module level in `standard_1d_analyzer.py`, line 21). Importing matplotlib in a spawned subprocess that has no `QApplication` causes the subprocess to terminate abruptly before returning any result — hence `BrokenProcessPool`.

There is a secondary issue: `load_image` sets `self._device_in_filename` as instance state (line 107 of `line_stitcher.py`). With `ProcessPoolExecutor`, this mutation happens only in the worker's copy of the object; a different worker running `analyze_image` would see `None` for that attribute. Threads fix this too, since they share the same object.

This bug was latent because the stitcher always received `no_data` before (due to a separate config directory name mismatch). It became visible only after that config issue was resolved and the stitcher started actually running.

## Proposed fix

Add `run_analyze_image_asynchronously = True` as a **class attribute** on `LineStitcher`. This overrides the instance attribute set by `Standard1DAnalyzer.__init__` and switches the executor to `ThreadPoolExecutor`, which:
- Shares the parent process's already-initialized Qt + matplotlib state (no re-import crash)
- Shares the same object instance (fixes the `_device_in_filename` state issue)
- Is appropriate for LineStitcher's workload, which is I/O-bound (reading TSVs + numpy ops)

**File to change:** `ImageAnalysis/image_analysis/offline_analyzers/line_stitcher.py`

Add before `__init__`:
```python
# Threads (not processes) are required here: ProcessPoolExecutor spawns fresh
# subprocesses that re-import matplotlib and crash when no Qt app is present.
# Threads also share the parent's object state, which matters because load_image
# writes self._device_in_filename for use by analyze_image on the same shot.
run_analyze_image_asynchronously = True
```

## Environment

- Platform: Windows (livewatch server)
- Entry point: `ScanAnalysis/LiveWatchGUI/main.py`
- Analyzer: `LineStitcher` via `HTT-MagSpectStitcher.yaml`

antanthonylu:
## Summary
- Moves `super().__init__()` to the top of `Standard1DAnalyzer.__init__()` so base-class defaults are set before instance attributes, not after
- Changes `self.run_analyze_image_asynchronously` default from `False` to `True`, making all `Standard1DAnalyzer` subclasses (including `LineStitcher`) use `ThreadPoolExecutor` by default

## Why
`ProcessPoolExecutor` spawns fresh subprocesses on Windows that must re-import `matplotlib.pyplot` (module-level import in `standard_1d_analyzer.py`). Without a `QApplication` in the worker process, matplotlib crashes on backend initialization, causing workers to terminate abruptly — the `BrokenProcessPool` error seen in livewatch.

`ThreadPoolExecutor` shares the parent process's already-initialized environment, avoiding the crash. It is also appropriate for `Standard1DAnalyzer`'s workload (I/O-bound file reads + numpy ops that release the GIL).

Moving `super().__init__()` first is the correct pattern: base-class defaults should be established before subclass values override them, not after.

Closes #341

## Test plan
- [ ] CI pre-commit checks pass
- [ ] Deploy to livewatch server and run one HTT MagSpec scan
- [ ] Confirm livewatch log shows `Using ThreadPoolExecutor` for the stitcher task
- [ ] Confirm `HTT-C23_MagSpecAnalysis/` directory is created with stitched TSV files
- [ ] Confirm stitcher task shows `done` in the status dialog

### DIFF
--- a/ImageAnalysis/image_analysis/offline_analyzers/line_stitcher.py
+++ b/ImageAnalysis/image_analysis/offline_analyzers/line_stitcher.py
@@ -41,12 +41,6 @@ class LineStitcher(LineAnalyzer):
         Passed through to LineAnalyzer for metric naming.
     """
 
-    # Threads avoid spawning subprocesses that re-import matplotlib and crash
-    # in GUI contexts where no QApplication exists in the worker process.
-    # Threads also share instance state, so load_image's write to
-    # self._device_in_filename is visible to analyze_image on the same shot.
-    run_analyze_image_asynchronously = True
-
     def __init__(
         self,
         line_config_name: str,
--- a/ImageAnalysis/image_analysis/offline_analyzers/line_stitcher.py
+++ b/ImageAnalysis/image_analysis/offline_analyzers/line_stitcher.py
@@ -41,6 +41,12 @@ class LineStitcher(LineAnalyzer):
         Passed through to LineAnalyzer for metric naming.
     """
 
+    # Threads avoid spawning subprocesses that re-import matplotlib and crash
+    # in GUI contexts where no QApplication exists in the worker process.
+    # Threads also share instance state, so load_image's write to
+    # self._device_in_filename is visible to analyze_image on the same shot.
+    run_analyze_image_asynchronously = True
+
     def __init__(
         self,
         line_config_name: str,
--- a/ImageAnalysis/image_analysis/offline_analyzers/standard_1d_analyzer.py
+++ b/ImageAnalysis/image_analysis/offline_analyzers/standard_1d_analyzer.py
@@ -64,7 +64,6 @@ class Standard1DAnalyzer(ImageAnalyzer):
 
         # Store analyzer state
         self.line_config_name = line_config_name
-        self.run_analyze_image_asynchronously = False
 
         # Storage for metadata from read_1d_data
         self.data_metadata: Optional[Dict[str, str]] = None
--- a/ImageAnalysis/image_analysis/offline_analyzers/standard_1d_analyzer.py
+++ b/ImageAnalysis/image_analysis/offline_analyzers/standard_1d_analyzer.py
@@ -53,6 +53,9 @@ class Standard1DAnalyzer(ImageAnalyzer):
         line_config_name: str,
     ):
         """Initialize the standard 1D analyzer with external configuration."""
+        # Initialize base class first so any defaults it sets can be overridden below.
+        super().__init__()
+
         # Load line configuration
         try:
             self.line_config = load_line_config(line_config_name)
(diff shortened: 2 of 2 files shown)


skbarber (approved):
Looks fine. You could get away with leaving standard1danalyzer untouched by just setting the "asynchronous" part in the constructor of the line stitcher. Generally, that would be the much safer path. If you change a parent class like that, you generally need to check that all other users of that class are not all of a sudden broken. This definitely points us to the more immediate urgency of full integration tests that run all of our defined analyzers on test data...

I looked at the parent class of standard1danalyzer and it already sets has the asynchronous part set to false, so I don't think this will break anything. But, at a minimum, you should go through the examples in image/scan analysis docs and make sure they execute without error.

antanthonylu:
**Implementation pivot based on review feedback**

The original approach added `run_analyze_image_asynchronously = True` as a class attribute on `LineStitcher` and removed the forced `False` instance assignment from `Standard1DAnalyzer.__init__`. While correct, this was narrow — it only fixed `LineStitcher` and left the broader design issue in place.

Per Sam's review comment, the fix is now made at the `Standard1DAnalyzer` level:

1. `super().__init__()` moved to the **top** of `__init__()` — the correct pattern so base-class setup can't stomp values we set afterward
2. `self.run_analyze_image_asynchronously` changed from `False` → `True` — all `Standard1DAnalyzer` subclasses now default to `ThreadPoolExecutor`

This is a more general fix: any current or future subclass of `Standard1DAnalyzer` gets safe thread-based execution without needing to explicitly override the flag. Individual subclasses can still opt back into `ProcessPoolExecutor` by setting the flag to `False` in their own `__init__` after calling `super()`.

PR #439, opened 2026-06-24
Fix camera asset readback filename resolution

skbarber:
## Summary
- resolve existing camera assets from either legacy ScanNNN_Device_001.png filenames or Bluesky/native Device_<acq_timestamp>.png filenames
- allow archived readback callers to pass ACQ_TIMESTAMP explicitly, while preserving SHOT_NUMBER lookup for legacy files
- fall back to timestamp-sorted native camera files so notebook readback works for Bluesky-created scan folders

## Validation
- pytest GeecsBluesky/tests/test_assets.py -q
- ruff check GeecsBluesky/geecs_bluesky/assets GeecsBluesky/tests/test_assets.py
- ruff format --check GeecsBluesky/geecs_bluesky/assets GeecsBluesky/tests/test_assets.py
- pydocstyle GeecsBluesky/geecs_bluesky/assets GeecsBluesky/tests/test_assets.py --convention=numpy
- compileall -q GeecsBluesky/geecs_bluesky/assets GeecsBluesky/tests/test_assets.py
- executed GeecsBluesky/notebooks/external_asset_readback.ipynb with archived lookup disabled
- verified an archived Bluesky camera lookup resolves a native timestamp filename and fills the image as a uint16 array

## Note
When ACQ_TIMESTAMP is omitted for native timestamp files, SHOT_NUMBER is interpreted as a 1-based index into timestamp-sorted files. That keeps the notebook useful during the scanner transition, but exact event semantics should eventually come from the Bluesky/Tiled document stream.

### DIFF
--- a/GeecsBluesky/CHANGELOG.md
+++ b/GeecsBluesky/CHANGELOG.md
@@ -4,6 +4,16 @@ All notable changes to `geecs-bluesky` are documented here.
 
 Format follows [Keep a Changelog](https://keepachangelog.com/en/1.1.0/).
 
+## [0.12.3] - 2026-06-24
+
+### Fixed
+
+- Existing-scan camera asset readback now resolves both legacy
+  `ScanNNN_Device_001.png` files and Bluesky/native `Device_<acq_timestamp>.png`
+  files. The notebook helper accepts an explicit acquisition timestamp or maps
+  a shot number to timestamp-sorted native files when no legacy shot-number file
+  exists.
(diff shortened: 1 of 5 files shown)


skbarber:
Closing as superseded. The folder-only filename fallback is not the desired archive-access path; the next work should use Tiled as the source of truth for shot/event metadata and fill native assets locally from the event-derived acquisition timestamp.

PR #449, opened 2026-07-04
GEECS access layer: CA gateway, CA-only Bluesky backend, optimization as a scan

skbarber:
# The GEECS access layer: CA gateway + CA-only Bluesky backend + optimization as a scan

This PR has grown into the full arc it was always heading toward: **GeecsCAGateway becomes THE GEECS access layer** (wire protocol, DB, PV naming, caproto CA server), **GeecsBluesky becomes a pure EPICS/Bluesky consumer** (direct UDP/TCP backend deleted after verified live parity), and **the GUI's config-driven optimization stack runs on the new backend as a normal scan**. Net diff is negative despite the added features — the deletion paid for everything.

## Arc (in merge order)

1. **GeecsCAGateway package** — caproto soft-IOC serving GEECS devices as PVs (`[Exp:]Device:Variable` readbacks + `:SP` setpoints riding GEECS's blocking set), dynamic from the experiment DB. Deployed on abmx (0.5.1, 83 devices).
2. **CA device family in GeecsBluesky** (`devices/ca/*`, ophyd-async ≥0.19.3 over aioca): triggerable/detector/contributor/snapshot/settable/motor, all consuming the gateway as a *service, never an import*. Verified live equivalent to the direct backend (Scans 007–015, 26_0703/04: NOSCAN, STANDARD, full-output images, multi-role, strict single-shot).
3. **Direct backend deleted** (−2,310 lines); domain tests ported to a hermetic CA-mock harness. `BlueskyScanner` reduced to a thin GUI adapter (990→666 lines) over the headless `GeecsSession`.
4. **Access-layer migration**: `transport/`, `db/`, `pv_naming`, fake device server, and wire-level exceptions moved into the gateway package; dependency flipped (geecs-bluesky → geecs-ca-gateway). Slim deploy verified on abmx.
5. **EPICS citizenship** (gateway 0.5.x): read-only readback PVs (caput → `ECA_PUTFAIL`; verified live), per-device `CONNECTED` PVs (MAJOR alarm when the GEECS link drops), `CAGateway:{VERSION,UPTIME,HEARTBEAT,DEVICES_CONNECTED}` status PVs, long-string path channels, enum label semantics.
6. **Optimization as a scan**: `GeecsSession.optimize` (iteration = bin, one scan number, one Tiled run, schema-v1 rows — no Badger-style side channel) + `geecs_scanner.optimization.session_bridge`, which runs the existing BaseOptimizer/evaluator/ScanAnalysis stack (Xopt 3.1) against session bin rows through the new `EvaluatorDataSource` seam in `BaseEvaluator`. Legacy DataLogger path unchanged and pinned. Full artifact parity: `ScanInfo`, `scan.log` (now capturing the analyzer/evaluator story), s-files, native images, `optimization.json`, `xopt_dump.yaml`.
7. **Live hardening from end-to-end optimization runs** (Scans 001–007, 26_0705):
   - **ScanAnalysis 1.14.0**: shots join to files by the device's own `acq_timestamp` (deterministic ms-canonicalized lookup) when the auxiliary frame carries it — Bluesky-native filenames are timestamp-named; MC shot-number mapping remains the fallback. Makes *any* Bluesky scan folder analyzable by existing tooling.
   - **Stale SMB directory listings** defeated by probing deterministic expected filenames with direct stat (analyzer + bridge pre-analysis wait).
   - **Gateway 0.5.1 data-integrity fix**: the readback monitor deadband no longer inherits the DB `tolerance` (a set-convergence criterion, 0.05 A on magnet PSUs) — it was hiding real sub-tolerance motion from PVs and therefore from recorded scan rows. Verified live. The archiver-volume intent is preserved in `DESIGN.md` (MDEL/ADEL-style plan).
   - Fixed stale lockfiles (GUI + root): a fresh install now yields a working CA-backend environment.

## Versions
GeecsCAGateway 0.5.1 (new package) · geecs-bluesky 0.19.0 · geecs-scanner 0.30.0 · scan-analysis 1.14.0. Full changelogs in each package.

## Verification
- Suites (all green, run 2026-07-05): gateway 66, geecs-bluesky 155, scanner-gui 159, scan-analysis 153.
- Live: acquisition parity scans (007–015 on 26_0703/04), foundation stamp NOSCAN vs gateway 0.5.1, optimization end-to-end with finite image-analysis objectives (Scans 005–007 on 26_0705), deadband fix (sub-tolerance readback tracking).

## Known residuals (documented in `Planning/geecs_session/00_overview.md` → "Open threads")
- Feed Xopt the measured readback instead of the proposal (GEECS set tolerance means they differ by up to ~tolerance).
- Analyzer-device auto-provisioning for optimization scans (legacy merges `device_requirements` into the save set; Bluesky path doesn't yet).
- Per-shot `Objective:`/`Observable:` s-file columns + BAX results-file rooting — parked with the "analysis artifacts in Tiled" design discussion.
- Live re-verifications owed: strict mode post-migration, GUI-launched (PyQt) scan, warm start from `xopt_dump.yaml`.
- Configs repo: migrate pre-DiagnosticConfig flat-schema diagnostic YAMLs (e.g. `UC_Amp2_IR_input.yaml`).
- Composite scan variables on CA (pseudo-positioner design, device layer).
- Ops: `poetry install` in `GeecsCAGateway/` on abmx to refresh the `VERSION` PV metadata label (behavior already 0.5.1).


skbarber:
## Code review — 15 confirmed findings

Max-effort multi-agent review of the full branch diff against merge-base `499b5d12` (12 finder angles → 9 adversarial verification passes → gap sweep; static analysis only). Every item below was **verified**: the trigger and failure were traced in code, not inferred. Ranked most severe first, as a remediation checklist. Lower-confidence and cleanup material is in the collapsed sections at the end.

> **Status (2026-07-06): all 15 findings fixed** in the fix wave `df67c584..d156dc48` on this branch, each with a pinning test; all six package suites green. The first two below-cap items were fixed in the same wave. Deferred material (plausibles, gap-sweeps, cleanups, docs) is recorded with forcing functions in `Planning/geecs_session/00_overview.md`. **Both live checks verified 2026-07-06.** #1: legacy MC re-analysis (Undulator 26_0219 Scan010) — timestamp join attempted, loud fallback, all shots mapped. #2: unreachable free-run reference from the GUI on Windows (first-ever GUI/Windows Bluesky scans) — healthy camera scans normally; off camera aborts loudly at trigger (claimed-folder failure logged, never deleted); dead contributor fails t0-sync loudly. No unpaced rows anywhere. Follow-up commits on the branch: legacy-only probe skip, `GEECS_USE_BLUESKY`, variable-less sync devices, config-driven `[epics] ca_addr_list`, t0-sync stale-device naming.

### Critical — silent data integrity

- [x] **1. Legacy scan re-analysis silently maps zero files** — `ScanAnalysis/scan_analysis/analyzers/common/single_device_scan_analyzer.py:320`
  The acq_timestamp mapping strategy is selected whenever a `<Device> acq_timestamp` column exists, with no fallback — but every legacy GUI-produced scan has both that column (`device_manager.py` force-appends it; `_make_sFile` writes it) **and** shot-number-renamed files (`file_mover`). The timestamp probes never match, `_data_file_map` stays empty, and the analyzer produces nothing for all existing scans. The new tests never cover the ts-column + shot-number-filenames combination.
- [x] **2. Contributor silently promoted to pacemaker → duplicate rows** — `GeecsBluesky/geecs_bluesky/scanner_bridge/bluesky_scanner.py:781`
  `_build_session_devices` skips connect failures without reclassifying roles; if the reference device fails, `detectors[0]` becomes a non-Triggerable `CaTimestampedReadable` that passes the plan's `ShotIdSupport` guard. `trigger_and_read` skips the wait when nothing is Triggerable, so each iteration reads immediately — N unpaced rows of the same cached frame instead of a clean abort.
- [x] **3. UDP replies uncorrelated to commands** — `GeecsCAGateway/geecs_ca_gateway/transport/udp_client.py:301`
  Exe/ACK futures are bare one-shots resolved by the first datagram, and `_parse_exe_response` never compares the reply's variable name to the request. After a >10 s timeout on `set('A')`, A's late reply resolves the next command's future — `set('B')` "succeeds" returning A's value, and B's real reply desyncs subsequent exchanges.

### High — operational failures

- [x] **4. Comma-valued variables dropped from TCP push frames** — `GeecsCAGateway/geecs_ca_gateway/transport/tcp_subscriber.py:33`
  `_VAR_PATTERN` uses `[^,]+` with no escaping; a `localsavingpath` containing a comma never matches, so its PV stays stale forever (one warning at subscribe, silent after), and the fragment after the comma can corrupt the adjacent variable.
- [x] **5. String variables corrupted by float round-trip** — `GeecsCAGateway/geecs_ca_gateway/transport/tcp_subscriber.py:216`
  `_coerce` numerifies numeric-looking strings and discards the raw text: `'007'`→`'7'`, `'1.10'`→`'1.1'`, `'1e5'`→`'100000'` on every CA client and recorded scan row.
- [x] **6. Image-analysis config repoint breaks existing deployments; `image_config_dir` is dead** — `GEECS-Data-Utils/geecs_data_utils/config_roots.py:50`
  `IMAGE_ANALYSIS_CONFIG_DIR` / ini `Paths.config_root` are dropped with no fallback or warning. Separately, ImageAnalysis's loader now resolves through `scan_analysis_config`, so `LiveTaskRunner(image_config_dir=...)` (`live_task_runner.py:122`) and two `__main__` demos set `base_dir` on a manager nothing consults.
- [x] **7. CaMotor's 30 s move budget defeated by the gateway's 10 s exe timeout** — `GeecsCAGateway/geecs_ca_gateway/gateway.py:319`
  A legitimate 10–30 s GEECS blocking set raises `GeecsConnectionError` in `_GeecsSetpoint.write`, failing the caput before the readback-tolerance poll designed for slow axes ever runs — the scan aborts while the hardware completes the move.
- [x] **8. Stop can't abort a scan pre-RunEngine; timed-out join drops the thread ref** — `GeecsBluesky/geecs_bluesky/scanner_bridge/bluesky_scanner.py:358`
  `RE.abort()` on an idle engine raises (swallowed at debug) and `_abort_requested` is never checked before `RE(plan)`; after `join(timeout=15)` the handle is cleared, so `is_scanning_active()` reports False while the scan runs — a second start then hits `RE(plan)` on a running engine, and `reinitialize()` can disconnect devices under the live scan.
- [x] **9. One transient Tiled failure permanently disables persistence** — `GeecsBluesky/geecs_bluesky/tiled_integration.py:70`
  `SafeDocumentCallback` latches `_enabled=False` on the first exception; the subscription lives on the persistent RunEngine and nothing re-enables it, so every subsequent scan silently writes nothing to Tiled and `write_scalar_files_from_tiled` fails until app restart.
- [x] **10. Unbounded `_shot_queue` + no `disconnect()` on any CA device** — `GeecsBluesky/geecs_bluesky/devices/ca/triggerable.py:128`
  The acq_timestamp monitor enqueues every update into a queue only `trigger()` drains — contributors/idle detectors grow it one float per machine shot forever. And no CA device defines `disconnect()` (nor does ophyd-async), so `_disconnect_devices_sync` swallows AttributeError at debug level: per-scan teardown is a silent no-op and stale devices leak monitors across an all-day GUI session.

### Moderate

- [x] **11. Strict-mode fail-fast gaps** — `GeecsBluesky/geecs_bluesky/session.py:467`
  `optimize()` never calls `require_strict_single_shot()`; the validator checks only ARMED, never SINGLESHOT (`shot_controller.py:171`); `over_ca` builds setters with no reachability check (`shot_controller.py:96`). All three cases previously raised `GeecsConfigurationError` pre-scan; now they fail mid-run with quiescence/trigger timeouts after the folder is claimed.
- [x] **12. `_await_bin_assets` blocks the RE event loop; suffixed diagnostics burn the full 10 s per bin** — `GEECS-Scanner-GUI/geecs_scanner/optimization/session_bridge.py:233`
  `time.sleep(0.5)` polling runs inside plan code on the RunEngine's loop thread (CA callbacks, TiledWriter, Stop/pause stall), and the expected paths ignore the registry's `directory_suffix` and the analyzer's `data_device_name`, so FROG/magspec-style devices always hit the timeout.
- [x] **13. One device's UDP bind failure aborts the whole gateway** — `GeecsCAGateway/geecs_ca_gateway/gateway.py:474`
  `connect()` runs before `run()`'s try/finally with no per-device handling; only `EADDRINUSE` is tolerated (`EADDRNOTAVAIL` from the documented empty-local-IP fallback re-raises), exiting with N−1 healthy devices unserved and transports leaked.
- [x] **14. Early exits leave claimed ScanNNN folders without ScanInfo, and leak connected settables** — `GeecsBluesky/geecs_bluesky/scanner_bridge/bluesky_scanner.py:605`
  `claim_scan` creates the folder on disk, VOCS settables open persistent CA monitors, then the no-detector return fires before either the ini write (now inside `session.scan`) or the settables joining `self._detectors` — an empty ScanInfo-less folder plus monitors leaked per attempt. Same folder state from the step-scan/strict raise paths (lines 672–688).
- [x] **15. `optimization.json` is invalid JSON when any objective fails** — `GeecsBluesky/geecs_bluesky/session.py:618`
  Failed objectives record `float('nan')` and `json.dumps` emits bare `NaN` tokens, which jq/`JSON.parse`/strict parsers reject — exactly the runs you'd inspect after a partial failure.

<details>
<summary><b>Confirmed, below the cap</b> (6 lower-severity)</summary>

- ~~`session.py:695` — `Path(None)` TypeError~~ **fixed in the wave**: when passing `scan_number` without `scan_folder` to `scan()`/`optimize()`.
- ~~`session.py:276`~~ **fixed in the wave**: `shot_control({})` → `None.variables` AttributeError (`from_information` returns None for a falsy dict); `shot_control(None)` detaches cleanly.
- `transport/udp_client.py:241` — `{:.12f}` truncates setpoints below ~5e-13 to `0.000000000000`; device confirms 0 with no error. Realistically precision loss more than the exact-zero case.
- `scanner_bridge/bluesky_scanner.py:399` — `_completed_shots` counts every event document; the free-run tail flush makes it N+1 for N shots (display clamped, so low impact today).
- `notebooks/bluesky_hardware_smoke.ipynb` — stale `from geecs_bluesky.db.geecs_db import GeecsDb`; module moved to `geecs_ca_gateway.db`.
- `geecs_data_utils/io/array1d.py:364` — header sniff drops a first row starting with `.` or `+` (pre-existing logic moved from ImageAnalysis, not a regression).

</details>

<details>
<summary><b>Plausible — mechanism verified, trigger unconfirmed</b> (8)</summary>

- `gateway.py:257` — push-frame fan-out doesn't write `acq_timestamp` last (device-payload order, await per write): `trigger()` can complete while the frame's data variables are unposted, pairing a new shot id with the previous shot's values in strict mode. Currently safe only if devices echo subscription order — worth ordering deliberately.
- `channels.py:118` — enum numeric fallback maps a numeric *label* to an *index* when wire text differs from the label after float round-trip (`2.0` for label `'2'` → index 2).
- `transport/tcp_subscriber.py:126` — `.encode('ascii')` on the subscribe command: a non-ASCII variable name raises per supervise attempt, retried forever as a generic connection error.
- `gateway.py:175` — a device variable named exactly `CONNECTED` collides with the status PV in `_register` and raises during `__init__`, killing the whole gateway.
- `assets/tiled_readback.py:776` — `//` collapse corrupts UNC roots in Resource docs; the Filler root map normalizes differently and can't compensate. Live only if a deployment uses UNC rather than a mapped drive.
- `assets/readback.py:242` — drive-root mapping `Z:/` strips to drive-relative `Z:`; known deployments use `Z:/data`, which is unaffected.
- `analysis/camera.py:202` (+ `assets.py:203`) — `int(scan_event_index)` aborts filtered analysis on a None/NaN row; `shot_numbers=[]` is conflated with "no filter". Current-schema runs always carry the int; pre-schema archives are the exposure.
- `single_device_scan_analyzer.py:444` — duplicate acq_timestamps double-map one file to two shots silently; same-millisecond files overwrite in `files_by_ms`. The `valid`-column skip mitigates; strict-mode scans without it are exposed.

</details>

<details>
<summary><b>Gap-sweep candidates</b> (4, self-verified by the sweep pass)</summary>

- `session.py:104` — `GeecsSession()` defaults `tiled=True` and blocks in `from_uri(...)` during construction; off the lab network, building a session hangs until the HTTP timeout.
- `config.py:353` — `subscribed_only=True` omits devices with zero get-variables entirely (no readbacks, no CONNECTED, no save/localsavingpath setpoints), contradicting the `include_settable` docstring's always-exposed control surface.
- `devices/ca/settable.py:65` (family-wide) — PV names never pin `ca://`; with `p4p` importable but `aioca` missing, ophyd-async defaults to PVA and every connect times out with generic errors against the CA-only gateway.
- `devices/ca/triggerable.py:161` — cold-cache race in `trigger()`'s "no blind window" guarantee: when `_last_acq` is None, the baseline get and queue drain happen after the status returns, so the first strict shot after connect/INVALID can be consumed as the baseline and time out.

</details>

<details>
<summary><b>Cleanup (verified) and conventions</b></summary>

- The `{device}_{ts:.3f}{ext}` filename contract now lives in three packages (`assets/registry.py`, `single_device_scan_analyzer.py`, `session_bridge.py`) and is already drifting — only ScanAnalysis has the ±1 ms canonicalization; the bridge hardcodes `.png` and ignores `directory_suffix`. Belongs once in geecs-data-utils.
- `analysis/camera.py` duplicates ~150–180 of its 295 lines from `analysis/assets.py`, where `TiledCameraAsset = TiledGeecsAsset` is a literal alias — could be a ~20-line wrapper over the generic path.
- `assets/tiled_readback.py:240` — one MySQL connect/query/close per event when `device_type` is omitted; the batch loops call it per shot for a constant device. Resolve once per run or cache.
- `config.py:279` — ~166 sequential MySQL connections at gateway startup (two per device); `get_subscribed_variables` in the same file is the one-query precedent to extend.
- `session.py:474` — `optimize()` copy-pastes `scan()`'s run preamble/epilogue; already drifted once (`set_reference` — behaviorally harmless since the plans re-wire it).
- Stale docstring examples import the deleted direct backend (`plans/step_scan.py:28` and ~10 cross-references incl. `ca/settable.py` saying "a future CaMotor" while CaMotor ships in this PR); two identically-named `read_tiled_config` functions in one package; five hand-rolled `config.ini` readers; 0.1 s polling loops in the gateway supervisor and `CaMotor._set_and_wait` where monitors/events exist; per-shot full-table Tiled reads in `load_asset_from_tiled` where the `iter_filled_*` batch iterators already exist.
- Conventions: GeecsCAGateway has no CLAUDE.md (root CLAUDE.md promises one per subpackage); GeecsBluesky/CLAUDE.md omits the new `analysis/` subpackage and still says "as of 0.8.0"; root dependency graph misses the new optional GeecsBluesky→ImageAnalysis edge; `shot_controller.py` plan stubs lack return annotations; bare `Any` in new public signatures. All five packages did bump versions + changelogs correctly.

</details>

Two candidates were refuted during verification and are not listed (a falsy-zero `elapsed_time` guard that is unreachable in both data paths, and the descriptor in-place mutation, which TiledWriter deep-copies before patching).

PR #452, opened 2026-07-07
Strict-shot hardening: frame-complete PV ordering, numeric enum labels, cold-cache trigger race

skbarber:
Three shot-control-adjacent items from the PR #449 max-effort review's "plausible — mechanism verified, trigger unconfirmed" list, now code-verified and fixed. All three gate the owed strict-mode live re-verification.

## Fixes

1. **Gateway push-frame fan-out writes timestamp variables last** (`gateway.py`). PV writes followed device-payload order with an `await` per write, so `acq_timestamp` could post before the frame's data variables — a strict-mode `trigger()` would then read shot N−1's values under shot N's id. Frame completeness is now an explicit ordering guarantee.
2. **`enum_index` matches numeric labels by value, not index** (`channels.py`). `"2.000000"` streamed for label `"2"` resolved to *choice #2* via the index fallback. Resolution is now: exact label → numeric-value match against numeric labels → index interpretation only when no label is numeric (preserving `["Off","On"]` + `"1"` → On). DG645-style configs are full of numeric labels.
3. **Cold-cache race in `CaTriggerable.trigger()`** (`triggerable.py`). The baseline get + queue drain ran inside the coroutine — after `trigger()` returned — so the first strict shot after connect/INVALID could be drained or consumed as the baseline, then time out. On a cold cache any queued positive value is provably the shot (the monitor filters placeholders); the drain is gone and the `0.0` placeholder baseline normalizes to `None`. The narrow residual window is documented.

## Verification

Each new pinning test was confirmed to **fail with its fix reverted** (the cold-cache test reproduces the exact production symptom, `GeecsTriggerTimeoutError`). Suites: GeecsCAGateway 91 passed (package env), GeecsBluesky 194 passed, Scanner-GUI 184, Data-Utils 107. Versions: GeecsCAGateway 0.5.2, GeecsBluesky 0.19.1, changelogs updated.

Follow-up (lab): strict-mode live re-verification, owed since the access-layer migration — these three were its known code-side gates.


skbarber:
Added the two approved riders: (1) `subscribe_tiled` does a 2 s TCP reachability pre-check — off-network session construction warns and proceeds instead of hanging on the Tiled HTTP timeout; (2) all device PVs are pinned `ca://` (verified against ophyd-async 0.19.3's `split_protocol_from_pv`), so an aioca-missing environment fails at construction with a clear install hint instead of timing out via PVA. GeecsBluesky suite: 202 passed (+10 new). CHANGELOG 0.19.1 extended; no extra version bump.

skbarber:
Review follow-up addressed: the cold-cache path now takes **no CA-get baseline at all** (the review's first suggested option — the second, re-checking the queue after the get, can't distinguish "the shot the get already saw" from a monitor replay of the same value, so the shot could still be eaten). `trigger()` drains synchronously on both paths — on a cold cache the drain can only discard a stale CA subscribe replay, which lands at connect time, seconds before any trigger; nothing fires before `trigger()` returns, and the monitor callback shares the RE loop with `trigger()`'s sync body, so no interleave window exists. The first positive update after `trigger()` is the shot: the race is structurally gone, not narrowed. GeecsBluesky suite: 202 passed.

skbarber:
**Strict-mode live verification passed** (2026-07-06, Undulator Scan007, laser-off DG645 config, 2 cameras): ARMED → quiescence confirmed (2.5 s) → 10 plan-owned single shots at ~1 Hz with zero trigger timeouts or retries — including the first shot after device connect on a cold cache, the exact case the race fix on this branch protects. Clean STANDBY disarm and Tiled→s-file export. This was the live re-verification owed since the access-layer migration; both acquisition modes are now verified on the final PR state.

PR #454, opened 2026-07-07
Optimization parity: analyzer-device auto-provisioning + measured-readback-to-Xopt

skbarber:
Two legacy-parity features the maintainer requested for the Bluesky optimization path, both documented open threads in `Planning/geecs_session/00_overview.md` (now marked landed).

## 1. Analyzer-device auto-provisioning

Legacy `ScanManager._initialize_optimization` merges `optimizer.device_requirements` (auto-generated from the evaluator's `analyzers`: `{synchronous: True, save_nonscalar_data: True, variable_list: ["acq_timestamp"]}` per diagnostic device) into the save-device set. The Bluesky path used only the GUI save list — forget to add the objective's camera manually and every iteration finds no data.

`BlueskyScanner._run_optimization` now merges the requirements before building devices: absent devices are added per the template; overlapping devices keep their GUI config and gain only missing variables (legacy semantics). The requirements cross the GUI→bluesky boundary as a duck-typed `bridge.device_requirements` property — same pattern as the existing `on_finish`/`finish`, so `geecs_bluesky` still never imports `geecs_scanner`.

## 2. Measured readback to Xopt

GEECS set convergence is tolerance-bounded (±0.05 A on magnet PSUs), so the evaluated point can differ from the proposal. `observe()` now substitutes each VOCS variable's proposal with the bin-mean measured readback from the current bin's rows (NaN rows excluded, proposal fallback when the column is missing). The model learns from where the hardware actually sat. Tell-frame shape, `xopt_dump.yaml`, and history structure are unchanged.

## Tests / versions

9 new tests across both packages (auto-provision add/merge/absent-attribute; readback substitution/fallback/NaN handling; bridge property population). Suites: GeecsBluesky 195, GEECS-Scanner-GUI 190 — green. GeecsBluesky 0.20.0, GEECS-Scanner-GUI 0.31.0.

**Merge order:** last of the three open PRs (#452 → #453 → this); I'll rebase and reconcile the version/changelog stacking before it lands.

Follow-up (lab): first real GUI optimization campaign — which also covers the warm-start live test.


skbarber:
Two live-driven additions from tonight's first GUI optimization campaign:

1. **Device-name case tolerance** — the diagnostic YAML spelled `UC_Amp4_IR_Input` vs the DB's `UC_Amp4_IR_input`; CA PVs are case-sensitive, so auto-provisioning failed to connect (and duplicated the GUI device). The bridge now canonicalizes requirement names against the experiment DB (best-effort, cached) and the scanner merges case-insensitively with the GUI spelling winning.
2. **Bounded single-shot refire (default 2)** — Scan011 aborted mid-campaign on a dropped camera frame. Forensics: 84 fires vs 83 frames, the missing frame belonged to the aborting fire, fire→frame offset median 0.21 s (max 1.06 s) vs the 3.0 s timeout — a genuine ~1.2% Basler frame drop, not latency. At that rate a 100-shot strict campaign aborts with ~70% probability. `geecs_single_shot` now re-fires with a WARNING naming the frameless device; strict semantics preserved (failed attempts record nothing; orphan frames drained at the retrigger baseline). Abandoned-status behavior verified against bluesky 1.15.1 source.

GeecsBluesky suite: 211 passed.

skbarber:
**Refire live-verified** (2026-07-07, artificially shortened trigger timeout to force first-attempt misses): the WARNING fires naming the frameless device, the shot re-fires, and the scan continues to a clean finish — the exact failure that aborted Scan011 reduced to a log line. With this, everything on the PR is live-verified: auto-provisioning, case tolerance, measured-readback path, and refire.

PR #458, opened 2026-07-07
Bluesky progress events + dead-contributor pre-flight dialog

skbarber:
The two GUI items recommended by the stewardship assessment (`Planning/gui_stewardship/00_overview.md`), now landed.

## 1. Progress bar works in Bluesky mode (the audit's "best value-per-line change")

`BlueskyScanner._on_document` emits a shot-level `ScanStepEvent` per event document — `shots_completed` clamped at `total_shots` (absorbs the free-run tail-flush overcount), schema-v1 `bin_number` as `step_index`. The GUI's existing Block-7 handler consumes it as-is: **zero GUI changes for the bar**, both backends now speak the same event vocabulary. Optimization scans emit progress too. No `DeviceCommandEvent` (no consumer — deliberate).

## 2. Dead-contributor pre-flight dialog (free-run only)

At the pre-claim seam: any synchronous device whose `acq_timestamp` cache is >10 s stale (2 s re-check grace for fresh connects) raises an operator dialog through the **legacy `DialogRequest` channel** — same event bridge, same Qt rendering path, production-tested. Outcomes:

- Stale contributor(s): **Drop && Continue** (disconnected, removed, logged loudly) or **Abort** — pre-claim, no scan number burned.
- Stale **reference**: abort-only v1 (the second button is a clearly-labeled "Try Anyway" that preserves today's fail-loud t0-sync path; no reference promotion — deferred per the doc).
- **All** stale: the message blames the trigger ("may be off / not free-running"), not the cameras.
- Headless (`on_event=None`), geecs_scanner not importable, or dialog unanswered after 30 s: **proceed unchanged** — a scan can never hang on a dialog nobody will answer. `GeecsSession` untouched; dependency direction preserved via defensive imports.

`DialogRequest` gains optional `title`/`continue_label`/`abort_label` (legacy positional construction pinned byte-identical).

## Tests / versions

21 new tests (15 GeecsBluesky, 6 GUI incl. legacy-rendering pins). Suites: GeecsBluesky 226, Scanner-GUI 202 — green. GeecsBluesky 0.21.0, GEECS-Scanner-GUI 0.32.0. Stewardship doc updated to "landed"; stale docstrings fixed per its ledger.

Follow-up (lab): tune the 10 s staleness threshold; live-check the dialog by starting a free-run scan with one camera off — expect the dialog instead of the t0-sync abort observed 2026-07-06.

**Coordination note**: bumps GeecsBluesky — if the gateway-as-a-service branch also touches that package, whichever merges second needs the usual changelog reconcile.

### DIFF
--- a/GEECS-Scanner-GUI/CHANGELOG.md
+++ b/GEECS-Scanner-GUI/CHANGELOG.md
@@ -3,6 +3,32 @@
 All notable changes to this package will be documented here.
 Format follows [Keep a Changelog](https://keepachangelog.com/en/1.1.0/).
 
+## [0.32.0] — 2026-07-07
+
+### Added
+
+- **`DialogRequest` grew optional `title`, `continue_label`, and
+  `abort_label` fields** so a request author can own the dialog wording —
+  used by GeecsBluesky's free-run pre-flight dead-contributor dialog
+  ("Drop && Continue" vs "Abort Scan") emitted through the existing
+  `ScanDialogEvent` channel. Unset fields keep the legacy device-command
+  rendering byte-for-byte; existing positional `(exc, context)`
+  construction is pinned unchanged. See
+  `Planning/gui_stewardship/00_overview.md` §4.
+
+### Changed
+
+- **`show_device_error_dialog` honors the request's custom title/labels**
+  via a new pure helper `_resolve_dialog_content` (requests with custom
+  fields render `str(exc)` verbatim as the body; plain requests keep the
+  device-command message construction). With this plus GeecsBluesky
+  0.21.0's `ScanStepEvent` emission, the Bluesky backend drives the GUI
+  progress bar and operator dialogs with **no changes to the GUI event
+  handlers** (`_handle_scan_event` / `_on_step_event` / `_on_dialog_event`).
+- `engine/dialog_request.py` module docstring updated — it described the
+  removed 200 ms QTimer/`dialog_queue` pattern (flagged in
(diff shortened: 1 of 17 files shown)


skbarber:
Live test round 1 (2026-07-07, Scan006): progress bar **verified working**; the dialog didn't appear because the scan ran in strict mode and the v1 check was free-run only — the OFF camera instead burned all three refires and aborted post-claim, with a misleading "~1% intermittency" warning for a camera that was simply off.

Extended: the pre-flight now runs in **both modes**, with one deliberate asymmetry — strict dialogs only on *differential* staleness (some devices fresh, some not), since all-stale is a legitimate strict pre-scan state (trigger OFF; ARMED starts it) and strict has no pacemaker. Free-run behavior unchanged. Suites: GeecsBluesky 228, GUI 202.

skbarber:
Reworked per maintainer review: **liveness now comes from the gateway's `CONNECTED` status PV, not staleness inference** — CA-connect success never implied device liveness (the gateway serves PVs for every DB device, so an OFF camera connects fine). Pre-flight checks CONNECTED in both modes and dialogs immediately, pre-claim; the staleness check survives only for free-run's trigger-must-be-running requirement (now unambiguous); the strict differential heuristic is gone. Refire is correctly scoped: trigger timeout + device CONNECTED → refire (real frame drop); + device DISCONNECTED → immediate "went down mid-scan" failure, one fire. Fail-open on unreadable status PVs so older gateways never block scans. Suites: GeecsBluesky 234, GUI 202.

PR #460, opened 2026-07-08
Vision M1: geecs-schemas — models, converters, operator docs

skbarber:
First milestone of the target-architecture build (`Planning/vision/00_target_architecture.md` §4). **Targets `feat/vision-v1`, not master** — nothing merges to master until the full vision is accepted.

## What this is

`GEECS-Schemas/` (`geecs_schemas`, v0.1.0, **pydantic-only** runtime dep): versioned models for the five config families — `ScanRequest` (incl. the `per_step:` actions slot — the architecture's acceptance test), `SaveSet`, `ScanVariables` (motors/setpoints/pseudo-composites, numexpr verbatim), `TriggerProfile` (device-agnostic states + explicit `variants` replacing parallel laser-on/off files), `ActionPlan`. Unknown fields are rejected everywhere, so config typos fail at load.

## Converters, tested against the entire real corpus

Every file in the sibling configs repo converts: **71/71 save elements, 77 scan variables, 10/10 shot-control configs, 22 action plans, 12/12 presets, 11/11 optimizer configs** — zero unexplained skips (multiscan presets and one app lookup table deliberately out of scope, enumerated). Both laser-on/off shot-control pairs fold losslessly into single profiles with a `laser_off` variant (round-trip equality asserted).

**Real corpus findings** (the converter as a linter):
- Two Thomson presets reference **stale save-element names** (9 dangling refs, e.g. `D_OAPLeakage` vs actual `D-Diag_OAPLeakage`) — genuine config-repo bugs.
- YAML 1.1 parses an unquoted `OFF:` state key as boolean `False` — the model normalizes it; legacy code depended on files remembering to quote.
- 10 elements carry the deprecated `post_analysis_class` (dropped with notes); one element is action-only (`WedgeInsert`).

## Operator docs as a structural property

Every field of every model (20 models, verified recursively) carries a plain-language operator `description`, **enforced by a CI test** — a future field without operator language fails the build. `docgen.py` renders Markdown reference sections from the models (single source of truth for future GUI tooltips + the manual). Hand-written operator overview at `docs/geecs_schemas/schemas_overview.md` (mkdocs nav registered, build green).

## Safety

Pure addition — no existing package imports this; zero behavior change anywhere. 155 tests (hermetic fixtures + golden snapshots; the corpus walk is `integration`-marked and auto-skips without the sibling checkout). CI gains one per-testpath line.

[...]

PR #465, opened 2026-07-08
vision M3b: actions in scans, multi-axis grids, multi-device triggers

skbarber:
Third engine milestone on the `feat/vision-v1` integration branch (stack: #460 M1 schemas → #463/#464 M2+M3a → this). Closes every documented v1 engine-pending seam in the ScanRequest runner — after this, a ScanRequest with actions, a multi-axis grid, and a multi-device trigger profile executes end-to-end.

## What executes now (was `NotImplementedError`)

- **Actions in scans** — request-level `setup` / `per_step` / `closeout`, SaveSet entry rituals, and ExperimentDefaults, compiled via the M3a action compiler and placed:
  - `setup`: inside the run wrapper, before quiesce/t0-sync and the first step — a failing setup still fires the full finalize chain
  - `per_step`: at every step boundary, after the move, before shots — always with the trigger disarmed (free-run) / quiescent (strict)
  - `closeout`: outermost `finalize_wrapper` — runs even on mid-scan abort (legacy ActionControl parity), always after trigger disarm
- **Multi-axis grids** — outer product, first axis slowest; only changed axes re-move (concurrently); every axis readback lands in every event row; `scan_axes`/`grid_shape` in run metadata
- **Multi-device trigger profiles** — new `ShotControlWrites` (per-state ordered `(device, variable, value)` writes, replayed sequentially in declared order); the legacy single-device `ShotControlConfig` path is byte-identical and untouched
- **Production action-signal factory** — `CaActionSignalFactory`: str-typed `:SP` signals with wire-string coercion, prefetched on the RE loop pre-claim so unreachable action targets fail fast before a scan number is claimed

## One design decision that changes a schema (flagging for review)

Closeout ordering is now **mirrored/LIFO** — setup runs defaults → entry rituals → request; closeout runs request → entries → defaults, nesting like context managers. This changed the `ExperimentDefaults` closeout merge rule from *prepend* to *append* → **GEECS-Schemas 0.2.0** with changelog rationale.

## Deliberately still refused (loudly, documented)

Pseudo scan variables, `all_scalars`, optimize without an injected objective/suggester, actions/rituals on optimize-mode requests (`GeecsSession.optimize` has no hooks yet), GUI-bridge staging of actions/multi-axis (points to `GeecsSession.run`; GUI submission is a later milestone).

## Tests

~150 new hermetic tests across runner slot-assembly/dedup/ordering, message-level grid plans, RE-level composition (setup→per-step→closeout incl. abort paths), ordered multi-device transitions, and the zero-refusal end-to-end fake-session flow. Plus `tests/test_scan_request_hardware.py` (integration-marked, CI-deselected, `GEECS_HW_*` env knobs) — the standing Gate-2 hardware script.

Suites: GeecsBluesky **245 passed / 12 skipped** (package env, CI parity) and **350 passed** (root env); GEECS-Schemas **196**; GEECS-Scanner-GUI **207**. Versions: GeecsBluesky 0.23.0 (changelog also folds in the deferred M3a entry), GEECS-Schemas 0.2.0.

Gate-1 context: this branch's base already passed live hardware smoke (Scan013 = first ScanRequest-driven scan; first live compiled ActionPlan cycled the Amp4 dump). Gate-2 = running the new hardware test with actions-in-scan on metal.

### DIFF
--- a/GEECS-Schemas/CHANGELOG.md
+++ b/GEECS-Schemas/CHANGELOG.md
@@ -5,6 +5,23 @@ All notable changes to GEECS-Schemas are documented here.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [0.2.0] - 2026-07-07
+
+### Changed
+
+- **`ExperimentDefaults` merge rule: closeout is now mirrored** (ratified
+  with the M3b action-execution milestone). Default *setup* plans still run
+  before the scan's own, but default *closeout* plans now run **after** the
+  scan's own — teardown mirrors setup, so the four setup layers (vision doc
+  §4.4b) nest like context managers (defaults → save-set entry rituals →
+  the scan's own on the way in; exact reverse on the way out) and an
+  experiment-wide "return the machine to standby" always runs last.
+  Previously the module documented defaults-first for both slots.
+  Documentation-contract change only (the merge itself is implemented by
+  resolvers, e.g. `geecs_bluesky.scan_request_runner.apply_experiment_defaults`,
+  updated in lockstep); field descriptions and the operator-language test
+  pin the new ordering.
+
 ## [0.1.0] - 2026-07-07
 
 ### Added
--- a/GEECS-Schemas/geecs_schemas/experiment_defaults.py
+++ b/GEECS-Schemas/geecs_schemas/experiment_defaults.py
@@ -7,15 +7,25 @@ mention them when it wants something *different*.  You would edit it when
 the experiment's routine changes — a new standard trigger profile, a new
 "always run this first" checklist.
 
-The merge rule, in plain terms: **defaults run first, then the scan's own.**
-A default trigger profile is used only when the scan names none; default
-setup plans run before the scan's own setup plans, and default closeout
-plans run before the scan's own closeout plans.
+The merge rule, in plain terms: **defaults run first on the way in and last
+on the way out.**  A default trigger profile is used only when the scan
+names none; default setup plans run before the scan's own setup plans, and
+default closeout plans run *after* the scan's own closeout plans — teardown
+mirrors setup, so the experiment-wide baseline is the outermost bracket
+around every scan.
 
 Developer notes
 ---------------
 There is no legacy YAML dialect behind this model — the legacy scanner kept
 these choices in GUI state — so there is no converter for it.
+
+The mirrored closeout ordering is deliberate (ratified with the action
+execution milestone): the four setup layers (vision doc §4.4b) nest like
+context managers.  On the way in the order is defaults → save-set entry
+rituals → the scan's own setup; on the way out it is the exact reverse —
+the scan's own closeout → entry rituals → defaults.  A defaults closeout
+like "return the machine to standby" therefore always runs last, after
+every scan-specific cleanup has finished.
 
 Resolvers MUST record the defaults they applied into the resolved request
 (provenance): a run's metadata has to show the trigger profile and action
@@ -35,8 +45,9 @@ from geecs_schemas._base import SchemaModel, VersionedSchemaModel
 class DefaultActions(SchemaModel):
     """The action plans every scan of the experiment runs by default.
 
-    These prepend to whatever the scan itself asks for: defaults run first,
-    then the scan's own plans.
+    These bracket whatever the scan itself asks for: default setup runs
+    first, then the scan's own plans; default closeout runs last, after the
+    scan's own — teardown mirrors setup.
     """
 
     setup: list[str] = Field(
@@ -49,8 +60,9 @@ class DefaultActions(SchemaModel):
     closeout: list[str] = Field(
         default_factory=list,
         description=(
-            "Names of action plans to run after every scan, ahead of any "
-            "closeout plans the scan itself lists."
+            "Names of action plans to run after every scan, after any "
+            "closeout plans the scan itself lists (teardown mirrors setup: "
+            "these are the outermost bracket)."
         ),
     )
 
@@ -61,8 +73,8 @@ class ExperimentDefaults(VersionedSchemaModel):
     Declares the trigger profile and routine action plans most scans share,
     so individual scan requests stay short.  Defaults never override what a
     scan says explicitly: a default trigger profile applies only when the
-    scan names none, and default plans run first, followed by the scan's
-    own.
+    scan names none, and default plans bracket the scan's own — defaults
+    run first on setup and last on closeout.
     """
 
     trigger_profile: Optional[str] = Field(
@@ -75,8 +87,9 @@ class ExperimentDefaults(VersionedSchemaModel):
     actions: DefaultActions = Field(
         default_factory=DefaultActions,
         description=(
-            "Action plans every scan runs by default — these run first, "
-            "then the scan's own plans."
+            "Action plans every scan runs by default — setup plans run first "
+            "(before the scan's own), closeout plans run last (after the "
+            "scan's own)."
         ),
     )
     apply_db_scan_defaults: bool = Field(
--- a/GEECS-Schemas/pyproject.toml
+++ b/GEECS-Schemas/pyproject.toml
@@ -1,6 +1,6 @@
 [tool.poetry]
 name = "geecs-schemas"
-version = "0.1.0"
+version = "0.2.0"
 description = "Versioned Pydantic schemas for GEECS scanner configs (scan requests, save sets, scan variables, trigger profiles, action plans) plus legacy-YAML converters."
 authors = ["Sam Barber <sbarber@lbl.gov>"]
 readme = "README.md"
--- a/GEECS-Schemas/tests/test_experiment_defaults.py
+++ b/GEECS-Schemas/tests/test_experiment_defaults.py
@@ -71,11 +71,20 @@ class TestExperimentDefaults:
 
     def test_merge_rule_documented_in_operator_language(self):
         # The resolver contract is part of the schema's documentation:
-        # defaults run first, then the scan's own.
+        # defaults run first on setup and last on closeout (teardown
+        # mirrors setup — ratified with the action-execution milestone).
         doc = inspect.getdoc(ExperimentDefaults) or ""
         module_doc = inspect.getmodule(ExperimentDefaults).__doc__ or ""
         assert "defaults run first" in (doc + module_doc).lower()
+        assert "mirrors" in (doc + module_doc).lower()
         actions_field = ExperimentDefaults.model_fields["actions"]
         assert "run first" in (actions_field.description or "")
+        assert "run last" in (actions_field.description or "")
+        # The mirrored closeout is pinned on the closeout field itself, so
+        # an editor tooltip states the ordering an operator will observe.
+        from geecs_schemas.experiment_defaults import DefaultActions
+
+        closeout_field = DefaultActions.model_fields["closeout"]
+        assert "after any" in (closeout_field.description or "")
         # provenance requirement is stated for resolver implementers
         assert "provenance" in module_doc
--- a/GeecsBluesky/CHANGELOG.md
+++ b/GeecsBluesky/CHANGELOG.md
@@ -115,6 +115,21 @@ Both items were found during Gate-2 hardware verification of this release
   `GeecsSession.optimize`'s adaptive scan still enables saving eagerly
   (known remaining window — the step-scan paths were the hardware-verified
   offenders).
+  **Tail closed too** (Gate-2 re-verify on 26_0708: strict Scan001 was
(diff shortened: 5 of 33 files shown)


skbarber:
**Gate-2 hardware verification passed tonight (over VPN), on this branch's exact code:**

- **Scan014** — the branch's own `tests/test_scan_request_hardware.py` (strict noscan via ScanRequest, `Amp4In` + `HTU-LaserOFF` converted on the fly, TriggerProfile→ordered-writes path): **1 passed**. One find: the test's default save-set name was the device name (`UC_Amp4_IR_input`) instead of the corpus element (`Amp4In`) — the pre-claim fail-fast rejected it exactly as designed; default fixed in 2ef2800b.
- **Scan015** — **first scan ever with ActionPlans executing in all three slots**: `closeout_amp4` (set `DO.Ch9='off'` → wait 3 → check `DI.Ch17=='off'`, the gaia-free corpus plan, physically verified by the limit switch each time) bound as setup + per_step + closeout around a strict 3-shot noscan. The log timeline confirms placement: setup before `ARMED`, per_step after quiescence / before the shots, closeout after the `STANDBY` disarm. Images + Tiled + s-files on disk.

skbarber:
**Laser-on verification (Scan016)** — completes the Gate-2 timing story across both trigger regimes, same branch code:

- `HTU-Normal` + oscillator on, strict noscan, 5 shots: **5 frames, spacings 1.000 / 1.000 / 0.999 / 1.001 s** — external single-shot phase-locked to the real 1 Hz timing edge.
- Contrast with laser-off (Scan014/015): `SINGLESHOT` is an internal software fire there, so shots return at the plan's own cycle time (~0.2 s). Strict mode is plan-paced by design; rep rate only paces free-run.
- Scan015's 6-images-for-3-rows is understood: native saving turns on before arming, so the ~5 s setup-action window saved three 1 Hz STANDBY free-run frames that join to no event row (orphans). Scan016 had no setup action → exact 1:1 frames-to-rows.

**Two follow-ups agreed for M3c scope** (not this PR):
1. Strict mode: enable native saving *after* arming (and disable before the final disarm) so pre-arm free-run frames never land in the scan folder.
2. Headless `GeecsSession` runs should attach a per-scan `scan.log` — today only the GUI bridge does (`BlueskyScanner._scan_log`), so ScanRequest scans 013–016 have no scan.log.

skbarber:
**Follow-up fixes hardware-verified (26_0708, over VPN)** — both Gate-2 findings are now fixed on this branch and re-verified on metal:

| Scan | Config | Before | After |
|---|---|---|---|
| Scan001 | strict, actions in all 3 slots (= Scan015 rerun) | 6 images / 3 shots | **3 / 3 exact** |
| Scan002 | free-run 5-shot (= Scan013 rerun), on 685f2b10 | — | 7 / 5: exposed the *tail* leak (STANDBY passes edges; save-off sat behind tail flush + Tiled writes) |
| Scan003 | same free-run config, on a82a5e77 | — | **5 / 5 exact**, state sequence `OFF → SCAN → STANDBY → OFF → STANDBY` |

`scan.log` now appears in headless `session.run()` scans and made both diagnoses trivial (scan-id-stamped timelines above came straight from it). Between-step STANDBY frames in multi-step free-run scans are documented as accepted legacy-parity behavior — deliberately not per-step save toggling.

Suites after both fixes: GeecsBluesky 364 (root) / 255+13skip (package-env CI parity), GEECS-Schemas 196, GEECS-Scanner-GUI 207.

skbarber:
## Automated code review — findings

Ran a multi-angle review (correctness / removed-behavior / cross-file / reuse / simplification / efficiency / altitude / conventions), deduped, and verified each survivor against the branch head. Ranked most-severe first. **Findings 1 and 2 are functional bugs with realistic triggers; the rest are edge-cases, metadata, and cleanup.** Verdicts: **CONFIRMED** = constructible from the code; **PLAUSIBLE** = reachable on a realistic-but-not-default path.

Nice work overall — this is a large, well-tested change with refusals documented. These are the things worth a second look before merge.

---

### 1. Free-run scan with a saving detector but no shot controller re-opens the orphan-frame leak — **CONFIRMED**
`GeecsBluesky/geecs_bluesky/plans/free_run_step_scan.py:217`

`enable_saving` is derived purely from `saving_detectors`, independent of `controller` (`orchestration.py:116`). When `controller=None`, the arm/disarm/quiesce hooks are all `None`, so `_quiesce` is `None` and nothing ever stops the free-running external trigger — yet `enable_saving()` still turns saving on. Frames get saved during t0-sync and every move: the Gate-2 Scan002 leak (7 images for 5 shots) the windowing was written to prevent, un-caught on this path.

### 2. Any experiment with a default action plan can no longer run *any* optimize-mode ScanRequest — **CONFIRMED**
`GeecsBluesky/geecs_bluesky/scan_request_runner.py:1251`

`resolve_defaults_for` (line 1237) merges `experiment_defaults` setup/closeout into `request.actions` **before** the optimize refusal, with no mode gating. The refusal at line 1251 then inspects `resolved_actions` (the merged set), not the request's own actions. So once an `experiment_defaults.yaml` declares any default setup/closeout plan, **every** OPTIMIZE-mode request raises `NotImplementedError` even when the submitter set no per-request actions. The check should inspect the request's *own* actions, or optimize mode should skip defaults injection.

### 3. The quiesce that closes the trigger is a silent no-op when it resolves to STANDBY or the profile has no OFF state — **PLAUSIBLE**
`GeecsBluesky/geecs_bluesky/plans/free_run_step_scan.py:214`

Two related paths: (a) `_quiesce = quiesce_trigger or disarm_trigger` falls back to `disarm_trigger` (STANDBY), and per the DG645 model, SCAN/STANDBY both pass external rising edges — only OFF is single-shot — so a direct caller of the public plan passing only `disarm_trigger` gets a quiesce that doesn't stop the trigger. (b) If a profile defines no OFF state, `set_state(OFF)` writes and logs nothing (`shot_controller.py:226`) — a silent no-op — while saving is enabled. Nothing enforces that a free-run profile defines OFF (`require_strict_single_shot` only checks ARMED/SINGLESHOT).

### 4. Bare `except Exception` in `_LazyResolverRegistry` masks real resolver faults as "plan not found" — **CONFIRMED**
`GeecsBluesky/geecs_bluesky/scan_request_runner.py:813`

`__getitem__`/`get` wrap `resolve_action_plan()` in a bare `except Exception`, converting any internal fault (AttributeError, TypeError, transient IO) on a valid nested run-step plan into a miss. Combined with `__iter__`/`__len__` reporting empty, the downstream `ActionPlanNotFoundError` lists zero candidates — misdirecting debugging away from the actual exception. Catch only the "not found" case; let unexpected errors propagate.

### 5. Multi-axis grid via a direct `session.scan()` call writes "Step size = 0" into ScanInfo — **CONFIRMED (metadata-only)**
`GeecsBluesky/geecs_bluesky/session.py:893`

`_write_scan_info` reduces grid tuples to their outer value (`[(0,4),(0,5),(1,4),(1,5)]` → `[0,0,1,1]`) then defaults `step = real[1]-real[0] = 0` because the inner axis varies fastest. `session.scan(motor=[outer,inner], positions=grid)` is a documented public API and `scan_info` defaults to `None`, so a direct caller records `Step size = 0` in `ScanInfoScanNNN.ini`. Masked on the `run_scan_request` path only because it injects an explicit step override. Execution is unaffected (grid truth is in `scan_axes`/`grid_shape`); the legacy ini is what's wrong.

### 6. `assemble_action_slots` recovers the defaults/own boundary by index arithmetic on already-merged lists — **CONFIRMED (fragility)**
`GeecsBluesky/geecs_bluesky/scan_request_runner.py:775`

It counts provenance entries (`n_setup_defaults`, `cut = len(merged_closeout) - n_closeout_defaults`) and slices the merged lists to splice save-set rituals into the middle — un-merging by count what `apply_experiment_defaults` merged. No assertion validates the boundary, so the moment that merge de-duplicates or changes prepend/append order, rituals splice into the wrong layer silently (e.g. a closeout ritual running after the experiment-default teardown it should precede). Consider carrying the three layers (defaults / rituals / own) separately and composing in nesting order.

### 7. `CaActionSignalFactory.disconnect` clears caches but never releases the connected CA channels — **PLAUSIBLE (bounded)**
`GeecsBluesky/geecs_bluesky/devices/ca/action_signals.py:152`

`disconnect` only does `self._settables.clear()` / `self._readables.clear()`. The factory is prefetched/connected pre-claim and placed in the runner's `created` cleanup list, so on a failed `session.scan` the "disconnect" leaves aioca channels open. Bounded rather than unbounded — aioca keys channels by PV name, so repeated identical failed scans reuse them — but the channels persist past the factory's lifetime with no teardown.

### 8. `session.shot_control` silently detaches on an empty-states `ShotControlWrites` — **CONFIRMED (low)**
`GeecsBluesky/geecs_bluesky/session.py:320`

`if config.states:` is falsy for a hand-built `ShotControlWrites(states={})` (the field has `default_factory=dict`, no non-empty validator), so the controller stays `None` and detaches with only an info log. A subsequent strict-mode scan then raises the generic "requires shot_control(...) first" rather than a profile-specific error. Production is safe (`trigger_writes_from_profile` guarantees non-empty), so this only bites a hand-built caller.

### 9. `move_changed_axes` gates re-moves on exact float equality — **PLAUSIBLE (latent)**
`GeecsBluesky/geecs_bluesky/plans/step_scan.py:139`

`target != prev` is safe *today* only because `itertools.product` reuses the identical outer-axis value object across inner iterations. The plan's docstring advertises accepting arbitrary floats/tuples, and nothing enforces product-origin, so a future caller building points arithmetically (e.g. `np.linspace` yielding `0.30000000000000004` vs `0.3`) would needlessly re-command a stationary axis every inner step, or skip an intended reset of equal-looking values. A tolerance-based compare would harden it.

### 10. Dead code left behind — **CONFIRMED (trivial)**
`GeecsBluesky/geecs_bluesky/session.py:749` — an unreachable `return self._last_run_uid, history` sits immediately after a `with scan_log(...)` block whose last statement already returns. And `collect_save_set_action_names` (`scan_request_runner.py:940`) has no non-test callers — superseded by `collect_save_set_rituals`. Both are safe to delete.

---

### Checked and cleared
- **Scan-folder invariant**: `scan_log.py` correctly checks `folder.is_dir()` and skips (never creates) — no violation; a test pins it.
- **Versioning/CHANGELOG**: both GeecsBluesky (0.23.0) and GEECS-Schemas (0.2.0) bumped with matching entries.
- **Refuted candidates**: the double end-of-scan quiesce (deliberate Gate-2 fix, guarded, tests assert two); the empty-string `:SP` write (two independent guards); `describe_target` AttributeError (unreachable + None-guarded); `CaSettable`/`CaPutSetter` "duplication" (datatypes genuinely differ — str-typed vs float-typed); the sequential per-write shot-control path (documented ordering-correctness tradeoff, not a defect).

<sub>🤖 Generated with [Claude Code](https://claude.com/claude-code)</sub>

skbarber:
**Review fixes pushed (`90ac47ef`).** Addressed findings 1, 2, 4, 10 — the two functional ones plus the two cheap cleanups. Verified each finding against branch head before fixing.

- **#2 (optimize + default actions) — fixed per design decision.** `apply_experiment_defaults` merges default setup/closeout into the request *before* the optimize check, so the old refusal would have blocked every optimization the moment an experiment defines default bracket actions (a future `experiment_defaults.yaml`), even with zero per-request actions. Since optimize has no action hooks yet, the actions (request + defaults + save-set rituals) are now **skipped, logged (WARNING), and recorded in run metadata** under `skipped_action_plans` — never silently. Unknown names still fail fast. (Decision: optimization should run; running safety actions *during* optimization is a new capability not present in the legacy scanner, so deferring it is fine.)
- **#1 (free-run saving without a controller) — warn at build time.** Windowing relies on the controller's quiesce to stop the trigger; with no controller there's no such point, so t0-sync/move frames may be orphaned. `build_step_scan_plan` now warns loudly rather than silently leaking or refusing the supported controllerless config.
- **#4 (registry masks faults) — fixed.** `_LazyResolverRegistry` now catches only `GeecsConfigurationError` (the genuine miss); unexpected resolver faults propagate instead of becoming a candidate-less "plan not found."
- **#10 (dead code) — removed.** Unreachable `return` in `GeecsSession.optimize`; superseded `collect_save_set_action_names`.

New tests: optimize-skip (actions + rituals recorded/logged), free-run controllerless warning (+ the no-warning-with-controller control), registry fault propagation. Suites green: GeecsBluesky **367** (root-env from package dir) / **258 + 13 skip** (package-env CI parity), GEECS-Schemas **196**. Folded into the unreleased 0.23.0 changelog (no version bump).

Findings 3, 5–9 (free-run OFF-state enforcement, grid Step-size-0 in ScanInfo, the `assemble_action_slots` index-arithmetic fragility, factory channel release, empty-states detach, float-equality axis compare) are accurate but non-blocking — deferring to a fast-follow; #6 (the slot-splice fragility) is the one I'd prioritize there.

PR #466, opened 2026-07-08
Add CA alarm limit overlay

skbarber:
## Summary

- add optional DB-backed `ca_alarm_limits` overlay for scalar CA value alarms
- apply LOLO/LOW/HIGH/HIHI severity/status from the curated overlay as the single value-alarm evaluator
- keep DB `min`/`max` as display metadata only; they are not interpreted as alarm limits
- do not export curated alarm thresholds as native DBR_CTRL alarm/warning metadata, avoiding caproto native limit-alarm conflicts
- add deployment SQL plus docs for applying the table and smoke-testing one configured row

## Review follow-up

- fixed native caproto limit-alarm interference by removing threshold metadata from channel construction and bypassing caproto numeric verification for overlay-managed writes
- removed separate steady-state `alarm.write()` fan-out by writing the value and computed alarm metadata together
- fixed HIHI/HIGH and LOLO/LOW hysteresis step-down behavior
- made missing table rollout fail-open at INFO, while unexpected DB lookup failures log at WARNING
- removed the optional overlay FK into the core `device` table
- added regression tests for high-only rows, hysteresis behavior, DB rollout errors, and the absence of native caproto threshold metadata

## Validation

- `poetry run pytest tests/test_channels.py tests/test_config_from_db.py tests/test_geecs_db.py tests/test_gateway.py -q` -> `76 passed`
- commit hooks passed `ruff`, `ruff-format`, and `pydocstyle`
- live gateway smoke test on this branch:
  - DB row: `Undulator`, `U_S1H`, `Current`, `lolo=-4.5`, `low=-2.5`, `high=2.5`, `hihi=4.5`, severities `MAJOR/MINOR/MINOR/MAJOR`, `hysteresis=0.05`
  - with `U_S1H:Current = 3 A`, CA/Phoebus reported `HIGH` / `MINOR_ALARM`, as expected

## Notes

- This intentionally keeps alarm validation local to `GeecsCAGateway`; it does not add a `geecs-schemas` dependency.
- The table is optional during rollout. If it is absent, the gateway starts with no curated value alarms.
- Derived-channel schema/evaluator work is in a separate branch and is not part of this PR.

### DIFF
--- a/GeecsCAGateway/CHANGELOG.md
+++ b/GeecsCAGateway/CHANGELOG.md
@@ -3,6 +3,27 @@
 All notable changes to `geecs-ca-gateway` are documented here, following
 [Keep a Changelog](https://keepachangelog.com/en/1.1.0/) and semantic versioning.
 
+## [0.7.0] - 2026-07-08
+
+### Added
+
+- **Optional `ca_alarm_limits` MySQL overlay for curated scalar value alarms.**
+  Enabled rows are keyed by `(experiment, device, variable)` and provide
+  `LOLO` / `LOW` / `HIGH` / `HIHI` thresholds with configured severities.
+  The table is optional during rollout: if it is absent or unreadable, the
+  gateway starts with no value alarms.  Rows attach only to served numeric
+  readbacks; DB `min` / `max` remain display metadata and are never treated as
+  alarm limits.
+- `deploy/ca_alarm_limits.sql`, a non-destructive DDL snippet for creating the
+  optional alarm-limit table. Dropping the table restores pre-0.7 behavior.
+
+### Changed
+
+- Live numeric readback writes now recompute curated value-alarm severity.
+  Device disconnects still take precedence: stale readbacks remain
+  `INVALID/COMM` until the next live frame, which then reports either
+  `NO_ALARM` or the active value alarm.
+
 ## [0.6.0] - 2026-07-07
 
 ### Added
--- a/GeecsCAGateway/PV_CONTRACT.md
+++ b/GeecsCAGateway/PV_CONTRACT.md
@@ -308,20 +308,26 @@ What severity means **today**:
 
 | Condition | PV | Severity / status |
 |---|---|---|
-| Device TCP subscription live | readbacks | `NO_ALARM` (plain value writes reset severity) |
+| Device TCP subscription live, no configured value alarm active | readbacks | `NO_ALARM` |
+| Device TCP subscription live, configured scalar limit crossed | that numeric readback | configured severity (`MINOR_ALARM`, `MAJOR_ALARM`, or `INVALID_ALARM`) with status `LOW`, `LOLO`, `HIGH`, or `HIHI` |
 | Device TCP subscription dropped / unreachable | all of that device's **readback** PVs | **`INVALID_ALARM`**, status `COMM` — data is stale, last value retained |
 | Same event | that device's `CONNECTED` PV | value `Disconnected`, **`MAJOR_ALARM`**, status `COMM` |
-| Recovery (first live frame / reconnect) | both | automatic return to `NO_ALARM` |
+| Recovery (first live frame / reconnect) | both | automatic return to live-value severity (`NO_ALARM` unless a configured value alarm is active) |
 
 So: **INVALID on a readback means "not live", not "bad value"** — the held
 value is the last known good one. `CONNECTED` is the explicit liveness signal
 for Phoebus/alarm layers; prefer it over inferring liveness from data severity.
 
+Value-based alarms are an explicit curated overlay from the optional
+`ca_alarm_limits` MySQL table, keyed by `(experiment, device, variable)`.
+Only non-null threshold columns apply. The gateway validates rows at startup,
+attaches them only to served numeric readbacks, and treats the table as optional:
+if it is absent during rollout, the IOC starts with no value alarms. DB `min` /
+`max` remain display metadata and are **never** interpreted as alarm limits.
+
 **Not yet implemented** (be honest with your displays):
 
-- No value-based alarms: DB min/max are display limits only; there are no
-  HIHI/HIGH/LOW/LOLO thresholds, so a readback never goes `MINOR`/`MAJOR` on
-  value. A NaN or out-of-range value arrives as a plain `NO_ALARM` update.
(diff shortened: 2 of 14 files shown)


skbarber:
**Review (agent-assisted; findings verified against the branch code, the critical one reproduced empirically against caproto 1.3.0 through the real gateway fan-out path). Verdict: needs changes before merge — one critical.**

The core design is right — curated opt-in table, DB min/max correctly kept display-only, fail-open on missing table, pure in-memory evaluation in the hot path, frame-ordering guarantee intact, per-channel alarms not shared with `:SP`. But:

### 1. CRITICAL — exporting thresholds as `ChannelNumeric` metadata activates caproto's *native* limit-alarm evaluation, which fights the overlay
`channels.py` (~L246-255) puts lolo/low/high/hihi into `lower_alarm_limit`/`lower_warning_limit`/`upper_warning_limit`/`upper_alarm_limit`. caproto's `ChannelNumeric.verify_value` evaluates those on **every write** (a limit pair is active whenever lo≠hi, and unset fields default to **0**), sets alarm state itself, and **publishes to CA monitors before** the overlay's `alarm.write()` corrects it. Two reproduced consequences:
- **Partial row → standing false alarm**: a legitimate "high-only" row leaves `lower_warning_limit=0`; any healthy readback ≤ 0 (a bipolar magnet at −1 A) publishes `MINOR/LOW` → `NO_ALARM` **every frame**. A Phoebus alarm server latches the transient and rings continuously.
- **Hysteresis hold → wire flapping**: a value inside the hold band gets native `NO_ALARM` then overlay `MINOR/HIGH`, per frame — hysteresis makes chatter *worse*.

The PR's tests and the live smoke test use full/benign limit sets where both evaluators agree, which is why this didn't show. Fix direction: keep the overlay as the **single** evaluator — either don't export thresholds into the alarm/warning metadata, or override `verify_value` on the read-only channel subclass to skip native limit evaluation (while preserving the existing plain-write→NO_ALARM recovery).

### 2. MEDIUM — unconditional `alarm.write()` doubles CA event fan-out at steady state
`gateway.py` (~L412): caproto's `ChannelAlarm.write` has no change detection, and a `status` kwarg sets DBE_VALUE — value-mask subscribers (archiver, ophyd-async) get a duplicate event per frame, forever, even in steady `NO_ALARM`. Skip the write when `(severity, status)` is unchanged.

### 3. MEDIUM — hysteresis doesn't apply to level→level transitions (HIHI→HIGH etc.)
`alarms.py` `evaluate` (~L270): direct evaluation wins immediately; hysteresis only guards alarm→normal. EPICS semantics hold HIHI until value < hihi−HYST. The exact live-tested row (hihi=4.5 MAJOR / high MINOR, hyst=0.05) with a readback jittering around 4.5 chatters MAJOR↔MINOR at frame rate. (The dead `current.level != previous` branch just above suggests this was intended and lost.)

### 4. MEDIUM-LOW — fail-open catches *all* DB errors at INFO
`get_ca_alarm_limits`: a permission error or transient DB failure silently disables the whole overlay with a log line invisible at WARNING. Treat "table doesn't exist" (MySQL 1146) as the expected rollout case; log everything else at WARNING/ERROR — silently-missing alarms are the twin of false alarms.

### 5. LOW — drop the FK into the core `device` table
`deploy/ca_alarm_limits.sql`: the FK can fail the migration (index/collation dependent), and `ON DELETE RESTRICT` makes GEECS's own device-admin tooling fail against a table it knows nothing about. The loader's unmatched-row warning is the right enforcement layer.

### Smaller
- `(device, variable)` matching is exact-case while MySQL accepts any case — a `u_s1h` row silently never attaches, and the "not served by the gateway" warning misleads. Case-fold or extend the message.
- Deadband-suppressed writes skip alarm evaluation — latent (deadband policy is 0.0 today) but a threshold crossed by less than the deadband never alarms.
- **No test covers hysteresis** — the only novel algorithm here — and none covers a partial row through the channel layer; either would have caught findings 1/3.
- PV_CONTRACT should state whether thresholds appear in DBR_CTRL/GR metadata (externally observable either way finding 1 is resolved).

Suite ran green (119 passed) in a review worktree; no file overlap with #465. With finding 1 fixed (+ ideally 2/3), this is a solid addition.

skbarber:
Addressed the review findings in `6ed5e004`.

- Removed curated thresholds from native caproto alarm/warning metadata so the overlay is the sole value-alarm evaluator.
- Wrote computed alarm state with the value update using `verify_value=False` for overlay-managed writes, avoiding native limit evaluation and separate steady-state `alarm.write()` fan-out.
- Fixed HIHI/HIGH and LOLO/LOW hysteresis step-down semantics.
- Changed DB fail-open logging so MySQL missing-table rollout remains INFO, while unexpected lookup/connect failures are WARNING.
- Removed the FK from `deploy/ca_alarm_limits.sql`.
- Added regression tests for high-only partial rows, hysteresis, missing-table/failing-query behavior, and non-exported native caproto threshold metadata.

Validation after the fix: `poetry run pytest tests/test_channels.py tests/test_config_from_db.py tests/test_geecs_db.py tests/test_gateway.py -q` -> `76 passed`; commit hooks passed `ruff`, `ruff-format`, and `pydocstyle`.

skbarber:
**Re-review of 6ed5e004: ✅ all blocking findings resolved, no new defects — clear to merge.**

Re-verified empirically through the same CA event-stream harness that caught the original critical: high-only partial row with healthy negative values now emits exactly one clean `NO_ALARM` event per frame (no MINOR/LOW transient, native GR limits confirmed unpopulated); hysteresis hold band shows no flapping and the full EPICS step-down matrix (HIHI→HIGH at hihi−hyst, LOLO→LOW symmetric, immediate escalation, immediate cross-side plunge) is correct; steady-state fan-out is back to one event per frame. `verify_value=False` mechanics are sound: applied per-write only on channels with curated limits, preprocessing still runs, plain-write→NO_ALARM recovery intact for everything else, and adding it to `_mark_device_invalid` was a necessary catch (native evaluation would otherwise stomp INVALID/COMM). Suite: 124 passed. Two of the three key new tests demonstrably fail on the old code (genuine regression pins).

Non-blocking nits, fine to defer: no test observes the actual event stream (a publish-count spy would pin the one-event-per-frame property permanently); the high-only test asserts final state only, which the old code also satisfied; case-fold matching + deadband-skip alarm evaluation consciously deferred as noted.

PR #467, opened 2026-07-08
Add gateway derived-channel overlays

skbarber:
## Summary

- add `DerivedChannels` schema to `GEECS-Schemas`
- add gateway support for schema-validated derived numeric readbacks
- auto-load `scanner_configs/experiments/<Experiment>/gateway/derived_channels.yaml` from the configs repo when present
- keep `--derived-channels PATH` as an explicit override
- document the production path and add a derived-PV smoke test

## Review follow-up

- fixed derived reconnect recovery by clearing `_derived_last_written` for source-device derived PVs on reconnect and INVALID transitions
- derived PVs now initialize as `INVALID/UDF` until the first successful compute, avoiding valid-looking placeholder `0.0` values
- missing or non-numeric derived inputs mark the output `INVALID/UDF`
- persistent expression/runtime failures mark the output `INVALID/CALC` instead of holding a stale good value
- repeated identical derived INVALID states are transition-only and do not re-publish on every failing frame
- bad derived-channel config loads now include the offending file path in the exception
- added regression tests for reconnect with unchanged value, missing input, division-by-zero/runtime failure, initial INVALID state, repeated INVALID suppression, and corrupt YAML path reporting
- documented derived PV alarm semantics in `PV_CONTRACT.md` and `CHANGELOG.md`

## Validation

- `GEECS-Schemas`: `poetry run pytest tests/test_derived_channels.py tests/test_docgen.py -q` -> `91 passed`
- `GeecsCAGateway`: `poetry run pytest tests/test_derived_channels.py tests/test_entrypoint.py -q` -> `20 passed`
- commit hooks passed `ruff`, `ruff-format`, and `pydocstyle`
- earlier full gateway suite on this branch: `123 passed`
- live gateway smoke test: configs repo branch loaded `Undulator:TargetChamberPressure:Pressure`, and Phoebus displayed pressure from `U_VacuumGauge:AI_mean.Channel 0`

## Configs repo companion

Companion config PR: https://github.com/GEECS-BELLA/GEECS-Plugins-Configs/pull/12

It adds the tested `scanner_configs/experiments/Undulator/gateway/derived_channels.yaml` file.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -105,7 +105,8 @@ GEECS-Data-Utils     →  (no intra-repo deps — foundational data layer)
 LogMaker4GoogleDocs  →  (no intra-repo deps — pure Google API wrapper)
 
 ImageAnalysis        →  GEECS-Data-Utils
-GeecsCAGateway       →  (no intra-repo deps — the GEECS access layer:
+GeecsCAGateway       →  GEECS-Schemas (schema-only vocabulary for optional
+                        derived-channel overlays; otherwise the GEECS access layer:
                         wire protocol, DB, PV naming, CA server)
 GeecsBluesky         →  GEECS-Data-Utils, GeecsCAGateway
                         (+ ImageAnalysis, optional via the `analysis` extra —
--- a/GEECS-Schemas/CHANGELOG.md
+++ b/GEECS-Schemas/CHANGELOG.md
@@ -5,6 +5,15 @@ All notable changes to GEECS-Schemas are documented here.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [0.3.0] - 2026-07-08
+
+### Added
+
+- `DerivedChannels` schema for CA gateway derived readbacks: operator-curated
+  read-only float PVs computed from one source device's numeric push-frame
+  values. Schema version 1 enforces same-source-device inputs so calculations
+  such as Convectron pressure from one DAQ analog input are frame-coherent.
+
 ## [0.2.0] - 2026-07-07
 
 ### Changed
--- a/GEECS-Schemas/geecs_schemas/__init__.py
+++ b/GEECS-Schemas/geecs_schemas/__init__.py
@@ -2,9 +2,9 @@
 
 Configs are schemas; YAML is just serialization.  This package is the single
 home of the models (vision doc §4): scan requests, save sets, scan variables,
-trigger profiles, and action plans — plus converters from every legacy YAML
-dialect (``geecs_schemas.convert``) and a Markdown reference generator
-(``geecs_schemas.docgen``).
+trigger profiles, action plans, and gateway derived channels — plus converters
+from every legacy YAML dialect (``geecs_schemas.convert``) and a Markdown
+reference generator (``geecs_schemas.docgen``).
 
 It depends on Pydantic only, so anything — engine, GUI, scripts, docs
 tooling — can import it without dragging in hardware or analysis stacks.
@@ -19,6 +19,11 @@ from geecs_schemas.action_plan import (
     RunPlanStep,
     SetStep,
     WaitStep,
+)
+from geecs_schemas.derived_channels import (
+    DerivedChannel,
+    DerivedChannels,
+    DerivedInput,
 )
 from geecs_schemas.experiment_defaults import DefaultActions, ExperimentDefaults
 from geecs_schemas.save_set import SaveRole, SaveSet, SaveSetEntry
@@ -65,6 +70,10 @@ __all__ = [
     "OptimizationSpec",
     "EvaluatorSpec",
     "GeneratorSpec",
+    # derived_channels
+    "DerivedChannels",
+    "DerivedChannel",
+    "DerivedInput",
     # save_set
     "SaveSet",
     "SaveSetEntry",
@@ -105,4 +114,5 @@ SCHEMA_REGISTRY: dict[str, type[VersionedSchemaModel]] = {
     "action_plan": ActionPlan,
     "action_plan_library": ActionPlanLibrary,
     "experiment_defaults": ExperimentDefaults,
+    "derived_channels": DerivedChannels,
 }
--- a/GEECS-Schemas/geecs_schemas/derived_channels.py
+++ b/GEECS-Schemas/geecs_schemas/derived_channels.py
@@ -1,0 +1,186 @@
+"""DerivedChannels — computed PV declarations for the CA gateway.
+
+This document declares operator-curated read-only PVs that the CA gateway
+computes from one source device's numeric push-frame values.  Use it for
+semantic quantities such as a Convectron pressure PV derived from a DAQ analog
+input voltage, where the result should be visible to Phoebus, archiving, and
+CA-backed clients like any other gateway readback.
+"""
+
+from __future__ import annotations
+
+from pydantic import Field, model_validator
+
+from geecs_schemas._base import SchemaModel, VersionedSchemaModel
+
+
+class DerivedInput(SchemaModel):
+    """One source variable bound to a symbol in a derived-channel formula."""
+
+    symbol: str = Field(
+        description=(
+            "Python-style symbol used in the expression, e.g. 'v' for a "
+            "voltage input. Must be a valid identifier and must not shadow a "
+            "reserved math function or constant."
+        )
+    )
+    device: str = Field(
+        description=(
+            "GEECS source device that provides this input variable, e.g. "
+            "'U_DaqPad1'. All inputs for one derived channel must come from "
+            "the same source device in schema version 1."
+        )
+    )
+    variable: str = Field(
+        description=(
+            "GEECS source variable on the input device, e.g. "
+            "'Analog Input 10'. The gateway subscribes to it even if it is not "
+            "exposed as its own raw readback PV."
+        )
+    )
+
+    @model_validator(mode="after")
+    def _validate_symbol(self) -> "DerivedInput":
+        if not self.symbol.isidentifier():
+            raise ValueError(f"input symbol {self.symbol!r} is not a valid identifier")
+        if self.symbol in {
+            "acos",
+            "asin",
+            "atan",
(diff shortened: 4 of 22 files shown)


skbarber:
**Review (agent-assisted; findings verified against the branch code). Verdict: needs two small fixes before merge; otherwise this is a strong PR.**

A lot verified good first: expression evaluation is AST-whitelisted (constants, bound symbols, arithmetic, 11 math funcs) with `__import__`/attribute-access/lambdas/subscripts all rejected at startup and runtime errors caught without killing the update loop; same-device frame coherence is correct (derived posts after raw data PVs and *before* the frame's timestamp PVs, carrying the source frame's device timestamp — pinned by test); cross-device inputs are schema-rejected in v1 with the vision-doc rationale documented; output PVs go through `normalize_pv_component` with startup collision detection; derived PVs are read-only with no `:SP`; config is whole-document-validated (no half-load), missing file is a clean no-op; zero per-update overhead for devices without derived channels; no cycles possible (inputs must name real devices); `poetry.lock` adds exactly geecs-schemas + pyyaml. Both suites green in a review worktree (schemas 209, gateway 123).

### 1. CONFIRMED BUG — derived deadband cache never cleared on reconnect → derived PV can stick INVALID/COMM forever
`gateway.py:587` clears `self._last_written[dev.name] = {}` on (re)subscribe precisely so "the first frame always posts (and clears any INVALID left from the drop)". The new `_derived_last_written` (created L161, checked L431-436) is **never cleared anywhere**. Scenario: `U_VacuumGauge` drops → `_mark_device_invalid` correctly sets `TargetChamberPressure:Pressure` INVALID/COMM → device reconnects with steady pressure → recomputed value equals the cached one → deadband `continue` suppresses the write → **the PV stays INVALID until the physical value changes**. Contradicts `_mark_device_invalid`'s own "recovery is automatic" docstring and PV_CONTRACT recovery semantics. Fix: in the resubscribe block next to L587, also pop the derived entries for that source device; add a reconnect test (the fake-server infra supports it — it's exactly the test that would have caught this).

### 2. Input variable names are never validated — a typo yields a permanently valid-looking 0.0
Source *devices* are validated loudly at startup (L224-228), but `inputs[].variable` is not checked against the device's DB variable set, and a missing input just skips computation while the PV sits at the 0.0 init placeholder with NO_ALARM. Derived inputs are the first hand-typed variable names to reach the subscription path (raw PV names all come from the DB and can't be wrong). `variable: "AI_mean.Channel O"` (letter O) → gateway starts clean, pressure PV reads 0.0 Torr, valid, forever. Fix: validate inputs against the DB variable list at startup, or at minimum init derived PVs at INVALID/UDF until first successful compute.

### 3. Worth deciding: persistent evaluation failure holds the last good value, looking live
Eval failure warns once and skips (L419-428) — the PV keeps its old value/timestamp with NO_ALARM as long as the failure persists. The vision coherence rules say failed → INVALID; either post INVALID/CALC on failure or document the hold in PV_CONTRACT.

### Merge order with #465 / #466 (the practical part)
- #465 touches GEECS-Schemas in 4 files; the trap is `pyproject.toml`: **both PRs set the identical `0.2.0` version string, which merges *cleanly*** — two different feature sets silently sharing one version. `CHANGELOG.md` conflicts textually at the same spot.
- #466 overlaps this PR on `gateway.py` / `config.py` / `channels.py` / PV_CONTRACT / CHANGELOG / pyproject, and this PR's `_make_callback` refactor sits exactly where #466's alarm publish lives — that rebase needs manual care.
- **Recommended order: #465 → #466 → #467** (matches the version numbering this PR already anticipated: gateway 0.8.0 above #466's 0.7.0). After the first two land, rebase this one and: bump GEECS-Schemas to **0.3.0** (move the derived-channels changelog entry under it), resolve the gateway overlaps, re-run `poetry lock`.

### Nits
- Wrap `load_derived_channels` errors with the file path (currently a bad YAML aborts startup without naming the file).
- `GEECS_PLUGINS_CONFIGS` is a brand-new env var that exists nowhere else — the Bluesky/GUI resolvers support only `GEECS_SCANNER_CONFIG_DIR` + config.ini, so DEPLOYMENT.md's "resolved the same way as scanner configs" overclaims; this is also a third hand-rolled copy of that resolution logic.
- Exact-duplicate derived spec is silently dropped (L232); log it. `_warn_once` key collides for two derived channels sharing an output variable name on one source device.
- Reserved-symbol list is duplicated between `geecs_schemas/derived_channels.py` and `geecs_ca_gateway/derived.py` (schemas can't import the gateway, so acceptable — cross-reference them in comments).
- Missing failure-mode tests: div-zero/missing-input at runtime, collision raise, corrupt-YAML startup, reconnect (finding 1).
- Root CLAUDE.md dep-graph edit is accurate for the gateway, but while touching it: the graph has no GEECS-Schemas node line yet and GeecsBluesky's line omits its geecs-schemas dep (pre-existing staleness on feat/vision-v1).

skbarber:
Addressed the review findings in `057048e8`.

- Cleared `_derived_last_written` for derived PVs on reconnect and when marking derived outputs INVALID, so unchanged values still recover from `INVALID/COMM`.
- Derived PVs now initialize `INVALID/UDF` until first successful compute.
- Missing or non-numeric inputs mark the output `INVALID/UDF` rather than leaving a valid-looking placeholder.
- Runtime expression failures mark the output `INVALID/CALC` rather than holding the last good value as live.
- `load_derived_channels` now wraps load/validation failures with the config file path.
- Added regression tests for reconnect with unchanged value, missing input, division-by-zero/runtime failure, initial INVALID state, and corrupt YAML path reporting.

Validation after the fix: `GeecsCAGateway` focused tests `19 passed`; `GEECS-Schemas` focused tests `91 passed`; commit hooks passed `ruff`, `ruff-format`, and `pydocstyle`.

skbarber:
**Re-review of 057048e8: ✅ findings resolved — two small items remain, neither needs another review round if done as described.**

Verified end-to-end: the reconnect trap is closed in both places (resubscribe pops the derived cache next to the existing raw-cache clear; `_mark_device_invalid` pops it for derived kinds before writing INVALID/COMM), and the new reconnect test exercises a real fake-server stop/restart with an *unchanged* value — reverting the source files while keeping the new tests confirms all three behavioral tests fail on the old code (genuine pins). The finding-1-in-a-new-costume trap is also closed: `_mark_derived_invalid` pops the cache as its first action, so an INVALID PV that recomputes to the same number always posts and clears — the deadband `continue` is only reachable while NO_ALARM. Initial `INVALID/UDF`, `INVALID/CALC` on failure, and path-named load errors all check out. Suites: gateway 127, schemas 209.

**Two pre-merge items (small):**
1. **Docs**: 057048e8 touches no docs — PV_CONTRACT §"Derived numeric channels" should document the new alarm semantics (INVALID/UDF until first compute — a deliberate divergence from the raw-readback 0.0-valid placeholder; UDF on missing/non-numeric input; CALC on expression failure; COMM on source disconnect) and the pinned-test table should gain the three new tests; CHANGELOG 0.8.0 entry should mention the operator-visible behavior.
2. **Per-frame INVALID re-publish**: `_mark_derived_invalid` runs on every failing frame and caproto publishes every write — a persistently broken derived PV (typo'd input, `''` pre-acquisition, div-zero regime) emits 1–5 events/s to all monitors + archiver with a wall-time-ticking timestamp. One-line fix: skip the write when the channel is already INVALID with the same status (transition-only).

Nit: the absent/empty-input path marks UDF with no log at all (only non-numeric inputs warn) — a one-time "input never seen in any frame" log would speed triage of typos.

Merge-order recommendation unchanged: **#465 → #466 → #467**, with #467's final rebase re-versioning GEECS-Schemas 0.2.0 → 0.3.0, regenerating the gateway poetry.lock, and resolving the gateway.py/config.py/CHANGELOG overlaps with #466.

skbarber:
Addressed the two pre-merge follow-ups in `f38024c6`.

- Added `PV_CONTRACT.md` coverage for derived PV alarm states: initial/missing/non-numeric input `INVALID/UDF`, expression failure `INVALID/CALC`, source disconnect `INVALID/COMM`, and recovery on the next successful compute even when the numeric value is unchanged.
- Added the new derived alarm tests to the pinned-test table and updated the `CHANGELOG.md` 0.8.0 entry for the operator-visible behavior.
- Changed `_mark_derived_invalid` to publish only the transition for repeated identical INVALID statuses, while still clearing `_derived_last_written` first so the next successful same-value compute posts and clears severity.
- Added `test_repeated_derived_failure_does_not_republish_invalid`.

Validation: `GeecsCAGateway` focused tests now `20 passed`; `GEECS-Schemas` focused tests `91 passed`; commit hooks passed `ruff`, `ruff-format`, and `pydocstyle`.

skbarber:
**Verified f38024c6: ✅ both pre-merge items done correctly — this PR is now clear pending the rebase after #465/#466.**

- **Transition-only guard**: exactly right — the deadband cache is still popped unconditionally *first* (preserving the same-value recovery guarantee), and the early-return triggers only when the channel is already INVALID with the same status, so a status *change* (UDF→CALC) still publishes. I traced the guard's reads through caproto: `ChannelData.severity/status` delegate to the alarm object, and `_collect_alarm` clears any pending native-evaluation override on every write — so the guard always compares the true published alarm state. The new test counts actual `write` calls across two consecutive div-zero frames — a genuine pin.
- **Docs**: PV_CONTRACT gains the full derived-alarm-state table (UDF/CALC/COMM + recovery-on-unchanged-value + transition-only), explicitly contrasted with the raw-readback 0.0 placeholder, and all five derived-alarm tests are in the pinned-test table. Changelog updated.

One piece of trivia for the record, no action needed: caproto's `_collect_alarm` has an upstream quirk (the severity condition tests `_status != alarm.status`, likely a copy-paste slip), which would drop a severity-only change with an unchanged status. Neither this PR nor #466 can hit it — every severity change here pairs with a status change (UDF↔CALC↔COMM; HIGH↔HIHI) — but worth knowing if a future overlay ever writes severity alone.

Remaining before merge, as previously agreed: land **#465 → #466 → this**, then rebase with GEECS-Schemas re-versioned 0.2.0 → 0.3.0, `poetry lock` regenerated, and the `gateway.py`/`config.py`/CHANGELOG overlaps with #466 resolved.

PR #468, opened 2026-07-08
vision M3c: DB-runtime tier (db_scalars, scan start/end writes, background telemetry) + s-file acq_timestamp

skbarber:
Fourth engine milestone on `feat/vision-v1` (stack: #465 M3b → #466/#467 gateway → this). Activates the DB-backed scan setup the schema declared back in M1, plus a small s-file fix Sam flagged from a live scan.

## What M3c adds (all gated by flags that already existed in the schema — schema untouched)

- **db_scalars resolution (Tier 1).** A SaveSet entry's recorded scalars = DB `get='yes'` ∪ its explicit `scalars` (`db_scalars=True`, default); `all_scalars=True` unions every DB variable; `db_scalars=False` (the legacy-converter pin) = explicit-only. Pure resolver behind a `ScalarPolicyProvider` seam; no provider (GUI bridge / off-network) = explicit-only, i.e. exact M3b behavior — strictly additive.
- **Scan start/end DB writes, participants-only.** For scan participants (save-set + scan-variable devices, **not** every experiment device), `expt_device_variable` `set='yes'` rows write `startvalue` at scan start / `endvalue` at scan end, through the same CA `:SP` path actions use. `at_scan_start`/`at_scan_end` overrides layer on (value replaces, explicit `null` suppresses, absence = DB value); `save`/`localsavingpath` always skipped. Start writes chain into the setup hook (before acquisition); end writes into the closeout hook (finalize → run on mid-scan abort). Gated on `ExperimentDefaults.apply_db_scan_defaults`; applied writes recorded in run metadata (`db_scan_writes`).
- **Background telemetry (Tier 2).** Every live `get='yes'` device not in the save set → soft read-only `CaTelemetryReadable` columns (`telemetry_<device>-<var>`): never waited on, a failed read is a NaN cell, a dead-at-start device is dropped with a log line. Gated on `ScanRequest.background_telemetry` else the experiment default. **Softness and synchronicity are mutually exclusive — telemetry never gates a shot.**
- New `GeecsDb.get_all_experiment_variables` + `get_scan_boundary_writes` (following the established `get_subscribed_variables` join pattern; `set` reserved-word escaped, values as wire strings). `GeecsDbScalarPolicy` caches three whole-experiment queries and degrades to empty+warning on any DB failure — a scan never aborts because the DB blipped.

Optimize mode resolves db_scalars but skips start/end writes + telemetry (no scan-boundary hook on `GeecsSession.optimize` yet) — skip-and-record (`db_scan_runtime` metadata).

## Also here (separate commit): s-file `acq_timestamp` fix

Sam observed a live images-only scan produced an s-file with only `Bin #/scan/Shotnumber` — no column tying the saved frames to rows. acq_timestamp reaches Tiled fine (it's the shot-id source) but was excluded from the legacy s-file as a companion column, and the detector filtered it out of its header map unconditionally. Now a device saving non-scalar data surfaces acq_timestamp as an s-file column (`<device> acq_timestamp`, matching legacy camera behavior); stays excluded for pure-scalar devices.

## Verification

Hermetic (DB mocked): GeecsBluesky **400 passed** / 1 skip (package env, `--extras ca`, CI-parity), GeecsCAGateway **141**, GEECS-Schemas **209** (untouched). The new `GeecsDb` SQL was reviewed against the established working query in the same file. Versions: GeecsBluesky **0.24.0**, GeecsCAGateway **0.9.0**.

## ⚠️ Owed before merge: live verification (some of it actuates hardware)

- **Safe / read-only:** the s-file `acq_timestamp` column on a real camera scan; db_scalars pulling a device's DB `get='yes'` scalars into the s-file; telemetry columns appearing for live non-save-set devices (and a dead one dropped, not aborting).
- **Actuates hardware (do deliberately):** the scan start/end DB writes physically move participant devices to their `startvalue`/`endvalue`. Needs a chosen-safe experiment/device and a confirmation that end-writes fire on abort. Not run yet.

### DIFF
--- a/GEECS-Data-Utils/geecs_data_utils/tiled_export.py
+++ b/GEECS-Data-Utils/geecs_data_utils/tiled_export.py
@@ -32,11 +32,14 @@ import pandas as pd
 
 logger = logging.getLogger(__name__)
 
-# Derived companion columns (event-schema v1) that must not appear in the legacy
-# scalar file.  Any event-stream column not named in ``geecs_scalar_headers`` is
-# dropped anyway; this set documents the intent.
+# Derived companion columns (event-schema v1) normally kept out of the legacy
+# scalar file.  Emission is driven entirely by ``geecs_scalar_headers`` (any
+# event-stream column not named there is dropped), so this set only documents
+# the intent — it is not an active filter.  Exception: ``-acq_timestamp`` is
+# deliberately surfaced (added to a device's headers) for file/image-saving
+# devices so saved files tie back to scan rows; it stays out for pure-scalar
+# devices.
 _COMPANION_SUFFIXES = (
-    "-acq_timestamp",
     "-t0_acq_timestamp",
     "-shot_id",
     "-shot_offset",
--- a/GEECS-Schemas/CHANGELOG.md
+++ b/GEECS-Schemas/CHANGELOG.md
@@ -5,6 +5,25 @@ All notable changes to GEECS-Schemas are documented here.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [0.4.0] - 2026-07-08
+
+### Changed
+
+- **Reserved the DB set-side scan-write fields — not honored; get-side
+  `db_scalars` / telemetry unchanged.**  `SaveSetEntry.at_scan_start` /
+  `at_scan_end` and `ExperimentDefaults.apply_db_scan_defaults` are kept in
+  the schema but their field descriptions and the surrounding docstrings now
+  state clearly that the DB **set-side** (scan start/end writes from the
+  `set='yes'` rows) is **not applied in this version**: the engine sets up
+  triggering via the TriggerProfile / shot controller and camera saving via
+  its own save-windowing, so DB boundary writes would race the shot
+  controller (on the DG645 the `set='yes'` rows are the very
+  trigger/amplitude variables it drives).  The fields are reserved for a
+  possible future re-enable.  No fields were removed, so existing configs
+  still validate.  The **get-side** — `SaveSetEntry.db_scalars` /
+  `all_scalars` (standard telemetry) and the `background_telemetry` defaults
+  — is unchanged and still honored.
+
 ## [0.3.0] - 2026-07-08
 
 ### Added
--- a/GEECS-Schemas/README.md
+++ b/GEECS-Schemas/README.md
@@ -57,13 +57,15 @@ schema-side only in M1), `PositionRange` / `PositionList`, `ActionBindings`
 `EvaluatorSpec`, `GeneratorSpec` — covers the legacy Xopt VOCS surface),
 `SaveSetEntry` / `SaveRole` (entries carry optional `setup` / `closeout`
 action-plan name references — the device's ritual travels with it through
-composition — plus the DB scan-default surfaces: `at_scan_start` /
-`at_scan_end` per-variable overrides of the DB's start/end writes, where a
-value replaces the DB's, an explicit `null` suppresses the write, and
-absence inherits; and `db_scalars`, on by default for new configs, making
-the DB's scan-logging telemetry the standard scalar source with `scalars`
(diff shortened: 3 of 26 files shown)


skbarber:
**Headless live verification (over VPN, read-only — M3c is get-only now so nothing actuates).**

**acq_timestamp fix ✓** — a headless strict noscan saving the Amp4 camera (`background_telemetry` off) produced run metadata `geecs_scalar_headers = {'uc_amp4_ir_input-acq_timestamp': 'UC_Amp4_IR_input acq_timestamp'}`. The s-file exporter emits exactly that map, so the s-file now carries the `UC_Amp4_IR_input acq_timestamp` column (the one scalar column an images-only save was missing). `db_scan_writes` metadata absent — set-side confirmed disabled live. (No on-disk s-file this run: the `hdna2` share unmounted mid-session; the metadata proof is definitive since the exporter is header-driven and its header→column mapping is unit-pinned.)

**Background telemetry (Tier 2) ✓ at scale + softness verified** — same scan with `background_telemetry` on selected **88 devices** (the full `get='yes'`-minus-save-set set: magnets, hexapod, DG645 readbacks, gauges, every camera's analysis scalars, PLC DI channels, …) recorded to Tiled. The scan **completed normally** (ARMED → 3 single-shots → STANDBY) — softness holds.

**⚠️ Finding: telemetry is float-only.** `U_VisaPlungers` (whose `DigitalOutput.Channel N` vars are enum/string) failed float-signal construction and the **whole device was dropped** with a log line — correctly *not* aborting the scan (the safety property working), but it means a device with any non-float `get='yes'` var contributes no telemetry, even its float vars. Not a blocker (soft, logged), but worth deciding: make `CaTelemetryReadable` dtype-aware (capture float vars, skip/represent non-float per-variable instead of dropping the device). Recommend a small follow-up.

Suites remain green (GeecsBluesky 388, gateway 141, schemas 209).

skbarber:
**Telemetry dtype fix — committed (`e512c106`) and live-verified.**

`CaTelemetryReadable` now builds signals with `datatype=None` (per-variable CA type inference) instead of forcing float. Numerics stay float; enum/string PVs are captured as their label. Per-variable, so one non-float channel never drops the device.

Re-ran the telemetry scan (uid `ef64bc54`): **`U_VisaPlungers` is now logged, not dropped** — its 8 `DigitalOutput.Channel N` enum channels appear in the Tiled event stream as string labels:

```
telemetry_u_visaplungers-digitaloutput_channel_0 = ['off' 'off' 'off']  (dtype <U3)
```

The only devices dropped now are `U_BCaveICT` / `U_UndulatorExitICT` — genuinely unreachable at scan start (the correct softness path), not type failures. 88 telemetry devices selected, scan completed normally.

So the full get='yes' telemetry set is captured regardless of variable type, numerics keep their numeric type, and the soft tier still never aborts. Hermetic suite: GeecsBluesky 392.

skbarber:
**Codex review addressed (`205cda4c`).** All three items were real; verified against the code and fixed (with the tests the review noted were missing).

- **[P2] telemetry metadata accuracy** — `build_telemetry_readables` now returns the **connected-only** device map, so `md["background_telemetry"]` records only devices that actually produced columns. A device dropped as unreachable at scan start is no longer advertised in the start doc (matches the EVENT_SCHEMA.md contract). New test `test_telemetry_metadata_excludes_dropped_devices` pins it (mixed live+dead → md excludes the dead one).
- **[P3] optimize-mode reserved warning** — `warn_if_reserved_boundary_overrides` is now called in `_run_optimize_request` too, so a reserved `at_scan_start`/`at_scan_end` warns in every mode that resolves a save set, not just scan/noscan. New test `test_optimize_warns_on_reserved_boundary_fields` pins it.
- **[doc] telemetry acq_timestamp/validity drift** — corrected the `SaveSetEntry` docstring: soft-tier telemetry columns are plain per-row snapshots and do **not** carry `acq_timestamp`/validity labeling (those are Tier-1 synchronous-device features). Now consistent with `EVENT_SCHEMA.md` and `telemetry.py`.

Suites: GeecsBluesky **394**, GEECS-Schemas **209** (docgen no-drift intact), pre-commit clean.

PR #469, opened 2026-07-09
Add cross-device derived channel freshness

skbarber:
## Summary
- add stale_after to derived-channel schema and require it for cross-device inputs
- evaluate cross-device derived PVs with latest-value semantics, missing/stale input invalidation, and source-disconnect COMM invalidation
- extend restricted expressions with comparisons and boolean operators for status PVs, emitted as 1.0/0.0
- update deployment/PV contract docs plus schema and gateway changelogs

## Notes
- same-device derived channels remain frame-coherent and keep source-frame timestamps
- cross-device derived channels intentionally use gateway receive-time freshness, not shot-coherent alignment
- stale PVs become invalid when an affected source update triggers recomputation; this PR does not add a proactive timer for completely quiet sources

## Tests
- GEECS-Schemas: poetry run pytest tests/test_derived_channels.py tests/test_docgen.py -q
- GeecsCAGateway: poetry run pytest tests/test_derived_channels.py tests/test_entrypoint.py -q

### DIFF
--- a/GeecsCAGateway/PV_CONTRACT.md
+++ b/GeecsCAGateway/PV_CONTRACT.md
@@ -223,20 +223,28 @@ the configs repo convention
 `scanner_configs/experiments/<Experiment>/gateway/derived_channels.yaml` (or an
 explicit `--derived-channels PATH` override). That overlay exposes additional
 read-only float PVs computed from one source device's numeric push-frame
-values. The v1 schema is intentionally narrow:
+values, or from multiple source devices with explicit latest-value freshness
+semantics.
 
 - A derived channel has one output PV (`device`, `variable`, optional
   `experiment`/`pv` override) and one arithmetic expression.
-- Inputs bind expression symbols to `(device, variable)` sources, but **all
-  inputs must come from the same source device**. The gateway evaluates the
+- Inputs bind expression symbols to `(device, variable)` sources.
+- Same-device inputs are **frame-coherent**. The gateway evaluates the
   expression once per source frame, after raw data PVs are posted and before
   that frame's timestamp PVs are posted. Therefore a client latching on the
   source device's timestamp observes raw and derived values from the same
   completed frame.
+- Cross-device inputs use **latest-value semantics**. A cross-device derived
+  PV recomputes when any input source updates, using the latest cached numeric
+  values from the other sources. Such entries must declare `stale_after`
+  seconds; if any input is missing or older than that window, the output PV is
+  `INVALID_ALARM` / `UDF`.
 - The expression subset is numeric arithmetic only: literals, input symbols,
-  `+ - * / % **`, unary `+/-`, and a small whitelist of `math` functions such
-  as `sqrt`, `exp`, `log`, and `log10`. Arbitrary Python, attributes,
-  indexing, comparisons, and conditionals are rejected during gateway startup.
+  `+ - * / % **`, comparisons, boolean `and` / `or` / `not`, unary `+/-`, and
+  a small whitelist of `math` functions such as `sqrt`, `exp`, `log`, and
+  `log10`. Arbitrary Python, attributes, indexing, and conditionals are
+  rejected during gateway startup. Boolean/status expressions are stored as
+  `1.0` or `0.0` on the float PV.
 - Inputs may be subscribed solely for the calculation; they do not need to be
   exposed as their own readback PVs.
 
@@ -272,8 +280,9 @@ Repeated failures with the same INVALID status do not re-publish on every
 source frame; the gateway publishes the transition and keeps the PV invalid
 until a successful computation clears it.
 
-Cross-device latest-value expressions, staleness windows, enum/string bad-state
-expressions, and shot-synchronized analysis products are outside v1.
+Enum/string bad-state expressions and shot-synchronized analysis products are
+outside v1. If inputs need a shared shot id, compute the result in analysis,
+not as a gateway derived PV.
 
 ---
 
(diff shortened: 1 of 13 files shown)


GeecsCAGateway/PV_CONTRACT.md
```diff
@@ -223,20 +223,28 @@ the configs repo convention
...
+- Inputs bind expression symbols to `(device, variable)` sources.
+- Same-device inputs are **frame-coherent**. The gateway evaluates the
   expression once per source frame, after raw data PVs are posted and before
   that frame's timestamp PVs are posted. Therefore a client latching on the
   source device's timestamp observes raw and derived values from the same
   completed frame.
+- Cross-device inputs use **latest-value semantics**. A cross-device derived
+  PV recomputes when any input source updates, using the latest cached numeric
+  values from the other sources. Such entries must declare `stale_after`
+  seconds; if any input is missing or older than that window, the output PV is
```
skbarber, line 260: **Contract overstates the staleness guarantee for completely-quiet sources.**

This line promises: *“if any input is missing or older than that window, the output PV is `INVALID_ALARM` / `UDF`.”* But staleness is only ever re-evaluated inside `_write_derived_channels`, which runs *only when some source device pushes a frame* (`_make_callback` → callback). If **all** input sources go completely quiet after a valid computation, nothing recomputes, `stale_after` never fires, and the PV holds its last value at `NO_ALARM` indefinitely.

The PR description acknowledges this (“does not add a proactive timer for completely quiet sources”), but the normative contract here doesn't carry that caveat — a reader/operator will trust the blanket promise.

This matters most because the worked example is a **`LaserPermit`** (`pressure < 1e-5 and ready > 0`). A permit-style PV that fails *stale-but-reads-`NO_ALARM`* is exactly the fail-open mode you don't want. Two acceptable resolutions:
1. Add the caveat here (and rename the example away from “permit”, since a gateway derived PV explicitly cannot be a safety interlock without the timer), or
2. Add the proactive staleness sweep (e.g. tick the cross-device runtimes from `_status_loop`) so the promise becomes true.

Either is fine; the current state where the contract promises more than the code delivers, on a permit example, is the thing to fix.

GeecsCAGateway/geecs_ca_gateway/derived.py
```diff
@@ -172,6 +178,20 @@ def _validate_node(self, node: ast.AST) -> None:
             self._validate_node(node.left)
             self._validate_node(node.right)
             return
+        if isinstance(node, ast.BoolOp):
```
skbarber, line 191: **`and`/`or` with non-comparison operands publishes a raw magnitude, not `1.0`/`0.0`.**

The CHANGELOG and PV_CONTRACT.md both state boolean/status expressions “publish as float values `1.0` or `0.0`.” That holds when the top-level result is a comparison or `not` (which return `bool`). But Python's `and`/`or` return an *operand*, not a bool: with the allowed grammar an operator could write `ready and enabled` (bare numeric inputs), and `evaluate()` does `float(result)` on line ~161 — so `float(2.0 and 5.0)` → `5.0`, published verbatim.

The only test (`test_expression_evaluator_supports_status_logic`) uses comparison operands, so it passes while the general promise doesn't hold. Either coerce boolean-typed results (`float(bool(result))` only when the top node is `BoolOp`/`Compare`/`Not`, preserving raw numeric channels like the Convectron formula), or narrow the documented promise to “comparison-based expressions.”

GeecsCAGateway/geecs_ca_gateway/gateway.py
```diff
@@ -453,62 +475,119 @@ async def _write_stream_value(
...
-                    break
-            if missing:
-                await self._mark_derived_invalid(channel, pv, AlarmStatus.UDF)
+        """Evaluate derived PVs affected by this source frame."""
+        runtimes = self._derived_by_source.get(device_name, [])
+        if not runtimes:
+            return
+        now = asyncio.get_running_loop().time()
+        self._update_derived_input_cache(device_name, update, runtimes, now)
+        evaluated: set[str] = set()
```
skbarber: **Minor / cleanup:** the `evaluated: set[str]` dedup appears to be dead code. `_build_derived_pvs` appends each runtime once per *unique* source device (`for source in sorted(spec.source_devices)`, and `source_devices` is a set), so within a single `_derived_by_source[device_name]` list a given runtime PV can appear at most once. The `if runtime.pv in evaluated: continue` guard therefore never triggers. If it's intended as defensive insurance against a future where a runtime could be listed twice for one device, a one-line comment saying so would help; otherwise it can go.

skbarber (commented):
Code review (medium effort). The cross-device machinery is clean — the `_derived_by_source` fan-out correctly invalidates a derived PV when *any* input source disconnects (not just `inputs[0].device`), same-device frame coherence is preserved through the cache, and the monotonic-clock staleness math is sound. Three findings below, ranked. #1 is the one I'd want resolved before merge because it's a contract-vs-behavior gap on a safety-adjacent example PV.

(Note: my automated finder agents hit an infra error on this run, so this is a direct read of the head source rather than the usual fan-out — coverage is the three changed source files + PV_CONTRACT.md.)

GeecsCAGateway/geecs_ca_gateway/derived.py
skbarber: Fixed in 914c0894. ExpressionEvaluator now treats top-level BoolOp, Compare, and not-expressions as boolean-intent outputs and coerces them through bool() before float(), so expressions like `ready and enabled` publish 1.0/0.0 instead of the selected operand magnitude. Added test_expression_evaluator_bool_ops_publish_binary_floats.

GeecsCAGateway/geecs_ca_gateway/gateway.py
skbarber: Fixed in 914c0894. Removed the evaluated set/guard since each runtime is already registered once per unique source device, so there is no duplicate to filter in the per-source runtime list.

GeecsCAGateway/PV_CONTRACT.md
skbarber: Fixed in 914c0894. I added a proactive freshness sweep from the gateway status loop so cross-device derived PVs are rechecked even when all input sources go quiet; stale inputs now drive INVALID/UDF without requiring a new source frame. I also updated PV_CONTRACT.md/DEPLOYMENT.md and pinned this with test_cross_device_quiet_sources_mark_stale_invalid.

PR #472, opened 2026-07-09
Scan-variable & PV-metadata vocabulary (confirm, .DESC, PseudoComponent)

skbarber:
Scoped work on the scan-variable schema and gateway PV metadata, branched from and targeting `feat/vision-v1`.

## Schema (`geecs-schemas` 0.6.0)
- **Rename `PseudoTarget` → `PseudoComponent`** — a `PseudoScanVariable` *contains* components; the old name collided with the element's own `target` field. Field names / serialized output unchanged.
- **`ScanVariable.confirm`** — optional `Device:Variable` naming the *measured* variable when it differs from the *set* variable (topology-C, e.g. the EMQ triplet's `Current_Limit.ChN` vs measured `Current.ChN`). Additive, defaults `None`, declared-but-not-yet-enforced; carries no tolerance (that's a device fact, stays in the DB).

## Gateway (`geecs-ca-gateway` 0.10.0)
- **Per-PV `.DESC`, resolved from the DB and served over CA.** DB `variable.description` → capability inheritance chain → `VariableSpec.description` → a read-only `<pv>.DESC` string channel (readbacks + derived channels). Verified end-to-end with a live CA client. Clipped to the 40-char DBR_STRING limit at both the DB column and the gateway. `PV_CONTRACT.md` §4 + pinned test.
- Merges in `master`'s variable-capability inheritance-chain fix (the branch was behind).
- `deploy/variable_description.sql` — DDL record (column narrowed to `VARCHAR(40)`, applied; optional type-level default left for follow-up).

## Bluesky (`geecs-bluesky` 0.24.1)
- **CaSettable / CaMotor docstring honesty** — CaSettable already rides GEECS blocking convergence; CaMotor's poll only helps with an independent same-variable readback; neither covers topology-C.

## Live DB changes already applied (2026-07-09)
- EMQ `Current.Ch{1,2,3}` tolerance `0 → 0.05` A.
- `U_S1H:Current` description populated (proof-of-concept).
- `variable.description` narrowed `VARCHAR(1000) → VARCHAR(40)` (0 rows affected).

## Config (separate repo)
Undulator migrated to new-schema `scan_variables.yaml` with EMQ `confirm:` — committed on `GEECS-Plugins-Configs@codex/undulator-derived-channels` (own PR if wanted).

## Tests
152 gateway + 214 schemas passing. Design note: `Planning/scan_variable_metadata/00_overview.md`.

## Follow-ups (documented, not gaps)
- Optional type-level `devicetype_variable.description` default (DDL + coalesce) — owner: Sam.
- Populating descriptions/aliases broadly; wiring `alias` into a catalog name-default chain.

### DIFF
--- a/GEECS-Schemas/CHANGELOG.md
+++ b/GEECS-Schemas/CHANGELOG.md
@@ -5,6 +5,29 @@ All notable changes to GEECS-Schemas are documented here.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [0.6.0] - 2026-07-09
+
+### Changed
+
+- **`PseudoTarget` renamed to `PseudoComponent`.** A `PseudoScanVariable`
+  *contains* a `list[PseudoComponent]`; the old name collided conceptually
+  with the element's own `target` field, and the legacy composite-variable
+  dialect already called these *components*. Field names (`targets`,
+  `forward`) and all serialized output are unchanged — this is a class rename
+  only. The export in `geecs_schemas.__all__` moves accordingly.
+
+### Added
+
+- **`ScanVariable.confirm`** — an optional `Device:Variable` naming the
+  variable that *measures* the result when it differs from the variable being
+  *set*. It documents the "topology C" split (e.g. the EMQ triplet, whose
+  catalog entry sets `Current_Limit.ChN` while the measured current is a
+  separate variable that GEECS's set-completion never checks). Additive,
+  defaults to `None` (⇒ prior behavior), validated as a `Device:Variable`.
+  **Declared but not yet enforced by the engine** — and deliberately carries
+  no tolerance field (the match tolerance is a device fact that stays below
+  the configs). See `Planning/scan_variable_metadata/00_overview.md`.
+
 ## [0.5.0] - 2026-07-08
 
 ### Added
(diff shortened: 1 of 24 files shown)


skbarber (commented):
**Finding**

- **P2: `PseudoTarget` import compatibility is broken.** `GEECS-Schemas/geecs_schemas/scan_variables.py` replaces the public `PseudoTarget` class with `PseudoComponent`, and `GEECS-Schemas/geecs_schemas/__init__.py` stops importing/exporting `PseudoTarget`. Serialized configs stay compatible, but any downstream Python importing `PseudoTarget` now fails on a minor version bump, which conflicts with the repo’s “minor = backwards-compatible” guidance. Keep `PseudoTarget = PseudoComponent` as a deprecated alias and export it, with a small import-compatibility test.

I did not run the local test suite; the Poetry env here had no installed dependencies. GitHub showed the previous `pre-commit` and `unit-tests` runs passed, with the latest rerun still queued at review time.

skbarber:
Re: **P2 — `PseudoTarget` import compatibility.**

Acknowledged but **not actioning** — the backward-compatibility premise doesn't apply here:

- `GEECS-Schemas` **does not exist on `master`** — it's a `feat/vision-v1`-only package, never released, never tagged.
- **Zero importers** of `PseudoTarget` anywhere in `GEECS-Plugins` or `GEECS-Plugins-Configs` (grep-verified, excluding the two files that define/export it).
- `PseudoTarget` only ever existed on the unmerged vision-v1 line, so no shipped version could import it.

The "breaks downstream on a minor bump" scenario needs a downstream, and there is none — this is renaming a symbol that never shipped. The "minor = backwards-compatible" convention is about not breaking *released* API; a deprecation alias here would be guarding against breakage of code that doesn't exist. When vision-v1 becomes a real release line, *that's* when the compat clock starts.

Keeping the clean rename (no alias) to avoid reintroducing the two-names-for-one-thing ambiguity the rename removed.

PR #474, opened 2026-07-09
Add ScanAnalysis scalar sidecars

### LINKED ISSUE #473: add sidecar files with scalars after scananalysis runs
Right now, a scananalyzer appends the scalar results to the analysis/sxxx.txt file. This file is dynamic and can get deleted/regenerated. So that we don't have to re-run the analyzer, can we add a simple sidecar file indexed by shotnumber that contains the generated scalar values that lives in the same path as the other post analyzed results?

skbarber:
## Summary
- write analyzer-generated scalar updates to per-analyzer scalar_results.tsv sidecars indexed by Shotnumber
- keep the existing mutable analysis/sNNN.txt append path unchanged
- bump ScanAnalysis to 1.15.0 and document the new sidecar output

Closes #473

## Tests
- poetry run pytest tests/test_scalar_sidecar.py tests/test_use_injected_data.py
- pre-commit hooks during git commit


skbarber (commented):
Low-effort diff review (1 pass, no findings).

Reviewed the changed hunks in `single_device_scan_analyzer.py` and `base.py`. No runtime-correctness bugs found:

- The `mkdir(parents=True, exist_ok=True)` in `write_scalar_sidecar` targets the `analysis/` output tree, not `scans/ScanNNN/`, so it doesn't trip the scan-folder-creation invariant.
- `write_scalar_sidecar` is called once with the full pending update set before `append_to_sfile`, so the full-overwrite `to_csv` (drop-duplicates keep=last, sorted by Shotnumber) is correct.

LGTM. 🤖 Automated low-effort review — no blocking issues.

skbarber:
@ag6520 requested the sidecar files land with .txt extension but use tab delimiters. Also, add the ScanXXX_ prefix to the sidecar file name

skbarber:
one more comment from @ag6520: he wants the sidecar file to land in the analysis/ScanXXX dir rather than nested down into the specific analyzer folders. This also means we need to replace the generic 'scalar_results' file stem that we were using. This should be replaced with the analyzer name. Though, let's be a little careful about which name we are using here. a diagnotic config has a filename e.g. amp2out that creates and analyzer for a device UC_Amp2_IR_input. My feeling is that the name should come from the config file name rather than the device. But, this is an ambiguity that should be worked out and clearly documented

PR #475, opened 2026-07-10
docs: record EMQ confirm-device characterization from interrupted session

skbarber:
The previous session characterized \`U_EMQTripletBipolar:Current.ChN\` live (no beam) before losing filesystem access, and flagged that the result needed to be written into \`Planning/scan_variable_metadata/00_overview.md\` before it was lost. Recovering it here from the session's final message.

Adds to Deferred #5 (the topology-C confirm-device):
- **Characterization**: jitter 0.01 A, response lag <1 s, settles within ~3 frames — supply electronics behavior with no beam present.
- **Derived defaults**: confirm-poll tolerance 0.05 A (5× jitter), global poll timeout ~10 s.
- **Build spec**: new branch \`feat/scanvar-confirm-device\` off \`feat/vision-v1\`; analog match by tolerance / discrete by equality; wire into \`scan_request_runner\`'s movable-construction path via \`ScanVariable.confirm\`; add the three EMQ \`confirm:\` catalog entries once built.

Docs-only; no code/schema change. Building the actual device is the next PR.

### DIFF
--- a/Planning/scan_variable_metadata/00_overview.md
+++ b/Planning/scan_variable_metadata/00_overview.md
@@ -237,3 +237,24 @@ Two consequences that revise the earlier plan:
    `Current.ChN` in the DB** — otherwise an exact-float confirm-poll never
    settles. This is the same "how much PV metadata can the DB serve" thread as
    Deferred #1.
+
+   **Live characterization of `U_EMQTripletBipolar:Current.ChN` (2026-07-09,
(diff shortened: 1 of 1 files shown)


skbarber:
Closing without merging — the content is superseded/duplicated now that #477 exists: the build spec paragraph describes work #477 already implements (with more accurate detail in its own code/PR description), and the characterization numbers already live in `confirm.py`'s module docstring. Not worth carrying a stale echo in `Planning/`.

PR #477, opened 2026-07-10
feat(GeecsBluesky): CaConfirmSettable — implement ScanVariable.confirm (topology-C)

skbarber:
Implements the device #472 laid schema/analysis groundwork for but did not build: a scan variable where the variable you *set* differs from the variable that *measures* the result (declared in the schema since M1 via `ScanVariable.confirm`, unenforced until now).

## What this adds

- **`CaConfirmSettable`** (`devices/ca/confirm.py`): writes `variable`, then polls a *different* `confirm_variable`'s streamed readback for the physical result. Layer 1 (the `:SP` put) rides GEECS's own blocking convergence on the written variable — this is the EMQ triplet's problem: `Current_Limit.ChN` converges trivially (the limit register echoes what was written) while the real `Current.ChN` is never touched. Layer 2 polls the confirming variable: analog (float) match by tolerance, discrete (string/enum) match by exact equality — unifying with a future `CaShutter`.
- **Defaults from a live characterization**, not guessed: `tolerance=0.05`, `timeout=10.0`, sized from a no-beam measurement of `U_EMQTripletBipolar:Current.ChN` (jitter 0.01 A, response lag <1 s, settles within ~3 frames — recorded in #475 / `Planning/scan_variable_metadata/00_overview.md` Deferred #5).
- **`GeecsSession.confirm_settable(...)`** constructs it, matching the `motor`/`settable` factory pattern.
- **Resolver wiring**: `resolve_movable_target` now returns `(device, variable, kind, confirm)` (was a 3-tuple); a new `build_movable(session, device, variable, kind, confirm)` helper dispatches confirm-first (a confirm target wins over `kind`) and replaces the inline motor/settable branches in both `run_scan_request`'s grid-axis construction and its optimize-mode path. The GUI bridge's staging-only call site just unpacks the new tuple shape (it doesn't build movables yet, so no behavior change there).
- New `GeecsConfirmTimeoutError` — same "operator attention, don't auto-retry" posture as `GeecsMotorTimeoutError`.

## Tests

4 hermetic device tests (writes target/reads confirm PVs on separate PVs, completes within tolerance, times out when confirm never matches, discrete equality match) + 1 resolver-level test (a scan variable with `confirm` dispatches to `confirm_settable`, not motor/settable) via a new `emq1_current` catalog entry in the test fixtures.

GeecsBluesky 0.25.0. Suite: **399 passed**.

## Next (not this PR)

- Add the three EMQ `confirm:` catalog entries (`Current_Limit.ChN → Current.ChN`) to the real Undulator scan-variable catalog.
- Live-verify on hardware (no beam is sufficient — the characterization this PR's defaults are based on was also no-beam).


skbarber:
**Live hardware verification (over VPN, on-network, 2026-07-10).** All three EMQ channels tested with a small reversible move (0 → 0.5 A → 0 A), confirming on the separate \`Current.ChN\` measured variable (not the \`Current_Limit.ChN\` register the write itself targets).

| Channel | Result | Confirm-time deviation from target |
|---|---|---|
| Ch1 | ✅ confirmed both directions | 0.0000 / 0.0000 |
| Ch2 | ✅ confirmed both directions | +0.0002 / +0.0254 |
| Ch3 | ⏱️ **timed out after 10s (correctly)** | \`Current_Limit.Ch3\` converged to 0.5 (register write succeeded); \`Current.Ch3\`, \`Voltage.Ch3\`, \`Power.Ch3\` all stayed ~0 |

**Ch3 is a live, unplanned demonstration of exactly the gap this device closes.** Ch3 has \`enable=off\` at the hardware level (confirmed with Sam) — the channel accepts and holds any \`Current_Limit\` write (it's just a command register) but physically outputs nothing. The pre-existing \`CaSettable\`/\`CaMotor\` devices would have reported this scan step as a successful set, because GEECS's own blocking convergence only checks the register that was written. \`CaConfirmSettable\` correctly refused to confirm and raised \`GeecsConfirmTimeoutError\` instead — silent-failure prevention working as designed, on the very device the schema's \`confirm\` field was written for.

(\`Current_Limit.Ch3\` was restored to 0 immediately after diagnosis; no lasting hardware state change.)

Also resolved a question from earlier review prep: the three EMQ \`confirm:\` catalog entries already exist in the real Undulator \`scan_variables.yaml\` (landed via a separate branch, \`codex/undulator-derived-channels\`) — so once this PR merges, those entries start building \`CaConfirmSettable\` automatically with no config change needed.

skbarber (commented):
**Findings**

- **P2: Optimization `on_finish` bypasses `CaConfirmSettable.set()`.** `GeecsBluesky/geecs_bluesky/session.py` moves final/restore targets by calling `m._setpoint.set(v)` directly, and the optimize path uses that helper for `on_finish="initial"` and `"best"`. During adaptive iterations, `bps.mv` calls the movable `set()` and gets the new confirm semantics, but the final best/initial move skips the `Current.ChN` confirmation entirely. For the exact EMQ failure mode this PR is closing, an optimization can finish by writing `Current_Limit.ChN`, log only a warning if the raw put fails, and never raise `GeecsConfirmTimeoutError` if the measured current does not follow. Use the movable protocol here, e.g. await `m.set(v)` / status completion in the RE loop, so motors and confirm-settables keep their completion semantics.

- **P3: Discrete exact equality is not actually exact for numeric-looking strings.** `GeecsBluesky/geecs_bluesky/devices/ca/confirm.py` tries `float(current)` and `float(target)` before falling back to equality. That means a `datatype=str` confirm target can accept `"1.04"` as matching `"1.0"` with the default tolerance, despite the docstring and PR description saying string/enum confirms match by exact equality. This is probably harmless for the EMQ float case, but it weakens the future shutter/enum behavior the class explicitly claims. Branch on datatype/type instead of parseability, and add a numeric-string discrete test.

- **P3: New public exception is omitted from `__all__`.** `GeecsBluesky/geecs_bluesky/exceptions.py` lists the package exported exception surface, but adds `GeecsConfirmTimeoutError` without adding it there. Direct imports work, but star imports and any tooling keyed off `__all__` miss the new scan-level exception.

**Open question**

Is it intentional that a confirm scan axis records the written variable readback (`Current_Limit.ChN`) as the motor column, while the confirming readback (`Current.ChN`) is only recorded if it appears separately in the save set? The code comment says yes, but for an operator-facing EMQ current scan this may be surprising.

**Verification**

I reviewed PR #477 plus the PR #472/#475 context and the live hardware note. GitHub CI is green for `pre-commit` and `unit-tests`. I did not run the local suite; this checkout Poetry env does not have dependencies installed.

skbarber:
**All three review findings fixed and pushed (\`d91ac5cc\`).** Verified each against the code before fixing; all are real.

**[P2] optimize \`on_finish\` bypassed confirm semantics — fixed and live-reverified.** \`_move_movables\` (used for \`on_finish="initial"\`/\`"best"\`) called \`m._setpoint.set(v)\` directly, skipping every movable's own completion protocol — for \`CaConfirmSettable\` that's exactly the failure this device exists to catch. Fixed to \`await m.set(v)\`. New unit test pins it against a fake confirm-shaped movable (confirmed it fails on the old code). I also live-reverified the exact fixed code path against real hardware on the disabled Ch3: \`_move_movables\` now correctly raises \`GeecsConfirmTimeoutError\` (logged as a WARNING with traceback) instead of silently "succeeding."

**[P3] discrete match coerced numeric-looking strings — fixed.** \`_matches\` tried \`float()\` before falling back to equality, so a \`datatype=str\` target could tolerance-match \`"1.04"\` against \`"1.0"\`. Dispatch is now on the device's declared \`datatype\` (stored on the instance), never on parseability. New test: a numeric-looking near-miss string now correctly times out instead of matching.

**[P3] \`GeecsConfirmTimeoutError\` missing from \`__all__\` — fixed.**

**Open question (recorded column)** — documented rather than left implicit: the confirm axis's event-row column is the *written* variable (same shape as \`CaSettable\`/\`CaMotor\`), not the confirming one. Added to \`confirm.py\`'s module docstring and \`GeecsBluesky/CLAUDE.md\` with the guidance to include the confirming variable in the save set separately when its measured value matters to analysis, not just pass/fail.

Suite: 401 passed (399 + 2 new tests). Pre-commit clean.

PR #479, opened 2026-07-10
M4 step (0): ScanRequest references a list of named save sets (union)

skbarber:
First build of the M4 GUI arc (`Planning/gui_retrofit/00_overview.md` step 0). A scan names **`save_sets: list[str]`** instead of one `save_set`, so operators mix and match named diagnostic groups (laser cams, aux, e-beam profiles) per scan; the engine resolves each named `SaveSet` and **unions** their devices.

## Change
- **Schema:** `ScanRequest.save_set` → `save_sets: list[str]`, before-validator coerces a bare string to `[str]` (single-set ergonomics). Clean rename. Step/noscan still requires a non-empty list.
- **Runner:** `merge_save_sets` unions device-by-device (scalars union, images/db_scalars/all_scalars OR, first role, rituals unioned once); rituals collected across all sets deduped by name; `md["save_sets"]` recorded.
- **Telemetry (M3c):** Tier-2 excludes devices from **any** named set.
- **Bridge:** `resolve_save_sets_checked` (union; still refuses actions/multi-axis — step (i)).
- **Converter:** `convert_scan_preset` emits `save_sets = referenced element names`.
- Goldens + schema reference regenerated (no-drift green). GEECS-Schemas 0.8.0, GeecsBluesky 0.25.0.

## Verified
- Hermetic: GEECS-Schemas **218**, GeecsBluesky **284** (7 new union tests).
- **Hardware** (free-run noscan, `save_sets=["Amp4In","Amp4Out"]`): run `25222169` — `md['save_sets']=['Amp4In','Amp4Out']`, both cameras built as detectors (2 in t0-sync), both present in the event stream. The union lands.

Next: M4 step (i) — bridge parity via `run_scan_request` delegation.

### DIFF
--- a/GeecsBluesky/geecs_bluesky/scan_request_runner.py
+++ b/GeecsBluesky/geecs_bluesky/scan_request_runner.py
@@ -16,7 +16,8 @@ This module owns the mapping:
   (:func:`merge_save_sets`) so operators mix and match named diagnostic
   groups per scan.  The per-device union rule: ``scalars`` union
   (order-preserving, deduped), ``images`` / ``db_scalars`` / ``all_scalars``
-  OR together (True wins), the first non-``None`` ``role`` is kept, and the
+  OR together (True wins), the single non-``None`` ``role`` is used (a device
+  the sets require with *conflicting* explicit roles is an error), and the
   entry-level ``setup`` / ``closeout`` ritual name lists union (deduped).
   Entry-level rituals are collected across *all* named sets, deduped by plan
   name so a shared ritual runs once (:func:`resolve_save_sets_and_rituals`).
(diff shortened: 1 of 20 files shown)


GeecsBluesky/geecs_bluesky/scan_request_runner.py
```diff
@@ -762,6 +774,132 @@ def resolve_save_set_and_rituals(
...
+            if item not in merged:
+                merged.append(item)
+        return merged
+
+    return SaveSetEntry(
+        device=existing.device,
+        scalars=_union(list(existing.scalars), list(addition.scalars)),
+        all_scalars=existing.all_scalars or addition.all_scalars,
+        images=existing.images or addition.images,
+        role=existing.role if existing.role is not None else addition.role,
```
skbarber, line 830: [P2] Refuse conflicting explicit roles instead of letting request order decide. If two named save sets include the same device with different explicit roles, this keeps whichever non-null role appears first in save_sets. That means save_sets=["snapshot_set", "reference_set"] can demote the device to an async snapshot, while the reverse order makes it the reference detector, even though the user selected the same two sets. Role controls the acquisition guarantees and pacemaker/contributor/snapshot wiring in save_set_to_devices_config, so silently picking by list order can produce a scan that records a required device with the wrong synchronization semantics. The legacy preset composer in this PR already treats contradictory role overrides as an error; the runtime merge should do the same for overlapping entries with incompatible explicit roles.

skbarber (commented):
Reviewed PR #479. I found one issue that should be addressed before merge.

skbarber:
Addressed the review (`f0374a61`). You're right — resolving `role` by `save_sets` list order silently gave a required device the wrong synchronization semantics.

`merge_save_sets` now **raises `GeecsConfigurationError`** when two named sets require the same device with *different* explicit roles (matching the legacy preset composer's behavior for the same conflict). Same role, or one side leaving `role` unset, still merges cleanly. New test `test_merge_save_sets_conflicting_roles_raise` pins that **both orders raise** (so the outcome is order-independent) and the unset case survives. Docstrings, CHANGELOG, and CLAUDE.md updated to state the rule.

Suites still green (schemas 218, bluesky 284 + the new test).

PR #485, opened 2026-07-10
feat(bridge): M4 step (i) — bridge delegates ScanRequest execution to run_scan_request

skbarber:
## What this is

M4 step (i) of the GUI retrofit plan (`Planning/gui_retrofit/00_overview.md` §3–4): the GUI bridge reaches **ScanRequest parity** with the headless engine. GeecsBluesky 0.27.1 → **0.28.0**.

## What delegation replaces

`BlueskyScanner._reinitialize_from_scan_request` used to synthesize a legacy scan-config `SimpleNamespace` (plus the `_request_step` side channel) and run the exec_config scan-thread body, refusing actions, entry rituals, and multi-axis grids with `NotImplementedError` pointers at `GeecsSession.run`. Now it only:

1. **Validates fail-fast** on a local post-defaults copy (defaults → actions → save sets + rituals → trigger profile → scan variables), discarding every result, so the GUI gets an immediate error pre-thread; and
2. **Stores the original pre-defaults request** + resolver (`run_scan_request` applies defaults itself — storing the post-defaults copy would prepend default action plans twice; pinned by `test_defaults_not_applied_twice`).

The scan-thread body then calls `run_scan_request(session, request, resolver, preflight=..., on_scan_start=...)` — so actions, entry rituals, multi-axis outer-product grids, db_scalars, and background telemetry all run through the **one engine definition**. The bridge never pre-claims on this path: `session.scan` claims the number and self-attaches `scan.log`. The legacy `exec_config` path is untouched.

## The two new runner hooks

Both keyword-only, default `None` — headless callers are unchanged. Neither runs on the optimize path (an optimize preflight is a later seam).

- **`preflight(detectors, strict) -> list | None`** — the scanner-layer operator-dialog seam. Called after the detector list is fully assembled (save-set devices + telemetry readables), pre-claim by construction. `None` aborts (the runner's `finally` still disconnects everything it created); a reduced list is honored. The bridge threads its existing liveness/staleness pipeline through with a new `disconnect_on_drop=False` (the runner owns disconnection on this path; exec_config path byte-identical, default `True`), and sets `_abort_requested` on abort so the thread cleanup reports ABORTED.
- **`on_scan_start(total_steps, total_shots)`** — progress totals, called immediately before `session.scan`: `(1, shots_per_step)` for noscan, `(len(positions), len(positions) × shots_per_step)` for step (flat outer-product length for grids). The bridge uses it to prime `_total_steps`/`_total_shots` and emit INITIALIZING → RUNNING.

## Deleted

- Runner refusal cluster: `raise_if_actions_present`, `resolve_save_sets_checked`, `MULTI_AXIS_MESSAGE` (their only callers were the bridge refusals this PR removes) + their tests.
- Bridge: the request→legacy-namespace synthesis, the `_request_step` machinery, and `_run_request_step_scan`.

## Retained

- The **optimize-mode refusal** at `reinitialize(ScanRequest)` (message updated) — wiring the GUI's `optimization_loader` into the delegated path is GUI-submission step (iii).
- `resolve_and_validate_actions`, `resolve_save_sets_and_rituals`, `resolve_defaults_for` — the bridge still uses them for fail-fast validation.

## The new parity pin

The old byte-parity pin (request noscan ≡ synthesized exec_config) is intentionally retired — the request path now legitimately produces richer kwargs (compiled action stubs, `save_sets` provenance, `scan_request_mode` md). Its replacement: **a request submitted through the bridge produces `session.scan` kwargs identical to headless `run_scan_request` on an equivalent fresh fake session** (`test_scan_request_delegation_parity_with_headless`), plus a no-pre-claim assertion. New coverage: multi-axis delegation, action compile-and-delegate, unknown-name fail-fast, trigger profile via `session.shot_control`, preflight abort, assembled-detector/strict-flag threading, totals hook, defaults-once, and runner-side hook tests (preflight reduce/abort, noscan + 2×3-grid totals).

## Tests

`poetry install --extras ca && poetry run pytest -q` → **416 passed, 1 skipped, 3 deselected**. Hardware smoke (headless multi-axis + action request via the bridge) still owed per the plan — noted in the planning doc.

## Docs/bookkeeping

CHANGELOG 0.28.0 entry; `GeecsBluesky/CLAUDE.md` stale bridge-refusal passages rewritten; `Planning/gui_retrofit/00_overview.md` step (i) marked landed; `Planning/cleanup_vision_v1/00_overview.md` deferred-until-M4 section updated with done/partial/kept/still-open status.

### DIFF
--- a/GeecsBluesky/geecs_bluesky/scanner_bridge/bluesky_scanner.py
+++ b/GeecsBluesky/geecs_bluesky/scanner_bridge/bluesky_scanner.py
@@ -74,16 +74,12 @@ from geecs_bluesky.preflight import (
     run_preflight,
 )
 from geecs_bluesky.scan_request_runner import (
-    MULTI_AXIS_MESSAGE,
     ConfigResolver,
     ConfigsRepoResolver,
-    raise_if_actions_present,
     resolve_and_validate_actions,
     resolve_defaults_for,
-    resolve_movable_target,
-    resolve_save_sets_checked,
-    save_set_to_devices_config,
-    trigger_writes_from_profile,
+    resolve_save_sets_and_rituals,
+    run_scan_request,
 )
 from geecs_bluesky.scan_log import scan_log
 from geecs_bluesky.session import GeecsSession, _positions
@@ -211,8 +207,9 @@ class BlueskyScanner:
         self._optimization_loader = optimization_loader
         # ScanRequest entry (the acceptance seam of the schema milestone):
         # set by reinitialize(ScanRequest); None on the exec_config path.
+        # The scan thread delegates a stored request to run_scan_request.
         self._scan_request: ScanRequest | None = None
-        self._request_step: dict[str, Any] | None = None
+        self._request_resolver: ConfigResolver | None = None
 
         # Catch a stale env loudly rather than silently changing behavior.
         legacy_backend = os.environ.get("GEECS_BLUESKY_DEVICE_BACKEND")
@@ -277,7 +274,7 @@ class BlueskyScanner:
                 exec_config, resolver=_kwargs.get("resolver")
             )
         self._scan_request = None
-        self._request_step = None
+        self._request_resolver = None
         self._scan_config = exec_config.scan_config
 
         # Derive shots from rep_rate × wait_time (ScanOptions has no shots_per_step field)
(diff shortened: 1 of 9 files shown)


GeecsBluesky/geecs_bluesky/scanner_bridge/bluesky_scanner.py
```diff
@@ -315,113 +312,64 @@ def reinitialize(self, exec_config: Any, **_kwargs: Any) -> bool:
...
-                end=positions[-1],
-                step=(positions[1] - positions[0]) if len(positions) > 1 else 0.0,
-                wait_time=request.shots_per_step,
-                additional_description=request.description,
-                background=request.background,
-            )
-
+        resolve_save_sets_and_rituals(resolver, validated.save_sets)
+        if validated.trigger_profile:
+            resolver.resolve_trigger_profile(validated.trigger_profile)
```
skbarber: [P2] Validate the trigger variant during reinitialize. This fail-fast pass resolves the profile and discards it, but never calls trigger_writes_from_profile(profile, validated.trigger_variant) or equivalent. A request with trigger_profile="HTU" and trigger_variant="typo" now passes reinitialize and only fails later inside the scan thread when run_scan_request adapts the profile. The old bridge path validated the variant here because it built the ShotControlWrites during reinitialize, and the PR body promises immediate pre-thread validation; please restore that check in this validation pass.

GeecsBluesky/geecs_bluesky/scanner_bridge/bluesky_scanner.py
```diff
@@ -315,113 +312,64 @@ def reinitialize(self, exec_config: Any, **_kwargs: Any) -> bool:
...
+        if validated.mode is ScanRequestMode.STEP:
+            for axis in validated.axes:
+                resolver.resolve_scan_variable(axis.variable)
```
skbarber: [P2] Resolve each scan variable all the way to an executable target here, not just through resolver.resolve_scan_variable. PseudoScanVariable resolves successfully, but run_scan_request still raises NotImplementedError when resolve_movable_target sees it. Before this PR the bridge called resolve_movable_target during reinitialize, so pseudo requests failed before the scan thread was launched; now they start the thread and fail later. Please run the same executable-target validation in this fail-fast pass, e.g. resolve_movable_target(spec, axis.variable).

skbarber:
Both review findings addressed in 206e8452:

- **Trigger variant**: the validation pass now runs `trigger_writes_from_profile(profile, validated.trigger_variant)` (result discarded), so an unknown variant fails at `reinitialize` with the known-variants error — pinned by `test_unknown_trigger_variant_fails_at_reinitialize`.
- **Pseudo scan variables**: each axis is now resolved to an executable target via `resolve_movable_target(spec, axis.variable)`, so a `PseudoScanVariable` is refused at `reinitialize` (`NotImplementedError`) instead of after the scan thread launches — pinned by `test_pseudo_scan_variable_fails_at_reinitialize`.

This restores the exact fail-fast depth the pre-delegation bridge had (it used to build the writes / resolve the target inline during reinitialize). Full suite: **418 passed**.

PR #487, opened 2026-07-10
G1: delete the legacy scan engine (BlueskyScanner becomes the only backend)

skbarber:
Implements **G1** of `Planning/cutover_strategy/00_overview.md` exactly — the deletion-only first PR of the gut sequence on the greenfield branch. The GUI keeps working bluesky-only via its existing `ScanExecutionConfig` submission through `RunControl`. **No GeecsBluesky changes.**

**Net: −5,433 lines, +128 (29 files).**

## Deleted (verified scan-path-only by the 2026-07-10 consumer audit)

Under `GEECS-Scanner-GUI/geecs_scanner/engine/`:

| Module | Lines |
|---|---|
| `scan_manager.py` | 1,213 |
| `data_logger.py` | 661 |
| `scan_data_manager.py` | 450 |
| `device_manager.py` | 436 |
| `file_mover.py` | 417 |
| `scan_executor.py` | 409 |
| `trigger_controller.py` | 118 |
| `default_scan_manager.py` | 70 |
| `lifecycle.py` | 45 |

Plus:
- `engine/__init__.py` pruned **first** (it eagerly imported all of the above) — surviving exports: `ActionManager`, `DeviceCommandExecutor`, `DatabaseDictLookup`, the full `scan_events` vocabulary, `get_full_config_path` / `visa_config_generator`.
- `ScanExecutionConfig.to_device_manager_dict()` — only ScanManager called it.
- `RunControl`'s legacy branch: no `use_bluesky` argument, no `GEECS_USE_BLUESKY` consultation, no `get_database_dict` legacy else-branch, no `GeecsDeviceInstantiationError` import. The Bluesky path is the only path.
- `engine/backend_selection.py` **collapsed to an always-True stub** — `resolve_use_bluesky()` keeps its signature for import-surface compatibility and returns `True` unconditionally; `tests/app/test_run_control_backend_selection.py` reworked to pin the always-bluesky behavior (arg and env var ignored).
- `TYPE_CHECKING`-only `DataLogger`/`ScanDataManager` imports stripped from `optimization/base_evaluator.py` and `base_optimizer.py`; the injected-runtime hints are duck-typed `Any` with comments.
- Tests of the deleted modules: `tests/engine/test_data_logger.py`, `tests/engine/test_scan_executor.py`, and the ScanStepExecutor/ScanManager/ScanLifecycleStateMachine halves of `tests/engine/test_event_emission.py` (the DeviceCommandExecutor emission tests remain; the event vocabulary stays pinned in GeecsBluesky's suite and `tests/engine/test_event_shims.py`). `tests/engine/conftest.py` pruned to just `FakeScanDevice`.
- Repo-root legacy-engine tests (slightly beyond the package-only scope, flagged for review): `tests/integration/hardware/test_scan_manager_hardware.py` (hardware test solely of ScanManager — would break root-suite collection if left) and the five `tests/ScannerTools/` stubs importing the deleted modules (that directory is already `norecursedirs`-excluded stale stubs; its remaining non-legacy files are untouched).

## Deliberately kept (NOT legacy, despite living in `engine/` — audit is explicit)

`action_manager.py` + `device_command_executor.py` (ActionLibrary GUI), `database_dict_lookup.py`, the `scan_events.py`/`dialog_request.py` re-export shims, **all** of `engine/models/` (GUI→backend contract, also used by optimization), `utils/sound_player.py`, `app/multi_scanner.py` (dies with the old GUI at cutover, not now), `app/lib/action_control.py`, everything else in `app/`. `tests/optimization/conftest.py`'s `FakeDataLogger` stays — it's an actively-used duck-type, not an import of the real class.

## Test results

- **GEECS-Scanner-GUI**: 207 passed (baseline) → **162 passed** after; the 45 removed tests all targeted deleted modules. (Invocation from the package dir needs `--import-mode=prepend` — the repo-root importlib config breaks `tests.engine.conftest` imports; pre-existing quirk on the base branch, unchanged here.)
- **GeecsBluesky** (untouched, `poetry install --extras ca`): **418 passed, 1 skipped, 3 deselected** — exactly the expected count.
- Sanity import check: `run_control` (via the app_controller-style loader, Qt stubbed, DB patched), `base_evaluator`, `base_optimizer`, `session_bridge`, all surviving `engine` exports, `resolve_use_bluesky() is True`, models intact, `to_device_manager_dict` gone.

## Notes

- Version bump `0.32.1 → 0.33.0` (minor) + CHANGELOG entry.
- Two stale `See Also → engine.device_manager` docstring refs removed from the kept `models/save_devices.py`.
- Follow-up owed before cutover (already tracked in the strategy doc): verify GeecsBluesky s-file export covers every scan mode the legacy producer handled.


GEECS-Scanner-GUI/geecs_scanner/app/run_control.py
```diff
@@ -7,105 +7,69 @@
...
-            except GeecsDeviceInstantiationError as e:
-                logging.error(f"GeecsDeviceInstantiationError: {e.message}")
-                raise ConnectionError(e.message)
+        self.scan_manager = BlueskyScanner(
+            experiment_dir=experiment_name,
+            shot_control_information=settings,
+            on_event=on_event,
+            optimization_loader=load_session_optimization,
+        )
+        self.action_control = None
```
skbarber, line 65: [P1] This makes the kept Action Library execution path unusable. The PR explicitly keeps ActionManager/ActionControl as non-legacy, but RunControl now always initializes action_control to None, and GEECSScannerWindow.open_action_library passes get_action_control() without a refresh. ActionLibrary.execute_action later unconditionally calls self.action_control.perform_action(...), so enabling execution now raises AttributeError instead of running the manual action. The old legacy backend constructed ActionControl(experiment_name) at RunControl init; if manual actions are intentionally kept until G-actions, get_action_control() needs to lazily construct/return one for the current experiment (or the GUI needs to request the refresh before opening the action library).

skbarber:
Review P1 addressed in the latest commit: `get_action_control()` now lazily constructs `ActionControl(experiment_name)` on first use (cached; `experiment_name_refresh` still rebuilds) — restoring the legacy init-time behavior in lazy form. Net effect is better than pre-G1: manual actions now actually **work** under the bluesky-only RunControl (they were dark before, since bluesky mode always had `action_control = None`), pending the proper G-actions re-homing. Pinned by two new headless-safe tests (Qt stubbed, module file-loaded). GUI suite: 164 passed.

PR #488, opened 2026-07-10
GEECS-Console 0.1.0: package scaffold + main window (screen map v0.1)

skbarber:
New top-level package `GEECS-Console/` — the greenfield PySide6 operator console decided 2026-07-10 (`Planning/cutover_strategy/00_overview.md` §"Greenfield GUI — settled parameters", feature inventory in `01_gui_feature_inventory.md`, approved screen-map artifact). This is the "one working screen" half of the commit/abort checkpoint's criterion (c).

## What's implemented, per region

- **R1 session bar** — experiment combo (populated from the configs repo's experiment folders), rep-rate field, trigger-profile combo + variant combo (variants resolved on demand through `ConfigsRepoResolver`), gateway/tiled/db health chips.
- **R2 save sets** — available/selected lists with Add/Remove, live union preview ("union: N devices") computed via the engine's `merge_save_sets`, role-conflict / pinned-reference hint line.
- **R3 scan form** — mode radios (No-scan / 1D / Grid / Optimization / Background), variable picker + start/stop/step (second axis row enabled in Grid mode), shots-per-step, acquisition combo (**free_run default**, strict — the request declares intent per the M4 acquisition-toggle decision), live computed shot count with the `MAXIMUM_SCAN_SIZE = 1e6` guard (arithmetic count, never materializes positions), description field.
- **R4 presets** — combo + Apply + Save-as present; persistence stubbed (a preset IS a saved ScanRequest).
- **R5 submit row** — Stop (danger) + Start (primary). Start requires: not scanning, ≥1 selected save set, valid in-guard shot count, mode ≠ Optimization.
- **R6 Now panel** — state pill, progress bar (sized from the INITIALIZING `total_shots`, advanced by `ScanStepEvent.shots_completed`), "Scan NNN" label with 10 s expiry to "(previous)", compact log tail.
- **R7 device panel** — device:variable combo, readback label, set field + Set button; **backend stubbed** (the real one is gateway PVs: CA monitor readback + put to `:SP` — never the legacy python API's ScanDevice).
- Menu bar (Ops/Actions/Editors/Preferences/Help placeholders) + status bar (gateway addr, configs path, version).

## Architecture

- `request_builder.build_scan_request` — the core: pure `ConsoleFormState → geecs_schemas.ScanRequest` (1D/grid → `axes[].positions` as `PositionRange`/`PositionList`; Background → `noscan` + `background=True`; Optimization refused until an `OptimizationSpec` editor exists).
- `submission.Submitter` protocol (the four ScanManager-compatible methods) + lazy `make_bluesky_submitter` — the window opens without the `ca` extra or any network.
- `services/configs.py` — offline-safe: missing configs root → empty listings + a status message, never a crash. Listing scans folders; validation happens on demand through the resolver.
- `services/health.py` — `HealthProbe` protocol + all-unknown `StubHealth` (real CA/HTTP/DB probes later).
- `events_adapter.py` — engine `on_event` → Qt signals; dispatches on event class names so it imports without `geecs_bluesky` and tests hermetically.
- **`geecs_python_api` is never imported** — pinned by `tests/test_no_geecs_python_api.py` (source grep + sys.modules sweep over every module).

## Stubbed seams (intentional)

Health probes, device-panel backend, presets persistence, optimization submission, `ScanDialogEvent` modal rendering, scan-number event source. All listed in the package `CLAUDE.md`.

## Tests

- `GEECS-Console`: **50 passed** (`QT_QPA_PLATFORM=offscreen poetry run pytest -q`) — request-builder shape tests validated against the real schema (round-tripped through `model_validate(model_dump())`), main-window behavior with fake configs/submitter, events-adapter emissions, the never-import guard.
- `GeecsBluesky` regression: **418 passed, 1 skipped, 3 deselected** (unchanged baseline).
- Smoke: `MainWindow` constructs offscreen with zero network/configs; `poetry run python main.py` launches.


GEECS-Console/geecs_console/app/main_window.py
```diff
@@ -0,0 +1,647 @@
...
+            message = f"Cannot submit: {exc}"
+            self.statusBar().showMessage(message, 10_000)
+            self.append_log(message)
+            return
+        submitter = self._ensure_submitter()
+        if submitter is None:
+            return
+        try:
+            submitter.reinitialize(request)
+            submitter.start_scan_thread()
```
skbarber, line 572: [P1] Honor a failed reinitialize before starting the scan thread. The Submitter protocol says reinitialize returns a bool, and the real BlueskyScanner path can return False when resolving a ScanRequest fails. This handler currently ignores that result and still calls start_scan_thread(), which starts the scanner with whatever _scan_config was left from the previous successful run (or None on a fresh scanner). From the console this means a bad save set/trigger/profile can silently rerun stale hardware configuration instead of refusing to start. Please branch on the return value and leave the scan stopped/logged when reinitialize returns False.

skbarber:
Review P1 addressed: `_on_start_clicked` now branches on `reinitialize()`'s return — a `False` logs "Submission refused" to the status bar + log tail and never calls `start_scan_thread()`, so a refused request can't rerun stale scanner state. Pinned by `test_refused_reinitialize_does_not_start`. Console suite: 51 passed.

PR #489, opened 2026-07-11
fix(actions): CA-native puts + dtype-inferred reads in CaActionSignalFactory

skbarber:
Found by the first hardware set-step on a numeric variable (the M4 step-(i) smoke): `CaActionSignalFactory` built **str-typed** ophyd signals, which connect-fail on a float `:SP` (`Position.Axis 1:SP` → "inferred datatype float cannot be coerced to str"). M3b's action verification evidently never exercised a numeric set-step over CA.

**Fix** — follow the hardware-proven `CaPutSetter` convention the module's own docstring already claimed:
- Settables put the **wire string via raw CA** (`caput(pv, str(value))` — CA converts to the PV's native type server-side, so float/enum/string `:SP` all work), wrapped in `AsyncStatus` so put-completion still rides the GEECS blocking set.
- A **dtype-inferred probe signal** is connected at creation, preserving the pre-claim fail-fast (missing PV fails before a scan number is claimed).
- Readables switch `str` → dtype-inferred (telemetry's pattern); `values_match` handles native numerics with the legacy string-expected quirk intact.
- Factory gains `mock=` (threaded from the session) and its **first hermetic tests** (`test_action_signals.py`, 6 tests) — the absence of which is how this survived to hardware.

Suite: **425 passed** (424 prior + 6 new − 1 reworked assertion in `test_session`).

Hardware re-verification of the step-(i) smoke (grid + actions via the bridge) follows on this branch before merge.

### DIFF
--- a/GeecsBluesky/CHANGELOG.md
+++ b/GeecsBluesky/CHANGELOG.md
@@ -4,6 +4,22 @@ All notable changes to `geecs-bluesky` are documented here.
 
 Format follows [Keep a Changelog](https://keepachangelog.com/en/1.1.0/).
 
+## [0.28.1] - 2026-07-10
+
+### Fixed
+
+- **Action set/check steps now work on numeric PVs.** `CaActionSignalFactory`
+  built `str`-typed ophyd signals for `:SP` settables and readbacks; a float
+  PV (e.g. `Position.Axis 1:SP`) then failed at connect ("inferred datatype
+  float cannot be coerced to str") — found by the first hardware set-step on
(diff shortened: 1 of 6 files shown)


skbarber:
Now includes the root-cause fix for #490 (the ca:// scheme strip) — the PR is the complete action-signals repair: connect-type bug + naming bug, both regression-tested (7 tests), **hardware-verified end-to-end** (M4 step-(i) smoke passes with closeout set-steps executing on the jet stage, ScanState.DONE). Suite: 426 passed.

PR #495, opened 2026-07-11
feat(telemetry): gate background telemetry on the gateway CONNECTED PV (#494)

### LINKED ISSUE #494: Background telemetry: gate on the gateway CONNECTED PV before full-connect (dead devices burn 20s each)
**Symptom** (found driving GEECS-Console live over VPN, 2026-07-10): a scan with unreachable background-telemetry devices stalls ~20 s **per dead device, serially**, before reaching pre-flight. Two dead ICTs (`U_BCaveICT`, `U_UndulatorExitICT`) = ~40 s of dead air on Start. VPN aggravates it (slow CA search runs the full timeout) but the root cause is local.

**Root cause**: `GeecsSession.telemetry()` → `_connect(det)` does a full ophyd `device.connect()` with `future.result(timeout=20.0)` on **every** candidate, then drops the ones that fail. So each genuinely-dead device costs the full 20 s connect budget. The soft/best-effort telemetry tier is paying the hard-device connect budget it was explicitly designed not to gate on.

**The fix** (maintainer's call, and it reuses existing infrastructure): the gateway already serves a per-device `CONNECTED` PV, and the **save-device pre-flight already consumes it** (`GatewayLivenessCheck` reads `connected_status`; only exact `"Disconnected"` means down, fail-open otherwise). Background telemetry should do the same: **read `[Experiment:]Device:CONNECTED` first** — a single cheap caget the gateway answers from its own live state — and **drop `Disconnected` devices immediately**, only full-connecting the survivors. Fan the CONNECTED reads out concurrently so the whole gate is ~one round-trip even over VPN, instead of 20 s × N-dead.

Benefits over a short-timeout band-aid: authoritative (same liveness source the scanner already trusts), instant (no doomed connect attempt), and consistent (one liveness mechanism, not a second heuristic).

**Scope**: `db_runtime.build_telemetry_readables` / `GeecsSession.telemetry` (GeecsBluesky, feat/vision-v1 — engine fix, both futures benefit). Keep the soft-tier contract: a device unreachable at scan start is dropped with a log line, never aborts. Consider also exposing `background_telemetry: off` in the console for remote operators who want to skip the tier entirely (Tier-2 data is never required for a scan).

**Related**: the same CONNECTED-based gating pattern lives in `preflight.GatewayLivenessCheck`; factor toward one shared "is this device up per the gateway?" helper if convenient.

skbarber:
Closes #494. Found driving GEECS-Console live over VPN: a scan with unreachable background-telemetry devices stalled ~20 s **per dead device, serially**, before reaching pre-flight (~40 s of dead air with two dead ICTs). Root cause: `build_telemetry_readables` did a full ophyd `connect()` (20 s budget) on every candidate, then dropped the failures.

**Fix** (the maintainer's insight): the gateway already serves a per-device `CONNECTED` PV, and the save-device pre-flight already trusts it. Background telemetry now does the same — read `[Experiment:]Device:CONNECTED` **first**, skip the `Disconnected` ones before any full-connect.

- New `GeecsSession.live_devices(devices, *, timeout=3.0)` — reads every candidate's `CONNECTED` PV **concurrently** on the RE loop (one round-trip for the whole set), returns the live/indeterminate subset. **Fail-open**: only the exact `"Disconnected"` string excludes a device; unreadable PV / mock backend / gateway-without-status-PVs all keep it, so a missing status surface never silently drops telemetry.
- `build_telemetry_readables` gates `selected` through `live_devices` before calling `session.telemetry` — dead devices are skipped with a log line; survivors take the normal (still soft, still drop-if-connect-fails) path.

Result: a scan with N dead telemetry devices reaches pre-flight in ~one round-trip instead of 20 s × N. Directly fixes the "slow and unresponsive" GEECS-Console experience for remote/VPN operators.

**Tests**: full suite **443 passed**. New: `live_devices` mock no-op + real-path CONNECTED gating (patched aioca.caget: Disconnected excluded, Connected kept, timeout fail-open kept); telemetry-builder gating (dead device skipped *before* `telemetry()`) + fail-open (indeterminate device still attempted then dropped softly). The old `test_telemetry_dead_device_dropped_not_raised` was updated to the new fast path.

GeecsBluesky 0.30.0. Engine fix → `feat/vision-v1`; syncs up to greenfield so GEECS-Console inherits it.

### DIFF
--- a/GeecsBluesky/CHANGELOG.md
+++ b/GeecsBluesky/CHANGELOG.md
@@ -4,6 +4,29 @@ All notable changes to `geecs-bluesky` are documented here.
 
 Format follows [Keep a Changelog](https://keepachangelog.com/en/1.1.0/).
 
+## [0.30.0] - 2026-07-10
+
+### Changed
+
+- **Background telemetry now gates on the gateway `CONNECTED` PV before
+  full-connect (issue #494).** `build_telemetry_readables` first reads each
+  candidate device's `[Experiment:]Device:CONNECTED` status (one fast
+  concurrent read via the new `GeecsSession.live_devices` — the same
+  authoritative liveness source pre-flight uses) and skips the
+  `Disconnected` ones immediately, instead of paying a full ~20 s device
+  connect timeout per dead device (serially). A scan with several dead
+  telemetry devices — common when driving remotely over VPN — now reaches
+  pre-flight in ~one round-trip instead of 20 s × N-dead. Fail-open: an
+  unreadable status keeps the device (the soft tier still drops it if the
+  real connect then fails), so a gateway without status PVs never silently
+  drops telemetry.
+
+### Added
+
+- `GeecsSession.live_devices(devices, *, timeout=3.0)` — concurrent gateway
+  `CONNECTED` liveness probe returning the live/indeterminate subset
+  (fail-open; mock-mode returns all).
+
 ## [0.29.0] - 2026-07-10
 
 ### Changed
--- a/GeecsBluesky/geecs_bluesky/scan_request_runner.py
+++ b/GeecsBluesky/geecs_bluesky/scan_request_runner.py
@@ -824,9 +824,22 @@ def build_telemetry_readables(
     selected = select_telemetry_variables(
         save_set, scalar_policy.subscribed_by_device()
     )
+    # Gate on the gateway CONNECTED PV first (one fast concurrent read per
+    # device — the same liveness source pre-flight uses) so a dead telemetry
+    # device is skipped immediately instead of burning a full connect
+    # timeout. Fail-open: an unreadable status keeps the device (session.
+    # telemetry still drops it softly if the real connect then fails).
+    live = session.live_devices(list(selected))
     readables: list = []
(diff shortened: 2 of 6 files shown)


skbarber:
Closing — the diagnosis that motivated this was wrong. Driving GEECS-Console live surfaced a ~40 s telemetry stall, which I first mis-attributed to dead devices / a CONNECTED-gate, then to a soft-tier connect-timeout. Hardware probing + @skbarber's domain knowledge traced it to the real cause: **a stale orphan row in `expt_device_variable`** — PicoscopeV2's old `charge_pC` variable was deleted from the device type but its `get='yes'` row was never removed (the LV DB GUIs don't cascade devicetype_variable deletions into expt_device_variable — a one-way street). The gateway serves PVs from real variable names, so the orphan `Charge pC` has no PV → the connect hangs. It's a data-hygiene fix (delete the orphan row), not an engine change. Across all 89 Undulator devices an audit found exactly 2 such orphans (both `Charge pC` on the ICTs). Robustness follow-up (per-variable telemetry connect tolerance) tracked in #494.

Issue #497, opened 2026-07-11
GEECS-Console: tooltips on controls (short descriptions + deep links to readthedocs)


skbarber:
**Idea (for later, not blocking):** add hover tooltips to GEECS-Console controls — short plain-language descriptions of what each field/button does, and where it helps, a deep link into the published docs.

## Why it's a natural fit
- The console builds a `geecs_schemas.ScanRequest`, and **the schema fields already carry human descriptions** (`Field(description=...)`) that the docgen renders into the readthedocs schema reference. So a chunk of the tooltips can be **auto-sourced from the schema** rather than hand-authored — one source of truth, and the tooltip text stays in sync with the docs by construction.
- For controls that map 1:1 to a schema field (rep rate, acquisition mode, shots-per-step, trigger profile/variant, save sets, axes/positions), the tooltip = that field's schema description; a "learn more" affordance can deep-link to the field's anchor in the readthedocs schema reference.
- For app-level controls with no schema field (Start/Stop, health chips, device panel, presets), short hand-written tooltips.

## Rough scope
- A small helper that, given a widget's schema-field path, sets `widget.setToolTip(...)` from the schema description (Qt tooltips render limited rich text / links via `QLabel`+`linkActivated` or `QToolTip` with an anchor).
- Deep-link target: the readthedocs schema reference page anchors (docgen already emits per-field sections; confirm/син anchors are stable and linkable).
- Health chips: tooltip showing the last poll detail (heartbeat/devices-connected/version for gateway, URL for tiled, etc.) — pairs well with the health-chips work.

## Related
- Schema descriptions + docgen live in `GEECS-Schemas` (the docgen renders the reference).
- Health-chip tooltips fold in naturally with the health-probe seam (console 0.2.0).

Low priority, high polish — good "operator-friendliness" pass once the functional seams (dialogs, device panel, presets) are wired.

skbarber:
Closing (Sam's call, 2026-07-16): **phase 1 shipped** in #521 — tooltips on all editor form fields derived from the geecs-schemas `Field(description=...)` texts (single source of truth; a poor tooltip means fixing the schema description), hand-written operator-language tooltips on the main-window controls, gated by Preferences → Show tooltips (an event-filter installed only while off, so the suppressor costs nothing when tooltips are on).

[...]

PR #521, opened 2026-07-13
Console UX wave: tooltips, #510 worker routing, optimization dropdown (0.7.0)

### LINKED ISSUE #510: Console: route device set completion through worker QObject
## Context

Post-merge adversarial review of the merged `feat/greenfield-epics-bluesky-gui` branch, covering Console PRs #499/#501-#508.

## Finding

`MainWindow._run_device_set()` emits `MainWindow.device_set_finished` from the daemon set-dispatch thread. That signal is owned by the window, not by a worker QObject.

Relevant code:

- `GEECS-Console/geecs_console/app/main_window.py:300-306` defines `device_value_ready` and `device_set_finished` as `MainWindow` signals; the comments state they are emitted from CA monitor / set-dispatch daemon threads.
- `GEECS-Console/geecs_console/app/main_window.py:601-609` connects them with queued connections.
- `GEECS-Console/geecs_console/app/main_window.py:1547-1552` starts the `console-device-set` daemon thread.
- `GEECS-Console/geecs_console/app/main_window.py:1554-1565` emits `self.device_set_finished` directly from that daemon thread.
- `GEECS-Console/CLAUDE.md:222-226` says the blessed pattern is `BackgroundResult`, and explicitly warns that daemon threads must emit on the worker QObject, never on the window, because window-owned cross-thread signals race teardown and have segfaulted under offscreen pytest.

## Risk

This violates the load-bearing GUI threading invariant: no window-owned Qt signal should be emitted from a daemon thread. Closing the window while a set is in flight can race the daemon emission against widget teardown.

## Suggested fix

Route the blocking device-set operation through the existing `BackgroundResult` / worker-QObject pattern, or introduce a short-lived set worker QObject whose signal is emitted by the worker rather than `MainWindow`.

## Regression test

Add an offscreen Qt test that starts a device set with a blocking/fake backend, closes the window before completion, then releases the backend and verifies no window-owned signal emission, crash, or stale in-flight state.

skbarber:
One PR, three items on GEECS-Console (minor bump 0.6.3 → 0.7.0).

## A. Tooltips (issue #497 phase 1 — text only, no doc links)

- New `services/schema_tooltips.py`: `apply_schema_tooltips` maps pydantic model fields → widget tooltips from `Field(description=...)`. A mapping to a missing or description-less field **fails loudly at editor construction** — the schema stays the single source of truth.
- Applied in all four editors (save sets, scan variables, shot control, action library). The Device/Variable edit pairs share their joint schema field's description.
- **No GEECS-Schemas changes were needed**: an audit of every model in the package found zero fields without a description, so the planned schema backfill commit (and patch bump) has no content. GEECS-Schemas is untouched by this PR.
- Main-window operator controls (mode radios, acquisition combo, shots/step, save-set lists, device panel, Start/Stop, health chips, presets row, R6 panel, R1 combos) get hand-written operator-language tooltips in `_apply_operator_tooltips`.
- Tests (`tests/test_tooltips.py`): editor tooltips compared verbatim against the schema descriptions; ~30 representative window controls asserted non-empty and not label restatements; helper unit tests (KeyError / LookupError contracts).

## B. Issue #510 — device-set completion via worker QObject

- The window-owned `device_set_finished(bool, str)` signal (emitted from the set-dispatch daemon thread) is gone. The blocking `backend.set` now dispatches through a dedicated `BackgroundResult` worker; its `result_ready` delivers `(ok, message)` queued to the GUI thread. The worker owns the cross-thread emission and survives window teardown, so the 0.6.2 `RuntimeError` swallow is deleted rather than papered over.
- `closeEvent` disconnects the set worker like the other `BackgroundResult` workers.
- The deterministic regression test (blocked fake backend released after window C++ deletion, no unhandled thread exceptions) stays green with its docstring updated.
- Closes #510.

## C. Optimization dropdown (GUI half — engine-gated)

- Selecting the Optimization radio shows an optimizer-config combo in R3, listing YAML stems from the experiment's `optimizer_configs/` folder (the legacy GEECS-Scanner-GUI folder name). Offline ⇒ empty combo and Start stays disabled; a selected config is the only optimize-specific Start gate.
- `ConsoleConfigs.optimization_spec(name)` loads a config as a validated `OptimizationSpec`: new-schema documents directly, and the legacy `vocs` dialect through `geecs_schemas.convert.convert_optimizer_config` (recognized unambiguously by the `vocs` key).
- `build_scan_request` stays pure: the window resolves name → spec when snapshotting the form (`ConsoleFormState.optimization`), and optimize forms map to `mode: optimize` requests. `form_state_from_request` round-trips optimize instead of raising, so optimize presets save/apply; Apply matches the preset's inline spec against the listed configs by content (no match ⇒ status-bar error, form untouched).
- **No pre-blocking**: the current engine still refuses optimize submissions; the console submits and surfaces the engine's answer in the status bar, so the parallel engine-side work lands with zero GUI changes.
- Tests: optimize round-trips (incl. YAML serialization parity), combo population/visibility/gating, submission shape, engine refusal and exception surfacing, preset content-matching, and `ConsoleConfigs` against a tmp configs tree (both dialects + error paths).

## Verification

`QT_QPA_PLATFORM=offscreen poetry run pytest -q`: **613 passed, exit 0**, across repeated runs (baseline 553).

Heads-up: a pyproject/CHANGELOG version conflict with parallel console PRs is expected — whoever merges second takes the next free number.

### DIFF
--- a/GEECS-Console/CHANGELOG.md
+++ b/GEECS-Console/CHANGELOG.md
@@ -22,7 +22,15 @@ semantic.
   hand-written operator-language tooltips in `_apply_operator_tooltips`.
   Pinned by `tests/test_tooltips.py`: editor tooltips compared against the
   schema field descriptions, representative window controls asserted
-  non-empty and not label restatements.
+  non-empty and not label restatements.  A **Preferences → Show tooltips**
+  toggle (checkable, persisted via `ConsoleSettings.show_tooltips`,
+  default on) turns them all off for operators who know the console: an
+  application-level `ToolTipSuppressor` event filter swallows
+  `QEvent.ToolTip` for every console widget — editors included — and is
+  installed **only while tooltips are off**, so the default path adds no
+  per-event filter overhead (an always-installed filter measurably slowed
+  the offscreen suite).  Parented to the window and removed on close, so
+  no dangling suppressor can outlive it.
 - **Optimization mode dropdown (GUI half)**.  Selecting the R3
   Optimization radio now shows an optimizer-config combo listing the
   YAML files in the experiment's `optimizer_configs/` folder (the legacy
--- a/GEECS-Console/CHANGELOG.md
+++ b/GEECS-Console/CHANGELOG.md
@@ -4,6 +4,56 @@ All notable changes to GEECS-Console are documented here.  Format follows
 [Keep a Changelog](https://keepachangelog.com/en/1.1.0/); versioning is
 semantic.
 
+## [0.7.0] - 2026-07-12
+
+### Added
+
+- **Tooltips everywhere** (issue #497 phase 1, text only — no doc links).
+  Editor form fields derive their tooltips from the geecs-schemas pydantic
+  `Field(description=...)` texts via the new
+  `services/schema_tooltips.py::apply_schema_tooltips` helper, applied in
+  all four editors (save sets, scan variables, shot control, action
+  library) — single source of truth; a widget mapped to a description-less
+  or renamed field fails loudly at editor construction.  (No schema
+  descriptions needed backfilling — every geecs-schemas field already
+  carries one, so GEECS-Schemas is untouched.)  The main window's operator
+  controls (mode radios, acquisition combo, shots/step, save-set lists,
+  device panel, Start/Stop, health chips, presets row, ...) get
+  hand-written operator-language tooltips in `_apply_operator_tooltips`.
+  Pinned by `tests/test_tooltips.py`: editor tooltips compared against the
+  schema field descriptions, representative window controls asserted
+  non-empty and not label restatements.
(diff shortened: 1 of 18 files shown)


skbarber:
Follow-up commit ad4100f0 adds **Preferences → Show tooltips** (checkable, persisted via `ConsoleSettings.show_tooltips`, default on) per Sam's request — always-on tooltips get annoying once you know the console. One application-level `ToolTipSuppressor` event filter gates every console widget (editors included); it is installed only while tooltips are off, because an always-installed per-window app filter measurably slowed the offscreen suite (~6s → ~27s on the window-heavy test files). Parented to the window and removed on close. Suite: 621 passed, exit 0, twice.

PR #525, opened 2026-07-13
Console: wire the optimization_loader — optimize runs end-to-end (0.9.0)

skbarber:
## Summary

Closes the last gap of M4 step (iii): after the engine half (#520, GeecsBluesky 0.31.0, already on this branch via the v1 sync) and the dropdown GUI half (#521), the console still constructed `BlueskyScanner` **without** an `optimization_loader`, so every optimize submission hit the engine's needs-a-loader refusal. This PR wires the loader; optimize-mode ScanRequests now run end-to-end from the console.

### GEECS-Console 0.8.0 (~230 LOC src, rest tests/lock)

- **`services/optimization.py`** (new, the only module in the console that touches `geecs_scanner`):
  - `optimizer_config_from_spec` — pure mapping from the request's `OptimizationSpec` onto the `BaseOptimizerConfig` dict shape. Pinned by a round-trip test as the **exact inverse** of `geecs_schemas.convert.convert_optimizer_config` (`generator.options` ↔ `xopt_config_overrides[generator.name]`). `max_iterations` deliberately unmapped — the engine consumes it from the spec.
  - `load_console_optimization` — the loader: `BaseOptimizer.from_config(mapped_config)` wrapped in `SessionOptimizationBridge` (the same bridge shape the engine binds: `bind(devices=, scan_tag=, scan_folder=) → (objective, suggester)`, plus `finish()`).
  - `make_optimization_loader` — availability-gated by a light `find_spec` probe (no Xopt import at submitter construction; heavy imports happen inside the loader call on the scan thread). Unavailable → `None` → the engine's explicit refusal shows in the status bar; every other mode unaffected.
- **`submission.py`** — `make_bluesky_submitter` passes `optimization_loader=make_optimization_loader()`.
- **`pyproject.toml`** — new opt-in **`optimization` extra** pulling `geecs-scanner-gui` (path dep, optional). Judgment call, flagged for veto: this is the first appearance of the legacy package in the console's dependencies. It's extras-gated, confined to one module, and explicitly temporary — deleted together with `services/optimization.py` when the redesigned hook (bluesky-adaptive direction) replaces the stack. The `no_geecs_python_api` pin is untouched and still green (hermetic env never imports the stack).

### GEECS-Scanner-GUI 0.34.0 (~60 LOC)

- **`BaseOptimizer.from_config(config_dict, config_dir=None, ...)`** — dict-shaped twin of `from_config_file`, which now delegates to it (identical behavior; relative `seed_dump_files` resolve against `config_dir`, i.e. the config file's directory on the file path, and are left as-is with the existing warning when there is no file origin). Needed because the console builds the config programmatically from the spec — no YAML file involved.

### Deliberately not included

- Optimizer `device_requirements` auto-provisioning on the delegated path (engine decision in #520): the request's save sets must name the objective's diagnostics.
- Relative `seed_dump_files` resolution on the spec path (no config-file directory once the spec is inline in the request) — use absolute paths in warm-starting configs; documented.

## Testing

Hermetic (no hardware/DB/network; the `optimization` extra is NOT installed in the test env — fakes/`sys.modules` injection cover the loader construction):

- **GEECS-Console**: `QT_QPA_PLATFORM=offscreen poetry run pytest -q` → **629 passed** (621 baseline + 8 new in `tests/test_optimization_loader.py`: full-field mapping, overrides-omitted case, converter round-trip inverse pin, availability gating both ways, bridge construction via fake modules, submitter injection parametrized over availability).
- **GEECS-Scanner-GUI**: `tests/optimization` → **120 passed** (117 baseline + 3 new in `test_base_optimizer_from_config.py`: from-dict construction, file→dict delegation equivalence, seed-path resolution only-with-config_dir).
- `poetry lock` regenerated for the optional dep (resolution verified).


skbarber:
Merged `feat/greenfield-epics-bluesky-gui` back in to resolve the version collision with the scan browser (#523), which took 0.8.0 — this PR is now **console 0.9.0**. Conflicts were confined to `pyproject.toml` (kept both dependency additions), `CHANGELOG.md` (entries stacked, 0.9.0 above 0.8.0), and `poetry.lock` (regenerated). Full merged suite: **659 passed** (browser + optimization-loader tests together).

Issue #540, opened 2026-07-13
Background telemetry serializes ~87 CA reads per shot — halves 1 Hz scan rate over VPN


skbarber:
Found while investigating slow 1 Hz scans (Undulator 2026-07-13, Scans 1–3, submitted from GEECS-Console over VPN).

## Symptom
- Free-run NOSCAN: camera `acq_timestamp` deltas are *exactly* 2.000 s — the 1 Hz trigger and cameras are healthy, but every other frame is skipped (row cycle > 1 trigger period).
- Strict NOSCANs: SINGLESHOT cadence 1.6–2.0 s/shot (directly measures the software cycle).
- Scans that previously kept up at 1 Hz. What changed: submission moved to GEECS-Console → the ScanRequest path → `background_telemetry` inherits the experiment default **True** (`geecs_schemas/experiment_defaults.py`). The legacy exec_config path never ran telemetry.

## Mechanism
`run_scan_request` appends the Tier-2 telemetry readables to the detector list (`scan_request_runner.py` ~1022–1028). Today that is **87 `CaTelemetryReadable` devices / 376 columns** (785 columns per event row total, verified in Tiled). Each event row reads the detector list device-by-device — 87 sequential awaited CA gets. Devices are never staged, so ophyd-async's caching monitors are off and every read is a real network round trip.

Measured on the VPN link: ~6–8 ms per warm caget, ~1 RTT for a concurrent batch. 87 × (RTT + RE message overhead) ≈ 0.6–0.9 s/shot from telemetry alone; with the blocking SINGLESHOT put and the 785-column event emission the cycle lands at the observed 1.6–2.0 s. On a control-room machine (sub-ms RTT) the same loop costs ~0.1 s — which is why this was invisible in earlier lab sessions.

The `CaTelemetryReadable` "soft" guarantee covers *failures* (null cells, never gates a shot) but not *latency*; the schema description's "read from the gateway's always-on monitor cache … cannot slow or stall the scan" is not true as implemented — the gateway answers from its cache, but the client pays one round trip per device per shot, serialized.

## Operator workaround
Set `background_telemetry: false` on the ScanRequest (also the confirmation test: cadence should return to 1.0 s).

## Fix options (not mutually exclusive)
[...]

PR #541, opened 2026-07-13
GeecsBluesky 0.32.0: read-path hardening phase 1 — stage per-row devices (#540)

skbarber:
Phase 1 of the read-path hardening plan (full audit + phases 2–4 in `Planning/device_read_path/00_overview.md`, added here). Fixes the root cause behind #540: no plan ever staged a device, so every per-shot read was one uncached network CA get per signal, serialized across ~87 telemetry devices by the RunEngine — a 1 Hz scan ran at exactly 0.5 Hz.

Per-concern breakdown (+579/−14, of which ~230 is the Planning doc and ~250 is tests):

**The fix (~12 LOC)** — `build_step_scan_plan` wraps the composed plan in `bpp.stage_wrapper` over every per-row read device. Staged signals get one caching CA monitor for the scan's duration; per-shot reads become local cache hits. Correctness rests on the gateway's frame-ordering contract (data variables post before the timestamp ladder, PV_CONTRACT §3) — the full coherence chain was verified against installed caproto 1.3.0 / aioca 2.1 / ophyd-async 0.19.3 sources, and the conditions it stands on are written down in CLAUDE.md "Read path: staging & shot coherence".

**Rate-derived bounds (1 Hz behavior unchanged; 5 Hz is the design target, verified at 1 Hz per operator constraints):**
- contributor grace wait capped at half a trigger period (was fixed 0.3 s — 1.5 periods at 5 Hz, serialized per lagging contributor)
- t0-sync window capped at `0.4 / rep_rate_hz`; effective value recorded in the start doc (a window wider than a period silently seeds devices one shot apart)
- telemetry per-signal read budget 2 s (was ophyd's 10 s default — one hung PV held the row; also covers the staged first-read-never-delivers hazard)
- shot queue 32 → 128

**Observability** — DEBUG-level phase timing: strict fire/frame-wait durations, free-run row duration.

**Tests** — `tests/test_read_path_staging.py` (8 new): zero backend gets per row (counting-proxy pin), stage/unstage bracket the run including the abort path, grace/t0 caps, hung-telemetry bound. Full suite: **465 passed**.

**Hardware verification at 1 Hz** (next step, at the lab): free-run NOSCAN row cadence must return to 1.00 s from today's exactly-2.00 s. Framework headroom at the 5 Hz / 50-device target was benchmarked at ~16 ms per 200 ms row.

### DIFF
--- a/GeecsBluesky/CHANGELOG.md
+++ b/GeecsBluesky/CHANGELOG.md
@@ -4,6 +4,48 @@ All notable changes to `geecs-bluesky` are documented here.
 
 Format follows [Keep a Changelog](https://keepachangelog.com/en/1.1.0/).
 
+## [0.32.0] - 2026-07-13
+
+### Fixed
+
+- **Per-shot reads no longer issue one network CA get per signal per row**
+  (issue #540 — scans at a 1 Hz trigger ran at 0.5 Hz once background
+  telemetry's ~87 devices joined every event row).
+  `build_step_scan_plan` now stages every per-row read device
+  (`bpp.stage_wrapper`, unstaged on every exit path), so each signal gets a
+  caching CA monitor for the scan's duration and per-shot reads are served
+  locally.  Cached reads are shot-coherent by the gateway's frame-ordering
+  contract (data variables post before the timestamp ladder,
+  PV_CONTRACT.md §3) — chain verified against installed caproto/aioca/
+  ophyd-async sources; conditions documented in CLAUDE.md "Read path:
+  staging & shot coherence".  Pinned by `tests/test_read_path_staging.py`
+  (zero backend gets per row; stage/unstage bracket the run, abort path
+  included).
+- **One hung telemetry PV can no longer hold an event row for 10 s**:
+  `CaTelemetryReadable` bounds each signal read at 2 s
+  (`_read_timeout_s`); a timeout degrades to the existing null cell.
+
+### Changed
+
+- **1 Hz-era constants now scale with the configured rep rate** (5 Hz
+  system-limit readiness; behavior at 1 Hz unchanged):
+  - the free-run contributor grace wait is capped at half a trigger period
+    (`FreeRunContributorSupport._effective_grace_wait_s`; was a fixed
+    0.3 s — 1.5 shot periods at 5 Hz, serialized per lagging contributor);
+  - the t0-sync acceptance window is capped at `0.4 / rep_rate_hz`
+    (a window wider than one period silently accepts caches seeded one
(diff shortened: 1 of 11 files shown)


skbarber:
**Live verification (partial — read path only), 2026-07-13 evening, on the lab VPN against the production gateway:**

62 devices (the ~telemetry-shaped load, one `MaxCounts` signal each over the two Amp4 IR cameras), read sequentially like the RunEngine's per-event loop — read-only, no puts, no scan:

```
UNSTAGED row read: 527, 551, 606, 516, 492 ms   (one CA get per signal, serialized)
STAGED  first row:  41.7 ms                      (one-time monitor warmup)
STAGED  row read:  7.2, 6.1, 6.7, 6.7, 6.8 ms   → 79x
```

The unstaged cost (~0.5 s/row) reproduces exactly the telemetry term that halved today's 1 Hz scans (#540); staged reads collapse it to ~7 ms — comfortably inside the 5 Hz budget with the measured ~16 ms framework cost on top.

**Still owed:** an end-to-end free-run NOSCAN cadence check (expect 1.00 s rows, was exactly 2.00 s). The trigger is currently quiescent and firing it means putting the DG645 into SCAN — which gates the gas jet at Undulator, so that's an operator action. A/B script is ready (`verify_staging_live.py` in the session scratchpad; runs `save_data=False`, no folder claim, staging bypassed vs enabled).

skbarber:
**End-to-end live verification complete** (2026-07-13 evening; DG645 on internal trigger at 1 Hz, operator-enabled; free-run NOSCAN, `save_data=False`, 120 snapshot devices over real gateway PVs as read load — heavier than the morning's 87-device telemetry shape):

**Run A — staging bypassed (the pre-PR code path):**
```
camera acq_timestamp deltas: [3.0, 2.0, 2.0, 2.0, 2.0, 2.0, 1.0]   median 2.0 s/row
```
The bug reproduces exactly: rows land on every *second* 1 Hz tick — the same integer-quantized signature as this morning's Scans 1–3.

**Run B — staged (this PR), same load, same trigger:**
```
camera acq_timestamp deltas: [1.0, 1.0, 1.0, 1.0, 1.0, 1.0, 1.0]   median 1.0 s/row
wall-row deltas: 0.99–1.01 s
```
Every trigger tick captured, at ~2× the device load of the morning's scans.

Together with the earlier isolated measurement (62-device row read: 527 ms unstaged → 6.7 ms staged, 79×), the hardware-verification item for this PR is closed. 5 Hz remains designed-for/untested per the operator's constraints — the per-row read cost is now ~7 ms against a 200 ms budget.

PR #544, opened 2026-07-13
GeecsBluesky 0.35.0: telemetry shot context (read-path phase 4, owner-decided)

skbarber:
Implements phase 4 exactly per the owner decisions (design note updated to DECIDED in this PR):

- **D1 → (a):** async devices carry no derived labels — and no `systimestamp` column either (Sam's refinement: it advances with the device loop even when nothing is acquiring, so it can be meaningless as an alignment signal). Every telemetry device does record its own raw `telemetry_<device>-acq_timestamp` (the gateway serves the PV for every device; async devices honestly show the `0.0` placeholder).
- **D2 → tabled:** no classification stage, no observation window, no cache, no DB toggle. Strict synchronization remains a save-set concept.
- **D3 → full companion set under the D2 constraint:** on free-run scans, a telemetry device whose cached `acq_timestamp` is **positive at the quiesced t0 snapshot** (observed to have actually fired — classification collapses into the seed itself, for free) is seeded like a contributor and emits `shot_id`/`shot_offset`/`valid`, computed with no grace wait (telemetry never gates a shot). Everything else — async, never-fired, strict-mode scans — stays value-columns + raw acq only. The start doc lists seeded devices under `telemetry_shot_seeded`.

Mechanics: `CaTelemetryReadable` gains `ShotIdSupport` + an optional `shot_rep_rate_hz` (session factories pass the session rep rate); the free-run plan anchors group members to the reference and soft-seeds them via `bps.rd` right after t0-sync (staged cache reads — zero added latency, no trigger authority). `EVENT_SCHEMA.md` gets the additive telemetry-shot-context section (explicitly *not* a schema version bump), including the consumer rule: key attribution on the `acq_timestamp` column, never a data column's reading timestamp (deadband makes those last-change times).

+308/−5; 3 new tests (fired member gets companions matching the reference; never-fired member gets the honest 0.0 and no labels; session wiring); full suite **480 passed**.

Live verification on the 2-camera rig to follow as a comment (save set with one camera → the other becomes triggered telemetry, per the design note's verification plan).

### DIFF
--- a/GeecsBluesky/CHANGELOG.md
+++ b/GeecsBluesky/CHANGELOG.md
@@ -4,6 +4,26 @@ All notable changes to `geecs-bluesky` are documented here.
 
 Format follows [Keep a Changelog](https://keepachangelog.com/en/1.1.0/).
 
+## [0.35.0] - 2026-07-13
+
+### Added
+
+- **Telemetry shot context (read-path phase 4, per the owner decisions in
+  `Planning/device_read_path/01_telemetry_attribution.md`).**  Every
+  telemetry device now records its own
+  `telemetry_<device>-acq_timestamp` column (the gateway serves the PV
+  for every device; async devices honestly show the `0.0` never-acquired
+  placeholder — no derived labels are manufactured for them).  On
+  free-run scans, telemetry devices **observed to have fired** (positive
+  cached `acq_timestamp` at the quiesced t0 snapshot — no classification
+  stage, no config flag, no DB toggle) are seeded like contributors and
+  additionally emit the standard `shot_id`/`shot_offset`/`valid`
+  companions, computed without any grace wait (telemetry never gates a
+  shot).  The start doc records the seeded device names under
+  `telemetry_shot_seeded`.  Additive column convention — no event-schema
+  version bump (EVENT_SCHEMA.md updated).  Strict-mode scans record the
+  raw acq column only (no t0 stage).
+
 ## [0.34.0] - 2026-07-13
 
 ### Changed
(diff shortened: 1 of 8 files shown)


skbarber:
Hermetic verification complete (480 passed; the three phase-4 pins cover fired-gets-companions, never-fired-stays-plain, session wiring).

**Live verification attempted 13:06 but the trigger is no longer firing under SCAN** (the rig state changed since the 12:44 scans — consistent with the system shutdown in progress). Scan012 was claimed before the trigger timeout and is a folder with scan.log and no event data (loudly logged, left in place per the invariant).

**One-scan acceptance test when the trigger is back** (script-ready): free-run NOSCAN with save set `Amp4In` only → `UC_Amp4_IR_output` lands in telemetry. Expect in the run: `telemetry_uc_amp4_ir_output-acq_timestamp` advancing at 1 Hz, `-shot_id` matching the reference's, `-shot_offset == 0`, `-valid == True`; start doc `telemetry_shot_seeded` includes the output camera (plus any other telemetry device with a stale positive acq_timestamp from earlier sessions — those will show honestly huge negative offsets and `valid=False`, which is the design working, not a bug). Leaving this PR unmerged until that run.

skbarber:
**Live acceptance PASSED — 2026-07-15, Scan009** (run uid `3ac1308d`), exactly per the closing comment's recipe (free-run NOSCAN, save set Amp4In only, 10 shots @ 1 Hz, run headless from this branch's worktree via `session.run(ScanRequest)`):

- `telemetry_shot_seeded`: 41 of 88 devices, **UC_Amp4_IR_output included**
- `telemetry_uc_amp4_ir_output-acq_timestamp`: advancing at exactly 1.00 s
- `-shot_id`: equals the reference camera's shot_id on every row
- `-shot_offset`: all 0 · `-valid`: all True
- Never-fired example `uc_hiresmagcam`: honest 0.0 acq placeholder, no companion labels

This PR's base branch (`feat/vision-v1`) was deleted in the branch collapse, so it cannot be reopened — the branch will be rebased onto `dev` and re-filed once the in-flight engine PRs land (re-file precedent per /land).

PR #548, opened 2026-07-14
tooling: check.sh CI-mirror runner + /check skill (stacked on #546)

skbarber:
**Stacked on #546 — merge that first** (the diff shows #546's commits until it lands; only `654807c9` + `b370ec02` are new here). No package code changes; repo tooling only. Retitled scope after an adversarial pass: **/lab-day was removed** — it consumed an OWED hardware-verification ledger that has zero entries until the PR template sees real use; it can be re-proposed after the first post-template lab day, against observed reality.

Per-concern breakdown:

**`scripts/check.sh` (~230 LOC)** — runs lint + unit tests exactly the way `unit-tests.yml` does, scoped to what changed. Owns the env/marker mapping (root env vs package env, `not integration and not fake_server` for GeecsBluesky, `QT_QPA_PLATFORM=offscreen` for Console, exit-5 tolerance for root `tests/`); when CI changes, the runner table here changes in the same PR. Auto-picks the diff base (nearest of vision-v1 / greenfield / master by merge-base distance), flags touched packages that have **no** CI-mirrored suite instead of silently skipping them, and points env-shaped failures at `/env-doctor`. Modes: `--lint`, default (changed packages), `--all`, explicit package list, `--base`, `--dry-run`.

Hardening from the adversarial pass: NUL-delimited change detection (the repo tracks 48 filenames with spaces — proven live against the `Bella Normal.yaml` fixture, which the first version word-split into two bogus paths); untracked files are deliberately invisible until `git add` (scratch files inside a package no longer trigger its suite); two `[ -d ] && …` guards rewritten as `if` — under `set -e` they killed the script, so `--all` silently died on any branch without GEECS-Console (i.e. this one).

Verified: `bash -n`; `--dry-run` maps this branch's diff → GeecsBluesky; spaces simulation (modified `Bella Normal.yaml` → GEECS-Schemas suite selected, lint passes with the quoted path); `--all --dry-run` on this Console-less branch plans 8 suites without dying; suite mode from the main checkout: GEECS-Data-Utils **197 passed**, GEECS-Schemas **211 passed, 7 deselected**.

**`/check` skill (~50 LOC)** — when to run which tier (lint before commit / changed before push / `--all` before PR) and how to read the output; explicitly forbids re-deriving the CI mapping in prose.

**`/env-doctor` dedup (−13 LOC)** — its "which env runs which tests" table shrunk to a pointer at `check.sh --dry-run`. One canonical mapping; no prose copy to drift (the same critique that consolidated the branch topology into CONTRIBUTING.md in #546).

**`tests/test_skill_frontmatter.py` (~55 LOC, runs in the CI root-tests line)** — pins frontmatter well-formedness, `name` == directory, and a non-empty `description` for every `.claude/skills/*/SKILL.md`. A malformed skill drops out of agent sessions with no error anywhere; this makes it loud. **2 passed** against the five current skills.

**Pointer updates (~10 LOC)** — `/land` step 4 → `./scripts/check.sh`; CLAUDE.md + CONTRIBUTING.md skill lists.


skbarber:
Added `/lab-status` + `scripts/lab_status.sh` to this PR (commit 47b20184) per Sam — bounded lab-reachability probes (the recurring 75-second-hang / is-the-trigger-firing stumble, now a 2-8 s answer). Two tiers: network (TCP/HTTP/mount, watchdog-bounded — `nc`'s timeout flags don't reliably bound a SYN into a blackholed VPN route, so `/dev/tcp` + explicit kill) and read-only CA gateway liveness behind `--hardware`. Skill prose = the state→capability judgment table + standing authorization rules; script = the canonical endpoints (derived from config.ini). Frontmatter test passes (2 passed). Live-verified in both network states today — including an 8 s correct "hermetic work only" verdict against a half-down VPN just now.

Issue #552, opened 2026-07-14
G-actions v2: pause scan → operator decides → execute action → resume


skbarber:
Sam's design (2026-07-13), deferred from G-actions v1 (which refuses actions during a scan):

When an operator triggers an ad-hoc action while a scan is running, instead of refusing: **pause the scan** (`RE.request_pause()`, suspends at the next checkpoint), show a dialog — *"Action '<name>' was triggered. Scan paused. Execute the action and continue, or ignore it and continue the scan?"* — then execute the compiled action steps **outside the RE** (the paused RE still holds the scan plan, so execution goes through the same compiled `GatewaySetpointPut` steps directly) and `RE.resume()`.

Scope notes:
- Applies ONLY to ad-hoc operator actions; setup/per-step/closeout actions are part of the scan plan and unaffected.
- Design work owed before code: shot-controller state during the pause window (jet gating, strict-mode ARMED state), pause-during-step vs between-steps semantics, what the data row spanning the pause looks like, abort-during-pause.
- Builds on G-actions v1's `run_action`/compiled-steps executor — same steps, different dispatch.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

skbarber:
# Design note — G-actions v2: the pause/decide/resume engine flow (#552) and the action firing surface (#575)

Status: draft for decision. Written against `dev` @ b920dfcc. Sources: `GeecsBluesky/CLAUDE.md`, `GEECS-Console/CLAUDE.md`, `geecs_bluesky/shot_controller.py`, `plans/action_compiler.py`, `plans/step_scan.py`, `plans/free_run_step_scan.py`, `plans/single_shot.py`, `session.py` (`run_action`, `scan`), `scanner_bridge/bluesky_scanner.py`, and bluesky 1.15 `run_engine.py` (the installed copy).

---

## 1. Goals and non-goals

**Goals.** When an operator fires an ad-hoc action while a scan is running, the system should pause the scan at a safe boundary, put the machine into a safe quiescent state (on Undulator: gas jet not flowing), let the operator decide — execute the action, ignore it, or abort the scan — then restore the machine state and resume, with no data row corrupted, duplicated, or silently replayed, and with the pause recorded. Separately, the everyday firing surface for actions should make "wrong action" and "accidental Run" misfires structurally hard, not just guarded by one toggle. Both halves should land in small, independently verifiable PRs, hermetically testable except for a bounded list of lab checks.

[...]

PR #554, opened 2026-07-14
GEECS-Console: Actions menu with execute gating + dry-run preview (G-actions v1)

skbarber:
The console half of G-actions v1: the Actions menu goes from a disabled "(not wired yet)" placeholder to the full arm-preview-run flow, built against the engine contract with fakes (the engine half lands in its own PR).

## What's in the menu

- **Action-plan entries**: the current experiment's plan names from `ActionLibraryStore.list_names()` (the same `action_library/actions.yaml` the Action Library editor edits), fetched on a `BackgroundResult` daemon worker (never blocking the GUI thread on the configs repo), refreshed on experiment change, stale results dropped by experiment tag. Empty/offline renders one disabled "(no actions)" entry.
- **Execute gating (the accidental-click guard)**: a checkable "Enable action execution" item at the top. **Default OFF at every launch, deliberately NOT persisted** — a fresh session never starts armed. While unchecked, every action dialog is preview-only. Toggles are pushed into already-open dialogs.

## The per-action dialog (`app/action_dialog.py`)

Clicking a plan opens a non-modal `ActionRunDialog` (kept in `self._open_action_dialogs` — the PySide6 GC hazard): the plan name, a read-only steps table from `describe_action` (#, kind, device, variable, value, wait (s), from plan — execution order), and [Run]/[Close]. Run (armed-only) dispatches the blocking `run_action` on a dialog-owned worker: in flight the button disables and the status bar shows "running action '<name>'…"; success → "action '<name>' done"; failure/refusal → the raised message in the status bar **and** inline in the dialog, so the scan-in-progress refusal is never silent. `closeEvent` disconnects the workers — closing during a slow describe returns immediately.

One design fix found by the tests themselves: the async preview and the async run outcome originally shared one label, and a slow `describe_action` landing after a run could clobber the run's failure/refusal message (an intermittent test failure that was a real silent-refusal bug). They now render on **separate labels**, pinned by `test_late_preview_never_clobbers_the_run_outcome`.

## The engine contract (built against fakes — signatures unchanged)

`Submitter` gains exactly the two agreed methods, mapped by `make_bluesky_submitter` to `BlueskyScanner`'s same-named methods (structural — no adapter):

- `run_action(name: str) -> None` — blocking; raises operator-readable messages; during a scan exactly `RuntimeError("scan in progress — action not started")`
- `describe_action(name: str) -> list[dict]` — dry-run step dicts with keys `kind`/`device`/`variable`/`value`/`wait_s`/`from_plan`

No signature deviations to report.

## Per-concern LOC breakdown

| Concern | Files | ~LOC |
|---|---|---:|
| Actions menu wiring (fetch, rebuild, gating, dialog opening, teardown) | `app/main_window.py` | +150 |
| Run/Preview dialog | `app/action_dialog.py` (new) | +280 |
| `Submitter` protocol extension | `submission.py` | +25 |
| `BackgroundResult` extraction (judgment call, see below) | `services/background.py` (new), `main_window.py` (−42) | +62/−42 |
| Tests | `tests/test_actions_menu.py` (new) | +330 |
| Version/CHANGELOG/CLAUDE.md | | +75 |

**Judgment call for cheap veto**: `BackgroundResult` moved from `app/main_window.py` to `services/background.py` — the dialog needs its own workers and importing them from `main_window` would be circular. This is the shared extraction already recorded on issue #510; the browser's `BrowserWorker` twin is *not* swapped here (kept out of scope; CLAUDE.md updated to say the swap is now unblocked).

## Tests

`tests/test_actions_menu.py` (16 new; suite 663 → 679): menu population, experiment-change refresh, empty placeholder, stale-fetch drop, gating default-off, gating not persisted across sessions (real `ConsoleSettings` on tmp INI), arm-toggle pushed into open dialogs, preview table renders the step dicts, describe failure inline, Run calls `run_action` with the right name, refusal surfaced (status bar + inline + log), in-flight disable, late-preview no-clobber, close-during-slow-describe returns fast (<0.3 s), dialog referenced on the window.

Full suite: **679 passed, exit 0** — 8 consecutive clean runs after the label fix (the pre-fix flake was the clobber bug above), no segfaults, `QT_QPA_PLATFORM=offscreen`.

## Left for live verification

- Real `BlueskyScanner.run_action`/`describe_action` through the console against hardware (menu → preview → armed run; the scan-in-progress refusal during a live scan).
- Action-plan names appearing for a real experiment's `actions.yaml` over the network configs mount.
- #552 (pause-the-scan flow for actions) noted as future work in CLAUDE.md.

### DIFF
--- a/GEECS-Console/CHANGELOG.md
+++ b/GEECS-Console/CHANGELOG.md
@@ -4,6 +4,40 @@ All notable changes to GEECS-Console are documented here.  Format follows
 [Keep a Changelog](https://keepachangelog.com/en/1.1.0/); versioning is
 semantic.
 
+## [0.10.0] - 2026-07-14
+
+### Added
+
+- **Actions menu (G-actions v1, console half)**: the menu now lists the
+  current experiment's action-plan names (from the same
+  `action_library/actions.yaml` the Action Library editor edits), refreshed
+  on experiment change via a `BackgroundResult` worker; empty/offline shows
+  a disabled "(no actions)" entry.  Clicking a plan opens an
+  `ActionRunDialog` (`app/action_dialog.py`): a dry-run steps table from the
+  engine's `describe_action` (kind / device / variable / value / wait /
+  from-plan per row, execution order) plus Run/Close.  Run dispatches the
+  blocking `run_action` on a daemon worker — in flight the button disables
+  and the status bar shows "running action '<name>'…"; success reports
+  "action '<name>' done"; a failure or refusal (e.g. the engine's
+  "scan in progress — action not started") shows in the status bar AND
+  inline in the dialog.  Preview and run outcomes render on separate lines
+  so a slow preview can never clobber a refusal (pinned by test).
+- **Execute gating**: a checkable "Enable action execution" item tops the
+  menu — the accidental-click guard.  Default OFF at every launch and
+  deliberately NOT persisted (a fresh session never starts armed); while
+  unchecked every action dialog is preview-only.
+- `Submitter` protocol extended with the engine contract's
+  `run_action(name) -> None` (blocking; raises operator-readable messages)
+  and `describe_action(name) -> list[dict]` (dry-run step dicts) — mapped
+  by `make_bluesky_submitter` to the scanner's same-named methods.
+
+### Changed
+
+- `BackgroundResult` extracted from `app/main_window.py` to
+  `services/background.py` (the shared home recorded on issue #510) so the
+  action dialog can own its workers; the browser's `BrowserWorker` twin
+  still awaits its mechanical swap.
+
 ## [0.9.2] - 2026-07-13
 
 ### Changed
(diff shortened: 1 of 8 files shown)


skbarber:
**Hardware verification (retroactive OWED ledger — the PR template section this body should have carried):**

- OWED: Actions menu populated from a real experiment's actions.yaml (network mount)
- OWED: run an action end-to-end from the dialog on live hardware; refusal surfaced during a live scan
- VERIFIED (hermetic): 679 tests incl. gating default-off/unpersisted, preview table, late-preview-vs-refusal race

skbarber:
**Hardware verification update (Sam, live, 2026-07-16)** — Actions executed on real hardware from the Actions menu: works ('a great start'). Remaining from the OWED list: the mid-scan refusal repro (untested, low risk — the exact-message path is pinned hermetically). Field UX feedback recorded separately: menu-firing feels misfire-prone despite the gate; the *capability* is the accepted deliverable, the firing surface is an open design question (issue to follow).

PR #570, opened 2026-07-16
GeecsBluesky 0.38.2: silence upstream RequestAbort traceback on operator abort

skbarber:
The #563 follow-up its PR body promised, now field-driven: after #563, an operator abort still shows upstream bluesky's `ERROR bluesky: Run aborted` + full `RequestAbort` traceback (observed live this morning). Upstream logs that message from **two** sites — the RequestAbort path and a genuine-exception path — so the fix filters by **exception type, never message**: a `logging.Filter` on the session RE's underlying logger drops RequestAbort-typed records only. Real failures' identically-worded records pass through with tracebacks intact (pinned by a type-discrimination test); an end-to-end live-abort test asserts zero RequestAbort tracebacks reach the stream.

## Tests
`./scripts/check.sh`: lint green, GeecsBluesky **533 passed** (2 new).

## Hardware verification
OWED: one live abort — expect exactly one geecs INFO + one calm WARNING and no traceback blocks at all (rides today's optimize-repro session).

### DIFF
--- a/GeecsBluesky/CHANGELOG.md
+++ b/GeecsBluesky/CHANGELOG.md
@@ -4,6 +4,19 @@ All notable changes to `geecs-bluesky` are documented here.
 
 Format follows [Keep a Changelog](https://keepachangelog.com/en/1.1.0/).
 
+## [0.38.2] - 2026-07-16
+
+### Fixed
+
+- **Upstream bluesky's `Run aborted` traceback silenced for operator
+  aborts** (the #563 follow-up, observed live 2026-07-16): the session
+  attaches a log filter to its RunEngine's logger that drops records
+  whose exception is `RequestAbort` — the abort *mechanism*, not an
+  error. The filter keys on the exception type, never the message, so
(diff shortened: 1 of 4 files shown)


skbarber:
**Adversarial review (fresh-context subagent, per /land) — findings dispositioned:**

1. *(low)* Filter attaches to the process-global `bluesky` logger, not per-engine as the comment/CHANGELOG claimed; attachment was non-idempotent (one duplicate filter per session construction). — **FIXED** in the follow-up commit: idempotence guard (`isinstance` scan of the logger's filters), comment and CHANGELOG reworded to state the process-wide scope honestly (which the reviewer notes is arguably the better behavior — any RE in the process gets quiet aborts), plus a pinning test (`test_abort_log_filter_attaches_once_across_sessions`).

Reviewer verified independently: type discrimination against both upstream logging sites, `RequestStop`/halt paths unaffected, tests fail-if-wrong, no repo duplication (only same-shaped code is the gateway's package-local `_QuietMissingVariables`), placement in session.py correct. No other surviving findings.

PR #572, opened 2026-07-16
GeecsBluesky 0.39.0: delete the legacy exec_config path root-and-stem (G3, executed early)

skbarber:
# GeecsBluesky 0.39.0 — delete the bridge's legacy exec_config path root-and-stem (G3, executed early)

## What + why

`BlueskyScanner.reinitialize` now accepts **only** a `geecs_schemas.ScanRequest` and raises a clear `TypeError` for anything else. The entire duck-typed `exec_config` arm — the legacy GUI's submission shape — is deleted, along with every helper, state field, constructor parameter, and test that existed solely for it.

The cutover doc (`Planning/cutover_strategy/00_overview.md`, status 2026-07-13) had re-timed G3 to M6. This PR executes it early with explicit owner approval — Sam, 2026-07-16:

> "we've totally abandoned the old gui path… if it makes sense, we should remove 'legacy' code in favor of the new approach… We can break the legacy scanner on the dev branch. I mean, we've already deleted the whole engine."

Consequences, accepted and documented (not patched):
- **The legacy GEECS-Scanner-GUI is dead on `dev`** — it breaks at *engine construction*: `RunControl` passes the deleted `shot_control_information` kwarg during main-window init and gets a generic kwarg `TypeError` (uncaught), before any submission could reach the curated `reinitialize` error. Accepted and documented (`GEECS-Scanner-GUI/CLAUDE.md` carries a dead-on-dev banner + 0.35.1 docs-only changelog entry; no GUI code changed — it stays in-tree for its `optimization` module until M6).
- **`master`'s legacy scanner line is untouched** — the fallback story is unchanged.
- **The `optimization_loader` contract is `OptimizationSpec`-only.** The legacy calling convention (optimizer-YAML path string) is gone with the exec_config arm. The console's loader already speaks `OptimizationSpec` (GEECS-Console ≥ 0.9.0), so no console change is needed.
- **`GEECS_BLUESKY_ACQUISITION_MODE` env override is gone** — acquisition comes from `ScanRequest.acquisition`; a request declares intent (this was already the delegated-path rule).
- **`shot_control_information` constructor parameter is gone** — trigger profiles are resolved per request by the runner; the console never passed it.

## What died (deletion inventory)

In `geecs_bluesky/scanner_bridge/bluesky_scanner.py` (1657 → 943 lines, −838/+125):

| Deleted | Was |
|---|---|
| exec_config arm of `reinitialize` | duck-typed `.scan_config`/`.options`/`.save_config` storage |
[...]

Issue #511, opened 2026-07-12
GeecsBluesky: ignore foreign RunEngine event docs in scanner progress


skbarber:
## Context

Post-merge adversarial review of the merged `feat/vision-v1` branch, covering GeecsBluesky PR #500.

## Finding

`BlueskyScanner` guards foreign start documents for scan-number pickup, but it still accepts every RunEngine event document for GUI progress.

Relevant code:

- `GeecsBluesky/geecs_bluesky/scanner_bridge/bluesky_scanner.py:188-194` creates a `GeecsSession`, stores `self._RE = self._session.RE`, and subscribes `self._on_document` to the RunEngine.
- `GeecsBluesky/geecs_bluesky/scanner_bridge/bluesky_scanner.py:547-553` handles every `"start"` and `"event"` document. Event docs always increment `self._completed_shots` and call `_emit_step_progress(doc)`.
- `GeecsBluesky/geecs_bluesky/scanner_bridge/bluesky_scanner.py:555-584` guards only scan-number pickup: `_note_claimed_scan_number()` returns unless the scanner is currently `RUNNING`.
- The test added for foreign runs, `GeecsBluesky/tests/test_bluesky_scanner_scan_request_seam.py:612-627`, only exercises a foreign start document. It does not cover event documents from that foreign run.

## Risk

The session RunEngine is shared. A headless or otherwise foreign run driven directly on the scanner's session can still mutate GUI scanner progress while this scanner is idle, done, or otherwise not the owner of that run. The GUI can receive stale/spurious `ScanStepEvent`s and `estimate_current_completion()` can be contaminated.

## Suggested fix

Make `_on_document` ignore event documents unless they belong to the active scanner-owned run. At minimum, gate events on `self._current_state == RUNNING`; preferably also track the active start uid and ignore event descriptors/run_start values that do not match.

## Regression test

Extend the foreign-run test to send a start document while idle followed by one or more event documents, and assert that `_completed_shots` remains unchanged and no `ScanStepEvent` is emitted. Add a second case for a foreign event after this scanner has completed.

skbarber:
Fixed in #515 (merged to feat/vision-v1): full run-uid ownership tracking — start docs claimed only when this scanner owns the run, descriptor→run mapping filters event docs, stop clears the claim. Regression tests for foreign runs while idle and after completion.

PR #573, opened 2026-07-16
GeecsBluesky 0.40.0 + GEECS-Console 0.15.0: native stop lifecycle — Stop during init is honoured, never blocks the GUI (#571)

### LINKED ISSUE #571: Stop during scan initialization: GUI pinwheels and the stop is ignored (delegated path)
Field report (Sam, 2026-07-16): pressing Stop shortly after Start — during initialization — pinwheels the GUI for ~a minute while the terminal keeps logging init steps, and then the scan starts anyway.

**Root cause (two stacked defects, confirmed in source):**

1. **GUI freeze**: `MainWindow._on_stop_clicked` calls `stop_scanning_thread()` on the GUI thread; that method ends in `thread.join(timeout=_THREAD_JOIN_TIMEOUT)`. During init the scan thread is inside device connects (20 s timeouts each), so the join blocks the GUI thread → pinwheel.
2. **Stop ignored on the delegated path**: `stop_scanning_thread` sets `_abort_requested` and correctly skips `RE.abort()` while the engine is idle — the pre-RE stop is *supposed* to be covered by `_abort_before_acquisition()` checkpoints, but those exist only on the legacy exec paths (bluesky_scanner.py ~1003, ~1357). `_run_delegated_request → run_scan_request` never consults the flag during resolution/connect/preflight/claim, and nothing re-issues the abort once the RE starts. The flag is set; nobody reads it; the scan runs.

**Fix shape (two small PRs):**
- Console: dispatch Stop through a `BackgroundResult` worker (never join on the GUI thread); Stop button → disabled + "stopping…" state until the worker reports; re-enable on completion.
- Engine: thread a `should_abort` callable (bridge supplies `lambda: self._abort_requested`) through `run_scan_request`'s init stages — checked between resolution, device connect, preflight, and **before claim** (a stop during init should burn no scan number), plus once immediately before `session.run`/RE start to close the race where the abort lands between the idle-check and plan start.

Severity: UX-medium (operator loses control for ~a minute; scan runs against intent). Workaround: wait for the scan to start, then Stop works (quiet, since 0.38.2).

skbarber:
## What

Fixes #571 the RunEngine-native way, post-G3: pressing Stop during scan
initialization used to (1) pinwheel the console for up to a minute and
(2) be silently ignored — the scan started anyway.

**Engine (GeecsBluesky 0.40.0)**

- `run_scan_request` gains a `should_abort: Callable[[], bool] | None`
  hook (the bridge supplies `lambda: self._abort_requested`), consulted
  between init stages: after configuration resolution, after device
  connect, after the preflight hook, and immediately before the
  scan-number claim (optimize path: after device connect + before the
  runner's own pre-bind claim). **Every runner checkpoint is pre-claim —
  a stop caught there burns no scan number.** (Gate-path caveat, from
  review finding 1: on `session.scan` the claim precedes the in-plan
  gate, so a stop landing in the claim→plan-start window still burns a
  number — folder with ScanInfo, no data, INFO "stopped before the
  plan started". Only the runner checkpoints carry the burns-nothing
  promise; the session docstring and CHANGELOG say so explicitly.) On trip: one INFO line
  ("stopped during initialization … nothing claimed"),
  `session.last_run_aborted = True` (the #563 quiet aborted-outcome
  contract, so the bridge reports ABORTED and optimize `finish()` is
  skipped), created devices disconnected by the runner's existing
  `finally`, `None` returned.
- **Idle-window race closed with an in-plan stop gate** (design choice
  below): `GeecsSession.scan`/`optimize` accept `should_abort` and wrap
  the plan in `_stop_gated`, which re-reads the probe as the plan's very
  first instruction. The engine sets `self._state = "running"` *before*
  fetching the plan's first message (bluesky 1.15.0
  `run_engine.py::_run`, line 1503), so a stopper that observed the
  engine `idle` (and therefore skipped `RE.abort()`) set the flag
  strictly before the gate runs — the gate trips, the engine processes
  zero messages, opens no run, and settles back to idle. A stopper that
  observed `running` used `RE.abort()` directly, which is verified
  correct from another thread while running: `abort()` dispatches
  `_abort_coro` onto the engine loop via `__interrupter_helper`'s
  `call_soon_threadsafe` and, from `running`, cancels the plan task
  without waiting on cleanup (bluesky 1.15.0 `run_engine.py`
  `abort`/`_abort_coro`/`__interrupter_helper`). The alternative — the
  bridge issuing `RE.request_pause()` once the engine reports running —
  was rejected: the `state_hook` fires on the engine loop thread, where
  `request_pause()`'s `run_coroutine_threadsafe(...).result()`
  self-deadlocks, and a pause would additionally strand the engine in
  `paused` needing a second abort step. The gate is pure plan
  composition, no engine internals.
- `stop_scanning_thread` no longer waits for completion: the 15 s join
  became a 2 s bookkeeping-only join (`_STOP_JOIN_TIMEOUT`); a
  still-finishing thread is INFO (normal during init stops), not ERROR;
  the thread handle is still kept so `is_scanning_active()` stays honest.
  Completion is announced natively by the scan thread's terminal
  ABORTED/DONE `ScanLifecycleEvent` (STOPPING is emitted at request
  time, as before).
- Cleanup rider (waived finding from #572's review, explicitly in scope
  here): removed the write-only `_last_run_uid` bridge remnant and the
  caller-less `_request_operator_decision` (its behavior lives on in
  `operator_channel.EventStreamOperator`; docstring cross-ref updated).

**Console (GEECS-Console 0.15.0)**

- `_on_stop_clicked` dispatches `stop_scanning_thread` through a
  `BackgroundResult` worker (the blessed daemon-thread → queued-signal
  shape; no result slot needed — completion arrives via the lifecycle
  stream). The Stop button disables and shows "Stopping…" until
  `_on_scan_state` sees a terminal state (aborted/done — exactly the
  engine's terminal vocabulary; `paused_on_error` deliberately
  excluded — resumable), then the label
  restores and normal gating resumes. The state pill knows the engine's
  STOPPING state (amber).

## Per-concern LOC (production code)

| Concern | Files | ~LOC |
|---|---|---|
| Runner init-stage checkpoints | `scan_request_runner.py` | +74/−5 |
| Session in-plan stop gate | `session.py` | +81 |
| Bridge stop rework + should_abort wiring | `scanner_bridge/bluesky_scanner.py` | +70/−48 |
| Bridge dead-code removal (rider) | `bluesky_scanner.py`, `operator_channel.py` | −40 |
| Console async stop + button state | `app/main_window.py`, `submission.py` | +50/−6 |
| Tests | 6 files | +410 |

## Tests

- **GeecsBluesky: 519 passed, 3 deselected** (was 505; +14 new,
  including the review-fix pin that a stop with no active scan emits no
  STOPPING repaint):
  runner checkpoint tests for each init stage — the pre-claim one
  asserts `session.scan`/`claim_scan` is never called, devices built
  before the trip are disconnected, INFO logged, `last_run_aborted` set;
  probe threaded through to `session.scan`/`optimize` pinned; real-RE
  session-gate tests (gate trips → zero documents emitted, quiet aborted
  outcome; never-tripping probe → normal run untouched; optimize
  gate → `(None, [])`, nothing moved); bridge stop returns promptly
  (timing assertion), still-finishing join is INFO not ERROR, live
  `should_abort` probe wiring. The #563 running-phase quiet-abort pins
  are untouched and green.
- **GEECS-Console: 748 passed** (was 745; +3 new, 1 updated to the async
  contract): stop click returns in <0.3 s with a 0.5 s-blocking
  submitter (the close-during-slow-read pattern), button shows
  "Stopping…" and disables, STOPPING lifecycle event keeps the hold,
  terminal event releases it and restores gating, stop with no active
  scan is a no-op.
- Remaining suites (untouched packages), run the way CI does via
  `scripts/check.sh` after installing the fresh worktree's envs per
  /env-doctor — all green: ImageAnalysis 224 passed / 1 skipped,
  ScanAnalysis 171 passed / 1 skipped, GEECS-Data-Utils 198 passed,
  GEECS-Schemas 211 passed, GeecsCAGateway 181 passed, GEECS-LogTriage
  53 passed; changed-files lint clean.

## Hardware verification

**OWED** (next lab session):
1. Click Stop during scan initialization (within the first seconds after
   Start, while device connects are logging): GUI stays responsive (no
   pinwheel), Stop shows "Stopping…", the scan never starts, terminal
   state ABORTED in the pill. Expected log: INFO "stopped during
   initialization … nothing claimed" and **no `ScanNNN` folder** in the
   common case; a stop landing in the narrow claim→plan-start window
   instead logs INFO "stopped before the plan started" with a claimed
   folder holding ScanInfo and no data — both are correct outcomes, do
   not score the second as a failure.
2. One running-phase stop stays quiet (#563 parity): scan aborts with
   cleanup (disarm, save-off), no ERROR blocks in scan.log, GUI returns
   to normal gating.

Closes #571.

### DIFF
--- a/GEECS-Console/CHANGELOG.md
+++ b/GEECS-Console/CHANGELOG.md
@@ -4,6 +4,22 @@ All notable changes to GEECS-Console are documented here.  Format follows
 [Keep a Changelog](https://keepachangelog.com/en/1.1.0/); versioning is
 semantic.
 
+## [0.15.0] - 2026-07-16
+
+### Fixed
+
+- **Stop no longer pinwheels the GUI** (#571 — the console half; the
+  stop-is-ignored engine half is GeecsBluesky 0.40.0).
+  `_on_stop_clicked` used to call `stop_scanning_thread()` on the GUI
+  thread, which could block up to 15 s (the engine joined a scan thread
+  sitting in 20 s device connects) — the window froze while the terminal
+  kept logging init steps.  Stop now dispatches through a
+  `BackgroundResult` worker: the click returns immediately, the Stop
+  button disables and shows "Stopping…", and normal gating resumes when
+  the lifecycle event stream reports the terminal state (ABORTED/DONE) —
+  the same stream the window already consumes.  The state pill knows the
+  engine's STOPPING state (amber).
+
 ## [0.14.0] - 2026-07-16
 
 ### Changed
--- a/GEECS-Console/CLAUDE.md
+++ b/GEECS-Console/CLAUDE.md
@@ -32,7 +32,13 @@ are prefixed by region (`r3_radio_1d`, `r5_start_button`, …).
   selected sets is valid there), valid shot count within the guard, and in
   Optimization mode a selected optimizer config.  The GUI never pre-blocks
   an optimize submission beyond that — the engine's accept/refuse answer is
-  surfaced in the status bar.
+  surfaced in the status bar.  **Stop is asynchronous** (#571):
+  `stop_scanning_thread` can block briefly engine-side, so the click
+  dispatches it through a dedicated `BackgroundResult` stop worker (no
+  result slot — completion arrives via the lifecycle stream); the button
+  disables and shows "Stopping…" until a terminal lifecycle state
+  (`_TERMINAL_SCAN_STATES` = aborted/done, exactly the engine's terminal
+  vocabulary) releases the hold and normal gating resumes.
 - **R6 now panel** — state pill, progress bar, "Scan NNN" with 10 s expiry
   to "(previous)" (**live**: driven by `ScanLifecycleEvent.scan_number`
   through the adapter's `scan_number_known` signal), compact log tail.
--- a/GEECS-Console/geecs_console/app/main_window.py
+++ b/GEECS-Console/geecs_console/app/main_window.py
@@ -124,10 +124,12 @@ _STATE_DOT_COLORS = {
     "failed": _COLOR_RED,
 }
 
-# Lifecycle states that settle an in-flight Stop request: the scan is over
-# (or the engine failed), so the Stop button's "Stopping…" hold releases and
-# normal gating resumes.
-_TERMINAL_SCAN_STATES = frozenset({"aborted", "done", "error", "failed", "idle"})
+# Lifecycle states that settle an in-flight Stop request, releasing the Stop
+# button's "Stopping…" hold.  Exactly the engine's terminal vocabulary: the
+# bridge's scan-thread cleanup always ends in ABORTED or DONE
+# (geecs_bluesky.events.ScanState; BlueskyScanner._run_scan).  Deliberately
+# NOT paused_on_error — that state is resumable, the scan is not over.
+_TERMINAL_SCAN_STATES = frozenset({"aborted", "done"})
 
 
 def _default_completions_factory(experiment: str) -> CompletionsProvider:
--- a/GEECS-Console/geecs_console/app/main_window.py
+++ b/GEECS-Console/geecs_console/app/main_window.py
@@ -118,10 +118,16 @@ _STATE_DOT_COLORS = {
     "complete": _COLOR_GREEN,
     "completed": _COLOR_GREEN,
     "initializing": _COLOR_AMBER,
+    "stopping": _COLOR_AMBER,
     "aborted": _COLOR_RED,
     "error": _COLOR_RED,
     "failed": _COLOR_RED,
 }
+
+# Lifecycle states that settle an in-flight Stop request: the scan is over
+# (or the engine failed), so the Stop button's "Stopping…" hold releases and
+# normal gating resumes.
+_TERMINAL_SCAN_STATES = frozenset({"aborted", "done", "error", "failed", "idle"})
 
 
 def _default_completions_factory(experiment: str) -> CompletionsProvider:
@@ -851,6 +857,13 @@ class MainWindow(QMainWindow):
         self._actions_worker.result_ready.connect(
             self._apply_action_names, Qt.ConnectionType.QueuedConnection
         )
+        # Stop dispatch (issue #571): stop_scanning_thread may block briefly
+        # (engine bookkeeping join), so it runs on a worker — never the GUI
+        # thread.  No result slot: completion is announced by the terminal
+        # lifecycle event (_on_scan_state clears the in-flight hold).
+        self._stop_worker = BackgroundResult()
+        self._stop_in_flight = False
+        self._stop_button_label = self.stop_button.text()
 
         self.events.state_changed.connect(self._on_scan_state)
         self.events.totals_known.connect(self._on_totals_known)
@@ -1720,7 +1733,9 @@ class MainWindow(QMainWindow):
             and (not optimize or bool(self.optimization_combo.currentText()))
         )
         self.start_button.setEnabled(ready)
-        self.stop_button.setEnabled(scanning)
+        # While a stop is in flight the button stays disabled ("Stopping…")
+        # until the terminal lifecycle event lands (_on_scan_state).
+        self.stop_button.setEnabled(scanning and not self._stop_in_flight)
 
     def _ensure_submitter(self) -> Optional[Submitter]:
         """Return the injected submitter, or lazily build the real engine."""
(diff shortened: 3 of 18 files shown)


skbarber:
# Adversarial review (fresh-context subagent) — report + dispositions

Per the /land ritual: a fresh-context reviewer subagent was given the PR number, the branch, and the standard three-lens brief (correctness / redundancy / placement) — none of the author's rationale. Its report, verbatim, then the disposition of every finding. Fixes went back to the **same** reviewer for confirm/deny (round 2, commit 889df3fe).

## Round 1 report (on a325e8f3)

> Review complete. I read the full PR diff, the changed files in the worktree, the surrounding modules (bridge, session, runner, console window, events adapter, `BackgroundResult`), and verified every bluesky-internals claim against the installed `bluesky 1.15.0` source.
>
> **Verified before findings (things I tried to break and couldn't)**
>
> - **The idle-window happens-before argument is sound.** `run_engine.py:1503` sets `self._state = "running"` before the message loop fetches the plan's first message; the stopper writes `_abort_requested` before reading `RE.state`; under the GIL, a stopper that observed `idle` is guaranteed visible to the gate. `abort()`/`_abort_coro`/`__interrupter_helper` behave exactly as the PR claims (thread-safe dispatch via `call_soon_threadsafe`, `TransitionError` on idle is caught by the bridge's `except`, task-cancel without cleanup wait).
> - **No redundancy.** `_stop_gated` has no existing equivalent (checked `plans/`, `run_wrapper.py`, `_abort_before_acquisition` — the latter remains complementary, covering stop-before-thread-start). The console reuses the blessed `BackgroundResult` rather than growing a twin, and `_stop_in_flight` mirrors the existing `_device_set_in_flight` pattern. The dead-code rider is complete (no remaining `_last_run_uid`/`_request_operator_decision` references outside the intentional docstring cross-ref).
> - **Tests would fail on wrong code**: the <0.3 s timing test fails against the old sync stop; the gate tests run a real RunEngine and assert zero documents; checkpoint tests assert `claim_scan`/`session.scan` never ran. Init-order in `MainWindow.__init__` is safe.
> - Stale-worker-kills-next-scan and stale-abort-flag scenarios: refuted (flag reset in `start_scan_thread`; the hold isn't released until a terminal event, so no new scan can start under a pending stop).
>
> **Findings (ranked)**
>
> **1. Low-medium — the "burns no scan number" promise is false on the gate path, and the PR's own hardware-verification criterion can false-fail.** The runner checkpoints are pre-claim, but the in-plan gate trips *after* `session.scan` claims (claims and writes `ScanInfo` before the gate wraps the plan). The burn window is not microseconds: it spans `claim_scan_number`'s network I/O plus the `ScanInfo` write over NetApp/VPN — seconds, exactly where a Stop clicked "within the first seconds after Start" can land. Result: `ScanNNN/` folder with `ScanInfo` and no data, and the INFO line is "stopped before the plan started" (session), not "stopped during initialization … nothing claimed" (runner). The `session.py` docstring is honest about this; the PR body headline, the GeecsBluesky changelog, and hardware-verification item 1 are not. Behavior is fine (scanner-side folder creation is allowed); the fix is wording.
>
> **2. Low — new cosmetic race: pill can end stuck on amber STOPPING after a natural completion.** Scenario: the scan's last shot completes between the GUI-thread `is_scanning_active()` check in `_on_stop_clicked` and the worker's `stop_scanning_thread` run. `_run_scan`'s finally reads `_abort_requested` still False → emits DONE; the worker then emits STOPPING unconditionally. The hold releases correctly at DONE, but the pill then repaints STOPPING and stays amber on an idle scanner until the next scan. The unconditional STOPPING emission is pre-existing, but ending-on-STOPPING presentation is newly possible (the old sync call emitted STOPPING ahead of the queued terminal event). Cheap guard: early-return in `stop_scanning_thread` when `not is_scanning_active()`.
>
> **3. Nit — `_TERMINAL_SCAN_STATES` contains states that can never arrive.** The adapter only ever delivers `ScanState` values, and the bridge only emits INITIALIZING/RUNNING/STOPPING/ABORTED/DONE. `"error"`, `"failed"`, `"idle"` in the new frozenset are unreachable; this set is load-bearing gating, and a reader can't tell real vocabulary from padding.
>
> **4. Nit — GEECS-Console/CLAUDE.md not updated.** That file maintains a per-worker inventory of every `BackgroundResult` use and describes R5 Stop; the new stop worker, the "Stopping…" hold, and the terminal-release contract are absent, while GeecsBluesky/CLAUDE.md *was* updated for its half.
>
> One observation, not a finding: `GEECS-Scanner-GUI/.../run_control.py` still calls `stop_scanning_thread` and silently inherits the changed semantics — moot since the legacy GUI is un-launchable on `dev` per G3.
>
> No correctness failures survive in the code itself.

## Dispositions

| # | Finding | Disposition |
|---|---|---|
| 1 | Burns-no-number wording overreach (gate path) | **Fixed** — 889df3fe: CHANGELOG gate-path caveat; PR body headline + hardware-verification item 1 reworded to list both correct outcomes. Reviewer round 2: **confirmed**. |
| 2 | STOPPING repaint after natural completion | **Fixed** — 889df3fe: `stop_scanning_thread` early-returns when not scanning (no STOPPING emission, no parked abort flag, dead handle still reaped); pinned by `test_stop_with_no_active_scan_is_a_stateless_no_op`. Reviewer round 2: **confirmed** (notes the remaining TOCTOU window is microseconds, cosmetic, and would need a lock — not worth it; also notes the guard now protects foreign headless runs on the shared RunEngine from a no-op stop's `RE.abort()`, a #511-adjacent improvement). |
| 3 | `_TERMINAL_SCAN_STATES` padding | **Fixed** — 889df3fe: trimmed to exactly `{"aborted", "done"}` with the vocabulary + `paused_on_error` rationale in the comment. Reviewer round 2: **confirmed**. |
| 4 | Console CLAUDE.md missing the stop worker | **Fixed** — 889df3fe: R5 section documents the async Stop, the dedicated worker, the hold, and the terminal-release contract. Reviewer round 2: **confirmed**. |
| — | Round-2 nit: no-op path logs "abort requested" then "no active scan" | **Waived** — reviewer marked it take-or-leave ("good to land from my side"); the pair is accurate as a narrative (an abort *was* requested) and reordering costs another CI round pre-merge. Cheap to veto: say the word and it moves in a follow-up. |

Suites after the fixes: GeecsBluesky **519 passed, 3 deselected**; GEECS-Console **748 passed**.

skbarber:
**Hardware verification: PASSED (Sam, live, 2026-07-16)** — Stop during the initialization phase 'closed quickly and cleanly': responsive GUI, no pinwheel, prompt termination. Running-phase quiet abort was verified live earlier the same day (post-#570: single INFO, no tracebacks). OWED section closed.

PR #576, opened 2026-07-16
GeecsBluesky 0.41.0: validate_scan_request — one fail-fast definition for bridge + runner

### LINKED ISSUE #529: scan_request_runner: validate-only entry point so bridge fail-fast can't drift from execution
From the post-#483 audit (PRs #484–#526).

`BlueskyScanner._reinitialize_from_scan_request` (rewritten by #485/#520) re-implements the resolution sequence of `run_scan_request` in validate-and-discard form: defaults, action bindings, save sets + rituals, trigger-profile writes, step-axis movable targets, and optimize VOCS catalog names — each mirrored by hand with "resolve and discard" comments.

Any new resolvable field added to `run_scan_request` (pseudo variables, a new action slot, per-axis confirm changes) must be manually mirrored in the bridge, or GUI submitters get their error in the scan thread instead of at submission time.

Proposal: extract a `validate_scan_request(request, resolver)` dry-run function in `scan_request_runner` that performs exactly the runner's fail-fast resolution, called by both the bridge (`reinitialize`) and — as its own first phase — `run_scan_request`, so there is one definition of "what must resolve".

🤖 Filed from the post-483 cleanup audit (Claude Code)

skbarber:
Closes #529.

## What + why

`BlueskyScanner.reinitialize` re-implemented `run_scan_request`'s resolution sequence in hand-maintained "resolve and discard" form — defaults, action names, trigger profile/variant, save sets + rituals, step-axis movable targets, optimize VOCS names. Any new resolvable field added to the runner had to be manually mirrored in the bridge, or GUI submitters got their error in the scan thread instead of at Start. The drift had already happened once: the optimize save-set-less rule (0.38.0) is encoded in both places today.

This extracts **`validate_scan_request(request, resolver)`** in `scan_request_runner` — pure resolution, no session, no hardware — and makes it THE one definition: the bridge calls it for submission-time errors (its whole 37-line mirror collapses to one call), and `run_scan_request` runs it as its own first phase. New resolvable fields get added there, never to a caller.

**One deliberate behavior improvement** (flagged for veto): the runner now refuses a bad request *before* touching any session state — previously it attached the trigger profile (`session.shot_control`) before resolving save sets and axis targets, so a doomed request could still mutate session shot-control state. Pinned by a test that runs `run_scan_request` against a session object where *any* attribute access is a failure.

## Per-concern LOC breakdown

- `validate_scan_request` + docstring (`scan_request_runner.py`): +80
- Runner phase-1 rewiring (head of `run_scan_request`; duplicate save-set raise removed — one definition): +10 / −9
- Bridge mirror deletion (`bluesky_scanner.py`: 37-line block → 1 call; 5 now-unused imports dropped): +7 / −40
- Tests (`tests/test_validate_scan_request.py`: 14 tests — every resolvable category refused through the one function, the optimize exemptions, defaults provenance + copy-not-in-place, the session-untouched fail-fast pin, and a structural spy pinning that `reinitialize` calls the shared function): +290
- Version bump + CHANGELOG (0.41.0 minor): +30

## Tests

- `./scripts/check.sh` (scoped): lint clean; **GeecsBluesky 533 passed, 3 deselected** (519 existing + 14 new). The existing bridge seam suite — which pins every fail-fast behavior through `reinitialize` (unknown names, pseudo refusal, defaults-not-applied-twice, store-original semantics) — passes unchanged through the shared function, pinning behavior preservation.

## Hardware verification

N/A — pure pre-claim name resolution; no scan execution, device I/O, or claim path touched. The only ordering change moves failures *earlier* (before `session.shot_control`), which has no hardware-observable surface on a request that was going to be refused anyway.

### DIFF
--- a/GeecsBluesky/CHANGELOG.md
+++ b/GeecsBluesky/CHANGELOG.md
@@ -4,6 +4,33 @@ All notable changes to `geecs-bluesky` are documented here.
 
 Format follows [Keep a Changelog](https://keepachangelog.com/en/1.1.0/).
 
+## [0.41.0] - 2026-07-16
+
+### Added
+
+- **`validate_scan_request(request, resolver)`** (issue #529) — THE one
+  fail-fast definition of "what must resolve" for a ScanRequest, in
+  `scan_request_runner`: experiment defaults, action names (nested `run`
+  references included), trigger profile + variant, the save-set rule
+  (non-optimize needs ≥1; optimize may run save-set-less), save sets +
+  entry rituals, step-axis movable targets (pseudo refused), and optimize
+  VOCS catalog names.  Pure resolution — no session, no hardware.
+
+### Changed
+
+- `BlueskyScanner.reinitialize` and `run_scan_request` both route through
+  `validate_scan_request` (the bridge for submission-time errors, the
+  runner as its own first phase), replacing the bridge's hand-maintained
+  "resolve and discard" mirror of the runner's resolution sequence — the
+  drift class where a new resolvable field surfaced its error in the scan
+  thread instead of at Start (it had already bitten once: the optimize
+  save-set-less rule of 0.38.0 was encoded twice).  New resolvable fields
+  are added to `validate_scan_request`, never to a caller.  The runner now
+  refuses a bad request **before** touching any session state (previously
+  the trigger profile was attached before some name resolutions); pinned
+  by `tests/test_validate_scan_request.py`, including a structural
+  no-drift spy on the bridge and a session-untouched fail-fast pin.
+
 ## [0.40.0] - 2026-07-16
 
 ### Fixed
--- a/GeecsBluesky/geecs_bluesky/scan_request_runner.py
+++ b/GeecsBluesky/geecs_bluesky/scan_request_runner.py
@@ -826,9 +826,14 @@ def validate_scan_request(
 
     Returns
     -------
-    tuple[ScanRequest, dict]
-        The **post-defaults** validated copy of *request*, and the applied-
-        defaults provenance record (as :func:`apply_experiment_defaults`).
+    tuple[ScanRequest, dict, Any]
+        The **post-defaults** validated copy of *request*, the applied-
+        defaults provenance record (as :func:`apply_experiment_defaults`),
+        and the raw defaults object itself — returned so execution reads
+        the defaults file exactly once per run (flags that are not request
+        fields, e.g. ``background_telemetry``, must come from the same
+        snapshot the validation applied, not a second read that could see
+        a concurrent edit).
 
     Raises
     ------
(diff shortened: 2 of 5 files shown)


skbarber:

The reviewer independently verified the PR's core claims before reporting: the session-untouched fail-fast test genuinely fails against pre-PR code (the old runner attached `session.shot_control` before save-set/axis resolution), execution-time error ordering is preserved (the new function's internal order matches the old runner's), `noscan`-with-axes is refused by the schema itself so its absence from the STEP-only axis check is not a gap, the optimize dispatch receives `applied_defaults`/`skipped_actions` identically with the VOCS `:` pass-through exactly mirroring execution dispatch, and no external caller of the five imports dropped from the bridge exists (checked `session.py`, console `submission.py`/`services/configs.py`, all GeecsBluesky tests). Placement judged correct (validator beside the runner it fronts).

**2 surviving findings, both LOW:**

### Low 1 — the defaults file was read twice per run, and the two reads feed one run's semantics
`validate_scan_request` applied defaults from one `resolve_experiment_defaults` call, then `run_scan_request` immediately called it again solely for the `background_telemetry` flag — and `ConfigsRepoResolver.resolve_experiment_defaults` re-reads and re-validates `experiment_defaults.yaml` on every call. Concrete scenario: an operator saves an edit to the file in the window between the two reads at scan start — the run's applied defaults and provenance come from the old file while the telemetry flag comes from the new one; a mid-write/malformed file at read #2 raises after phase 1 declared the request valid.

### Low 2 — fourth near-verbatim `_StubResolver` copy; the spy test hand-rolls a scanner-attribute list
The new test module's `_StubResolver` is a line-for-line duplicate of the seam suite's; the structural spy test reconstructs a `BlueskyScanner` via `__new__` with a hand-maintained attribute list that is a subset of the seam suite's `_make_scanner`. Per-file stubs are the established local pattern; this is the copy that makes a `conftest.py` promotion worth doing as a follow-up.

Everything else attempted (double-application of defaults, stale `applied_defaults` on optimize, merge/role-conflict coverage, hygiene, CHANGELOG consistency) was refuted on inspection.

skbarber:
## Disposition

| # | Finding | Disposition |
|---|---|---|
| 1 | Defaults file read twice per run (torn-edit window between validation's read and the telemetry-flag read) | **Fixed** in 7ce54818 — `validate_scan_request` reads once and returns the raw defaults snapshot as a third tuple element; `run_scan_request` unpacks it. Verifier: **CONFIRMED** — "no second read, so a concurrent edit or mid-write file can no longer split one run's semantics or raise after phase-1 validation passed"; also verified no remaining production `resolve_experiment_defaults` call on any run path (the optimize path hardcodes its flag), the 3-tuple breaks no caller, and 117 tests pass across the two suites. *(Process note: the original reviewer agent became unreachable after repeated API-overload terminations, so the confirm/deny ran on a fresh-context verifier given the finding text and the fix commit — same independence, none of the author's rationale.)* |
| 2 | Fourth near-verbatim `_StubResolver` copy + hand-rolled scanner attributes in the spy test | **Waived** for this PR, per the reviewer's own assessment ("reasonable as a follow-up rather than scope creep") — per-file stubs are the established local pattern. Follow-up noted: promote a shared stub resolver + scanner factory to `GeecsBluesky/tests/conftest.py` next time a fifth copy would appear. |

CI green on both commits (unit-tests, console-windows, pre-commit). Merging.

PR #578, opened 2026-07-16
GeecsBluesky 0.42.0: deferred-pause checkpoints + last_state, delete hard-pause API

skbarber:
PR-1 of the G-actions v2 sequence (issue #552; design note + owner decisions recorded on the issue — all defaults accepted, decision 11 reversed to refuse shot-control-device actions, which lands in PR-3).

## What + why

The design note's key verified finding: **our step plans emitted zero checkpoint messages**, so any pause + resume would rewind to plan start and replay the entire scan — re-opening the run, re-driving every trigger transition, re-firing every shot. The existing `pause_scan`/`resume_scan` bridge methods (hard pause) were exactly that trap, with no callers.

This PR lays the groundwork all later PRs build on:

- **Checkpoints in both step plans** — one before each step's move, one before every row. `request_pause(defer=True)` now lands at a row/step boundary with an empty rewind cache: resume replays nothing. Zero behavior change for an unpaused scan (a checkpoint is a no-op outside pause/suspend handling).
- **`ShotController.last_state`** — records the last *standing* state actually driven (never the momentary SINGLESHOT fire, so a later re-assert can never refire a shot). The PR-3 pause supervisor reads this to restore the trigger state after an operator action.
- **`pause_scan`/`resume_scan` deleted** with a tombstone comment; absence pinned by test (design decision 2).

## Per-concern LOC breakdown

- Checkpoints (`step_scan.py` +9, `free_run_step_scan.py` +9, comments included)
- `last_state` (`shot_controller.py`: attribute + `_record_state` + two call sites): +22
- Hard-pause API deletion (`bluesky_scanner.py`): −16 / +6 tombstone
- Tests (`tests/test_pause_checkpoints.py`, 8 tests): +250 — checkpoint count/placement, never-inside-an-open-event-bundle, an RE-level deferred pause + resume yielding every row exactly once (deterministic: the plan itself yields `deferred_pause` via the per_step hook), `last_state` standing-states-only semantics, and the deleted-API pin
- One existing pin updated: `test_grid_plans_message_level.py` per-step placement now accepts the row checkpoint as part of the row boundary (+3/−1)
- Version bump + CHANGELOG (0.42.0 minor): +30

## Tests

`./scripts/check.sh` (scoped): lint clean; **GeecsBluesky 541 passed, 3 deselected** (533 + 8 new). The RE-level test exercises real deferred-pause semantics against bluesky 1.15 (RunEngineInterrupted → paused → resume → exact event count, monotone scan_event_index).

## Hardware verification

**N/A for this PR alone** — checkpoints are inert without a pause request, and nothing requests one yet (the supervisor arrives in PR-3). The full pause-window hardware checklist (jet stops in free-run OFF, strict stays quiescent, pause latency at real rep rate, abort-during-pause end state) is enumerated in the design note §6.5 and owed at the end of the sequence, before the feature is declared usable.

### DIFF
--- a/GeecsBluesky/CHANGELOG.md
+++ b/GeecsBluesky/CHANGELOG.md
@@ -8,10 +8,12 @@ Format follows [Keep a Changelog](https://keepachangelog.com/en/1.1.0/).
 
 ### Added
 
-- **Deferred-pause checkpoints in both step plans** (issue #552, PR-1 of
-  the G-actions v2 sequence): one `bps.checkpoint()` before each step's
-  move and one before every row, in `geecs_step_scan` and
-  `geecs_free_run_step_scan`.  `request_pause(defer=True)` now lands at a
+- **Deferred-pause checkpoints in all three scan plans** (issue #552,
+  PR-1 of the G-actions v2 sequence): one `bps.checkpoint()` before each
+  step's move / iteration and one before every row, in `geecs_step_scan`,
+  `geecs_free_run_step_scan`, and `geecs_adaptive_scan` (optimize — added
+  after review: without them a pause requested during an optimize scan
+  would never land and the console would sit in PAUSING).  `request_pause(defer=True)` now lands at a
   row/step boundary with an empty rewind cache — resume replays nothing
   (no re-move, no re-fire).  Zero behavior change for an unpaused scan.
   Pinned by `tests/test_pause_checkpoints.py` (checkpoint placement,
--- a/GeecsBluesky/CHANGELOG.md
+++ b/GeecsBluesky/CHANGELOG.md
@@ -4,6 +4,33 @@ All notable changes to `geecs-bluesky` are documented here.
 
 Format follows [Keep a Changelog](https://keepachangelog.com/en/1.1.0/).
 
+## [0.42.0] - 2026-07-16
+
+### Added
+
+- **Deferred-pause checkpoints in both step plans** (issue #552, PR-1 of
+  the G-actions v2 sequence): one `bps.checkpoint()` before each step's
+  move and one before every row, in `geecs_step_scan` and
+  `geecs_free_run_step_scan`.  `request_pause(defer=True)` now lands at a
+  row/step boundary with an empty rewind cache — resume replays nothing
+  (no re-move, no re-fire).  Zero behavior change for an unpaused scan.
+  Pinned by `tests/test_pause_checkpoints.py` (checkpoint placement,
+  never inside an open event bundle, and an RE-level deferred
+  pause + resume yielding every row exactly once).
+- **`ShotController.last_state`** — records the last *standing* state
+  actually driven (OFF/SCAN/STANDBY/ARMED; never the momentary
+  SINGLESHOT fire, so a later re-assert can never refire a shot).  Read
+  by the upcoming #552 pause supervisor to restore the trigger state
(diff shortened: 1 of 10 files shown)


skbarber:

The reviewer verified against the installed bluesky 1.15 source and ran the new tests 5× for flake-stability. **3 surviving findings:**

### Medium — free-run checkpoints completely unpinned
Every placement test exercised only `geecs_step_scan`; deleting both checkpoints from the free-run plan (the production acquisition mode, whose pause behavior is the most load-bearing) left the suite green. Verified consequence in the RE source: `_deferred_pause_requested` is consulted **only** in `_checkpoint` — no checkpoint, no pause, scan runs to completion with the console stuck in PAUSING.

### Medium-low — `geecs_adaptive_scan` (optimize) got no checkpoints, and nothing recorded the decision
Same loop shape, same bridge/session, zero checkpoints → a pause requested during an optimize scan would never land (PAUSING forever, operator-facing hang). Plus a smaller contract note: a pause requested after the *last* checkpoint (disarm/quiesce/tail-flush/closeout) never lands and the plan completes normally — PR-3's supervisor must treat that as a normal outcome.

### Low — the loosened grid-plan assertion only looked one message past
`after in ("checkpoint", "trigger", "create")` let anything hide behind the row checkpoint; the original "nothing between per_step and the first shot" pin lost its teeth.

**Checked and refuted** (verified in run_engine.py, not recalled): checkpoint legality across run_decorator/stage_wrapper/tail-flush (raises only while bundling; both plans' checkpoints sit outside bundles), non-rewindable handling (caching gated, `_checkpoint` isn't), no suspenders/Pausable/Stoppable devices anywhere so pause has zero device side effects, pause-after-arm jet-state consistency with the §2.3 supervisor contract (drive OFF unconditionally, `last_state == SCAN` is the right restore), `last_state` staleness (impossible at a checkpoint; inert on abort since the abort path never restores), RE-level test determinism (plan-driven pause request; 5× stable) and pinning power (fails without checkpoints), no other message-level consumers, no callers of the deleted API, no redundancy. Bonus observed: per-row checkpoints also stop `_msg_cache` growing unboundedly over long scans.

skbarber:
## Disposition (all findings fixed in 19f1e57d, reviewer-confirmed)

| # | Finding | Disposition |
|---|---|---|
| 1 | Free-run checkpoints completely unpinned | **Fixed** — RE-level `msg_hook` test (count + never-in-bundle incl. tail flush). Reviewer: **confirmed** — "the concrete regression I constructed — delete both checkpoints, suite stays green — now fails on the count assertion" |
| 2 | Optimize plan had no checkpoints; decision unrecorded | **Fixed** — both checkpoints added to `geecs_adaptive_scan` (pre-iteration sits before `propose_pool.submit` — reviewer: "correct spot"), pinned by a `2×(1+2)+1` msg_hook test; the end-of-plan-window supervisor contract recorded on #552. Reviewer: **confirmed, exceeds the asked-for minimum** |
| 3 | Loosened grid assertion lost its teeth | **Fixed** — checkpoint-transparent scan restores the original two-command pin. Reviewer: **confirmed**; `_NamedFake`/`_collect` dedup stays the standing conftest-promotion follow-up |

Reviewer's closing: "No new findings introduced by the fix commit. Nothing outstanding." CI green on both commits. Merging.

PR #581, opened 2026-07-17
GEECS-Console 0.16.0 + GeecsBluesky 0.44.1: G-actions v2 console half + #575 dialog hardening

### LINKED ISSUE #575: Actions: rethink the firing surface (menu feels misfire-prone)
Field feedback (Sam, 2026-07-16, first live use of the Actions menu): execution works and is 'a great start,' but **the dropdown menu is the wrong firing surface** — 'waaay too easy to accidentally hit the wrong action,' even with the Enable-execution gate. Verdict: the *capability* (G-actions v1, #553/#554) is the accepted deliverable; how operators *use* actions is deliberately unsettled — 'I don't have a great feeling about how we use these actions just yet.'

Context for whoever picks this up:
- The legacy tooling fired actions from the **ActionLibrary editor** — and was 'fairly lightly used,' so don't over-invest.
- Candidate directions (none chosen): execute-from-editor (legacy parity — the editor already exists, and its per-plan view naturally shows *what* you're about to fire); the dialog-only flow (menu opens the preview dialog, Run lives only there — the menu itself never fires; NOTE: this is already true today, the menu opens the ActionRunDialog, so the complaint may be about proximity/selection in the menu list itself — clarify with Sam); the deferred v1.1 quick-access buttons (persisted, deliberately-placed, arguably *harder* to misfire than a scrolling menu).
- Related: #552 (pause/decide/resume during-scan flow) — any redesign should keep that in view.

No action until the usage patterns firm up — this issue is the parking spot for the design feeling, not a work order.

skbarber:
PR-4 (final) of the G-actions v2 sequence (issue #552; engine half #578/#579/#580 merged). This makes the pause/decide/resume flow reachable from the UI and closes #575's misfire complaint.

## What + why

**During-scan action flow (console half):** the Actions Run/Preview dialog now drives the pause path. While a scan is active the Run button reads **"Pause scan & run (N steps)"** and calls the new `Submitter.request_action_during_scan` member (pause → decide → run) instead of idle `run_action`; the window pushes live scan state into open dialogs (`set_scanning`) so the label flips automatically. The three-way **action-decision modal** (Execute / Ignore / Abort) is rendered by `_on_action_decision` from the engine's `ActionDecisionRequest`, routed through the *existing* `ScanDialogEvent`/`dialog_requested` transport — `_on_operator_dialog` duck-types on the `verdict` attribute, so the binary pre-flight dialog is untouched. New `PAUSING`/`PAUSED` states render amber on the R6 pill.

**The #552 PR-3 contract, honoured:** the modal is **dismissed on a terminal ABORTED state** (`self._decision_box.reject()` in `_on_scan_state`). A Stop during the pause window aborts the scan without the engine answering, and the engine has no dialog-cancel signal — so the console tears the dangling modal down itself; the programmatic dismiss reads as `abort`.

**#575 misfire hardening (Option A from the design note):** action name in larger type on its own line; **Run disabled until the dry-run preview loads** (no firing beside an empty table — this also *eliminates by construction* the old late-preview-clobbers-run race, so that test became a positive pin for the gate); step count on the Run button.

## Per-concern LOC breakdown

- `submission.py` (Submitter member): +14
- `action_dialog.py` (during-scan path + preview gate + step-count label + name hardening): +85 / −20
- `main_window.py` (dialog wiring, scan-state push to dialogs, three-way modal, terminal dismiss, pill colors): +90
- `events_adapter.py` (log-line duck-type for the two request shapes): +8 / −3
- Tests (`test_actions_menu.py`: existing 16 updated for the preview gate + 7 new — during-scan path, refusal surfaced, the three verdicts, terminal dismiss): +130
- CLAUDE.md + CHANGELOG + version: +55

**Judgment-call cross-package addition (cheap to veto):** a one-line GeecsBluesky fix (`0.44.1`) — `_make_pause_supervisor` passes `ask=None` when the bridge has no `on_event` consumer, so a console-less bridge that somehow reached a pause window defaults to `ignore` instead of parking forever (the supervisor's safe posture). Reachable only by misuse (`request_action_during_scan` is console-only), but it closes a real park-forever hole in the just-merged engine. Pinned by a bridge test. Split it out if you'd rather keep this PR console-only.

## Tests

- `./scripts/check.sh`: **GEECS-Console 753 passed** (23 in the actions suite: 16 existing + 7 new), **GeecsBluesky 571 passed** (+1 park-gap pin). The 2 `*_without_the_extra` optimization failures are the known local-venv-has-the-extra issue (pass in CI).

## Hardware verification

**OWED** — the whole feature is now code-complete, so the checklist of record (issue #552 comment) is runnable: free-run pause = jet off, strict quiescent, pause latency, full pause→execute→resume integrity, abort-from-dialog, Stop-during-pause single-abort, decision-11 refusal live. One Undulator session.

### DIFF
--- a/GEECS-Console/CHANGELOG.md
+++ b/GEECS-Console/CHANGELOG.md
@@ -4,6 +4,30 @@ All notable changes to GEECS-Console are documented here.  Format follows
 [Keep a Changelog](https://keepachangelog.com/en/1.1.0/); versioning is
 semantic.
 
+## [0.16.0] - 2026-07-16
+
+### Added
+
+- **G-actions v2 console half** (issue #552 PR-4; #575 dialog hardening).
+  The Actions Run/Preview dialog now drives the during-scan pause flow:
+  while a scan is active the Run button reads **"Pause scan & run (N
+  steps)"** and calls the engine's `request_action_during_scan` (pause →
+  decide → run) instead of the idle `run_action`; the window pushes live
+  scan state into open dialogs so the button flips automatically.  The
+  three-way **action-decision modal** (Execute / Ignore / Abort) renders
+  from the engine's `ActionDecisionRequest` (routed through the existing
+  `ScanDialogEvent` transport, duck-typed on the `verdict` attribute) and
+  is **dismissed on a terminal ABORTED state** — a Stop during the pause
+  window aborts the scan without the engine answering, and the engine has
+  no dialog-cancel signal (the #552 PR-3 review contract).  New
+  `Submitter` protocol member `request_action_during_scan`; new
+  `ScanState.PAUSING`/`PAUSED` render amber on the R6 pill.
+- **Actions dialog misfire hardening** (issue #575): the action name in
+  larger type on its own line, **Run disabled until the dry-run preview
+  loads** (no firing beside an empty table — this also eliminates the old
+  late-preview-clobbers-run race by construction), and the **step count on
+  the Run button** ("Run (N steps)").
+
 ## [0.15.1] - 2026-07-16
 
 ### Changed
--- a/GEECS-Console/CLAUDE.md
+++ b/GEECS-Console/CLAUDE.md
@@ -269,7 +269,22 @@ are prefixed by region (`r3_radio_1d`, `r5_start_button`, …).
   describe arriving late must never clobber a refusal (pinned by test).
   Both engine methods are `Submitter` protocol members
   (`submission.py`), mapped to `BlueskyScanner`'s same-named methods.
-  A pause-the-scan flow for actions (#552) is future work.  Pinned by
+  **During-scan (G-actions v2, #552, 0.16.0):** the same dialog drives the
+  pause flow — with a scan active the Run button reads "Pause scan & run
+  (N steps)" and calls the `Submitter.request_action_during_scan` member
+  (pause → decide → run) instead of idle `run_action`; the window pushes
+  live scan state into open dialogs (`set_scanning`) so the button flips.
+  The three-way **action-decision modal** (Execute / Ignore / Abort) is
+  rendered by `_on_action_decision` from an engine `ActionDecisionRequest`
+  (routed through the same `ScanDialogEvent`/`dialog_requested` transport
+  as the binary pre-flight dialog — `_on_operator_dialog` duck-types on the
+  `verdict` attribute), and is **dismissed on a terminal ABORTED state**
+  (`self._decision_box.reject()` in `_on_scan_state`): a Stop during the
+  pause window aborts without the engine answering, and the engine has no
+  dialog-cancel signal (the #552 PR-3 review contract).  **Misfire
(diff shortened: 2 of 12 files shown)


skbarber:

The reviewer deep-traced the GUI failure modes and **defended most of the risky ones as correct** — queued main-thread delivery of both `_on_action_decision` and `_on_scan_state`; the terminal-dismiss nested-loop timing (`reject()` only under an `is not None` guard, `_decision_box` cleared only in the `finally` after `exec()` returns); re-entrancy precluded (the engine thread is parked, issues no second dialog); both during-scan click races degrading to a surfaced engine refusal; and `BlueskyScanner` structurally satisfying the new `Submitter` member. **3 surviving findings:**

### 1. LOW-MEDIUM — engine-unblock not guaranteed against a render exception
`_on_action_decision` reaches `response_event.set()` only by falling through the whole body; the lone `try/finally` guards only `_decision_box = None`. If anything in the render path raises, the paused engine thread never gets its verdict — recoverable (the supervisor's sliced abort-probe wakes it on a Stop) but the operator faces a silently-paused scan with no prompt. `_on_operator_dialog` has the identical pre-existing gap.

### 2. LOW (redundancy) — `_on_action_decision` duplicates the scaffold + the byte-identical response-event tail of `_on_operator_dialog`
A shared `_answer`/tail helper removes the duplication and is the natural home for #1's guarantee (fixes both call sites at once).

### 3. LOW (cosmetic) — during-scan Run label shows "(0 steps)" before the preview loads
The idle path suppresses the count until `_preview_loaded`; the during-scan branch appended it unconditionally.

skbarber:
## Disposition (all findings fixed in 78317f68, reviewer-confirmed)

| # | Finding | Disposition |
|---|---|---|
| 1 | LOW-MEDIUM: engine-unblock not guaranteed on a dialog render exception (same gap in the pre-existing binary handler) | **Fixed** — both handlers wrap the render body in try/except/finally; the finally calls a shared `_unblock_engine` helper, so the parked engine thread is released even if `exec()` raises (on the first supervisor slice, not relying on an operator Stop). Safe defaults: action-decision → `ignore` (resume, run nothing); pre-flight → `abort` (don't proceed past an unshowable warning). Reviewer: **confirmed** |
| 2 | LOW: `_on_action_decision` duplicated the byte-identical response-event tail | **Fixed** — factored into `_unblock_engine`, fixing both call sites. Reviewer: **confirmed** |
| 3 | LOW cosmetic: during-scan Run label showed "(0 steps)" before preview loaded | **Fixed** — step-count suffix gated on `_preview_loaded` uniformly on both paths. Reviewer: **confirmed** |

Reviewer defended as correct (deep-traced): queued main-thread delivery, the terminal-dismiss nested-loop timing, re-entrancy precluded, both click races, protocol conformance, the GeecsBluesky one-liner. "No new concerns introduced by the fix commit." CI green on both commits. Merging.

PR #582, opened 2026-07-17
GeecsBluesky 0.45.0 + GEECS-Console 0.17.0: standalone operator Pause/Resume button

skbarber:
Follow-up to the G-actions v2 sequence (issue #552): the **standalone Pause button** the operator actually wanted. The four-PR sequence built the pause *engine* but wired its only trigger to "run an action mid-scan"; this adds a bare Pause control that reuses all of it.

## What + why

The pause supervisor, deferred-pause checkpoints, safe-state drive, and PAUSED lifecycle are all already in place — a bare pause just needed a trigger with **no action** and a way to hold until Resume.

- **Engine (GeecsBluesky 0.45.0):** `BlueskyScanner.request_pause()` / `request_resume()` + `PauseSupervisor.arm_manual_pause()` / `request_resume()`. A manual pause drives the mode-safe state (free-run → `OFF`, jet off; strict → nothing) and **parks non-modally** on a resume/stop signal — no dialog, so the GUI stays fully usable while the scan holds (read device panels, etc.). On Resume it re-asserts the pre-pause shot-control state; Stop aborts out (finalize chain owns the end state). This is *why* `on_pause` no longer immediately resumes an actionless pause.
- **Console (GEECS-Console 0.17.0):** a **Pause** button on the R5 submit row (new `r5_pause_button` in `main_window.ui`), between the spacer and Stop. Enabled while RUNNING (reads "❚❚ Pause"); becomes "▶ Resume" while PAUSED; disabled during the transitional PAUSING/STOPPING and when idle/terminal. New `Submitter` members `request_pause` / `request_resume`.

**Naming:** the new methods are `request_pause`/`request_resume` — deliberately *not* the tombstoned `pause_scan`/`resume_scan` (the deleted hard-pause trap), so the PR-1 tombstone test stays valid and no one confuses the safe deferred pause with the old bare-`RE.resume()` replay trap.

## Per-concern LOC breakdown

- Supervisor manual-pause branch (`arm_manual_pause`/`request_resume`/`_park_manual_pause` + on_pause dispatch): +45
- Bridge `request_pause`/`request_resume`: +45
- Console R5 button (.ui widget, bind, `_on_pause_clicked`, `_refresh_pause_button`, state tracking): +75
- Submitter protocol members: +14
- Tests (bluesky: 3 supervisor + 3 bridge; console: 6 button-lifecycle): +150
- Version + CHANGELOG × 2: +40

## Tests

- `./scripts/check.sh`: **GeecsBluesky 577 passed** (incl. the tombstone test — still green after the deliberate rename), **GEECS-Console 761 passed** (+6 button tests; the 2 `*_without_the_extra` optimization failures are the known local-extra issue, pass in CI). Verified the button renders + cycles (disabled → Pause → Resume) by driving the real window headlessly.

## Hardware verification

**OWED** — folds into the same Undulator session as the rest of #552 (the checklist of record). This makes step 1 a two-click test: Start, then Pause — no action library needed. Behavior is identical to an action-triggered pause minus the action, so the free-run-OFF / strict-quiescent / abort-end-state checks cover it.

### DIFF
--- a/GEECS-Console/CHANGELOG.md
+++ b/GEECS-Console/CHANGELOG.md
@@ -4,6 +4,21 @@ All notable changes to GEECS-Console are documented here.  Format follows
 [Keep a Changelog](https://keepachangelog.com/en/1.1.0/); versioning is
 semantic.
 
+## [0.17.0] - 2026-07-16
+
+### Added
+
+- **Pause / Resume button on the submit row (R5)** — the standalone
+  operator pause (GeecsBluesky 0.45.0).  While a scan is RUNNING the
+  button reads "Pause" and calls the engine's `request_pause` (pause at
+  the next safe point; the machine goes quiescent — jet off in free-run);
+  while PAUSED it becomes "▶ Resume" and calls `request_resume`; Stop still
+  aborts out of the pause.  Non-modal by design — the device panels and
+  everything else stay usable while the scan holds.  Disabled during the
+  transitional PAUSING/STOPPING states and when idle/terminal.  New
+  `Submitter` members `request_pause` / `request_resume`; the `r5_pause_button`
+  widget added to `main_window.ui`.  Pinned by `tests/test_main_window.py`.
+
 ## [0.16.0] - 2026-07-16
 
 ### Added
--- a/GEECS-Console/geecs_console/app/main_window.py
+++ b/GEECS-Console/geecs_console/app/main_window.py
@@ -417,6 +417,7 @@ class MainWindow(QMainWindow):
         # R5 submit row
         self.stop_button: QPushButton = self._child(QPushButton, "r5_stop_button")
         self.start_button: QPushButton = self._child(QPushButton, "r5_start_button")
+        self.pause_button: QPushButton = self._child(QPushButton, "r5_pause_button")
         # R6 now panel
         self.state_pill: QLabel = self._child(QLabel, "r6_state_pill")
         self.progress_bar: QProgressBar = self._child(QProgressBar, "r6_progress")
@@ -556,6 +557,7 @@ class MainWindow(QMainWindow):
         )
         self.start_button.clicked.connect(self._on_start_clicked)
         self.stop_button.clicked.connect(self._on_stop_clicked)
+        self.pause_button.clicked.connect(self._on_pause_clicked)
         self.apply_button.clicked.connect(self._on_preset_apply)
         self.save_as_button.clicked.connect(self._on_preset_save_as)
         self.delete_button.clicked.connect(self._on_preset_delete)
@@ -605,6 +607,11 @@ class MainWindow(QMainWindow):
         self._stop_worker = BackgroundResult()
         self._stop_in_flight = False
         self._stop_button_label = self.stop_button.text()
+        #: Last lifecycle state word (lowercase) — drives the Pause/Resume
+        #: button (G-actions v2 operator pause).
+        self._scan_state_text = "idle"
+        self._pause_button_label = self.pause_button.text()
+        self._refresh_pause_button()
 
         self.events.state_changed.connect(self._on_scan_state)
         self.events.totals_known.connect(self._on_totals_known)
@@ -1479,6 +1486,31 @@ class MainWindow(QMainWindow):
         # While a stop is in flight the button stays disabled ("Stopping…")
         # until the terminal lifecycle event lands (_on_scan_state).
         self.stop_button.setEnabled(scanning and not self._stop_in_flight)
+        self._refresh_pause_button()
+
+    def _refresh_pause_button(self) -> None:
+        """Set the Pause/Resume button from the current lifecycle state.
+
+        Pause while RUNNING, Resume while PAUSED; disabled otherwise
+        (transitional PAUSING/STOPPING, idle, or terminal).  The engine's
+        operator-pause holds the scan non-modally, so the GUI stays usable
+        while paused.
+        """
+        state = self._scan_state_text
+        if state == "paused":
+            self.pause_button.setText("▶ Resume")
+            self.pause_button.setEnabled(not self._stop_in_flight)
+            self.pause_button.setToolTip("Resume the paused scan.")
+        elif state == "running":
+            self.pause_button.setText(self._pause_button_label)
+            self.pause_button.setEnabled(not self._stop_in_flight)
+            self.pause_button.setToolTip(
+                "Pause the scan at its next safe point (the machine goes to "
+                "its quiescent state; Resume or Stop from here)."
+            )
+        else:
+            self.pause_button.setText(self._pause_button_label)
+            self.pause_button.setEnabled(False)
 
     def _ensure_submitter(self) -> Optional[Submitter]:
         """Return the injected submitter, or lazily build the real engine."""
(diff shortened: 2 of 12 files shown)


skbarber:
## Adversarial review — PR #582 (operator Pause/Resume)

Three lenses (correctness / redundancy / placement). Findings below survived my own refutation; scenarios that didn't are noted at the end so you can see they were checked.

### 1. [Correctness — moderate] A manual pause colliding with an action-during-scan silently drops the action *and* leaks its factory

`on_pause` reads both flags under the lock, then the manual branch returns early:

```python
with self._lock:
    pending, self._pending = self._pending, None          # the staged action
    manual_pause, self._manual_pause = self._manual_pause, False
...
if manual_pause:
    outcome = self._park_manual_pause()
    return outcome          # pending taken, never executed, never cleaned up
```

`pending.cleanup()` lives only in the inner `try/finally` of the `pending is not None` branch, which the early `return` skips. `pending` was already cleared from the slot, so the scan-end `take_unconsumed_pending()` recovery in `bluesky_scanner.py:1079` returns `None` — the connected `CaActionSignalFactory` is never disconnected (violating `PendingAction.cleanup`'s documented "always called exactly once" contract), and the operator's requested action vanishes with no dialog and no log line beyond the generic manual-resume message.

**Reachable through the console**, both orderings:
- Operator clicks **Pause** (`request_pause` → `arm_manual_pause`, state → PAUSING). While still PAUSING (before PAUSED), the action dialog's Run button reads "Pause scan & run" — `set_scanning(True)` fires for `"pausing"` — so a click routes to `request_action_during_scan`, which `set_pending`s successfully (slot empty).
- Or the reverse: stage the action first, then click Pause.

Neither `request_pause` nor `request_action_during_scan` guards against the other (`_manual_pause` and the `_pending` slot are independent), so both end up set and `on_pause`'s manual branch wins. Minimum fix: run `pending.cleanup()` on the manual-wins path (and decide the intended precedence — either refuse one while the other is armed, or process the action after the manual park).

### 2. [Docs — minor] `sed resume_scan→request_resume` collateral in `GeecsBluesky/CHANGELOG.md`

The rename over-applied to the changelog and misapplied in two spots:

- **Line 12 (new 0.45.0 entry):** ``BlueskyScanner.request_pause()` / `resume_scan()`` — the bridge method is `request_resume()`; `resume_scan()` was *missed* by the sed here.
- **Line 117 (historical "Removed" tombstone):** now reads ``**BlueskyScanner.pause_scan / request_resume** — the old hard-pause API``. The sed *corrupted* history: that entry documented removing `resume_scan`; it now claims the brand-new `request_resume` was removed in a prior version — self-contradictory against the 0.45.0 "Added" entry.

Impl comment (`bluesky_scanner.py:417`) and the tombstone test (`test_pause_checkpoints.py:251-254`, asserting the OLD `pause_scan`/`resume_scan` absent) are both correct and intact. `BlueskyScanner` satisfies the console `Submitter` — both `request_pause`/`request_resume` present, `(self) -> None`, matching the protocol.

### 3. [Redundancy — minor] `request_pause` duplicates the PAUSING+deferred-pause scaffold

`request_pause` and `request_action_during_scan` share the same tail: `is_scanning_active` guard → arm/stage → `_set_state("PAUSING")` → `self._RE.request_pause(defer=True)` in a `try/except`. A small `_begin_deferred_pause()` helper would centralize it. The error handling differs (the action path withdraws the pending on failure), so it's a genuine-but-minor extraction, not a bug.

---

### Refuted (checked, no finding)

- **(a) resume-before-park race:** `arm_manual_pause` clears `_resume_event` *before* the pause is requested, and the button only enables as "Resume" once `on_pause` has emitted PAUSED (i.e. after it consumed `_manual_pause` and is about to park). So `request_resume`'s `set()` always follows the `clear()`; `_resume_event.wait()` returns immediately — no lost or spurious resume. Cross-window leak is also handled: the next `arm_manual_pause` re-clears. (The only lost-resume ordering — a second `arm_manual_pause` clearing an already-set event mid-park — needs a Pause click while PAUSED, which the button, reading "Resume", never issues; direct-API misuse only.)
- **(c) button stuck as Resume after abort:** `_on_scan_state` writes `_scan_state_text` on every event and `_refresh_pause_button` gates on it; `"aborted"`/`"done"` fall to the disabled branch (pinned by `test_terminal_state_disables_the_pause_button`), and the `_stop_in_flight` guard disables Resume during a Stop. The paint/click race in `_on_pause_clicked` is bounded by the `is_scanning_active()` guard — worst case one engine-side no-op pause/resume.
- **(e) `request_resume` with nothing paused:** sets `_resume_event`, but `arm_manual_pause` clears it before the next window and the action-pause path waits on `response_event`, not `_resume_event` — no contamination. `request_pause`/`request_resume` are both None-supervisor guarded.
- **Placement:** the resume `Event` must be set on the GUI thread and waited on the scan thread inside the pause window the supervisor already owns — the supervisor is its natural home, not a bridge flag.

skbarber:

**3 findings** (the reviewer refuted an extensive probed set on inspection — the resume-before-park race is safe because `arm_manual_pause` clears the event before the button ever enables as Resume; button-stuck-after-abort is handled by the `_scan_state_text` gating + terminal disables; the stray-resume leak is cleared by the next `arm_manual_pause`; and the supervisor is the correct home for the resume Event):

### 1. MODERATE — manual pause + a staged action drops the action and leaks its factory
In `on_pause`, both `_pending` and `_manual_pause` are read under the lock, then the manual branch returns early — skipping the `pending.cleanup()` in the inner `try/finally`. A staged action's connected `CaActionSignalFactory` is never disconnected (violating cleanup's exactly-once contract), the action is silently dropped with no dialog, and it's unrecoverable at scan end (`take_unconsumed_pending` returns None — the slot was already cleared). Reachable through the console in both orders.

### 2. MINOR (docs) — `sed resume_scan→request_resume` collateral in `GeecsBluesky/CHANGELOG.md`
Line 12 (0.45.0 entry) said `resume_scan()` for the method actually named `request_resume()`; line 117 (historical tombstone) was corrupted to claim the brand-new `request_resume` was removed in a prior version.

### 3. MINOR (redundancy) — `request_pause` duplicates the PAUSING + `RE.request_pause(defer=True)` scaffold of `request_action_during_scan`
A `_begin_deferred_pause()` helper would centralize it (error handling differs slightly).

skbarber:
## Disposition (findings fixed in 8cdccefb, reviewer-confirmed)

| # | Finding | Disposition |
|---|---|---|
| 1 | MODERATE: manual pause + a staged action drops the action and leaks its factory | **Fixed** — `on_pause` restructured so a staged action wins (`if pending is not None` runs the decision first, never dropped), and `pending.cleanup()` moved to the outer `finally` (guarded `if pending is not None`) so the factory disconnects exactly once on every branch/exception path. Pinned by a both-armed test. Reviewer: **confirmed** |
| 2 | MINOR: CHANGELOG sed-rename collateral | **Fixed** — line 12 now `request_resume()` (real method), line 117 tombstone restored to `resume_scan` (old deleted name). Reviewer: **confirmed** |
| 3 | MINOR: `request_pause` duplicates the deferred-pause scaffold | **Waived** — the two error-handling shapes (validate/stage/rollback vs log-only) genuinely differ; a shared helper trades ~4 lines for more coupling. Optional follow-up. Reviewer: **reasonable** |

CI green on both commits. Merging.

PR #583, opened 2026-07-17
GeecsBluesky 0.45.1: emit RUNNING on resume — multiple pauses per scan

skbarber:
Operator-pause follow-up (issue #552). **Bug found in live testing:** after Resume, the console's Pause button stays on "Resume" and the pill on PAUSED, so you only get **one pause per scan** — a second click hits the no-op resume path.

## What + why

The **engine already supports repeated pauses** — `_run_supervised`'s loop re-enters `on_pause` on every deferred pause. The bug is purely that the GUI never learns the scan resumed: the pause supervisor emits `on_state("paused")` when a window opens but had **no symmetric close**, so nothing flipped the console back to RUNNING. The state pill and the Pause/Resume button both key off the lifecycle state word, so both stuck on PAUSED.

Fix: the supervisor emits `on_state("running")` when a pause window ends in a **resume** (not on abort — ABORTED comes from the bridge's abort path). One symmetric line in the `finally`. Covers both the bare operator pause (#582) and the action-triggered pause (#581) — both stuck identically.

Why the hardware checklist missed it: item 4 verified *data* integrity through a pause→execute→resume (shot_offset labels, s-file) — all unaffected by a stale pill. The stuck UI state corrupts nothing; it just caps you at one pause.

## Per-concern LOC breakdown

- Supervisor `on_state("running")` on resume (`pause_supervisor.py`): +12 (mostly the why-comment)
- Tests: bluesky +2 (resume-emits-running, abort-does-not); console +1 (full running→paused→running cycle + a second pause click works)
- Version bump + CHANGELOG (0.45.1 patch): +16

Console is test-only (a new pin against the engine's emit) — no console code changed, so unbumped.

## Tests

- `./scripts/check.sh`: **GeecsBluesky 580 passed** (+2), **GEECS-Console 762 passed** (+1; the 2 `*_without_the_extra` optimization failures are the known local-extra issue). The console test drives the full cycle and asserts a *second* Pause click reaches `request_pause`.

## Hardware verification

**N/A for correctness** — pure lifecycle-event emission; no scan execution, device, or timing path changed. Worth an eyeball at the beamline next session (pause → resume → pause again → the button cycles), but nothing data-affecting.

### DIFF
--- a/GEECS-Console/tests/test_main_window.py
+++ b/GEECS-Console/tests/test_main_window.py
@@ -782,6 +782,25 @@ class TestSubmission:
         assert window._submitter.resumes == 1
         assert window._submitter.pauses == 0  # resume, not a second pause
 
+    def test_button_returns_to_pause_after_resume_running_event(self, window):
+        """Multiple pauses per scan: the engine emits RUNNING on resume, so
+        the button flips back to Pause and a second Pause click works."""
+        from fake_events import ScanLifecycleEvent
+
+        select_save_set(window, "Amp4In")
+        window._on_start_clicked()
+        window.events.handle(ScanLifecycleEvent(state="running"))
+        window.pause_button.click()  # pause
+        window.events.handle(ScanLifecycleEvent(state="paused"))
+        assert "Resume" in window.pause_button.text()
+        window.pause_button.click()  # resume
+        # The engine's resume emits RUNNING; the button must flip back.
+        window.events.handle(ScanLifecycleEvent(state="running"))
+        assert "Pause" in window.pause_button.text()
+        assert window.pause_button.isEnabled()
+        window.pause_button.click()  # a SECOND pause must work
+        assert window._submitter.pauses == 2
+
     def test_pause_button_disabled_during_pausing_transition(self, window):
         from fake_events import ScanLifecycleEvent
 
--- a/GeecsBluesky/CHANGELOG.md
+++ b/GeecsBluesky/CHANGELOG.md
@@ -4,6 +4,21 @@ All notable changes to `geecs-bluesky` are documented here.
 
 Format follows [Keep a Changelog](https://keepachangelog.com/en/1.1.0/).
 
+## [0.45.1] - 2026-07-16
+
+### Fixed
+
+- **Multiple pauses per scan** (operator-pause follow-up): the pause
+  supervisor now emits `on_state("running")` when a pause window ends in a
+  resume — the symmetric close of the `on_state("paused")` it emits at
+  window open.  Without it the console's state pill and Pause/Resume button
+  stayed stuck on PAUSED/"Resume" after resuming, so a second Pause click
+  hit the no-op resume path — one pause per scan.  The engine already
+  supported repeated pauses (the supervised-run loop re-enters `on_pause`
+  per deferred pause); this just lets the GUI keep up.  Covers both the
+  bare operator pause and the action-triggered pause.  Pinned by
+  `tests/test_pause_supervisor.py`.
+
 ## [0.45.0] - 2026-07-16
 
 ### Added
--- a/GeecsBluesky/geecs_bluesky/pause_supervisor.py
+++ b/GeecsBluesky/geecs_bluesky/pause_supervisor.py
@@ -250,8 +250,18 @@ class PauseSupervisor:
             # The abort outcome deliberately skips the restore: RE.abort()'s
             # finalize chain (quiesce → save-off → disarm → closeout) owns
             # the end state.
-            if drove_safe_state and entry_state and outcome == "resume":
-                self._drive_state(session, controller, entry_state)
+            if outcome == "resume":
+                if drove_safe_state and entry_state:
+                    self._drive_state(session, controller, entry_state)
+                # The scan is about to resume — flip the consumer back from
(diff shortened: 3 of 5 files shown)


skbarber:
## Adversarial review — PR #583 (emit RUNNING on resume)

Traced correctness, redundancy, and placement. **No surviving substantive findings** — the fix is correct, necessary, and well-placed. Details of the refutation below, plus one low-severity robustness note.

### Correctness — all clear

- **Every resume exit emits exactly one `running`; every abort emits none.** `outcome` starts `"resume"` and only the two abort branches (`_decide_and_execute` → `"abort"`, `_park_manual_pause` → `"abort"`) flip it. The `finally` gate `outcome == "resume"` therefore fires on manual-resume, action-execute, action-ignore, failed-step→retry→resume, and the spurious no-pending path — and is skipped on every abort. `on_state("paused")` is emitted unconditionally at window open (line 220), so even the spurious path gets a symmetric open/close. Matches the two new supervisor tests.
- **Ordering (restore → emit) is right.** The machine is driven back to `entry_state` *before* the GUI is told it is running, so there is never a window where the pill says RUNNING while the hardware is still in its safe state.
- **Transient RUNNING before DONE is benign.** After the `finally` emits `running`, `_run_supervised` calls `RE.resume()`, which may run to completion → bridge emits DONE. The console sequence PAUSED→RUNNING→DONE is a correct, sub-perceptible pill flip; the RUNNING blip re-enables the Pause button only until DONE disables it. No wrong gating.
- **`_set_state("RUNNING")` is idempotent.** It only sets `_current_state` and emits a `ScanLifecycleEvent`; it does **not** re-claim the scan number (`_note_claimed_scan_number` runs solely from `_on_document` start docs, and early-returns once `_scan_number` is set) and does **not** touch `_completed_shots`/progress. Re-emitting RUNNING after the pre-pause RUNNING causes no duplicate claim, no progress reset.
- **`_stop_in_flight` interaction is safe.** `_refresh_pause_button` gates the button on `not self._stop_in_flight` in both the paused and running branches. A `running` event racing an in-flight Stop leaves the button disabled; and a Stop while paused takes the abort path (supervisor is sole abort owner), which emits no `running` at all.
- **Threading discipline respected.** The emit runs on the scan thread → `on_state` → `_set_state` → `_on_event` (the adapter), which marshals to the GUI thread via queued Qt signals. Same path as every other lifecycle emission; correct.

### Redundancy — none

`RE.resume()` (session.py:632) continues the same run and emits no new `start` document, so `_on_document`/`_note_claimed_scan_number` never re-emit RUNNING on resume. Nothing else in the resume path announces a state. The supervisor emit is the *only* source of post-resume RUNNING — necessary, not duplicative.

### Placement — correct

Supervisor `finally` (scan thread, knows the true outcome and owns the restore) is the right home, not the bridge's `_run_supervised` loop — the bridge cannot distinguish resume-vs-abort without duplicating the supervisor's decision logic, and the emit must land after the restore that only the supervisor performs.

### Console test — non-vacuous but seam-split (not a defect)

`test_button_returns_to_pause_after_resume_running_event` injects the `running` event manually, so it would **not** fail if the engine emit regressed — it pins the *console half* (RUNNING → button flips to Pause, second pause works). The *engine half* (resume actually emits `running`) is pinned separately by the new `test_pause_supervisor.py` tests. Correct two-package decomposition; together they cover the cycle.

### Low-severity robustness note (largely refuted)

The `running` emit sits after `self._drive_state(...)` in the `finally`; if that call raised, the emit would be skipped and the console left stuck PAUSED. In practice `_drive_state` is exception-safe by construction — it swallows per-write failures (`future.result` in try/except, returns a bool). It could only raise from exotic conditions (`controller.state_setters()` throwing, or a torn-down loop), and those same conditions would derail `RE.resume()` itself, so a stuck pill would be the least of the problems. Not worth a code change, but if a belt-and-suspenders guarantee were ever wanted, structuring the emit so it cannot be gated by a restore exception would make the "symmetric close" invariant unconditional.

skbarber:

**No surviving findings.** The reviewer traced all three lenses adversarially:

- **Correctness** — every resume exit (manual, action-execute, action-ignore, failed-step→retry→resume, spurious no-pending) emits exactly one `running`; every abort emits none (symmetric open/close via the `outcome == "resume"` gate against the unconditional `on_state("paused")` at open). Restore-then-emit ordering right; the transient PAUSED→RUNNING→DONE pill flip is benign. `_set_state("RUNNING")` idempotent — no scan-number re-claim (that lives in `_on_document`, which early-returns once `_scan_number` is set), no progress reset. The `_stop_in_flight` guard keeps Pause correctly disabled if a `running` races an in-flight Stop. Threading discipline respected (scan thread → adapter → queued signals).
- **Redundancy** — none; `RE.resume()` emits no new start document, so the supervisor is the sole RUNNING-on-resume source.
- **Placement** — supervisor `finally` is correct; the bridge couldn't distinguish resume-vs-abort without duplicating supervisor logic.
- **Console test** — non-vacuous (a deliberate seam-split: injects the `running` event to pin the console half; the engine emit is pinned by the new supervisor tests).

One low note, self-refuted: the emit sits after `_drive_state` in the finally, so a raise there would skip it — but `_drive_state` is exception-safe by construction, and anything exotic enough to make it raise would derail `RE.resume()` anyway. Not worth a change.

CI green. Merging.

PR #585, opened 2026-07-21
Lowercase PV naming contract, one :SP owner, safe_name unification

skbarber:
## What

Standardizes the naming boundary decided in today's design discussion: **GEECS-native names are the canonical vocabulary; PV names and event-column names are ONE derived lowercase encoding**, produced only by `pv_naming`.

- **`normalize_component` case-folds** — every PV component is lowercase (`Undulator:U_S1H:Current` → `undulator:u_s1h:current`). Case carries no meaning anywhere in GEECS; folding makes every derived name case-insensitive to operator input for free (any casing typed into the console R7 combo resolves to the same PV).
- **`pv_naming.setpoint_pv` + `SETPOINT_SUFFIX`** — the one place the `:SP` suffix is applied. Gateway pvdb build, GeecsBluesky (`shot_controller`, `settable`, `timestamped_readable`, `generic_detector`, `action_signals`, re-export via `devices/ca/_pv.py`), and the console R7 backend all use it; zero hand-built `f"...:SP"` strings remain.
- **`safe_name` delegates to `normalize_component`** (+ the non-empty `"var"` fallback) — event-document columns and PV components now mangle identically; the agreement is pinned by a new cross-check test. Column-name change: runs of specials collapse to one underscore (`Wavelength (nm)` → `wavelength_nm`, was `wavelength__nm`).
- **Contracts updated in the same PR** (per the gateway's contract rule): `PV_CONTRACT.md` §1 (case-fold step, lowercase examples/status namespace, canonical-vocabulary statement, `setpoint_pv`, test map), `EVENT_SCHEMA.md` (column-convention note; schema version stays 1 — runs are self-describing, pre-0.46.0 runs remain readable via `geecs_scalar_headers`), `DEPLOYMENT.md` smoke-test PVs, `docs/geecs_gateway/client_overview.md`, CLAUDE.md placeholders.

## Why

The console R7 investigation surfaced the naming conflation: three vocabularies (GEECS raw, gateway PV, event columns) with two independent lossy normalizers and case-sensitivity nobody wants. Owner decision (2026-07-20): lowercase everything, unify the normalizers, do it now while pre-production (Phoebus screens are play-only; Tiled may be purged, though it doesn't need to be).

## Per-concern LOC (production code)

| Concern | Files | ~LOC |
|---|---|---|
| Case-fold + `setpoint_pv` helper (`pv_naming.py`) | 1 | +40/−10 |
[...]

PR #586, opened 2026-07-21
GeecsCAGateway 0.13.5: extract DB connection layer in geecs_db

skbarber:
## What & why

Internal refactor of `GeecsCAGateway/geecs_ca_gateway/db/geecs_db.py`. Every
`GeecsDb` query classmethod repeated the same two blocks verbatim: the
`mysql-connector-python` import guard, and the connect / cursor / `finally: close`
dance (~14 lines of ceremony per method, ~8 methods). This PR extracts that into
a connection layer so each method is just its SQL plus its result shaping.

**No behavior change.** Return shapes, error semantics, and query text are all
identical; this is pure boilerplate extraction. `PV_CONTRACT.md` is unaffected
(no externally observable surface changes).

## Concerns (single concern, one file of logic)

- **New connection layer** (`geecs_db.py`): a `_cursor()` `@contextmanager`
  owns the import guard + pure-Python connection + guaranteed close; a thin
  `_query(sql, params, *, fetch=)` primitive wraps single-statement lookups.
- **8 methods de-boilerplated**: `find_device`, `get_device_type`,
  `get_subscribed_variables`, `get_all_experiment_variables`,
  `get_scan_boundary_writes`, `get_device_variables`, `get_experiment_devices`,
  `get_experiment_device_variables`, `get_fk_orphan_variables`,
  `get_ca_alarm_limits`. The two-statement batch methods (type + instance;
  count + sample) reuse **one** `_cursor()` block so both statements stay on a
  single connection — no extra connections opened.
- **`get_ca_alarm_limits` de-specialized**: previously it had bespoke
  try/except around both connect and execute. It now goes through `_query()`
  with a single fail-open `try/except` at the call site that preserves the
  exact prior contract — missing table (errno 1146) → INFO, any other DB
  failure → WARNING, both returning `{}` so IOC startup never aborts on the
  optional overlay table. `ImportError` deliberately re-raises (a missing
  connector is a deployment error, not an absent overlay).

Net: `3 files changed, 157 insertions(+), 210 deletions(-)` — ~130 lines of
method boilerplate removed (the file grows a ~55-line documented helper +
CHANGELOG in return).

## Tests

- `test_geecs_db.py`: **16 passed, unchanged** — the tests mock at
  `_connect_mysql` / the injected `mysql` module, both *below* the new
  `_cursor`/`_query` layer, so they exercise the refactor transparently. This is
  the strongest safety net for a "no behavior change" claim: the existing
  behavioral pins still hold with zero test edits.
- Full gateway suite: **181 passed** (13.25s), offline.
- `pre-commit` (ruff, ruff-format, pydocstyle) clean on the changed file.

## Hardware verification

N/A — internal refactor of DB-client plumbing, no scan-execution or device
path touched. Behavior is pinned by the offline suite; return shapes and error
semantics are byte-for-byte identical.

## ⚠️ Changelog / version coordination

Another agent's branch `gateway/lowercase-naming-contract` (worktree
`naming-lowercase`) is concurrently modifying `GeecsCAGateway` and bumps it to
**0.14.0** (minor, lowercase naming contract). It does **not** touch
`geecs_db.py`, so there is **no code conflict** — the only overlap is
`CHANGELOG.md` + `pyproject.toml` version.

This PR bumps to **0.13.5** (patch), correct against the current `origin/dev`
(0.13.4). If the 0.14.0 PR merges first, this one will need a trivial
re-number to **0.14.1** and a CHANGELOG re-stitch before merge — handled at
merge time, flagged here so it's not a surprise.

### DIFF
--- a/GeecsCAGateway/CHANGELOG.md
+++ b/GeecsCAGateway/CHANGELOG.md
@@ -3,6 +3,28 @@
 All notable changes to `geecs-ca-gateway` are documented here, following
 [Keep a Changelog](https://keepachangelog.com/en/1.1.0/) and semantic versioning.
 
+## [0.14.1] - 2026-07-20
+
+### Changed
+
+- `db/geecs_db.py`: internal refactor extracting a connection layer. A single
+  `_cursor()` context manager now owns the connector-import guard, the
+  pure-Python connection, and the guaranteed close; a thin `_query()` primitive
+  wraps single-statement lookups. All eight `GeecsDb` query methods drop their
+  repeated import-guard + connect/close boilerplate (~130 lines) and keep only
+  their SQL and result shaping; the two-statement batch methods reuse one
+  `_cursor()` block to stay on a single connection. No behavior change — the
+  full `test_geecs_db.py` suite passes unchanged, and the externally observable
+  surface (return shapes, error semantics) is identical, so `PV_CONTRACT.md` is
+  unaffected.
+- `get_ca_alarm_limits()` is no longer special-cased around its own
+  connect/execute error handling. It now goes through `_query()` like every
+  other method, with a single fail-open `try/except` at the call site that
+  preserves the exact prior contract (missing-table → INFO, other DB failure →
+  WARNING, both returning `{}` so IOC startup never aborts on the optional
+  overlay table). `ImportError` still propagates — a missing connector is a
+  deployment error, not an absent overlay.
+
 ## [0.14.0] - 2026-07-20
 
 ### Changed
--- a/GeecsCAGateway/geecs_ca_gateway/db/geecs_db.py
+++ b/GeecsCAGateway/geecs_ca_gateway/db/geecs_db.py
@@ -19,8 +19,9 @@ from __future__ import annotations
 import configparser
 import logging
 import os
+from contextlib import contextmanager
 from pathlib import Path
-from typing import Optional
+from typing import Iterator, Optional
 
 from pydantic import ValidationError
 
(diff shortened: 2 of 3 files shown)


skbarber:

Reviewed the diff through the three lenses (correctness / redundancy / placement). Report and dispositions below.

### Correctness — challenged, held
- **Connection close on all paths**: `_cursor()` closes the connection whether `conn.cursor()` raises, the `with` body raises, or `_connect_mysql` raises before a connection exists. ✓
- **Fetch semantics**: `_query(..., fetch="one")` → `fetchone()` (row or `None`); default → `fetchall()`. Matches every caller's prior expectation. ✓
- **`get_ca_alarm_limits` fail-open contract**: missing-table (errno 1146) → INFO + `{}`; any other DB error incl. connect failure → WARNING + `{}`; `ImportError` still propagates (not swallowed). ✓ — pinned by the three existing alarm tests, which pass unchanged.
- **Single-connection batching**: the two-statement methods (`get_device_variables`, `get_experiment_device_variables`, `get_fk_orphan_variables`) share one `_cursor()` block; the `test_get_*_batches_*` tests assert `len(queries) == 2` on one connection and would fail if a second connection were opened. ✓

### Redundancy — none found
Reviewer searched GEECS-PythonAPI's `geecs_database`, GEECS-Data-Utils, and the rest of the gateway; no existing connection-wrapper / query-helper this duplicates. GeecsCAGateway is the sole MySQL layer for the vision world.

### Placement — correct
`_cursor` / `_query` are private module-level helpers in the module that owns `GeecsDb`. No dependency-graph or boundary issue; GeecsBluesky imports only the library surface, unchanged.

### Surviving finding (1, minor) — dispositioned
**F1 (minor, diagnostic text):** the connect-failure branch of `get_ca_alarm_limits` previously logged `"ca_alarm_limits lookup failed before query"`; it now logs `"ca_alarm_limits lookup failed"`. Still WARNING level, still returns `{}` — no test keys on the text and no consumer parses it.

**Disposition: waived (accepted simplification), with accuracy fix.** Restoring the "before query" qualifier would require re-introducing the separate connect-vs-execute try blocks the refactor deliberately collapses, for a message-text nuance nothing reads. The observable contract (level + return) is unchanged. The commit message and this comment now state the text change explicitly so it's on the record rather than hidden behind "exact prior contract."

No other findings survived challenge.

---
**Note on version:** rebased onto `dev` after #585 (0.14.0) merged; re-numbered 0.13.5 → **0.14.1** and re-stitched the CHANGELOG. Gateway suite **182 passed** against the rebased base; `test_geecs_db.py` (16) unchanged.

PR #589, opened 2026-07-21
GEECS-Console 0.18.1: ActionsMenuController extraction — #534 step 3

skbarber:
Step 3 of the #534 plan of record (per the 2026-07-20 refresh comment there): the Actions menu moves out of `main_window.py` into `app/actions_menu.py::ActionsMenuController`.

## Per-concern breakdown

| Concern | LOC | What |
|---|---|---|
| **Extraction (the step itself)** | +273 module, −109 window (2,290 → 2,181) | Menu contents, arming switch, plan-name fetch (worker owned by the controller), `ActionRunDialog` bookkeeping, `set_scanning` push. Boundary rulings per the plan refresh: the action-decision modal and the scan-lifecycle fan-out **stay in the window**. Thin test surface kept: `enable_actions_action` / `_open_action_dialogs` properties; one test call site migrated to the controller. The controller severs its window references in `closeEvent` (`dispose()`) — a window↔controller reference cycle would otherwise defer dead-window teardown to the cyclic GC mid-event-processing. |
| **Bug fix surfaced by the extraction** | ~20 | **Startup double-fetch races the lazy `geecs_bluesky` import** — construction fires the plan-name fetch twice back-to-back (restore-fired experiment change + explicit call); two daemon threads first-importing the native extension chain **abort the process**. Latent on `dev` (timing luck); the extraction's tighter spawn timing made it a deterministic segfault under offscreen pytest. Fix: one in-flight fetch per experiment, failure delivers an empty answer (never a stuck tag). Pinned by `test_concurrent_fetches_for_one_experiment_are_deduplicated`. |
| **Test hermeticity fix** | ~25 (conftest) | Windows built without an injected `action_store` were reaching the **real** configs repo + real user config from fetch threads on every construction — the exact gap conftest's `_offline_window_defaults` already closes for the completions and idle-scan seams; the action store was missed when G-actions landed. Now neutralized identically. (An earlier ~3× suite-speed claim did not reproduce under review — locally ~8%; the win is determinism and hermeticity, with any speedup environment-dependent.) |
| **Flagged secondary concern: docstring trim** | docs-only, ~40 lines | 0.18.0-era docstrings in `editors/base.py` + editor tests trimmed to present-tense constraints; provenance (issue/PR numbers, review-finding citations) removed per owner feedback — history lives in commits/PRs. |

## Deliberately unchanged
[...]

Issue #534, opened 2026-07-13
Console: main_window.py accretion watch (already at the documented danger size)


skbarber:
From the post-#483 audit of GEECS-Console.

`app/main_window.py` is at 2,116 lines — the same size at which `geecs_scanner.py` was declared organizational debt in the root CLAUDE.md. It is in much better shape than its predecessor (engine logic, path resolution, request building, and persistence are all extracted behind seams; the region-comment structure is disciplined), but the growth vector is real: +~350 lines in 0.3 (device panel), +~250 in 0.5 (ops/beeps), +~300 in 0.7 (tooltips/workers).

Bounded extractions, cheapest first — don't do a big-bang split; let the next feature that touches a cluster pay for its extraction:

1. the module-level worker/filter classes (~130 lines) → `services/background.py` (see the background-extraction issue)
2. `_apply_operator_tooltips`'s 165-line dict → a data module (`app/operator_tooltips.py`)
3. if R7 grows composites support, the device-panel handler cluster (~170 lines) → its own controller

🤖 Filed from the post-483 cleanup audit (Claude Code)

skbarber:
Plan of record (owner-approved 2026-07-16; built from a structural scan at 2,390 lines — past the documented ~2,200 danger size after the optimization + stop-lifecycle work). Five mechanical extractions, each a single-concern PR; 2–5 deliberately wait for their forcing functions per repo policy:

- [x] **Step 1 — `HealthPoller` → `services/background.py` (both blessed worker shapes in one module) + `ToolTipSuppressor` and the operator-tooltip catalog → new `app/tooltips.py`** (pure attribute-name → text data, fail-loud on a missing widget). Landed in PR #577 (0.15.1): 2,390 → 2,124 lines, reviewer-verified semantically exact (AST-compared tooltips, diffed class bodies).
- [ ] **Step 2 — R3 scan form → `app/scan_form.py`** (~400 lines: mode radios, axis rows, optimization combo + iterations, shot-count guard, `form_state()`/`_apply_form_state()`). **Paired with #558** — the prerequisite that makes N-axis grids a bounded change.
- [ ] **Step 3 — Actions-menu cluster → `app/actions_menu.py`** (~110 lines). **Paired with #575/#552** — extract before reworking the firing surface.
- [ ] **Step 4 — R7 device panel → `app/device_panel.py`** (~170 lines). When R7 next changes (pseudo-variable runtime).
[...]

Issue #533, opened 2026-07-13
Console: editor scaffolding base class — extract before editor #5


skbarber:
From the post-#483 audit of GEECS-Console.

The four M5 editors (4,322 lines) share duplicated-shape scaffolding, roughly 700–900 lines: `_load_ui` + `_child` (5 copies incl. main window, ~125 lines), `_start_completions_fetch` (verbatim-identical 16 lines ×3 + a near-variant in `save_set_editor.py`, ≈150 lines with `_apply_completions` prefixes), the Enter-guard (4 variants, ~50 lines), dirty tracking / title-star composition, unsaved-changes prompting, and New/Duplicate/Rename/Delete list management.

**The prompts have already diverged behaviorally:** `save_set_editor._maybe_leave` and `shot_control_editor._prompt_unsaved` offer Save/Discard/Cancel, while `scan_variable_editor._confirm_discard_modal` and `action_library_editor._confirm_discard` offer only Discard/Cancel (no Save at the prompt). Same story for the name prompts (`_prompt_name` vs `_ask_name_modal` vs `_prompt_text` — one returns `Optional[str]`, one `""`-on-cancel). A user-visible inconsistency, which is the classic first symptom.

The root CLAUDE.md records that GEECS-Scanner-GUI grew to 2,200 lines precisely by NOT extracting editor scaffolding, and prescribes a base class for editor windows as the fix once a feature drives it. The forcing functions are already on the roadmap: the deferred config bootstrap/repair dialog and eventually an OptimizationSpec editor.

Recommend: extract `ConfigEditorDialog` (load-ui/child, completions plumbing, dirty/title, one unified Save/Discard/Cancel prompt, list-CRUD hooks) **before** editor #5 is written, with the four existing editor test suites (~1,900 lines) as the safety net. Depends loosely on the services/background.py extraction for the completions-fetch part.

🤖 Filed from the post-483 cleanup audit (Claude Code)

skbarber:
Closed by #588 (ConfigEditorDialog, GEECS-Console 0.18.0). **Deferred follow-ups recorded here** so they aren't lost with the close:

[...]

PR #590, opened 2026-07-21
GEECS-Console 0.18.2: NowPanelController extraction — #534 step 5

skbarber:
Step 5 of the #534 plan of record (per the 2026-07-20 refresh comment): the R6 now-panel rendering moves out of `main_window.py` into `app/now_panel.py::NowPanelController`.

## Per-concern breakdown

| Concern | LOC | What |
|---|---|---|
| **Extraction (the step itself)** | +262 module, −78 window (2,181 → 2,103) | State pill (+ the exact `_STATE_DOT_COLORS` dict, transplanted verbatim), progress-bar math (`_total_shots`), scan number + 10 s expiry timer, log tail, idle scan-number probe (worker owned by the controller; lookup resolved at probe time so conftest's module-level patch keeps working). Per the plan refresh: the **scan-lifecycle hub stays in the window** (it fans out to R5 gating and dialog teardown), as do the widget attributes (tests, tooltips), thin `append_log`/`set_scan_number` delegates, and the beep logic (Preferences cluster). #589's lifetime lessons applied up front: no Qt parent, `dispose()` in `closeEvent`. |
| **Idle-probe dedupe** | ~10 | The idle probe gets the same one-in-flight-per-experiment dedupe as the actions fetch (0.18.1) — startup double-spawns it identically, same native-first-import race shape. Of the three startup fetch pairs flagged in #589, **only the completions pair now remains** (R7 territory — #534 step 4's natural scope). |
| **Contract + comment hygiene** | docs | Console CLAUDE.md updated **in this PR** (the #589 finding-1 lesson): R6 seam bullets describe the controller shape, the worker-disconnect split names both controllers, the sibling-hazard note reflects the narrowed state. The startup hazard comment softened to "native init in the geecs_bluesky chain" per the #589 reviewer's terminology nit (EPICS specifically was never demonstrated). |

## Deliberately unchanged
Zero user-visible behavior intended. Test migrations: four private grips (`_apply_idle_scan_number` ×2, `_scan_number_timer`, `_expire_scan_number`) move to `window._now.…`; the widget attributes and public methods are untouched surfaces.

## Tests
Full console suite **770 passed**, verified **3× consecutively** (same flaky-crash caution as #589).

## Hardware verification
None owed — rendering/window plumbing only; nothing touches scan execution or devices.

### DIFF
--- a/GEECS-Console/CHANGELOG.md
+++ b/GEECS-Console/CHANGELOG.md
@@ -4,6 +4,25 @@ All notable changes to GEECS-Console are documented here.  Format follows
 [Keep a Changelog](https://keepachangelog.com/en/1.1.0/); versioning is
 semantic.
 
+## [0.18.2] - 2026-07-20
+
+### Changed
+
+- Main-window slimming step 5 (issue #534): the R6 now-panel rendering —
+  state pill, progress bar, scan number + 10 s expiry, log tail, and the
+  idle scan-number probe — moves to
+  `app/now_panel.py::NowPanelController` (2,181 → 2,103 window lines
+  against a 262-line module).  The window keeps the widget attributes,
+  thin `append_log`/`set_scan_number` delegates, the beep logic
+  (Preferences), and the scan-lifecycle hub.
+
+### Fixed
+
+- The idle scan-number probe gets the same one-in-flight-per-experiment
+  dedupe as the actions fetch (0.18.1) — startup double-spawned it, the
+  same native-first-import race shape; of the three startup fetch pairs
+  only the completions pair now remains undeduped.
+
 ## [0.18.1] - 2026-07-20
 
 ### Changed
--- a/GEECS-Console/CLAUDE.md
+++ b/GEECS-Console/CLAUDE.md
@@ -179,8 +179,8 @@ are prefixed by region (`r3_radio_1d`, `r5_start_button`, …).
 - **Scan number (R6)**: the engine's `ScanLifecycleEvent.scan_number`
   (`None` until the scan folder is claimed, then present on every lifecycle
   emission) is read duck-typed by `ScanEventsAdapter` and emitted as
-  `scan_number_known(int)`; the window connects it to `set_scan_number`
-  (10 s expiry to "(previous)").
+  `scan_number_known(int)`; the window connects it to `set_scan_number`,
+  which delegates to `NowPanelController` (10 s expiry to "(previous)").
 - **Ops menu**: four items, handlers in `main_window.py`, path resolution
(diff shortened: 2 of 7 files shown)


skbarber:

Reviewer verified every moved R6 method line-by-line against dev (state-pill markup + the full color dict, progress math including both clamps and the totals/steps fallback, scan-number formatting + expiry semantics, stale-drop and never-clobber rules), traced the dedupe state machine through experiment-switch permutations, checked the lookup-closure/conftest-patch wiring both ways, and ran the suite 3×. **3 surviving findings:**

1. **MODERATE** — the idle-probe dedupe (listed under Fixed) was pinned by **no test**: deleting the entire dedupe left all 770 tests green — the exact gap #589's own pin test exists for, one PR earlier.
2. **LOW** — no empty-experiment guard on the probe, and `dispose()`'s `lambda: ""` sentinel exactly **matches** an empty-experiment probe tag: a straggling `("",  None)` result delivered post-dispose threads the stale-drop guard by construction and touches a possibly-destroyed label.
3. **MINOR** — two doc inaccuracies: dispose() claimed to drop widget references it doesn't (only the closures are severed — which is correct and sufficient), and the startup hazard comment attributed the demonstrated native-init abort to the wrong lazy-import chain.

Verified-clean: zero drift in the moved code (colors, math, formats, guards all verbatim), beep split bit-identical, dedupe state machine correct across A→B→A switches, conftest patch still effective through the resolve-at-probe-time closure, no redundancy (browser B-regions checked), contract travel accurate.

## Disposition

| # | Disposition |
|---|---|
| 1 | **Fixed** (39fb3166) — the #589-shape pin transplanted. Reviewer **mutation-tested** the fix: deleting the dedupe guard now fails the test. **Confirmed** |
| 2 | **Fixed** (39fb3166) — empty-experiment probes answer inline (no thread, no "" tag; the needle-thread is unreachable), pinned by its own test. One accepted semantic delta noted by the reviewer: an injected lookup returning a number for "" would now render "No scans today" — no caller or test does this; matches the actions-controller precedent. **Confirmed** |
| 3 | **Fixed** (39fb3166) — dispose() docstring states exactly what is severed; hazard comment attributes the abort to the actions chain and names the idle probe's own. **Confirmed** |
| — | **Palette-constant duplication: waived** — bundling a 3-file palette consolidation into a review-fix commit stretches PR-scope discipline further than 4 duplicated hex constants justify; the #533 deferral + breadcrumbs keep it discoverable. Reviewer: "convinced, no objection." |

Reviewer verdict: no new findings at 39fb3166; clear to land (772 tests, reproduced independently).

PR #594, opened 2026-07-22
GeecsBluesky 0.47.0: pseudo (composite) scan variables execute on the ScanRequest path

skbarber:
## What & why

Pseudo (composite) scan variables — the legacy scanner's relative steering-bump / absolute-manifold (R56) capability — now **execute** on the Bluesky/ScanRequest path, closing the last big "validated, then refused" v1 gap. Design decisions confirmed with the owner before building: AST-whitelist evaluator (no numexpr dependency, full math freedom kept), `composite_var` + short alias `x` both accepted, setpoint-only completion per component in v1 (legacy `ScanDevice` parity), and optimize-mode pseudo axes included.

### The pieces (GeecsBluesky 0.47.0, minor)

- **`forward_expr.py` (new, 132 LOC)** — `compile_forward`: parses a `forward` formula with `ast`, validates every node against an explicit whitelist (arithmetic operators, parentheses, `sqrt`/trig/`exp`/`log`/`abs`, constants `pi`/`e`, the scanned value as `composite_var` or `x`) **before** anything is evaluated, then returns a `float -> float` callable carrying its source. Anything else — unknown names, attribute access, subscripts, smuggled calls — is `GeecsConfigurationError` at compile time. All 14 production corpus expressions (Undulator + Thomson `composite_variables.yaml`) are pinned expression-by-expression.
- **`devices/ca/pseudo.py` (new, ~170 LOC)** — `CaPseudoMovable`: `set(u)` evaluates each component's formula and puts every target's gateway `:SP` **concurrently** through the shared `GatewaySetpointPut` primitive — each put rides GEECS's native blocking set, so the move completes when the slowest target's exe response lands (identical completion semantics to `CaSettable`, per component). `absolute` = formula value; `relative` = offset from per-target baselines captured once per run at `stage()` (lazily on first `set()` for unstaged callers — the optimize path) and dropped at `unstage()`. The recorded event-row column is the demanded pseudo value (soft readback; header = the catalog friendly name).
- **Runner wiring (~90 LOC changed)** — `resolve_movable_target` now returns a typed `PlainMovableTarget | PseudoMovableTarget` union (formulas compiled fail-fast pre-claim — a bad expression can never burn a scan number) and `build_movable` dispatches to the new `GeecsSession.pseudo_movable` factory. Both call sites — step/grid axes and optimize movables — flow through it, so grids and Xopt optimization over pseudo axes work with no extra code. Provenance (mode + per-target forward sources) lands in run metadata under `pseudo_variables`.
[...]

PR #596, opened 2026-07-22
GEECS-Schemas 0.9.0 + GeecsCAGateway 0.14.3 + GeecsBluesky 0.47.1: one shared AST-whitelist expression core

skbarber:
## What + why

Two AST-whitelist expression evaluators had grown the same compile-then-restricted-eval skeleton with diverged whitelists: the gateway's derived-channel `ExpressionEvaluator` (`geecs_ca_gateway/derived.py`) and the engine's pseudo-variable `compile_forward` (`geecs_bluesky/forward_expr.py`). Both are security-sensitive eval sites, so a hardening or semantics fix had to land twice. This PR extracts the shared skeleton into **`geecs_schemas.restricted_expr`** (stdlib-only: `ast` + `dataclasses`, honoring GEECS-Schemas' import-from-anywhere rule; both packages already depend on it) and rewrites both consumers on top of it. It executes the consolidation sketched in `forward_expr.py`'s 0.47.0 module docstring (that sketch is now replaced by a pointer to the core).

**This is a refactor, not a behavior change.** Each consumer keeps its own:

- **whitelist** — gateway: comparisons/bool-ops, `isfinite`, `tau`, no `abs`/`//`; engine: arithmetic incl. `//`, `abs`, no comparisons
- **symbol contract** — gateway: per-expression input symbols (bare function names stay compile-legal, a preserved legacy quirk, passed as extra symbols with a comment); engine: fixed `composite_var`/`x`
- **error contract** — `DerivedExpressionError` vs `GeecsConfigurationError` at compile and call time (incl. the engine's runtime-failure wrapping from PR #594)
- **result coercion** — gateway publishes booleans as `1.0`/`0.0` via the core's `is_boolean` flag; engine floats the result

Both consumers' existing test suites pass unchanged — they pin the parity. The core gets its own suite (`GEECS-Schemas/tests/test_restricted_expr.py`, 14 tests) pinning the security contract (refusal of smuggled calls, attribute access, subscripts, unknown names, non-numeric literals, comprehensions, f-strings; empty-builtins evaluation) and the parameterization points (operator gating, bool/compare gating, symbol shadowing, `is_boolean`).

One deliberate observable delta, called out in the gateway CHANGELOG: rejection-message *phrasing* for invalid derived-channel expressions now uses the core's wording (e.g. `unknown name 'foo'` instead of `unknown name in expression: foo`). No test pinned the old strings; the error types are unchanged.

## Per-concern LOC breakdown

| Concern | Files | LOC |
|---|---|---|
| The shared core (new) | `GEECS-Schemas/geecs_schemas/restricted_expr.py` | +188 |
| Core test suite (new) | `GEECS-Schemas/tests/test_restricted_expr.py` | +140 |
| Gateway consumer rewrite | `GeecsCAGateway/geecs_ca_gateway/derived.py` | +34/−111 |
| Engine consumer rewrite | `GeecsBluesky/geecs_bluesky/forward_expr.py` | +23/−53 |
| Version bumps + CHANGELOGs + doc notes | 3× `pyproject.toml`/`CHANGELOG.md`, `GEECS-Schemas/README.md`, `GeecsBluesky/CLAUDE.md` | +68 |

Judgment-call additions (cheap to veto): the README "one non-model module" note, the CLAUDE.md layout-line update, and keeping the gateway's bare-function-name quirk rather than tightening it (tightening would be a behavior change out of scope here).

## Tests

`./scripts/check.sh --all`, everything locally runnable green:

- GEECS-Schemas **236 passed**, 7 deselected (incl. the 14 new core tests)
- GeecsBluesky **617 passed**, 3 deselected
- GeecsCAGateway **182 passed**
- ImageAnalysis 224 passed / ScanAnalysis 171 passed (1 skipped each) / GEECS-Data-Utils 198 passed / GEECS-Console 772 passed / GEECS-LogTriage 53 passed / root-tests 2 passed
- lint (pre-commit, all hooks) green

## Hardware verification

**None owed.** Behavior-preserving refactor of pure expression compilation/evaluation — no wire-protocol, PV-surface, or scan-execution semantics change (`PV_CONTRACT.md` untouched). The parity is pinned by the two consumers' existing suites: the legacy `composite_variables.yaml` corpus tests and the gateway's derived-channel suite (incl. the fake-server end-to-end derived tests), all green.

### DIFF
--- a/GEECS-Schemas/CHANGELOG.md
+++ b/GEECS-Schemas/CHANGELOG.md
@@ -5,6 +5,26 @@ All notable changes to GEECS-Schemas are documented here.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [0.9.0] - 2026-07-21
+
+### Added
+
+- `restricted_expr.py` — the shared AST-whitelist expression core behind
+  both GEECS eval sites (the gateway's derived channels and the engine's
+  pseudo-variable forward formulas): `compile_expression` parses with
+  `ast`, validates every node against a consumer-supplied
+  `ExpressionWhitelist` (functions, constants, operator sets, optional
+  bool/compare ops) **before** anything is evaluated, and returns a
+  frozen `CompiledExpression` whose `evaluate()` runs with empty
+  builtins and returns the raw result (`is_boolean` flags top-level
+  boolean intent for consumers that publish floats).  Stdlib-only
+  (`ast` + `dataclasses`) per this package's import-from-anywhere rule;
+  each consumer keeps its own whitelist, error wrapping, and result
+  coercion.  Extracted from `geecs_ca_gateway.derived.ExpressionEvaluator`
+  and `geecs_bluesky.forward_expr` so a hardening or semantics fix lands
+  once instead of twice; both consumers' behaviors stay pinned by their
+  existing test suites.
+
 ## [0.8.2] - 2026-07-21
 
 ### Changed
--- a/GEECS-Schemas/README.md
+++ b/GEECS-Schemas/README.md
@@ -70,6 +70,11 @@ fields are kept for a possible future re-enable),
 `TriggerVariant` / `TriggerState`, `DefaultActions`, and the four action
 step types.
 
+One non-model module: `geecs_schemas.restricted_expr` — the shared
+AST-whitelist core behind both GEECS expression eval sites (the gateway's
+derived channels and the engine's pseudo-variable forward formulas).
+Stdlib-only; each consumer supplies its whitelist and error wrapping.
+
 ## Converter usage
 
 Converters take a parsed dict or a YAML path (path input needs PyYAML, which
--- a/GEECS-Schemas/geecs_schemas/restricted_expr.py
+++ b/GEECS-Schemas/geecs_schemas/restricted_expr.py
@@ -1,0 +1,188 @@
+"""Shared AST-whitelist core for restricted math expressions.
+
+Two GEECS eval sites compile small operator-authored math expressions —
+the gateway's derived channels (``geecs_ca_gateway.derived``) and the
+engine's pseudo-variable forward formulas (``geecs_bluesky.forward_expr``).
+Both follow the same security-sensitive skeleton: parse with :mod:`ast`,
+validate every node against an explicit whitelist **before** anything is
+evaluated (so a config cannot smuggle attribute access, imports,
+subscripts, or arbitrary names), then evaluate the compiled code with an
+empty ``__builtins__``.  This module is that skeleton, factored out once
+so a hardening or semantics fix lands in one place.
+
+Each consumer supplies its own :class:`ExpressionWhitelist` (functions,
+constants, operator sets) and the set of symbol names its expressions may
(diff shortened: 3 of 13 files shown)


skbarber:

**Verification performed, not just read:** the reviewer resurrected the pre-PR `forward_expr.py` and `derived.py` from the merge-base (`ae3820ee`) and differential-tested old vs new over ~1,100 probes per consumer (chained comparisons, bool literals, NaN/inf, symbol-shadowing of `pi`/`sqrt`, missing inputs, syntax errors, every disallowed node shape) — **0 mismatches** in accept/reject decisions, result values, and exception types. Engine rejection messages old vs new: byte-identical; the gateway's disclosed phrasing delta is the only observable change, as the CHANGELOG claims. Redundancy sweep found no other eval/AST sites on the dev line (the legacy numexpr path in frozen GEECS-PythonAPI is correctly excluded); placement in GEECS-Schemas judged correct (the alternative — the gateway's library part — would couple engine config semantics to the CA server package).

### Findings and dispositions

|---|---|---|---|
| 1 | Refusals never pinned in *nested* logic positions (comparator/bool-op/unary recursion) — exactly where the refactor restructured the recursion; a dropped recursion loop would let `ready and payload.__class__` compile and evaluate with every suite green | Medium-low | **Fixed** in bfa13493: `test_nested_logic_positions_refused` (4 cases against the logic whitelist). Reviewer re-verified by mutation: dropping the comparator loop, the bool-op values loop, or the unary operand validation each makes the corresponding case(s) fail. |
| 2 | `test_evaluation_namespace_has_no_builtins` didn't test what its name claimed (empty-builtins is defense-in-depth, not black-box observable) | Low | **Fixed** in bfa13493: renamed to `test_declared_but_unbound_symbol_is_nameerror` with an honest docstring. Reviewer confirmed no overclaim remains. |
| 3 | Latent namespace-order flip: old engine code bound the scanned value first (whitelist would shadow a colliding token), the core binds values last — unobservable only while `SCAN_VALUE_NAMES` stays disjoint from `_FUNCTIONS`/`_CONSTANTS` | Note | **Fixed** in bfa13493: disjointness invariant documented on `SCAN_VALUE_NAMES` + pinned by `test_scan_value_names_disjoint_from_whitelist_names`. Reviewer confirmed by mutation (adding a colliding constant `x` fails the test). |
| 4 | Preserved-but-unpinned quirk: `True + x` compiles (`bool` is `int`); a future innocent tightening of the constant check would silently change both consumers | Note | **Fixed** in bfa13493: `test_bool_literals_count_as_numbers` pins acceptance. Reviewer confirmed by mutation (tightening to `type(...) in (int, float)` fails the test). |
| 5 | Root `CLAUDE.md` dependency graph omits GEECS-Schemas entirely (says GeecsCAGateway has no intra-repo deps) — now understates a security-load-bearing edge | Note (preexisting) | **Waived**: not introduced by this PR; flagged as a follow-up task to fix the graph on dev. Reviewer had no objection. |

**Reviewer's post-fix verdict:** "all four fixes are real, and each new test demonstrably fails under the specific regression it exists to catch. No remaining blockers — PR #596 at bfa13493 is clear to land from the adversarial-review standpoint."

Suite counts after fixes: GEECS-Schemas 241 passed, GeecsBluesky 618 passed, GeecsCAGateway 182 passed, lint green.

PR #597, opened 2026-07-22
GeecsBluesky 0.48.0: move_variable — manual moves through the movable dispatch (live-verified on ALine composites)

skbarber:
## What & why

**`move_variable(name, value)` — on-demand manual moves** (GeecsBluesky 0.48.0, minor). PR-A of the manual-composite-set design discussed with the owner: the engine seam that lets GEECS-Console (PR-B, next) manually move any catalog scan variable — including the pseudo/composite variables that the PV-oriented device panel cannot express — with **scan-identical completion semantics**.

- `GeecsSession.move_variable` + `BlueskyScanner.move_variable` (the third member of the console Submitter contract, alongside `run_action`/`describe_action`; contract comment updated). *name* = catalog scan-variable name (plain / confirm / pseudo) or raw `"Device:Variable"`.
- Dispatches through the existing `build_movable` seam — so a manual move gets the blocking GEECS set, `CaMotor`'s readback poll, `CaConfirmSettable`'s confirming poll (**manual EMQ sets no longer bypass confirmation** the way raw PV puts do), and a pseudo's concurrent fan-out. No new movable semantics were invented.
- **A fresh movable per call** — deliberate: a relative pseudo re-baselines from where its targets are *at the moment of the move* (each manual bump offsets from current position — legacy composite manual-set semantics, per owner design decision).
- Refusals mirror `run_action`: exact message `"scan in progress — move not started"` (session + bridge), and a non-finite value is refused before any device is built.
- Returns `{"variable", "kind", "value", "targets"}` for operator feedback; `CaPseudoMovable` now publishes `last_commanded` (per-target values of its last completed set) to back it.

## LOC breakdown

- `session.py` `move_variable`: ~115 (mostly docstring)
- `bluesky_scanner.py` delegation: ~28
- `devices/ca/pseudo.py` `last_commanded`: ~8
- `tests/test_move_variable.py`: ~215 (10 tests)
- CLAUDE.md / CHANGELOG / pyproject: ~50

## Tests

`./scripts/check.sh` (scoped, as CI): lint green, GeecsBluesky **626 passed, 3 deselected** (up from 616 on dev). New pins: dispatch per kind (plain / confirm / pseudo), fresh-movable-per-call, exact refusal messages (session + bridge), non-finite guard, bridge delegation with the stored resolver.

## Hardware verification — LIVE, PASSED (2026-07-22, over VPN, owner-authorized)

Ran `move_variable` headless against the live Undulator gateway (108 devices connected) on the production ALine composites:

| Step | Commanded | Expected | Measured | Result |
|---|---|---|---|---|
| initial | — | — | S3H +0.0001 A, S4H −0.0000 A | baseline |
| `ALine_e_beam_position_offset_x` +0.05 | S3H +0.0501, S4H −0.0500 | +0.05 / −0.05 from baseline | +0.0500 / −0.0498 | **OK** |
| same, −0.05 (fresh baseline = restore) | S3H +0.00001, S4H +0.00025 | back to baseline | +0.0001 / +0.0002 | **OK** |
| `ALine_e_beam_angle_offset_x` +0.05 | S3H +0.0501, S4H −0.0998 | +0.05 / −0.10 (1:−2 ratio) | +0.0500 / −0.0995 | **OK** |
| same, −0.05 (restore) | — | back to baseline | +0.0001 / +0.0004 | **OK** |

This live-verifies: relative fan-out math on real magnets, both corpus ratio shapes (1:−1 and 1:−2), the **fresh-baseline-per-call semantics** (the −0.05 restore was computed against the *bumped* currents and landed back on the originals, Δ ≤ 0.0005 A), the returned `targets` feedback, and completion via the GEECS blocking set over the gateway.

**Still OWED** (unchanged from PR #594): the *scan-path* pseudo verification (event rows + `pseudo_variables` in a real scan's start doc — needs trigger/beam state an operator must confirm), and the **R56 absolute-mode live check** — attempted, but the guard skipped it: chicane inner reads **1.0 A** (a real operating point, not zero), so an absolute move couldn't be restored to an unknown-x state. Needs an operator to confirm chicane state first. Absolute fan-out math is pinned hermetically.

### DIFF
--- a/GeecsBluesky/CHANGELOG.md
+++ b/GeecsBluesky/CHANGELOG.md
@@ -25,9 +25,23 @@ Format follows [Keep a Changelog](https://keepachangelog.com/en/1.1.0/).
   `{"variable", "kind", "value", "targets"}` for operator feedback
   (`CaPseudoMovable` now publishes `last_commanded` — the per-target
   values of its last completed set — to back it).
-- Suite grows to 626 (manual-move contract: dispatch per kind, fresh
-  movable per call, exact refusal messages, non-finite guard, bridge
-  delegation).
+- **Mutual exclusion is symmetric** (review finding, PR #597): a manual
+  move is not a plan, so the RunEngine stays idle while it runs — a
+  session-level lock now makes the exclusion mutual.  `scan()` /
+  `optimize()` / `run_action()` refuse while a move holds it
+  (`"manual move in progress — <scan|optimization|action> not started"`),
+  and a second concurrent move is refused (the double-click race that
+  would have corrupted a relative pseudo's baselines).
+- **Timeout cancels the move** (review finding): on the wall-clock budget
+  expiring, the in-flight coroutine is cancelled and a documented
+  `TimeoutError` raised (a GEECS set already on the wire cannot be
+  recalled — stated in the contract); the bridge exposes `timeout` for
+  legitimately slow moves.  A non-numeric value is refused as
+  `GeecsConfigurationError` before any device is built.
+- Suite grows to 631 (manual-move contract: dispatch per kind incl.
+  motor, fresh movable per call, exact refusal messages both directions,
+  timeout-cancellation, non-finite/non-numeric guards, bridge
+  delegation with timeout passthrough).
 
 ## [0.47.1] - 2026-07-21
 
--- a/GeecsBluesky/CHANGELOG.md
+++ b/GeecsBluesky/CHANGELOG.md
@@ -4,6 +4,31 @@ All notable changes to `geecs-bluesky` are documented here.
 
 Format follows [Keep a Changelog](https://keepachangelog.com/en/1.1.0/).
 
+## [0.48.0] - 2026-07-22
+
+### Added
+
+- **`move_variable(name, value)` — on-demand manual moves** (session +
+  bridge; the third member of the console Submitter contract alongside
+  `run_action`/`describe_action`).  *name* is a catalog scan-variable name
+  (plain, confirm, or pseudo/composite) or a raw `"Device:Variable"`
+  string; the move dispatches through the one `build_movable` seam, so a
+  manual move carries **scan-identical completion semantics**: blocking
+  GEECS set, `CaMotor` readback poll, `CaConfirmSettable` confirming-poll
+  (manual EMQ sets no longer bypass confirmation the way raw PV puts do),
+  and pseudo concurrent fan-out.  A fresh movable is built per call —
+  deliberately, so a relative pseudo re-baselines from the targets'
+  *current* positions on every manual bump (legacy composite manual-set
+  semantics).  Refused while a scan is active with the exact message
+  `"scan in progress — move not started"`; a non-finite value is refused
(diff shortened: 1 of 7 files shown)


skbarber:

Reviewer read the diff plus every seam it touches (run_action, `_move_movables`, movable dispatch, optimize path, bridge guards, console Submitter + device panel). Placement clean (session-method-plus-thin-bridge exactly mirrors the G-actions pattern; `last_commanded` correctly on the device); redundancy checked and nothing disqualifying (the ~6-line threadsafe-move pattern and raw-name split are deliberate second/third copies with different error semantics — waived, noted).

### Findings & disposition (fixes in 557b1b05, re-verified by the same reviewer)

**1 (medium-high): mutual exclusion was one-directional** — a manual move is not a plan, so the RE stayed idle mid-move; a scan, an action, or a double-clicked second move could start mid-move (the second-move case would corrupt a relative pseudo's baselines).
**Fixed**: session-level `_manual_move_lock`; `move_variable` acquires non-blocking, and `scan()` (covers `noscan`), `optimize()`, and `run_action()` refuse while held with exact per-verb messages. **Reviewer re-verified every RE entry point is covered** (including `run(ScanRequest)` via the runner), that a refusal fires **before** `claim_scan_number` (no scan number burned, no folder created), and that both new tests would fail if the guards were removed.

**2 (medium): timeout abandoned a live move uncancelled, with undocumented exceptions.**
**Fixed**: future cancelled on timeout; documented builtin `TimeoutError` raised with the wire-level caveat (an already-sent GEECS set cannot be recalled); `GeecsError` propagation documented; bridge exposes `timeout` for slow moves. **Reviewer confirmed** the 3.11 `TimeoutError` aliasing makes the except/raise types exactly right, and the cancellation is observed by test.

**3 (low): non-numeric value escaped the error contract.** **Fixed** — wrapped as `GeecsConfigurationError`, pinned; reviewer confirmed the `"nan"` string still lands in the isfinite guard (no gap).

**4 (low): contract comment ahead of reality.** **Fixed** — the bridge comment now says move_variable "will join" the console Submitter protocol with PR-B. Reviewer judged the CHANGELOG wording fine as-is (bridge-side trio exists).

**5 (low): coverage gaps.** **Fixed** (motor-kind dispatch test — a real pin) / **waived** (the "scan in progress" wording when the RE is busy with an *action* matches run_action's own precedent; cosmetic, safe direction).

Residual notes recorded by the reviewer as fine-as-designed (not defects): the exclusion check is check-then-act, closed on the console path by the bridge's `is_scanning_active()`; same race class the pre-existing RE-idle checks accept.

Reviewer's verdict on 557b1b05: "No new issues introduced by the fixes... clean to land once CI re-runs green."

skbarber:
## Hardware verification close-out (2026-07-22)

- Relative manual moves (position + angle ALine composites): **live-verified in this PR** (table in the PR body).
- **Absolute-mode (R56) — ✅ VERIFIED by the owner** (2026-07-22), closing the item the zero-state guard had deferred.

No remaining OWED items on the manual-move seam.

PR #598, opened 2026-07-22
GEECS-Console 0.19.0: catalog-aware movable panel — composites settable, scan-variable auto-select

skbarber:
## What & why

**PR-B of the manual-composite design** (PR-A = the engine's `move_variable`, #597, merged + live-verified): the console's R7 device panel becomes a **catalog-aware movable panel**, closing the two operator gaps discussed with the owner — composites cannot be set manually from the console, and the legacy scanner's "scan variable auto-selects in the device panel" behavior was missing.

### The pieces (GEECS-Console 0.19.0, minor)

- **`app/movable_panel.py::MovablePanelController` (new, ~430 LOC)** — owns the whole R7 surface the window previously ran inline (selection commits, readback rendering, completions fetch, set dispatch), following the established #534 controller shape (no Qt parent, injected widgets + callables, own `BackgroundResult` worker per issue #510, queued `value_ready(int, object)` signal, idempotent `dispose()` from `closeEvent`). **`main_window.py` sheds ~170 lines** — this is the "rework the panel" forcing function the #534 accretion watch wanted.
- **Catalog-aware selection**: resolves against `ConsoleConfigs.scan_variable_specs()` (new offline-safe accessor). A plain `ScanVariable` monitors its target; a **pseudo/composite monitors every component** (`subscribe_many` — new `DevicePanelBackend` protocol member, implemented in `GatewayDevicePanel` with the same generation-counter straggler-drop, duck-typed fallback for older single-monitor backends) rendered compactly ("U_S3H 0.05 · U_S4H −0.0998"). Raw `device:variable` strings keep the exact historical path.
- **Sets fork by selection kind**: catalog names dispatch **`Submitter.move_variable(name, value)`** — the new protocol member mapping straight onto `BlueskyScanner.move_variable` (GeecsBluesky ≥ 0.48.0), so manual moves get scan-identical completion semantics (motor poll, confirm poll, pseudo fan-out with a fresh relative baseline per move) and the engine's refusals ("scan in progress — move not started") surface verbatim. Raw strings keep the direct gateway `:SP` put — no engine required, offline-first preserved.
- **Auto-select (the legacy behavior)**: both R3 axis combos' `currentTextChanged` now re-point the panel — composites included. Programmatic populate churn is signal-blocked, and unresolvable names never hijack the panel (offline-safety).
- The combo dropdown lists **catalog names first** (composites become discoverable), then the DB `device:variable` completions.

### Judgment calls (cheap to veto)

- The old single-target `subscribe` stays on the protocol (thin wrapper over `subscribe_many` in the gateway impl) — keeps older injected backends and the seam's simplest consumers working.
- Confirm-variable catalog entries monitor the *written* target's readback (same as what scans record); monitoring the confirming variable alongside is a possible follow-up.
- GeecsBluesky's bridge comment ("move_variable will join [the Submitter protocol] with PR-B") is now satisfied but not edited here — one-line follow-up rather than a cross-package touch in a console PR.

## Per-concern LOC breakdown

| Concern | Files | LOC |
|---|---|---|
| Controller (extraction + catalog awareness) | app/movable_panel.py | +430 |
| Window sheds R7 internals + wiring | app/main_window.py | +40/−210 |
| Backend multi-subscribe + formatter | services/device_panel.py | +120/−40 |
| Catalog accessor + protocol member | services/configs.py, submission.py | +35 |
| New capability tests | tests/test_movable_panel.py | +230 |
| Existing-test updates (surface rename, subscribe_many fake) | 3 test files | ~+60/−25 |
| Docs/versioning | CLAUDE.md, CHANGELOG, pyproject | ~+80/−30 |

## Tests

`./scripts/check.sh` (scoped, as CI): lint green, GEECS-Console **784 passed** (772 on dev → +12: movable-panel capability tests — catalog/pseudo selection, per-target rendering, move_variable dispatch + verbatim refusal, raw-path preservation, auto-select incl. the never-hijack guard — plus `subscribe_many` service tests). The pre-existing R7 suites (TestDevicePanel, parse feedback, completions, the #510 deletion test) pass against the controller surface, pinning the extraction as behavior-preserving.

## Hardware verification

**OWED** (GUI-level; the engine seam beneath is already live-verified on PR #597 against U_S3H/U_S4H): launch the console on the lab network, pick `ALine_e_beam_position_offset_x` in the R3 axis combo → panel auto-selects and shows both magnet readbacks live; Set 0.05 → status bar reports "Moved … (U_S3H:Current=…, U_S4H:Current=…)" and the readbacks move by +0.05/−0.05; Set −0.05 restores; a Set clicked during a scan shows "scan in progress — move not started".

### DIFF
--- a/GEECS-Console/CHANGELOG.md
+++ b/GEECS-Console/CHANGELOG.md
@@ -4,6 +4,37 @@ All notable changes to GEECS-Console are documented here.  Format follows
 [Keep a Changelog](https://keepachangelog.com/en/1.1.0/); versioning is
 semantic.
 
+## [0.19.0] - 2026-07-22
+
+### Added
+
+- **The R7 device panel is now the catalog-aware movable panel**, owned by
+  `app/movable_panel.py::MovablePanelController` (issue #534 controller
+  shape; the window sheds ~170 lines of inline R7 machinery).  A selection
+  resolves catalog-first against the experiment's scan-variable catalog:
+  - a plain `ScanVariable` monitors its target readback; a composite
+    `PseudoScanVariable` monitors **every** component (`subscribe_many` on
+    the `DevicePanelBackend` — one live readback per target, rendered
+    compactly by the new `format_target_readbacks`); a raw
+    `device:variable` string keeps the historical direct path;
+  - catalog sets dispatch **`Submitter.move_variable`** (new protocol
+    member, mapping to `BlueskyScanner.move_variable`, GeecsBluesky ≥
+    0.48.0) — manual moves carry scan-identical completion semantics
+    (motor poll, confirm poll, pseudo fan-out with a fresh relative
+    baseline per move); engine refusals surface verbatim; raw strings
+    keep the direct gateway `:SP` put (no engine required);
+  - the combo's dropdown lists catalog names (composites included) ahead
+    of the DB `device:variable` completions;
+  - **the R3 axis combos auto-select the panel** (the legacy scanner
+    behavior), composites included — populate churn is signal-blocked and
+    unresolvable names never hijack the panel.
+- `ConsoleConfigs.scan_variable_specs()` — the catalog as
+  `{name: ScanVariableSpec}` (offline-safe), backing both the panel and
+  the names listing.
+- Suite grows 772 → 784 (the movable-panel capability tests +
+  `subscribe_many` service tests; the pre-existing R7 suites pin the
+  extraction as behavior-preserving).
+
 ## [0.18.3] - 2026-07-20
 
 ### Changed
--- a/GEECS-Console/CLAUDE.md
+++ b/GEECS-Console/CLAUDE.md
@@ -57,14 +57,17 @@ are prefixed by region (`r3_radio_1d`, `r5_start_button`, …).
   timing and granularity; they will drift, and shot counts must never be
   reconciled across them (live-investigated 2026-07-16 — a "pause shows
   extra steps" report was mostly this drift).
-- **R7 device panel** — device:variable combo (editable, with
-  `device:variable` dropdown completions from `GeecsDbCompletions` — see
-  Implemented seams), readback label, set field + button.  **Backend
-  live** (see Implemented
-  seams): gateway PVs — CA monitor on the readback, put to `:SP` riding
-  GEECS's native blocking set — never `geecs_python_api`'s ScanDevice.
-  Scalar-only at birth (composites arrive with the pseudo-variable
-  runtime, if ever).
+- **R7 movable panel** — an editable combo (catalog scan-variable names
+  first, then `device:variable` completions from `GeecsDbCompletions`),
+  readback label, set field + button — owned by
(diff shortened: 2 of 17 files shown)


skbarber:

The reviewer read the diff plus every console threading/GC rule it touches. Verified clean before findings: the GeecsBluesky ≥ 0.48.0 dependency is real and the contracts match; closure binding in `subscribe_many` is correct; the #510/no-QThread/queued-connection rules and offline-first are honored; no duplication (the fetch/set machinery is a relocation, not a copy); no scan-folder or hygiene issues.

### Findings & disposition (fixes in fd2ba620, re-verified by the same reviewer)

**1 (HIGH): R3 auto-select was wired to `currentTextChanged` on editable combos — typing an axis name would hijack the panel and open a garbage CA monitor per keystroke** (from `U_Hexapod:y` onward, every keystroke parses), violating the module's own "never churn CA monitors while the operator types" rule.
**Fixed**: commit-only `textActivated` wiring + explicit preset-path follow (axis 1 owns the panel — also resolving the reviewer's preset-order note). **Reviewer mutation-verified the pin**: reverting the wiring makes `test_typing_in_the_axis_combo_never_hijacks_the_panel` fail. Noted asymmetry (typed-then-focus-leave doesn't auto-select — under-follow, the right side to err on).

**2 (moderate): `dispose()` missed the completions-provider closure back to the window** — the one surviving controller→window edge, the deferred-C++-teardown hazard class. **Fixed** with a capture-free inert lambda; **reviewer confirmed** the last edge is severed and the post-dispose call surface is safe.

**3 (low-moderate): a raising move job would wedge the Set button forever** (BackgroundResult swallows without emitting; result formatting sat outside the try). **Fixed** — whole job body guarded; pinned by the malformed-result test. **Reviewer confirmed** no masking.

**4 (low): the `_scan_variables_catalog` private reach-in deepened just as its debt-note trigger ("when next touching geecs-bluesky") fired.** **Fixed** — `ConfigsRepoResolver.scan_variable_catalog()` promoted public (GeecsBluesky 0.49.0, minor + CHANGELOG + pin test); console consumes it (getattr fallback kept — reviewer calls it harmless dead code in the monorepo, take-or-leave); debt note retired. Reviewer explicitly endorsed the bridge contract-comment update riding along ("leaving it stale would itself violate the contract-travels rule").

**5 (minor): waived** — colon-bearing catalog names stay unenforced (schema-side change, precedence documented); broken-catalog per-keystroke reload accepted (log spam only). **Reviewer: waivers sound.**

Reviewer's verdict on fd2ba620: "Nothing introduced by the fixes. All findings closed."

skbarber:
## Hardware verification close-out (owner-run, 2026-07-22)

The GUI-level OWED check is **✅ VERIFIED**: selecting a 1D scan variable in R3 auto-populates the movable panel (composites included), with per-target live readbacks.

Owner feedback from live use, spun into a follow-up UI PR: (a) noisy final digits in the readback make the window width jitter — fixed-width rendering + the label no longer driving layout; (b) R7 moves below the scan/presets panels in the middle third, giving the R6 now-panel more room.

PR #599, opened 2026-07-22
GEECS-Console 0.19.1: movable-panel polish — width-stable readbacks, R7 to the middle column

skbarber:
## What & why

Owner feedback from first live use of the 0.19.0 movable panel (both items requested explicitly):

1. **Width-stable readbacks** — the old `.6g` float format let noise in a streamed value's trailing digits change the string length, so the window width jittered at the stream rate. `format_readback` now renders floats with fixed decimals (4) in the ordinary magnitude range and fixed-mantissa scientific outside it, **and** the readback label no longer drives layout at all (`Ignored` horizontal size policy + 160 px minimum in the `.ui`) — the second half is the real cure; the first makes values read calmly.
2. **R7 moves to the middle column**, below the scan form and presets, where there's room; the right column becomes the R6 now panel's alone, giving the log tail the freed space. Widget object names (`r7_*`) are unchanged, so the controller, tests, and tooltips are untouched by the move — it's purely a `.ui` block relocation (plus the label sizing above).

## Tests

`./scripts/check.sh` (as CI): lint green, GEECS-Console **789 passed** (788 → net +5 vs 0.19.0: fixed-decimal + scientific-range format pins including the exact jitter pair `0.05`/`0.0498` asserting equal width, and a structural pin that `r7_group` sits between the middle column's spacer and `right_column` in the `.ui`). Existing readback tests updated to the new fixed-decimal expectations.

## Hardware verification

GUI-only change; nothing moves hardware. Visual confirmation on next console launch: no width jitter with a live noisy readback, R7 under presets, roomier now panel.

### DIFF
--- a/GEECS-Console/CHANGELOG.md
+++ b/GEECS-Console/CHANGELOG.md
@@ -4,6 +4,22 @@ All notable changes to GEECS-Console are documented here.  Format follows
 [Keep a Changelog](https://keepachangelog.com/en/1.1.0/); versioning is
 semantic.
 
+## [0.19.1] - 2026-07-22
+
+### Changed
+
+- **Movable-panel polish from first live use** (owner feedback on 0.19.0):
+  - **Width-stable readbacks** — `format_readback` renders floats with
+    fixed decimals (4) in the ordinary magnitude range (scientific with a
+    fixed mantissa outside it), and the readback label no longer drives
+    layout (`Ignored` horizontal size policy + minimum width): noise in a
+    streamed value's trailing digits made the window width jitter at the
+    stream rate.
+  - **R7 relocated to the middle column**, below the scan form and presets
+    (space it fits naturally); the right column keeps only the R6 now
+    panel, which gets the freed room.  Widget object names (`r7_*`)
+    unchanged — controller, tests, and tooltips untouched by the move.
+
 ## [0.19.0] - 2026-07-22
 
 ### Added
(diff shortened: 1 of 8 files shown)


skbarber:

Verified clean: `format_readback` across its full input domain (nan/±inf/−0.0/negatives/boundaries/ints-as-floats — the jitter pair `0.05`/`0.0498` genuinely renders equal-width), the `.ui` relocation (well-formed XML, all nine `r7_*` names intact, QUiLoader loads, nothing addresses R7 by column), test discrimination (the new pins fail under the old code), no redundancy (the browser's `.6g` formatter serves a different purpose), placement consistent.

### Findings & disposition (fixes in da8a02ec, re-verified by the same reviewer)

**1 (HIGH, confirmed CI-red): the new layout-pin test read the `.ui` with the platform codec** — Windows' cp1252 default dies on the ❚ pause glyph's 0x9D byte; both console-windows jobs on the PR failed with exactly that `UnicodeDecodeError`.
**Fixed**: `read_text(encoding="utf-8")`. **Reviewer confirmed against the new CI runs — both console-windows jobs pass.**

**2 (LOW): the same test resolved the `.ui` cwd-relative** (the only cwd-dependent test in the suite). **Fixed**: anchored off `geecs_console.app.__file__`, matching the production `_UI_PATH` convention. **Reviewer confirmed.**

**3 (cosmetic, waived with reviewer agreement)**: width can still wobble at the 1e-3/1e5 magnitude boundaries — harmless, because the `Ignored` size policy (not the string format) is what owns layout stability; and the fixed-width design trades possible clipping at minimum window width for zero jitter — a deliberate, now-documented choice.

Reviewer's verdict on da8a02ec: "Nothing outstanding from my review — PR #599 is clear."

skbarber:
## Visual confirmation close-out (owner, 2026-07-22)

The PR body's "visual confirmation on next console launch" item is **✅ confirmed** — no width jitter with live readbacks, and the R7-below-presets layout looks good. Nothing outstanding on this PR.

PR #600, opened 2026-07-22
GeecsBluesky 0.50.0: relative pseudo scan axes restore captured baselines at end of scan

skbarber:
## What & why

**Relative pseudo (composite) scan axes now return to their captured baselines at end of scan** (GeecsBluesky 0.50.0, minor) — owner request from the first live composite scans ("doesn't return to initial setting from start of scan, which would be nice"), and the one piece kept from the broader mark/return design discussion (explicitly tabled by the owner to avoid console feature creep; manual-move semantics are untouched — "Set 0" still means "no move from here", fresh baseline per manual move).

- `CaPseudoMovable.restore_baselines_plan()` — puts every component back to its captured baseline with **direct per-target puts**: exact and formula-independent (no `f(0) = 0` assumption), riding the same `GatewaySetpointPut` blocking-set completion as every other move.
- `build_step_scan_plan` runs it as a **finalize** — success *and* mid-scan abort — placed after closeout (operator rituals see the scan's final state first) and inside the stage wrapper (before `unstage()` drops the baselines; the nesting makes this ordering structural, not incidental).
- **Deliberate scope**: absolute-mode pseudos are *not* restored (their pre-scan x is unknowable without an inverse — restoring blindly is the NaN-command class of bug this feature must not introduce) and plain axes keep the legacy end-at-last-position behavior. Optimize mode already has its own `on_finish` restore semantics and is untouched.

## LOC breakdown

- `devices/ca/pseudo.py` (`restore_baselines_plan` + `_restore_baselines`): ~36
- `plans/orchestration.py` (finalize + `_restore_movables`): ~16
- `tests/test_pseudo_scan_restore.py`: ~175 (5 tests)
- CHANGELOG / CLAUDE.md / pyproject: ~30

## Tests

`./scripts/check.sh` (as CI): lint green, GeecsBluesky **637 passed, 3 deselected** (632 on dev → +5): end-of-scan restore through the full free-run composition, abort-path restore through the composition *and* directly at the finalize level, absolute-mode-not-restored, and restore-without-baselines no-op.

## Hardware verification

**OWED** (one check, rides the next lab session): run a relative composite scan (e.g. `ALine_e_beam_position_offset_x`) and confirm S3H/S4H return to their pre-scan currents at scan end — including once with a mid-scan Stop. Expected: readbacks equal the pre-scan values within readback jitter (the manual-move restore verified ≤0.0005 A on the same magnets).

### DIFF
--- a/GeecsBluesky/CHANGELOG.md
+++ b/GeecsBluesky/CHANGELOG.md
@@ -4,6 +4,26 @@ All notable changes to `geecs-bluesky` are documented here.
 
 Format follows [Keep a Changelog](https://keepachangelog.com/en/1.1.0/).
 
+## [0.50.0] - 2026-07-22
+
+### Added
+
+- **Relative pseudo (composite) scan axes return to their captured
+  baselines at end of scan** (owner request from the first live composite
+  scans): `CaPseudoMovable.restore_baselines_plan()` puts every component
+  back to its captured baseline — direct per-target puts, exact and
+  formula-independent (no `f(0) = 0` assumption) — and
+  `build_step_scan_plan` runs it as a finalize on **every** exit path
+  (success and mid-scan abort), after closeout, inside the stage wrapper
+  (before `unstage()` drops the baselines).  Deliberately scoped:
+  absolute-mode pseudos are NOT restored (their pre-scan x is unknowable
+  without an inverse) and plain axes keep the legacy end-at-last-position
+  behavior.  Manual `move_variable` semantics are unchanged (fresh
+  baseline per move; "Set 0" still means "no move").
+- Suite grows to 638 (`tests/test_pseudo_scan_restore.py`: end-of-scan
+  restore, abort-path restore through the full composition and at the
+  finalize level, absolute-not-restored, restore-without-baselines noop).
+
 ## [0.49.0] - 2026-07-22
 
 ### Added
--- a/GeecsBluesky/CHANGELOG.md
+++ b/GeecsBluesky/CHANGELOG.md
@@ -13,16 +13,24 @@ Format follows [Keep a Changelog](https://keepachangelog.com/en/1.1.0/).
   scans): `CaPseudoMovable.restore_baselines_plan()` puts every component
   back to its captured baseline — direct per-target puts, exact and
   formula-independent (no `f(0) = 0` assumption) — and
-  `build_step_scan_plan` runs it as a finalize on **every** exit path
-  (success and mid-scan abort), after closeout, inside the stage wrapper
+  `build_step_scan_plan` runs it as a finalize on success and mid-scan
+  abort (a `halt` skips finalizes by bluesky contract, same as
+  disarm/closeout), after closeout, inside the stage wrapper
   (before `unstage()` drops the baselines).  Deliberately scoped:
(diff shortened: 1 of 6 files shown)


skbarber:

The reviewer attacked every axis in the brief and refuted the hard ones with evidence: abort/pause-then-abort ordering (all exception paths take the same finalize `finally`), generator-instance finalize semantics (matches the existing `controller.disarm()` precedent; pause/resume replays never re-iterate a finalize), **no optimize double-restore** (`session.optimize` composes via `geecs_adaptive_scan`, never `build_step_scan_plan` — its `on_finish` restore is disjoint), NOSCAN/motorless no-op, the abort-mid-move FIFO race (same assumption every finalize write makes), and redundancy vs bluesky's `reset_positions_wrapper` (not equivalent — it would be formula-dependent via `set(0)` and would wrongly restore plain axes; the direct-put design is justified).

### Findings & disposition (fixes in b4741c84 + d999fb23, re-verified by the same reviewer)

**1 (medium-high): a failed restore was silently swallowed** — `bps.wait_for` never retrieves task exceptions (verified against the installed bluesky source), so a dead supply/gateway during the finalize left the magnets displaced while the scan ended looking clean, with an INFO line printed *before* the puts affirmatively claiming a restore that didn't happen. The one hardware-write path in the package that couldn't fail visibly.
**Fixed along the reviewer's suggested shape**: the restore now issues `bps.abs_set` per target on the `GatewaySetpointPut` movables + one `bps.wait` — the RE's own set machinery — the hand-rolled coroutine is deleted, and the success log/readback update run only after the waits. **Reviewer confirmed by execution** (`FailedStatus: ... RuntimeError('supply unreachable')` out of `RE(plan)`) and checked the fix introduces no new hazards (late binding of baselines, preserved concurrency, abort-time `stop()` tolerance, unique group name).

**2 (low): "every exit path" overstated by one** — `halt` skips finalizes by bluesky contract. **Fixed** in docstring + CHANGELOG (same caveat as disarm/closeout).

**3 (low, coverage): mixed grids untested.** **Addressed** — a real two-axis grid test (plain `CaMotor` + relative pseudo) pins both halves: pseudo restored, plain axis stays at its last position. Strict-mode-specific restore test waived per the reviewer's own refutation (the finalize is wired after the mode branch on the shared plan).

**Re-check flag (minor, test hygiene): the failure pin accidentally exercised "move failed AND restore failed"** (the dead put lacked the scan path's `put()`). **Fixed** in d999fb23 — the stub's scan-path `put()` succeeds, so the test proves the advertised property: a restore failure *alone* fails an otherwise-clean scan.

Reviewer's verdict: all dispositions confirmed; "the feature's structure, scope, and placement are sound."

skbarber:
## Hardware verification close-out (owner-run, 2026-07-22)

The OWED live check is **✅ VERIFIED**: a relative composite scan returns its targets (S3H/S4H) to their pre-scan currents at scan end, including with a mid-scan Stop.

**No remaining OWED items on this PR.** The pseudo-variable arc (#594 → #600) is fully hardware-verified except one explicitly **deferred-by-owner** optional item: an optimize run with a pseudo axis (will ride a future optimization session — tracked on #594's close-out).

PR #587, opened 2026-07-21
lab_status.sh: tier-2 probe speaks both PV naming contracts

skbarber:
## What

Follow-up to #585 closing a gap both that PR's sweep and its review missed: `scripts/lab_status.sh --hardware` hardcoded the legacy mixed-case gateway status PVs (`{Experiment}:CAGateway:HEARTBEAT` …), so the moment a geecs-ca-gateway ≥ 0.14.0 deploys, the probe would report a healthy gateway as DOWN — exactly the tool you'd trust while smoke-testing that deploy.

The tier-2 probe now tries the lowercase contract first and falls back to legacy, labeling which one answered — doubling as a deployed-generation indicator during the transition. The dual-contract probe is transition tooling; once the 0.14.0 deploy is verified, the legacy branch can be deleted at leisure.

One script, +32/−10; no package code changed, so no version bumps.

## Tests / verification

- `bash -n` clean; tier-1 behavior untouched.
- **Live-verified over VPN against the running 0.10.0 gateway**: lowercase attempt times out, probe falls back and reports `version=0.10.0 [legacy mixed-case (<0.14.0) naming]`.

## Hardware verification

OWED (bundled with the #585 deploy session): after the gateway restarts on 0.14.0+, the same command should report `[lowercase (>=0.14.0) naming]` with no fallback delay.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #601, opened 2026-07-22
tooling: lab_status.sh --hardware — drop the legacy mixed-case heartbeat fallback

skbarber:
## What + why

Removes the legacy mixed-case gateway-heartbeat fallback (and its
which-contract-answered labeling) from the `--hardware` tier of
`scripts/lab_status.sh`, leaving the lowercase naming contract
(`pv_name()`, PV_CONTRACT.md §1) as the only probe path.

The bilingual probe was added in PR #587 explicitly as a
deploy-generation indicator "until the deploy is verified". That
verification is complete: the PR #585 hardware close-out (2026-07-22)
confirmed the deployed gateway answers on the lowercase contract after a
week of live operation. The fallback is now dead code, and its removal
was already flagged as owed follow-up.

Single concern, one file: `scripts/lab_status.sh` (+15/−38). No package
code changed, so no version bump / CHANGELOG (scripts/ is repo tooling,
same as #480/#492).

## Tests

`./scripts/check.sh --all`: lint green (from the root env; a PATH-level
`pre-commit` without `jupyter` misreports the notebook hook — env quirk,
not this diff) and all locally runnable suites pass — root-tests 2,
ImageAnalysis 224, ScanAnalysis 171, GEECS-Data-Utils 198, GEECS-Schemas
244, GeecsBluesky 639, GeecsCAGateway 182, GEECS-Console 789,
GEECS-LogTriage 53 (plus 1–2 skips where noted by the suites).

`scripts/` has no CI-mirrored test suite; the change is covered by
`bash -n` plus `py_compile` of the extracted heredoc, both clean.

## Hardware verification

Not owed for this PR: it removes a fallback whose obsolescence was
itself established by hardware verification (PR #585 close-out,
2026-07-22 — deployed gateway serves the lowercase contract). The
surviving lowercase path is byte-identical to the one exercised live
during that verification week. If the gateway were ever rolled back to
<0.14.0, this script would now report DOWN instead of falling back —
that is the intended post-transition behavior.

### DIFF
--- a/scripts/lab_status.sh
+++ b/scripts/lab_status.sh
@@ -159,11 +159,10 @@ async def main():
     # PV_CONTRACT.md §1) — never hand-assemble them here.
     from geecs_ca_gateway.pv_naming import pv_name
 
-    prefix = pv_name(experiment, "CAGateway")
     try:
-        heartbeat = await read(f"{prefix}:heartbeat")
-        connected = await read(f"{prefix}:devices_connected")
-        version = await read(f"{prefix}:version")
+        heartbeat = await read(pv_name(experiment, "CAGateway", "heartbeat"))
+        connected = await read(pv_name(experiment, "CAGateway", "devices_connected"))
+        version = await read(pv_name(experiment, "CAGateway", "version"))
     except Exception as exc:  # noqa: BLE001 — a failed probe is a finding
         print(f"  [DOWN] gateway PVs unreadable ({type(exc).__name__}: {exc})")
         print("         network may be up while the gateway service is not")
--- a/scripts/lab_status.sh
+++ b/scripts/lab_status.sh
@@ -154,47 +154,24 @@ async def read(pv):
     return await asyncio.wait_for(aioca.caget(pv), timeout=timeout)
 
 
-async def probe(prefix, names):
-    heartbeat = await read(f"{prefix}:{names['heartbeat']}")
-    connected = await read(f"{prefix}:{names['connected']}")
-    version = await read(f"{prefix}:{names['version']}")
-    return heartbeat, connected, version
-
-
 async def main():
-    # geecs-ca-gateway >= 0.14.0 serves lowercase PV components (PV_CONTRACT.md
-    # §1); older deployments serve mixed case. Probe new-contract names first,
-    # fall back to the legacy ones — which contract answers doubles as a
-    # deployed-generation indicator during the transition. New-contract names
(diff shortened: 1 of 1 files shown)


skbarber:

**Scope reviewed:** the full diff, the resulting script state, `GeecsCAGateway/geecs_ca_gateway/pv_naming.py`, the gateway status-PV producer (`gateway.py::_build_gateway_status_pvs`), `GEECS-Console/geecs_console/services/health.py`, `PV_CONTRACT.md` §1, the `/lab-status` skill, and the script's git history (`6a719b33` → `14b01f35` → this PR).

### Finding 1 (low) — comment forbids hand-assembly while the code still hand-assembled the leaves

The new comment says "never hand-assemble them here", but `f"{prefix}:heartbeat"` applies the reserved `:` separator and literal leaves outside `pv_name`. The prefix-only split (from 14b01f35) was justified only by the legacy fallback this PR removes; the producer side uses the three-arg form (`gateway.py:331`). No failure today — the leaves are fixed points of `normalize_component`, so the strings are byte-identical — a consistency/contract-hygiene finding, not a bug.

**Disposition: fixed** in 5d4ad388 (three-arg `pv_name(experiment, "CAGateway", "heartbeat"/"devices_connected"/"version")`). Same reviewer confirmed: resolves the finding, no behavior change, no new issues.

### Attempted findings that did not survive

- **Rollback behavior change** (a <0.14.0 gateway now reports DOWN instead of falling back) — the PR's stated intent; transition closed by the #585 hardware close-out.
- **All-or-nothing probe vs the console's tolerance** (`health.py:166` keeps the gateway OK when only `DEVICES_CONNECTED` fails; lab_status marks DOWN if any read fails) — divergence predates this PR; diff preserves the surviving path's semantics exactly. Out of scope.
- **Redundancy** — net deletion; the only overlapping probe (console health) predates the PR and sits on the correct side of the GUI/tooling boundary.
- **Stale cross-references** — no doc/skill/contract references the bilingual probe; remaining uppercase `HEARTBEAT`/`DEVICES_CONNECTED` literals elsewhere all pass through `pv_name`/`ca_pv` normalization.
- **Placement/process** — `scripts/` is the canonical home; no-bump matches #480/#492 tooling precedent; no hygiene violations.

**Verdict:** one low finding, fixed and reviewer-confirmed; otherwise the diff does exactly what it claims and nothing else. No undispositioned findings remain.

PR #602, opened 2026-07-24
GeecsCAGateway 0.15.0: set-failure observability — :SP WRITE/INVALID alarm + sticky LAST_SET_ERROR PV

skbarber:
## What + why

A failed GEECS set (e.g. an out-of-bounds value the device ACKs `accepted` but rejects in the exe reply) was visible **only** through CA put-completion. No-callback writers — plain `caput`, default Phoebus writes, and any future agent tooling doing raw CA — saw the put as successful; the device's error message vanished into the gateway log. Live-diagnosed on `U_S1H` `Current` 2026-07-23 (put −100 A, limits ±5): the gateway correctly failed a put-callback write in 0.17 s, but nothing was published for anyone else.

This PR makes the outcome **published channel state** (the standard EPICS idiom), so every client class observes it:

1. **`:SP` channel alarm** (`channels.py`): a failed forward stamps `WRITE`/`INVALID` and re-raises (put still fails, value still unstored — no change for callback clients); the next successful set clears to `NO_ALARM`. Transition-only publishes. Pre-forward client errors (uncastable value) deliberately do not alarm; unresolvable enum labels keep their existing forward-verbatim semantics and alarm only when GEECS rejects them.
2. **Per-device `…:last_set_error` PV** (`gateway.py`): long-string, read-only, sticky record of the most recent failure message (e.g. `U_S1H/Current: … value not in range (5.000000,-5.000000)`); built only for devices with a settable variable.
3. **Docs**: PV_CONTRACT §1/§2/§7 (new "Set-failure observability" section + put-callback requirement), DEPLOYMENT §3 client recipe ("always write with put-completion"), CHANGELOG, version minor → 0.15.0.

## Per-concern LOC (net)

| Concern | Files | ~LOC |
|---|---|---|
| Alarm on failed set | `channels.py` | +45 |
| `LAST_SET_ERROR` PV + failure recording | `gateway.py` | +47 |
| Fake-server per-variable limits (real error phrasing) | `testing/fake_device_server.py` | +33 |
| Pinned tests | `tests/test_gateway.py`, `tests/test_pv_contract.py` | +136 |
| Contract/docs/version | `PV_CONTRACT.md`, `DEPLOYMENT.md`, `CHANGELOG.md`, `pyproject.toml` | +87 |

- **`WRITE`/`INVALID`** as the failure alarm (consistent with the gateway's INVALID-on-comm-loss idiom).
- **Per-device** (not per-variable) error PV — GEECS serializes one command per device; the message names the variable.
- **Sticky** error text (the alarm carries current/cleared state; the PV is the forensic record).
- Fake server keeps `set() -> bool` for compat; new `try_set()` returns the error detail.

## Tests

`./scripts/check.sh` (scoped): lint green, **GeecsCAGateway 186 passed** (was 182; +4 pinned: alarm transitions, pre-forward no-alarm, e2e out-of-range with ACK-accepted + exe-error, `LAST_SET_ERROR` only on settable devices).

## Hardware verification (done — live, 2026-07-23)

Patched gateway run locally (CA on 127.0.0.1) against the **real U_S1H** over VPN:

- Out-of-bounds put (−100): aioca put-callback **failed in 0.21 s**, value unstored, hardware untouched (readback 8e-05 throughout).
- `:SP` alarm → `INVALID/WRITE` after the failure; → `NO_ALARM` after a subsequent in-bounds put (0.0, the standing setpoint; succeeded in 1.14 s).
- `last_set_error` = `U_S1H/Current: Error occurred during setCurrent -  -100.000000000000 - value not in range (5.000000,-5.000000)` — exact hardware phrasing, sticky across the successful set.
- Note: caproto's *threading client* `write(wait=True)` times out instead of surfacing the server's `ECA_PUTFAIL` ErrorResponse (client-side quirk; libca/aioca — what our stack and EPICS-base `caput` use — fail promptly). Not a server regression; deployed-gateway behavior identical pre-patch.

### DIFF
--- a/GeecsCAGateway/CHANGELOG.md
+++ b/GeecsCAGateway/CHANGELOG.md
@@ -3,6 +3,29 @@
 All notable changes to `geecs-ca-gateway` are documented here, following
 [Keep a Changelog](https://keepachangelog.com/en/1.1.0/) and semantic versioning.
 
+## [0.15.0] - 2026-07-23
+
+### Added
+
+- **Set-failure observability** (PV_CONTRACT §7): a failed GEECS set — e.g.
+  an out-of-bounds value the device ACKs `accepted` but rejects in the exe
+  reply (live-verified on U_S1H `Current`, 2026-07-23) — now surfaces as
+  published channel state, not only as put-completion failure, so
+  fire-and-forget CA writers (plain `caput`, default Phoebus writes) are no
+  longer blind to it:
+  - the `:SP` channel alarm goes `WRITE`/`INVALID` on a failed set and
+    clears to `NO_ALARM` on the next successful one (transition-only
+    publishes; pre-forward client errors such as an uncastable value do
+    not alarm);
+  - every device with a settable variable gains a
+    `[experiment:]device:last_set_error` long-string PV holding the most
+    recent failure message, sticky until the next failure.
+- `FakeGeecsDevice` gains per-variable `limits` — an out-of-range set is
+  ACKed but fails in the exe reply with the real hardware's
+  `value not in range (hi,lo)` phrasing, for offline pinning of the above.
+- `DEPLOYMENT.md` §3: always write with put-completion; documented why a
+  plain caput cannot see set failures (CA protocol semantics).
+
 ## [0.14.3] - 2026-07-21
 
 ### Changed
--- a/GeecsCAGateway/DEPLOYMENT.md
+++ b/GeecsCAGateway/DEPLOYMENT.md
@@ -242,6 +242,14 @@ subset plus each device's settable control surface (§1), so a variable
 visible in GEECS Master Control is usually reachable here under its
 lowercased name.
 
+**Always write with put-completion** (`caput -c`, caproto-put's default,
+aioca `wait=True`, ophyd-async `set()`). A fire-and-forget write returns
+before the GEECS exchange starts, so a rejected set (e.g. out of the
+device's limits) *looks* successful to that client. A failed set also stamps
+the `:SP` PV with a `WRITE`/`INVALID` alarm and records its message on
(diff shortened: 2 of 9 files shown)


skbarber:

Reviewer ran correctness/redundancy/placement lenses over the full diff, traced the vendored caproto publish path to verify the alarm semantics the PR relies on, and ran the suite. Three findings, all **fixed** in 105d1c4a and **confirmed by the same reviewer**:

1. **Medium — pinned tests couldn't detect a publish regression** (the PR's entire deliverable: `publish=False` on the failure path would have kept all tests green while making the feature a no-op for monitor clients). Fixed: `test_failed_set_alarm_is_published_on_transition_only` wraps `channel.alarm.write` and pins failure-publishes / repeat-doesn't-republish / clear-rides-value-publish; PV_CONTRACT test-map row added. Reviewer re-verified each falsification now trips the test.
2. **Low — failure-path `alarm.write` unguarded inside `except`**, could mask the real set error with a publish-side exception. Fixed: best-effort try/except + debug log, matching `_record_set_error`'s idiom; the original error's `raise` sits outside the guard.
3. **Nit — test helper duplicated `channels._to_text`.** Fixed: test delegates to the package helper.

Refuted candidates the reviewer checked and cleared (recorded for posterity): alarm-mask-only monitors missing the `publish=False` clear (caproto fans every event to all subscribers; the value event carries the cleared alarm), pre-forward errors hitting caproto's own alarm stamping, UTF-8 truncation on the error PV, redundancy/placement across GeecsBluesky/Console, and `LAST_SET_ERROR` name collisions (same pre-existing `_register` guard as `CONNECTED`).

No waived findings. Suite: 187 passed.

PR #603, opened 2026-07-24
GeecsCAGateway 0.16.0: mirror DB set limits as :SP control limits — pre-forward CA rejection, GEECS backstop unchanged

skbarber:
## What + why

Restores control-limit enforcement removed (as a side effect) by `9396c61f` — but on **setpoint channels only**, keeping that commit's readback fix intact. `:SP` channels now serve the DB `min`/`max` span as standard EPICS **control (drive) limits** (`DBR_CTRL`) and reject out-of-range puts at the CA layer, mirroring GEECS's own enforcement per usual EPICS gateway practice. **GEECS device-side enforcement remains the guaranteed backstop** (owner-stated framing): interlocks, local mode, and DB-edit drift are still caught device-side and drive the 0.15.0 alarm/`LAST_SET_ERROR` machinery unchanged.

Two implementation findings that shaped the code (both verified against installed caproto source):

1. **caproto rejects (`CannotExceedLimits`), it does not clamp** — so a failed put is a real failure, not a silently altered value.
2. **caproto's own enforcement (`verify_value`) runs inside `super().write()` — i.e. *after* our GEECS forward.** Naive ctrl-limit kwargs alone would let the UDP set go out first; with DB limits tighter than the device's, the hardware would move and *then* the put would fail. The check is therefore explicit and **pre-forward** in the setpoint write path (raising caproto's own exception class so rejections are indistinguishable from a native IOC's). Pinned by an ordering test (counting setter never invoked).

**Data-quality guard**: ctrl limits are served only for well-formed spans (both bounds, `lo < hi`). DB audit today: 1480/1550 settable Undulator numerics qualify; 62 are degenerate `0,0` rows, 1 one-sided, 0 swapped — all of those (and future bad rows) leave the `:SP` unenforced rather than bricking the axis. Pre-forward rejection is a *client* error per PV_CONTRACT §7: no `:SP` alarm, no `LAST_SET_ERROR`.

**Honest premise note**: the original ticket framed this as fixing clients "misreading success" — that's not quite it (post-0.15.0, callback clients already get `ECA_PUTFAIL`, and no-callback clients see ctrl-limit rejections no better — CA delivers those asynchronously too). The real wins: fail-fast with zero device perturbation, and `DBR_CTRL` metadata that ophyd-style clients (incl. the OSPREY epics_connector) and Phoebus sliders use to validate/bound *before* putting.

## Per-concern LOC (net)

| Concern | Files | ~LOC |
|---|---|---|
| Ctrl-limit metadata (SP-only) + pre-forward check | `channels.py` | +54 |
| Pinned tests (limits split, ordering, degenerate spans, layered backstop) | `tests/test_pv_contract.py`, `tests/test_gateway.py`, `tests/test_config_from_db.py` | +125 |
| Contract/docs/version | `PV_CONTRACT.md`, `CHANGELOG.md`, `pyproject.toml` | +69 |

## Tests

`./scripts/check.sh` (scoped): lint green, **GeecsCAGateway 190 passed** (was 187; +3 pinned: pre-forward rejection ordering, missing/one-sided/degenerate spans stay unlimited, CA-gate + device-backstop layering e2e; 1 updated: `test_pvdb_built_from_db_spec_has_limits` now asserts the readback/setpoint split; NaN-readback regression test untouched and green).

## Hardware verification (done — live, 2026-07-24)

Patched gateway run locally (CA on 127.0.0.1) against the **real U_S1H** over VPN:

- `:SP` serves `DBR_CTRL` metadata: ctrl (−5, 5), disp (−5, 5), EGU `A`.
- `caput …current:SP 100` (put-callback) **failed in 0.002 s** — faster than one VPN round-trip (20 ms rtt), proving no UDP reached the device; `:SP` alarm stayed `NO_ALARM` and `last_set_error` stayed empty (client error, not device outcome).
- In-range `caput …current:SP 0.0` succeeded in 1.12 s — blocking GEECS set semantics unchanged; readback steady at 8e-05 A throughout (magnet never moved).

### DIFF
--- a/GeecsCAGateway/CHANGELOG.md
+++ b/GeecsCAGateway/CHANGELOG.md
@@ -9,9 +9,11 @@ All notable changes to `geecs-ca-gateway` are documented here, following
 
 - **Control (drive) limits on setpoint PVs** (PV_CONTRACT §2): `:SP`
   channels now serve the DB `min`/`max` span as enforced `DBR_CTRL` limits —
-  mirroring GEECS's own enforcement at the gateway per standard EPICS
-  gateway practice — and reject out-of-range puts **pre-forward**
-  (`CannotExceedLimits`, no UDP traffic, no device perturbation; the check
+  mirroring GEECS's own enforcement at the gateway the way native EPICS
+  drive limits (`DRVL`/`DRVH`) express a record's valid span — and reject
+  out-of-range puts **pre-forward** (`CannotExceedLimits`; rejection, never
+  the EPICS record's clamp-to-limit — clamping would move hardware to an
+  uncommanded value; no UDP traffic, no device perturbation; the check
   runs before the GEECS forward because caproto's own `verify_value` runs
   after it). Pre-forward rejection is a client error: no `:SP` alarm, no
   `LAST_SET_ERROR` (those keep mirroring device-side outcomes only).
--- a/GeecsCAGateway/CHANGELOG.md
+++ b/GeecsCAGateway/CHANGELOG.md
@@ -3,6 +3,27 @@
 All notable changes to `geecs-ca-gateway` are documented here, following
 [Keep a Changelog](https://keepachangelog.com/en/1.1.0/) and semantic versioning.
 
+## [0.16.0] - 2026-07-24
+
+### Added
+
+- **Control (drive) limits on setpoint PVs** (PV_CONTRACT §2): `:SP`
+  channels now serve the DB `min`/`max` span as enforced `DBR_CTRL` limits —
+  mirroring GEECS's own enforcement at the gateway per standard EPICS
+  gateway practice — and reject out-of-range puts **pre-forward**
+  (`CannotExceedLimits`, no UDP traffic, no device perturbation; the check
+  runs before the GEECS forward because caproto's own `verify_value` runs
+  after it). Pre-forward rejection is a client error: no `:SP` alarm, no
+  `LAST_SET_ERROR` (those keep mirroring device-side outcomes only).
+  Served only for well-formed spans (both bounds, `lo < hi`; DB audit
+  2026-07-24: 1480/1550 settable Undulator variables qualify) — missing,
+  one-sided, or degenerate rows leave the `:SP` unenforced so a bad DB row
+  can never brick an axis. **GEECS device-side enforcement remains the
+  guaranteed backstop** (interlocks, DB-edit drift): values inside the DB
+  span that the device rejects still fail with the device's error and the
+  0.15.0 alarm machinery. Readbacks are untouched — display limits only
+  (the 9396c61f NaN-readback behavior, now pinned per channel kind).
+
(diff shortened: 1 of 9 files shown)


skbarber:

Reviewer traced every downstream `:SP` consumer in GeecsBluesky (gateway_put, motor, settable, pseudo, confirm, action signals, ShotController) + Console, verified the pre/post-forward check consistency against installed caproto source, and confirmed no scan/optimization/action pathway relied on out-of-range puts reaching the device (where the DB span mirrors the device's, those paths failed device-side before and fail CA-side now — scan aborts either way). Five findings; four **fixed** in 2a9d8cdd, one **accepted with doc fix** — all confirmed by the same reviewer:

1. **Medium — accepted trade-off (owner-vetoable): fire-and-forget out-of-range writes lose their PV trace.** Pre-PR they were device-rejected and left the 0.15.0 alarm + `LAST_SET_ERROR` breadcrumbs; now they're pre-forward client errors, visible only to put-completion writers and the gateway service log (reviewer verified caproto logs the rejection for no-notify writes too). Mitigation: served `DBR_CTRL` metadata lets sliders/ophyd-style clients refuse pre-put. PV_CONTRACT §7's now-inconsistent "or monitor the alarm" advice was rewritten to state the trade-off explicitly. If the operator-facing hole matters, recording ctrl-limit rejections into `LAST_SET_ERROR` (sans alarm) is a small follow-up.
2. **Medium — DESIGN.md/DEPLOYMENT.md contradicted the new contract** ("UX hint", "a failed set also stamps… (e.g. out of the device's limits)"). Fixed: both now tell the two-layer story (CA mirror + GEECS guaranteed backstop).
3. **Low — stale-tight drift direction undocumented** (a wrong-tight/widened DB span now hard-blocks until gateway restart). Fixed: §2 "Drift cuts both ways" bullet with fix-the-row-bounce-the-gateway guidance. The adjacent `CaPseudoMovable` baseline-restore edge (readback-captured value outside DB span) is accepted under the same premise — where DB mirrors device, the restore would have failed device-side before.
4. **Low — overstated precedent claims** ("same rejection a native IOC would produce" — native `ao` records clamp; "standard EPICS gateway practice" — the stock CA Gateway does access security, not drive limits). Fixed: accurate rejection-vs-clamp language with the uncommanded-move rationale.
5. **Low — `caproto._data` private import under an uncapped pin.** Fixed: guarded import with a `CaprotoValueError`-subclass fallback whose client-observable failure is indistinguishable (reviewer verified the ErrorResponse carries matching class name + message).

Refuted/clean (recorded): gateway self-write hazard (no internal setpoint writes; ctrl limits don't validate the 0.0 initial value), pre/post double-check divergence (semantics match caproto's exactly), 0.15.0 tests not hollowed (device-path test has no DB span; layered test pins both sides), redundancy/placement clean.

No waived findings. Suite: 190 passed; lint clean.

PR #604, opened 2026-07-25
geecs-data-utils 0.12.1: IMAQ decode tail-anchored fallback (no repeated-name wrapper)

skbarber:
## What + why

`decode_imaq_image_string` anchors the pixel payload on the device-name string being repeated immediately before it. Some cameras flatten frames **without** that repetition (observed live: `UC_Amp2_IR_input`, whose wrapper carries a compression-mode field in that position), so every frame failed with `payload 518392 not divisible by 506 rows`.

When the name-anchored payload is inconsistent with the header, the decoder now falls back to tail-anchoring: geometry **and pixel type** come from the IMAQ struct (type code at header offset +36; `7` = Grayscale U16), and the pixel block is the last `rows × stride` bytes with the stride rounded to IMAQ's 64-byte row alignment. Unsupported pixel types and short payloads still raise `ValueError`.

Motivation: this decoder is the decode step for the distributed PVA image-server work (per `GeecsCAGateway/DESIGN.md`'s images-off-CA decision); the gap was found during the live pilot on a Windows camera server.

~50 LOC code, ~50 LOC tests, single concern.

## Tests

- 2 new: tail fallback decodes a no-name-repeat 16-bit frame (asserts the anchor is genuinely absent); unknown pixel type raises instead of misdecoding.
- Suite: `GEECS-Data-Utils/tests` **143 passed** (was 141), including all 8 pre-existing decode tests unchanged.

## Hardware verification

Verified live 2026-07-25: a raw frame captured from `UC_Amp2_IR_input` (500×500 uint16, border 3, type 7) decodes through the patched public API to the same array as a manual header dissection — min/max/mean 44/3202/56.1, borders zero — and the same fix ran in the pilot PVA bridge streaming this camera at 1 Hz and 5 Hz to a viewer over the VPN.

## Flagged for cheap veto

- The captured real blob (~506 KB) is **not** committed as a test asset (size); synthetic wrappers cover both layouts, and the live verification above is the ground-truth check.
- Known gap, deliberately out of scope: a *compressed* (JPEG-mode) frame from a no-name-repeat camera would reach the tail fallback and raise rather than decode — no such blob has been observed; handling it blind felt speculative.
- Pre-existing ambiguity, unchanged: if a no-anchor payload length happens to be divisible by the row count, the anchor path can still 'succeed' on garbage; tightening that heuristic is not attempted here.


skbarber:

**Finding 1 (High, correctness):** Fallback dispatched on size arithmetic (`payload % rows != 0`) rather than on structural absence of the name-repeat anchor. Two failure modes: (A) an anchored camera's truncated frame — previously a loud `ValueError` — would silently tail-decode to garbage; (B) a no-repeat camera whose message length happens to be divisible by the row count would never reach the fallback and would misdecode wrapper bytes as pixels via the main path.
**Disposition: FIXED** in 5e5376b1 — dispatch is now `data.find(name, 4 + name_len) == -1` (structural, before any arithmetic); both original strict-path raises restored verbatim; new test `test_decode_anchored_truncated_frame_still_raises` pins dispatch at a scale where the tail block would fit (508-byte message vs 384-byte block). Reviewer re-verified the diff and **confirmed resolved**, including the edge cases (anchored+truncated → loud raise; accidental divisibility on no-repeat → tail path unconditionally).

**Finding 2 (Low, correctness):** 64-byte stride alignment in the tail path is a single-camera extrapolation; the main path deliberately infers stride from payload length instead.
**Disposition: FIXED** in 5e5376b1 — `_decode_imaq_tail` docstring now names the provenance (UC_Amp2_IR_input, 500×500 u16 → stride 1024) and both failure modes for a camera with a different real stride. Reviewer **confirmed resolved**.

**Finding 3 (Trivial, correctness):** `_IMAQ_DTYPE_BY_TYPE` mapped IMAQ type 1 → int16, contradicting the public docstring's uint8/uint16 return contract, untested.
**Disposition: FIXED** in 5e5376b1 — type 1 dropped (map is `{0: u8, 7: u16}`); unknown types raise. Reviewer **confirmed resolved**.

**Redundancy:** none found (sole IMAQ flatten decoder in the repo; python-api's `camera.py` delegates here). **Placement:** confirmed correct (bottom-of-graph io module, established home). Reviewer's final verdict after the fix commit: **no surviving findings**.

PR #605, opened 2026-07-25
GeecsCAGateway 0.16.1: TCP subscriber latin-1 decode — binary payloads survive byte-for-byte

skbarber:
## What + why

`GeecsTcpSubscriber._listen_loop` decoded push payloads with `ascii, errors="replace"` — correct for its scalar-only design scope, but every byte >0x7f becomes U+FFFD, which irreversibly corrupts binary payloads (image frames). Switching to **latin-1** (the lossless byte↔str bijection) preserves binary exactly while leaving ASCII scalar frames byte-identical, so existing behavior is unchanged.

Motivation: the distributed PVA image-server workstream (DESIGN.md's images-off-CA decision) will reuse this subscriber on the camera servers as its loopback GEECS transport — GEECS-PythonAPI is deprecated and not an option. This is the same fix python-api received in PR #438; the gateway's parser is otherwise already binary-capable (name-anchored `nval`/`nvar` tokenization, values pass through `text_variables` verbatim).

1 line of code + comment; ~40 LOC test. Single concern.

## PV contract

No externally observable gateway behavior changes: ASCII decodes identically under latin-1, and the gateway continues to serve scalars only (no image PVs — the ground rule in this package's CLAUDE.md stands). `PV_CONTRACT.md` untouched accordingly.

## Tests

- 1 new: `test_binary_payload_survives_byte_for_byte` — a hand-built push frame carrying `bytes(range(256)) * 4` in a `text_variables` value round-trips exactly via `value.encode("latin-1")`. Fails against the old ascii/replace decode.
- Suite: `GeecsCAGateway/tests` **191 passed** (was 190), gateway env, offline.

## Hardware verification

Not required — no live-path behavior change for served scalars (ASCII ⊂ latin-1, byte-identical). The binary path this enables was exercised live 2026-07-25 through the equivalent python-api decode during the image-server pilot (UC_Amp2_IR_input, 1–5 Hz frames over PVA); the gateway-transport port will get its own live check in the image-server PR (B).

### DIFF
--- a/GeecsCAGateway/CHANGELOG.md
+++ b/GeecsCAGateway/CHANGELOG.md
@@ -13,8 +13,11 @@ All notable changes to `geecs-ca-gateway` are documented here, following
   every byte >0x7f became U+FFFD. latin-1 is the lossless byte↔str map, so a
   `text_variables` subscriber recovers the exact wire bytes via
   `value.encode("latin-1")`. Prerequisite for the distributed PVA image-server
-  workstream (DESIGN.md: images stay off CA, data stays at the edge); no
-  externally observable gateway behavior changes.
+  workstream (DESIGN.md: images stay off CA, data stays at the edge). Served
+  PVs are unchanged for ASCII frames; text values carrying bytes >0x7f (never
+  observed in production) now decode as latin-1 characters instead of U+FFFD
+  replacement chars. The UDP get/exe path keeps its ASCII decode — the TCP
+  push stream is the only binary-bearing path.
 
 ## [0.16.0] - 2026-07-24
 
--- a/GeecsCAGateway/CHANGELOG.md
+++ b/GeecsCAGateway/CHANGELOG.md
@@ -3,6 +3,19 @@
 All notable changes to `geecs-ca-gateway` are documented here, following
 [Keep a Changelog](https://keepachangelog.com/en/1.1.0/) and semantic versioning.
 
+## [0.16.1] - 2026-07-25
+
+### Fixed
+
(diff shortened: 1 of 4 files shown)


skbarber:

Reviewer verified the change is correct, the test genuinely fails against the old `ascii/replace` decode (checked by reasoning through the U+FFFD → `UnicodeEncodeError` path), the duplication with python-api's identical fix is by design (dependency graph forbids the import; python-api deprecated), and placement/versioning are right. Three findings, none blocking:

**Finding 1 (Low):** CHANGELOG's "no externally observable gateway behavior changes" was strictly too strong — text PVs carrying bytes >0x7f would now render as latin-1 characters instead of U+FFFD.
**Disposition: FIXED** in d268fd66 — the entry now states the exact scope (ASCII unchanged; >0x7f text latin-1 vs U+FFFD, never observed in production; UDP path unchanged). Reviewer re-read the entry clause-by-clause and **confirmed exact**.

**Finding 2 (Low):** `logger.debug("TCP rx: %r", msg)` becomes a multi-MB-per-frame log line at DEBUG once image frames ride this subscriber. Lazy formatting means non-DEBUG deployments are unaffected.
**Disposition: WAIVED here, owed to the image-server PR (B)** — the line predates this diff, and PR B is where the subscriber first carries image frames; it will get a truncated/length-only debug line there. Recorded so it isn't lost.

**Finding 3 (Informational):** the UDP get/exe path (`udp_client.py`) still decodes ascii/replace — binary-capable means the TCP push stream only.
**Disposition: documented** — the CHANGELOG wording now says this explicitly; no code change (UDP is a text command protocol; images ride the push stream per DESIGN.md).

Issue #512, opened 2026-07-12
GeecsCAGateway: align audit classifier with gateway dtype resolution


skbarber:
## Context

Post-merge adversarial review of the merged `feat/vision-v1` branch, covering GeecsCAGateway PR #498.

## Finding

The DB audit classifier does not use the same effective dtype resolution as the gateway config builder. It only checks `variabletype`, but the gateway treats some `choices` values as authoritative type descriptors.

Relevant code:

- `GeecsCAGateway/geecs_ca_gateway/audit.py:112-163` builds `by_name` as `name -> variabletype` and flags skipped variables only when `variabletype` is in the skip set.
- `GeecsCAGateway/geecs_ca_gateway/db/geecs_db.py:102-124` includes both `variabletype` and `choices` in metadata.
- `GeecsCAGateway/geecs_ca_gateway/config.py:191-223` resolves the effective type from `choices` first when `choices` is one of `image`, `1darray`, `numeric`, `string`, or `path`; this is documented in comments as authoritative even when `variabletype` says `choice`.
- `GeecsCAGateway/tests/test_config_from_db.py:286-302` pins the real gateway behavior: `variabletype='choice'` with `choices='image'` or `choices='1darray'` is skipped, not served as a one-option enum.

## Risk

The audit can report some real non-scalar subscribed variables as clean, especially camera/array variables encoded as `variabletype='choice'` with `choices='image'` or `choices='1darray'`. That defeats the tool's purpose: it is supposed to report every `get='yes'` variable the gateway will not serve.

## Suggested fix

Share or mirror the gateway's effective-type resolution in the audit classifier. The safest version is to extract a small pure helper from `DeviceSpec.from_db_metadata()` and use it in both config building and `audit_subscribed_variables()`.

## Regression test

Add an audit test with metadata rows like `{"name": "frame", "variabletype": "choice", "choices": "image"}` and `{"name": "trace", "variabletype": "choice", "choices": "1darray"}`. Both should be reported as `SKIP:image` / `SKIP:1darray`.

skbarber:
Fixed in #514 (merged to feat/vision-v1): effective-type resolution extracted to `effective_vartype()` in config.py and shared by DeviceSpec.from_db_metadata and the audit classifier. Regression tests pin choice/image → SKIP:image and choice/1darray → SKIP:1darray.

PR #606, opened 2026-07-25
GeecsPvaGateway 0.1.0: distributed PVA image gateway (NTNDArray) + GeecsCAGateway 0.17.0 transport additions

skbarber:
## What + why

**PR B of the image-server ladder** (A = #604 decoder fix, A2 = #605 binary-safe transport — both merged). New package `GeecsPvaGateway/`: the **PVA peer of GeecsCAGateway**, one instance per Windows camera server, serving that host's GEECS camera images as NTNDArray PVs. Implements the DESIGN.md decision (images stay off CA; data stays at the edge) with the architecture verified in the 2026-07-25 live pilot (UC_Amp2_IR_input, 1–5 Hz over VPN).

Core behaviors:
- **DB-scoped serve set** (no per-host config): enabled devices whose endpoint IP is this machine and that have image-typed variables (`effective_vartype`, incl. the #512 choice-descriptor quirk). Verified live against the Undulator DB: the pilot box scopes to exactly its 12 cameras; its timing box and MCC filter out.
- **Gated subscriptions**: GEECS TCP subscription starts on the first PVA client, stops on the last (p4p `onFirstConnect`/`onLastDisconnect`) — unwatched cameras cost LabVIEW nothing.
- **Supervised reconnect** (0.5→30 s backoff) on socket drops; silence is not a drop (CA-gateway doctrine).
- **Latest-wins frames**: decode (`decode_imaq_image_string`) off the event loop; a stalled consumer drops stale frames, never backlogs. Completeness is the file path's job.
- **GEECS timestamp ladder** (`acq_timestamp` → `systimestamp`, LabVIEW→Unix).
- **Naming via the shared contract** (`pv_naming`): `undulator:uc_amp2_ir_input:image`. Per-instance `{exp}:pvagateway:{host}:version|heartbeat` identity PVs for the fleet screen.
- CLI: `geecs-pva-gateway --experiment NAME [--host IP] [--devices A,B] [--list]`.

## Per-concern LOC breakdown

| Concern | Files | ~LOC |
|---|---|---|
| Package core (config + server + CLI) | `GeecsPvaGateway/geecs_pva_gateway/` | 480 |
| Tests | `GeecsPvaGateway/tests/` | 320 |
| Package docs (README/CLAUDE/DEPLOYMENT/CHANGELOG) | `GeecsPvaGateway/*.md` | 300 |
| Gateway transport additions: `wait_disconnected()` + debug-log truncation (owed from #605 review finding 2) | `GeecsCAGateway/` (0.16.1→0.17.0) | 15 |
| Cross-package docs: root CLAUDE map/graph + python-api deprecation wording (owner decision 2026-07-25), gateway DESIGN/CLAUDE/PV_CONTRACT pointers, GeecsBluesky images note | 5 `.md` | 60 |
| CI + check.sh wiring | `unit-tests.yml`, `scripts/check.sh` | 12 |
| `poetry.lock` | 1 file | (generated) |

## Tests

- New package: **7 passed** — config scoping/naming units (fake DB rows) + end-to-end over a binary wire-format fake camera and an isolated PVA server: frame flow + timestamp ladder, gating on/off, reconnect-after-drop, identity PVs.
- GeecsCAGateway: **191 passed** after the transport additions.
- Lint: `check.sh --lint` green. New CI step runs the package suite (mirrors GeecsBluesky's own-env step).

## Hardware verification

**OWED** (next VPN/lab session, ~10 min, box already provisioned): on 192.168.6.100 run `geecs-pva-gateway --experiment Undulator --list` (expect exactly its 12 cameras' image PVs), then serve and view `undulator:uc_amp2_ir_input:image` live from the Mac (expect 500×500 u16 at trigger rate, timestamps tracking `acq_timestamp`); confirm an unwatched camera holds no GEECS TCP connection.

## Flagged for cheap veto

- **All image-typed variables are served** per camera (`image`, `processed image`, `bakground image`), not just `image` — DB-driven symmetry, and gating makes unwatched PVs free. Veto → serve `image` only.
- **Instance-PV naming** `{exp}:pvagateway:{host_token}:version|heartbeat` (lowercase, mirrors the modern contract rather than the CA gateway's `CAGateway:*` uppercase style).
- **PV_CONTRACT placement**: the CA contract gains only a cross-reference; the image-side contract lives in the new package's docs. Alternative would be one combined contract doc.
- **CI note**: GeecsCAGateway's own suite still has no CI step (pre-existing); I added one only for the new package rather than scope-creeping.
- Compressed (JPEG-mode) cameras: decode handles the anchored wrapper; a *no-name-repeat compressed* camera would fail decode loudly (documented gap from #604, unchanged here).


skbarber:
## Hardware verification — live results (2026-07-25, from Mac over VPN)

Ran with `--host 192.168.6.100` (same code path as on-box; only the loopback placement differs):

- **DB scoping**: `--list` resolves the box to exactly its **12 cameras / 36 image PVs** (contract-named, e.g. `undulator:uc_amp2_ir_input:image`); the timing box and MCC on the same host are excluded. ✅
- **Live frames** (`UC_Amp2_IR_input`, real device): 6 real frames in 6 s, acquisition-timestamp gaps **1.000 ± 0.001 s** (the camera's 1 Hz trigger, via the GEECS timestamp ladder); frame age at the client **0.13 s**. ✅
- **Gating, both directions, observed via netstat**: no TCP connection to the camera with zero PVA clients → ESTABLISHED while a monitor is attached → connection dropped within seconds of client close. ✅
- **Identity PVs** served: `undulator:pvagateway:192_168_6_100:version|heartbeat`. ✅
- **Bonus**: the camera is currently in 8-bit/compressed mode (this morning's pilot had it in uint16) — the decode chain handled the mode change transparently, giving the other decode path live coverage.

**Remaining OWED (shrunk):** run the same on the camera server itself (auto host scoping + loopback subscription + eventually the NSSM service) at the next session with box access — no code-path difference, placement only.

skbarber:

A fresh-context reviewer attacked the diff (correctness / redundancy / placement), ran the suites, and re-verified each fix against the actual commits. Five findings; all fixed or reasonably waived, confirmed by the reviewer.

**Finding 1 — no PV-name collision guard (medium).** Lossy name normalization means two variables could land on one PV name and one would silently shadow the other; the CA gateway can't backstop this (it skips image vartypes entirely).
**Disposition: FIXED** across two rounds. `run()` first gained a cross-worker owners-map guard (`e0de2466`); round 2 (`46ff6a6a`) closed the within-*camera* case the reviewer reproduced live — `provider_entries()` now yields one `(pv_name, var, SharedPV)` row per variable so duplicates can't pre-collapse in a dict. Both cases pinned (`test_cross_camera_...`, `test_within_camera_pv_name_collision_raises`). Reviewer **confirmed fixed** with its own reproducer.

**Finding 2 — per-camera gating contradicted the "unwatched PVs are free" claim (medium-low).** Watching one variable subscribed and decoded all of a camera's image variables.
**Disposition: FIXED** — gating is now per image variable (each has its own gated, supervised subscription; decode runs only for watched variables). Pinned by `test_gating_is_per_variable`; docs updated to say per-variable; re-verified live (6 frames/6 s at 1 Hz). Reviewer **confirmed**.

**Finding 3 — root CLAUDE.md self-contradiction on GEECS-PythonAPI.** Repo-map row said "deprecated, never build on it"; the Known-debt bullet still said "being refactored, treat ScanDevice as public surface."
**Disposition: FIXED** — both the Known-debt bullet and the declaration-quirks paragraph now carry the deprecation/deletion framing. Reviewer **confirmed**.

**Finding 4 — timestamp ladder duplicated across the two gateways (low).**
**Disposition: WAIVED** — true single-sourcing means refactoring the CA gateway's per-device fan-out ladder, out of scope and riskier than 6 duplicated lines; `server.py` carries a comment naming the CA gateway as doctrine source. Reviewer **accepted**.

**Finding 5 — local-IP route probe duplicated with a hard-coded router IP (low).**
**Disposition: FIXED** — reuses the gateway's helper (promoted to public `detect_local_ip`); probe target now derives from a DB endpoint IP, no magic constant. Reviewer **confirmed**.

**Minors (all confirmed fixed):** `__version__` single-sourced; `_publish` catches decode+post broadly without eating `CancelledError`; a `stop()` dict-mutation-during-iteration bug (found by the new tests) fixed via snapshot-and-clear; a mid-shutdown retain race closed with a `_stopping` latch; docstring nit.

**Deliberately deferred (dispositioned, non-blocking):** CI runs this package's socket-opening `fake_server` tests while GeecsBluesky deselects them — will revisit if it flakes. `_instance_token()` picks the first camera's host IP — a multi-homed camera server could flip the identity-PV prefix across restarts; single-homed boxes are the norm. `--list` still prints a collapsed duplicate line for a colliding config (cosmetic; `run()` is the hard gate).

Final reviewer verdict: no surviving findings.

PR #607, opened 2026-07-25
GeecsPvaGateway 0.2.0: deployment machinery — :restart PV, NSSM bootstrap, pull-on-restart, fleet screen

skbarber:
## What + why

**PR C of the image-gateway ladder** (A #604, A2 #605, B #606 — all merged): the fleet deployment machinery, per the endorsed shape (wheel-on-share pull-on-restart, RESTART-caput rollouts, version/heartbeat fleet screen). No changes to the frame pipeline, gating, or config scoping.

- **`:restart` PV** (`{exp}:pvagateway:{host}:restart`): any put → clean shutdown, workers stopped, exit code 86 (`RESTART_EXIT_CODE`) — NSSM relaunches, which re-resolves DB config and re-pins the wheel. Mirrors the CA gateway's `CAGateway:RESTART` → exit 86 → systemd pattern.
- **`deploy/bootstrap.ps1`** — one-time per-box setup: layout, venv + install, idempotent firewall rules, NSSM fetch + service registration. Key design: the service runs LocalSystem with **`USERPROFILE` overridden to a service-owned profile dir** (`C:\geecs\pva-gateway\profile\` holding config.ini + `user data\Configurations.INI`) — solves both session-0 rules (no mapped drives, no LocalSystem home) with zero service-account passwords.
- **`deploy/launch.bat`** — pull-on-restart: pins to the wheel named by `CURRENT` on the share when reachable; falls through to the installed version otherwise (a restart never bricks an instance).
- **`deploy/fleet_status.bob`** — Phoebus fleet screen: version / heartbeat / confirm-dialog restart per host.
- **`DEPLOYMENT.md`** rewritten as the scripted runbook.

Dropped from the original C sketch: a GEECS-Schemas config schema — moot since B folded in DB-self-scoping (there is no config file to schematize).

## Per-concern LOC breakdown

| Concern | Files | ~LOC |
|---|---|---|
| `:restart` PV + exit-code plumbing | `server.py`, `__main__.py` | 55 |
| Test (`test_restart_pv_shuts_down_cleanly`) | `tests/test_server.py` | 25 |
| Deploy scripts | `deploy/{bootstrap.ps1,launch.bat}` | 105 |
| Fleet screen | `deploy/fleet_status.bob` | 70 |
| Docs (DEPLOYMENT rewrite, CLAUDE, CHANGELOG) | 3 `.md` | 120 |

## Tests

[...]

PR #609, opened 2026-07-26
GeecsPvaGateway 0.3.0: pull-on-restart from the shared GEECS-Plugins clone (Active Version pattern)

skbarber:
## What + why

Owner decision: the lab's `Active Version` share already carries a GEECS-Plugins clone — the same pattern GEECS device software launches from — so the fleet should install from **it**, not from a parallel wheel drop. This replaces the wheel/`CURRENT` machinery from #607 (which never reached a box) before it ships anywhere:

- `launch.bat`: `GEECS_PVA_SOURCE` = UNC of the share clone; each restart reinstalls `GEECS-Data-Utils` + `GeecsCAGateway` + `GeecsPvaGateway` from it with `--no-deps --no-build-isolation`. The clone's checked-out commit **is** the fleet pin: rollout = `git pull` there + restart PVs; rollback = `git checkout <rev>` + restarts. Unreachable share falls through to installed versions (never-bricks contract kept).
- `bootstrap.ps1`: `-WheelShare` → `-SourceShare`; installs **poetry-core** into the venv so restart-time source builds need no internet (`--no-build-isolation`).
- `DEPLOYMENT.md`: Rollout section rewritten (three constraints: PyPI deps frozen at bootstrap; machine-account share readability with the one-shot canary validation recipe; don't pull the clone mid-restart). Bootstrap section notes console sessions can use the share clone as `-Source` while SSH-driven bootstraps use a local GitHub clone (SSH sessions hold no share credentials).

This also kills the wheel design's worst residual (per #607 review finding 1: monorepo wheels embed unresolvable build-machine `file://` metadata + leak home paths) by never building wheels at all. ~40 LOC scripts, ~40 LOC docs, single concern.

## Tests

No Python runtime changes — suite unaffected (14 passed at base). Lint green.

## Hardware verification

**OWED, next canary session** (needs the share-clone UNC from `net use` + box access): set `GEECS_PVA_SOURCE` on the canary service, `:restart`, and confirm the `pull-on-restart: reinstalling from …` log line — which simultaneously validates machine-account share readability, the one open infrastructure question. The fall-through path (unset/unreachable source) is what the canary currently runs.

## Flagged for cheap veto

- Reinstall runs on **every** restart (not only when the clone changed) — simplicity over cleverness; costs a few seconds of pip per restart, and restarts are rollouts/reboots.
- The reinstall builds from a **live** share clone; the doc constraint "pull first, then restart" is procedural, not enforced.

### DIFF
--- a/GeecsPvaGateway/CHANGELOG.md
+++ b/GeecsPvaGateway/CHANGELOG.md
@@ -3,6 +3,19 @@
 All notable changes to this package will be documented here.
 Format follows [Keep a Changelog](https://keepachangelog.com/en/1.1.0/).
 
+## [0.3.0] — 2026-07-25
+
+### Changed
+
+- **Pull-on-restart installs from the lab's shared GEECS-Plugins clone instead
+  of a wheel drop** (owner decision: reuse the "Active Version" pattern GEECS
+  itself launches from). `launch.bat` reinstalls the three packages from
+  `GEECS_PVA_SOURCE` (`--no-deps --no-build-isolation`; poetry-core installed
+  at bootstrap so restarts need no internet); the clone's checked-out commit
+  is the fleet pin — rollout = `git pull` there + restart PVs, rollback =
+  `git checkout <rev>` + restarts. Replaces the never-deployed wheel/CURRENT
+  machinery; bootstrap's `-WheelShare` becomes `-SourceShare`.
+
 ## [0.2.1] — 2026-07-25
 
 ### Added
--- a/GeecsPvaGateway/CHANGELOG.md
+++ b/GeecsPvaGateway/CHANGELOG.md
@@ -9,7 +9,8 @@ Format follows [Keep a Changelog](https://keepachangelog.com/en/1.1.0/).
 
 - **Pull-on-restart installs from the lab's shared GEECS-Plugins clone instead
   of a wheel drop** (owner decision: reuse the "Active Version" pattern GEECS
-  itself launches from). `launch.bat` reinstalls the three packages from
+  itself launches from). `launch.bat` reinstalls the four intra-repo packages (incl. the
+  transitive `GEECS-Schemas`) from
   `GEECS_PVA_SOURCE` (`--no-deps --no-build-isolation`; poetry-core installed
   at bootstrap so restarts need no internet); the clone's checked-out commit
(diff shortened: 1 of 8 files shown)


skbarber:

The reviewer verified the pip semantics **empirically** (throwaway venv: same-version directory reinstalls work, downgrades work, read-only source trees build, fully-offline restart works with poetry-core preinstalled) before judging the design. Three findings:

**1 (Medium-High):** the reinstall set omitted **GEECS-Schemas** — a required transitive dep on the service's startup import path. It would freeze at bootstrap while its importer (GeecsCAGateway) rolled forward → a future rollout crash-loops the fleet, and the fix would have required re-bootstrapping every box (launch.bat is copied at bootstrap).
**Disposition: FIXED** (cec38fa7) — four-package reinstall, GEECS-Schemas first; comments/docs/CHANGELOG updated. Reviewer **confirmed**, including dependency ordering.

**2 (Low-Medium):** wheel-era text survived in five places, including an operator-visible Phoebus tooltip, contradicting the clone-source story.
**Disposition: FIXED** across two commits (cec38fa7 + a909ce9b): reviewer confirmed 4/5, caught the fifth (a bootstrap comment) on re-verify — now purged. The two remaining "wheel" mentions are legitimate (pip's wheel *cache*; the "no wheels" contrast in the rollout doc).

**3 (Low):** the one unquoted `%GEECS_PVA_SOURCE%` expansion inside the `if (…)` block — a share path containing `)` would abort the batch *before the server-launch line*, violating never-bricks.
**Disposition: FIXED** (cec38fa7) — quoted. Reviewer **confirmed** the brick scenario is closed and the log-line prefix the canary-validation recipe greps for is unchanged.

**Added mid-review (owner input, c1600ba3):** documented fallback for share access — running the service as the lab's shared domain account via `nssm ObjectName` (operator-typed password, never in scripts; coupled to no password rotation), with the clarification that the `USERPROFILE` override stays under any account (mapped drives are per-logon-session). Reviewer **checked it out**: correct NSSM syntax (SeServiceLogonRight granted automatically), technically correct session claims, honest caveats; noted the typed password lands in that box's console history (acceptable, operator can clear).

**Empirically verified non-findings:** `--upgrade --no-deps --no-build-isolation` semantics all behave as documented; `setlocal` removal correct; rename complete; hygiene clean.

PR #614, opened 2026-08-18
extras: live camera viewer for GEECS IMAQ streams

ag6520:
Standalone matplotlib app that subscribes to a camera device's TCP image stream, decodes each NI IMAQ "Flatten Image to String" payload via geecs_data_utils.io.decode_imaq_image_string, and displays the newest frame in a refreshing window.

Frames are produced on the device's TCP listener thread into a one-slot buffer and consumed by a FuncAnimation timer, so a camera that outruns the display drops intermediate frames rather than building a backlog; the drop count is shown in the title. Axes are labelled in sensor pixels so coordinates stay meaningful under --downsample, and ROI changes mid-run retarget the extent.

Lives in extras/ as an operational script: no package version bump or CHANGELOG entry, and extras/** is excluded from ruff/pydocstyle.

Verified against live hardware (Undulator): 56 cameras enumerated from the database and a 950x1000 uint8 frame decoded from UC_ALineEBeam3. The 16-bit uncompressed path and the tail-anchored no-name-anchor fallback are not yet exercised on a live camera.

### DIFF
--- a/extras/scripts/live_camera_viewer.py
+++ b/extras/scripts/live_camera_viewer.py
@@ -1,0 +1,410 @@
+"""Live viewer for a GEECS/LabVIEW IMAQ camera.
+
+Subscribes to a camera device's TCP image stream, decodes each incoming NI IMAQ
+"Flatten Image to String" payload, and displays the newest frame in a matplotlib
+window that refreshes in a loop.
+
+Usage
+-----
+    python extras/scripts/live_camera_viewer.py UC_ALineEBeam3
+    python extras/scripts/live_camera_viewer.py UC_ALineEBeam3 --experiment Undulator
+    python extras/scripts/live_camera_viewer.py --list-cameras
+
+    # pin the display scale and refresh slower
+    python extras/scripts/live_camera_viewer.py UC_VisaEBeam1 --scale fixed \
+        --vmin 0 --vmax 4095 --fps 5
+
+Interactive keys
+----------------
+    space   pause / resume the display (the subscription keeps running)
+    s       save the displayed frame as .npy + .png in --save-dir
+    q       quit
+
+Notes
+-----
+Frames arrive on the device's TCP listener thread and are handed to the GUI
+through a one-slot buffer: if the camera outruns the display, intermediate
+frames are dropped rather than queued, so what you see is always the newest
+frame. Dropped frames are reported in the window title.
+"""
+
+from __future__ import annotations
+
+import argparse
+import logging
+import sys
+import threading
+import time
+from collections import deque
+from datetime import datetime
+from pathlib import Path
+from typing import Optional
+
+import matplotlib.pyplot as plt
+import numpy as np
+from matplotlib.animation import FuncAnimation
+
+from geecs_python_api.controls.devices.camera import latest_image, on_image
+from geecs_python_api.controls.devices.geecs_device import GeecsDevice
+from geecs_python_api.controls.interface.geecs_database import (
+    GeecsDatabase,
+    load_config,
+)
+
+LISTENER_NAME = "live_viewer"
+
+
+class LatestFrame:
+    """One-slot, thread-safe hand-off of the newest frame to the GUI thread.
+
+    The producer (TCP listener thread) overwrites the slot on every frame; the
+    consumer (matplotlib timer) takes it. Overwriting an untaken frame counts as
+    a drop, which is the desired behaviour for a live view — never queue up a
+    backlog of stale frames.
+    """
+
+    def __init__(self) -> None:
+        self._lock = threading.Lock()
+        self._frame: Optional[np.ndarray] = None
+        self.received = 0
+        self.dropped = 0
+        self.last_arrival: Optional[float] = None
+
+    def put(self, image: np.ndarray) -> None:
+        """Store the newest frame, dropping any frame not yet displayed."""
+        with self._lock:
+            if self._frame is not None:
+                self.dropped += 1
+            self._frame = image
+            self.received += 1
+            self.last_arrival = time.monotonic()
+
+    def take(self) -> Optional[np.ndarray]:
+        """Return the pending frame and clear the slot, or None if empty."""
+        with self._lock:
+            frame, self._frame = self._frame, None
+            return frame
+
+
+def resolve_experiment(explicit: Optional[str]) -> str:
+    """Return the experiment name from the CLI flag or ~/.config config.ini."""
+    if explicit:
+        return explicit
+    config = load_config()
+    if config is None:
+        sys.exit(
+            "No ~/.config/geecs_python_api/config.ini found — pass --experiment "
+            "explicitly, or copy a known-good config onto this machine."
+        )
+    try:
+        return config.get("Experiment", "expt")
+    except Exception:
+        sys.exit(
+            "config.ini has no [Experiment] expt entry — pass --experiment explicitly."
+        )
+
+
+def list_cameras(exp_info: dict) -> None:
+    """Print every device in the experiment that exposes an ``image`` variable."""
+    devices: dict = exp_info.get("devices", {})
+    cameras = sorted(
+        name for name, variables in devices.items() if "image" in (variables or {})
+    )
+    if not cameras:
+        print("No devices with an 'image' variable found in this experiment.")
+        return
+    print(f"{len(cameras)} camera-like device(s):")
+    for name in cameras:
+        print(f"  {name}")
+
+
+def compute_clim(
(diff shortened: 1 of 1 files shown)


skbarber:
Thanks for the PR — and welcome! For a first pass at the repo this is a genuinely well-put-together script: the latest-wins buffer has the right semantics, the PR body is honest about what was and wasn't exercised on hardware, and you verified against a live camera. The feedback below is substantial, but most of it is about *where this sits in the architecture* rather than the code itself — which is exactly the kind of thing you can only learn by making a first PR, so this did its job.

## The big one: this is built on the layer we're deleting

The script's three imports all come from **GEECS-PythonAPI**, which the root `CLAUDE.md` marks *deprecated, slated for deletion — never build on it, do not adopt it in new call sites*. When that package goes, this tool dies with it — and because `extras/` is exempt from lint/tests/CI, nothing would even flag the breakage.

The reason the rule exists is that the problem this script solves — "stream an image from a camera" — already has a sanctioned answer: **GeecsPvaGateway** runs on every camera server and serves each camera's frames as standard NTNDArray pvAccess PVs. The docs page that landed last week covers exactly this case, including a section literally titled *"Why not just read the camera's TCP stream yourself?"*:

- [docs/geecs_gateway/image_pvs.md](https://github.com/GEECS-BELLA/GEECS-Plugins/blob/dev/docs/geecs_gateway/image_pvs.md)

The short version of that page's argument, applied to this PR:

1. **Fan-out cost lands on the camera.** GEECS TCP push is per-connection — every viewer instance makes the LabVIEW device flatten and send each frame *again*, during scans. The gateway subscribes once on loopback; PVA fans out to any number of clients for free.
2. **The IMAQ wire format is a hazard** (latin-1 mangling, wrapper variants that have drifted before). The gateway's decoder fixes it in one place; every direct-TCP consumer re-inherits all of it. Your own `--poll` fallback and the "aliased image variable" diagnostic are symptoms of fighting that hazard again.
3. **The equivalent client is tiny.** Against the gateway, this viewer is a `p4p` monitor + your matplotlib loop — no DB connection, no `GeecsDevice`, no config.ini, and the `LatestFrame` class disappears because latest-wins delivery is a gateway guarantee. Something like:

```python
from p4p.client.thread import Context

ctx = Context("pva")
sub = ctx.monitor("undulator:uc_alineebeam3:image", buffer.put)  # frames arrive decoded
```

Most of the 410 lines here exist to work around not using that layer.

## Real bugs found in review

These were verified against the `master` sources the script imports:

- **The friendly unknown-device path is unreachable** (`main`, ~L235): `GeecsDevice.__init__` raises `GeecsDeviceInstantiationError` for a bad name *or* an offline device, so a typo'd camera name — the most common operator error — produces a raw traceback, never the intended "run with `--list-cameras`" message.
- **Import-time hang hazard** (~L49): importing `geecs_python_api.controls.interface` can open a tkinter file dialog at import time (documented hazard in root `CLAUDE.md`). Over ssh with a misconfigured/unmounted config path, even `--help` blocks forever with no output.
- **Off-network behavior** (~L228): `collect_exp_info()` is unguarded — off the lab network it blocks ~75 s on the MySQL timeout then dies with a raw traceback. And the "pass `--experiment` explicitly" advice is a dead end: without config.ini the DB credentials are missing too, so the suggested workaround raises `AttributeError`.
- **`--poll` can't do what its help text says** (~L196): `latest_image()` reads the same literal `state["image"]` key that push mode uses, so an aliased camera fails identically in both modes — the flag's documented use case is unreachable.
- **Decode failures are invisible or fatal** (~L293): in push mode a decoder `ValueError` is swallowed inside the TCP dispatch (logged, not propagated), so the 10 s timeout diagnostic then points you at aliasing and suggests calling the very decoder that's failing; in poll mode the same error escapes uncaught and freezes/kills the animation. Relevant since the PR notes the 16-bit path is unexercised.
- **Pause loses the pending frame** (~L361): `update()` drains the buffer / advances `last_shot` *before* checking `paused`, so on a low-rep camera the shot that arrived during a pause is silently dropped on resume.
- **Title shot number is off-by-one-or-more in push mode** (~L397): `camera.py`'s docstring says the listener fires *before* the message is parsed into `state`, so the displayed shot label doesn't match the displayed pixels — worst exactly when frames are being dropped.
- **`--scale fixed` isn't fixed** (~L128): omit one of `--vmin`/`--vmax` and the missing bound autoscales per frame — the flicker the mode exists to prevent; inverted bounds are silently accepted as `(lo, lo+1)`.
- **Invariant nit** (~L147): `save_dir.mkdir(parents=True, ...)` on a user-controlled path. This repo has a hard rule (with production incident history — see "Cross-package invariants" in root `CLAUDE.md`) against `parents=True` anywhere that could traverse a `scans/ScanNNN/` path. Dropping `parents=True` costs nothing here.

Smaller polish items (per-frame `min/max/mean` full-resolution passes just for the title, `blit=False` full-figure redraws, the `s` hotkey colliding with matplotlib's built-in save dialog, `build_parser()` constructed twice, `state["clim"]` mirroring what `handle.get_clim()` already knows) all become moot if the script shrinks onto the gateway, so I won't itemize fixes.

## Recommendation

Let's not merge this one — not because the code is sloppy (it isn't, for a script), but because it adopts the layer we're actively removing and duplicates what the PVA gateway already guarantees. Two things worth keeping from it, though:

1. **The UX ideas are good** — latest-wins display with a drop counter, pause/save keys, sensor-pixel axes under downsampling. If you feel like a round two: the same viewer rebuilt as a `p4p` monitor against the gateway (branched off `dev`) would be a fraction of the size, survive the PythonAPI deletion, and cost the camera nothing. The docs page above has everything you need, and I'm happy to point you at PV names for live cameras.
2. **You found a real gap**: there's no casual "point a window at a camera" tool in the repo today — Phoebus covers it in the control room, but a pip-installable Python one-liner has a place. That's a conversation worth having before writing round two, which is the general lesson here: a couple of minutes pinning down *what's desired and what already exists* changes which 400 lines get written.

*(Review assisted by Claude Code — findings were verified against the `master` sources before posting.)*

skbarber:
Some extra context from me, because the review above leans hard on "this package is being deleted" and the real story is a bit more nuanced — and the nuance is actually the interesting lesson.

**Where geecs-python-api is really headed.** Long term, I want the PVA/CA gateways to be the *primary* way anything talks to GEECS controls. But a thin Python wrapper still has a place: it's a useful backup path alongside the EPICS layer, and — importantly — I haven't deployed the gateway stack at the other facilities yet, so today `GeecsDevice` is genuinely the only way Python reaches controls there. My real problem with the package isn't that it exists; it's that `GeecsDevice` does way too much. It carries its own transport (duplicating what the gateway's access layer now owns), and the way it talks to the GEECS database is clunky. The likely future is to keep geecs-python-api but shrink it to a truly thin wrapper — get, set, subscribe, and nothing else — with transport and DB access delegated to the gateway package instead of reimplemented.

So reaching for `GeecsDevice` wasn't a crazy instinct. The issue is that building polished tooling *on top of the current fat version* invests in exactly the shape we're trying to shed — and when the wrapper slims down, everything stacked on its current surface area has to move.

**Which brings me to the question I most want this PR to plant: where should a thing like this live?** That's the two-minute question that changes which 400 lines get written. For "stream an image from a camera," it unpacks into:

- **Which contract am I building against?** The gateway's PV contract (names + NTNDArray) is designed to be the stable thing — a client written against it survives every refactor underneath. The raw TCP stream and today's `GeecsDevice` surface are both things we intend to change.
- **Which layer owns the hard part?** Wire-format decode belongs in one place (the gateway), not re-inherited by each script. Same for camera discovery — "has a variable literally named `image`" is a heuristic; the DB's variable-type info is the real answer, and a shared layer should own that lookup.
- **Who pays at runtime?** A direct TCP subscriber makes the camera device do per-connection work during scans; a gateway client costs it nothing.
- **What maintains it?** `extras/` is a fine scratchpad, but it's exempt from lint, tests, and CI — a tool operators come to rely on needs a versioned home where breakage is caught.

None of this needs deep repo knowledge up front — it mostly falls out of skimming the root `CLAUDE.md` and the gateway docs page before writing, and asking "does the repo already have an opinion about this problem?" It usually does.

**On this PR specifically:** my stance is we shouldn't merge it, for the reasons in the review — it adopts the surface we're shrinking and duplicates gateway guarantees. But the exercise did its job, and the follow-ups are real if you want them: the p4p-based viewer against the gateway (where it's deployed), or — if the plumbing interests you more — the thin-wrapper refactor of geecs-python-api is a design conversation I'd genuinely welcome thoughts on.

PR #620, opened 2026-08-20
GeecsBluesky 0.51.0: scan.log = complete per-scan record (root capture + pre-claim buffer)

skbarber:
## What + why

The 2026-08-19 console test scans showed three surfaces telling three different stories: the terminal (complete), `scan.log` (a four-namespace allowlist attached only after the folder claim — missing the 13-second pre-claim window and every foreign namespace), and the GUI ticker (a progress event stream, fine as-is). The terminal being the only complete record is an accident that dies with the queueserver migration, where the engine's "terminal" becomes a machine-global systemd journal.

This PR makes `scan.log` the complete per-scan record:

- **Root-logger capture**: the per-scan file handler attaches at the root, so `bluesky` RE state changes, `ophyd_async` connect-failure tracebacks, and `geecs_data_utils` folder/export lines land in the file — terminal parity. Root level lowered to INFO for the scan and restored; `CAPTURE_LOGGER_NAMES` deleted (no external importers; it had already gone stale — `tiled_export` was never in it).
- **Pre-claim buffer**: `begin_pre_scan_capture()` starts a bounded buffer at submission (`reinitialize` / `session.run`) and `scan_log()` flushes it into the file at the claim, so the file opens with device connects, telemetry drops, and the claim itself. A buffer with no scan.log home (refused submission, `save_data=False`, pre-claim failure) is discarded on every path — never leaked into a later scan's file.

## Judgment-call flags (cheap to veto)

1. **`QUIET_LOGGER_PREFIXES = ("httpx", "mysql.connector")`** — live Scan001 showed ~15 lines/scan of Tiled HTTP request and MySQL auth-plugin chatter. Filtered below WARNING from the scan.log capture only (terminal untouched, WARNING+ still lands). If you want strict verbatim capture, deleting the filter is one line.
2. **Planning doc rides along** (separate commit): `Planning/cutover_strategy/02_queueserver_migration.md` — the 2026-08-19 scope doc (previously untracked) + 2026-08-20 amendments (scan.log worker story, `reinitialize` disposition).

## Per-concern LOC

- scan.log unification (GeecsBluesky): ~360 insertions / 58 deletions across 7 files (scan_log.py rewrite + 2 attach points + tests + version/CHANGELOG/CLAUDE.md)
- planning doc: +195 lines, docs only

## Tests

- `test_scan_log.py`: 11 pass (5 new: foreign-namespace capture, buffer flush ordering + stamp, no-claim discard, superseded-buffer reset, transport-chatter filter)
[...]

PR #622, opened 2026-08-20
Relocate the Xopt/evaluator stack to geecs_bluesky.optimization

skbarber:
## What + why

The queueserver migration doc's one **Relocate** item (`Planning/cutover_strategy/02_queueserver_migration.md`, maintainer decision 5 + Scope): the config-driven Xopt/evaluator stack moves whole from `geecs_scanner.optimization` to `geecs_bluesky/optimization/`. Rationale: the queueserver worker preamble will invoke the optimization loader, and the worker imports geecs_bluesky — the stack living there makes `pip install geecs-bluesky[ca,optimize]` the complete worker environment. It also unblocks the M6 whole-package deletion of GEECS-Scanner-GUI, whose optimization module was its last live consumer on dev. Maintainer approved the destination in-session (2026-08-20).

**Relocation, not rewrite** — the only behavioral code change is the one legacy wart the move forced:

- `SessionOptimizationBridge`'s best-effort device-name canonicalization now queries the gateway's `GeecsDb` (`get_experiment_devices`, experiment from `config.ini` via `GeecsPathsConfig`) through a patchable module seam (`_load_experiment_device_names`), replacing the legacy `geecs_scanner.engine.DatabaseDictLookup` — the stack's last legacy-engine (transitively geecs-python-api-adjacent) thread. Same fallback semantics, re-pinned by the adapted seam tests.

## Per-concern LOC

- **The move**: ~4,100 LOC of package code + ~2,900 LOC of tests, 36 files as git renames (>90% similarity on all but the import-flip files). Internal imports flipped `geecs_scanner.optimization` → `geecs_bluesky.optimization`.
- **Legacy models travel** (+ shims): `engine/models/actions.py` + `save_devices.py` (pure pydantic, 544 LOC) move as `optimization/_legacy_models_*.py`; `geecs_scanner.engine.models` re-exports them from the new home (correct dependency direction — Scanner-GUI already depends on GeecsBluesky), so the legacy engine and its 48 tests keep passing.
- **DB seam replacement**: ~40 LOC in `session_bridge.py` + adapted tests.
- **Packaging**: GeecsBluesky `optimize` extra gains `gest-api` (imported directly, previously undeclared — rode in transitively), `scananalysis`, `imageanalysis`. Console's `optimization` extra drops `geecs-scanner-gui` and mirrors those deps directly (Poetry cannot express a second extras-variant of the existing geecs-bluesky path dep — mirror noted in both pyprojects). Availability probe now gates on `find_spec("xopt")`.
- **Docs**: root/GeecsBluesky/Console/Scanner-GUI CLAUDE.md truth-ups; three CHANGELOGs; versions GeecsBluesky 0.52.0, Console 0.20.0 (minor — extra semantics changed), Scanner-GUI 0.36.0.

## Judgment-call flags (cheap to veto)

1. **New optional dependency edge**: GeecsBluesky →(optimize extra) ScanAnalysis. No cycle today (ScanAnalysis never imports geecs_bluesky); a future cycle risk exists if ScanAnalysis ever consumes `geecs_bluesky/analysis` contracts — accepted because this stack is explicitly transitional (bluesky-adaptive redesign deletes it).
2. **`_legacy_models_*` live inside `optimization/`** rather than geecs_schemas — they are the legacy dialect, kept private with their one consumer, dying with the stack, not promoted into the long-lived vocabulary package.
3. **Scanner-GUI keeps its (now-unused) hard `xopt` dep** — minimal churn on a package awaiting whole deletion; its `run_control` lazy import still resolves in that env.
4. **CI skip guard**: `tests/optimization/conftest.py` sets `collect_ignore_glob` when `xopt`/`scan_analysis` are absent — CI installs GeecsBluesky with no extras, so the directory skips there (note: CI never ran these tests in their old home either; local runs with the extra cover them).

## Tests

- GeecsBluesky: **764 passed** (644 prior + 120 moved), incl. the AST dependency-direction test, which now also guards the relocated stack against geecs_scanner imports.
- GEECS-Console: **787 passed, 2 skipped** (the availability-probe premise skips).
- GEECS-Scanner-GUI: **48 passed** against the model shims.
- Repo-wide `poetry run pre-commit run --all-files`: green. (`check.sh --all`'s bare-pre-commit jupyter failure is environmental/pre-existing — side-task chip filed.)

## Hardware verification

**OWED: one live optimization scan through the console** (any optimizer config, e.g. TopViewMax) to confirm the loader path end-to-end from the new home — expected: identical behavior to the pre-move stack (loader builds, device_requirements auto-provision with DB-canonicalized names, iterations run, xopt_dump written). The move is import-path-only on that path except the DB-seam swap, which is best-effort with a verbatim fallback, so the risk is a silently-skipped canonicalization (visible as a wrong-case device failing to connect — the original 2026-07-06 incident signature).


skbarber:

Reviewed with renames skimmed and non-rename hunks read closely, against the repo + the sibling GEECS-Plugins-Configs checkout. Six findings; redundancy and placement lenses cleared (GeecsDb.get_experiment_devices is reuse, not duplication; the AST dependency guard genuinely covers the relocated stack; the `_legacy_models_*` placement holds up).

**1. HIGH — Every deployed optimizer config still imports the old module path.** `EvaluatorConfig.module` is a dynamic import path stored in *config data*; all 11 deployed Undulator optimizer YAMLs (including TopViewMax, named in this PR's own verification plan) say `module: geecs_scanner.optimization.evaluators.*`, and nothing in the chain rewrote it. Post-merge, every console optimization submission would die at loader time with `ModuleNotFoundError` — and the queueserver worker preamble hits the same wall.

**2. MEDIUM — The new DB seam's body had zero coverage**: both test fixtures monkeypatched `_load_experiment_device_names` away wholesale, and the fake reimplemented the tested transformation — the casefolding, experiment resolution, None-experiment return, and key-iteration were production code no test executed.

**3. LOW — Availability probe gated on `xopt` alone** while the stack also needs `scan_analysis` at import — an env with one but not the other would advertise optimization available and die mid-scan instead of the clean refusal.

**4. LOW — "Same best-effort semantics" overstated**: the canonicalization map's membership changed (legacy: `default_experiment` variable-table keying, no enabled filter → new: enabled `expt_device` set).

**5. LOW — Five missed text truth-ups** (GeecsBluesky CLAUDE.md intro; a stale test comment block asserting a conftest patch that doesn't exist; a dangling cross-ref in geecs_data_utils/native_files.py; the moved package's docstring; a See-Also).

**6. INFO — A latent `scans/ScanNNN` `mkdir(parents=True)` now resides in geecs_bluesky** (`base_optimizer.py`): pre-existing verbatim, unreachable on every live path, but dormant inside the package whose invariant names `claim_scan_number` as the one creator — and the queueserver worker will grow callers here.

skbarber:
## Disposition (all findings re-verified by the same reviewer against `d30eebd2`, including a live run of the new tests)

| # | Finding | Disposition |
|---|---|---|
| 1 | Deployed configs' legacy module paths | **FIXED** (`d30eebd2`, option a): `EvaluatorConfig` field validator rewrites the `geecs_scanner.optimization.` prefix at the one validation seam every load path crosses — reviewer verified by construction that `BaseOptimizerConfig.model_validate` runs it before `importlib.import_module`. One INFO line per rewrite; deployed YAMLs, xopt dumps, and seed files all stay loadable with **no configs-repo flag day** (update at leisure). Pinned by `TestEvaluatorModulePathRewrite`; schema example + generated reference updated (GEECS-Schemas 0.9.1). Reviewer: CONFIRMED FIXED. |
| 2 | Untested seam body | **FIXED**: `TestLoadExperimentDeviceNames` exercises the real function via an import-time alias (dodges the autouse hermetic patch), fakes at the true sources, pinning key-iteration, experiment passthrough, and the None-experiment zero-DB-call return. Reviewer ran them: pass. CONFIRMED FIXED. |
| 3 | Probe gap | **FIXED**: probe requires `xopt` AND `scan_analysis`, matching the test-dir guard. CONFIRMED FIXED. |
| 4 | Semantics claim | **FIXED**: CHANGELOG now states same *fallback* semantics with the membership change spelled out (enabled `expt_device` set vs the legacy unfiltered variable-table keying). Reviewer verified against both queries. CONFIRMED FIXED. |
| 5 | Five text truth-ups | **FIXED**, all five verified in-tree (incl. GEECS-Data-Utils 0.13.6 docs-only bump). CONFIRMED FIXED. |
| 6 | Dormant mkdir | **FIXED** (breadcrumb naming the invariant + claim_scan_number; correctly indented after an initial mis-indent the suite caught). CONFIRMED FIXED. |

**OWED (unchanged, now dual-purpose):** the live console optimization scan also field-checks finding 1's rewrite — running a legacy-prefixed config (e.g. TopViewMax) should show the INFO rewrite line in the log.

Post-fix test state: GeecsBluesky **768**, Console **787**, Schemas+Data-Utils **201** — all green; scoped lint green.

skbarber:
**OWED hardware verification: CLOSED** (2026-08-20, post-merge, lab network).

One optimization scan through the console's real submission path (`make_bluesky_submitter` → availability-gated loader → delegated ScanRequest), config `bax_alignment_simulation` (synthetic objective, real hardware: U_S1V corrector current + EMQ Current_Limit.Ch1 soft limit), strict acquisition, HTU-LaserOFF profile, 3 iterations × 3 shots → **Scan004, terminal state DONE**.

- The finding-1 rewrite fired exactly as designed on the legacy-prefixed deployed config: `Evaluator module path 'geecs_scanner.optimization.evaluators.beam_position_evaluator_simulation' rewritten to 'geecs_bluesky.optimization...'` — and (via #620's pre-claim buffer) the line is recorded **in the scan folder's scan.log**, not just the terminal.
- Loader built the stack from the new home; BAX generator ran; per-shot `Observable:x_CoM` evaluator lines in scan.log; 13 lifecycle/step events emitted.
- Both variables restored to their pre-scan values afterward via `Submitter.move_variable`, confirmed by readback.

PR #623, opened 2026-08-20
scripts: prefer root poetry env's pre-commit; skip jupyter hook loudly on bare fallback

skbarber:
## What + why

`check.sh` and `commit.sh` picked a bare `pre-commit` on PATH over the poetry env's. The pre-commit config's `language: system` hook (`jupyter-notebook-clear-output`) resolves its `jupyter` executable from the *invoking* environment — so on machines where a system-level pre-commit is first on PATH (this one has `/Library/Frameworks/Python.framework/.../pre-commit` with no `jupyter`), every `.ipynb` in the run set fails with **`Executable \`jupyter\` not found`** and `check.sh --all` reports a phantom `FAILED: lint` while `poetry run pre-commit run --all-files` passes. This produced phantom lint failures during PR #620 and the xopt-relocation PR (2026-08-20).

Two layers, same in both scripts:

1. **Prefer the root poetry env's pre-commit** when it exists and actually runs (`command -v poetry` && `poetry env info --path` && `poetry run pre-commit --version` — self-verifying, so a package env without pre-commit or a poetry-less machine falls through cleanly; `poetry env info --path` is used because it never creates an env as a side effect).
2. **On the bare-PATH fallback without `jupyter` findable**, export `SKIP=jupyter-notebook-clear-output` (appending to any caller-set `SKIP`) with a printed note, instead of failing on every notebook. This matters because worktrees have no poetry env of their own (envs are keyed by directory path), so agent sessions in `.claude/worktrees/` hit the fallback; CI's pre-commit workflow still enforces the hook. In `commit.sh` the exported `SKIP` also covers the commit-time hook run, which would otherwise hard-fail the same way after the pre-run.

## Per-concern LOC

Single concern (lint-runner selection): `scripts/check.sh` +26/−6 lines, `scripts/commit.sh` +21/−6 (mostly comments explaining the mechanism).

**Judgment-call additions** (cheap to veto):
- The `SKIP` fallback layer (layer 2) — the memory note offered it as an alternative; I included it *in addition* because worktrees never have a root env, so preference alone leaves the phantom failure alive exactly where agent sessions run.
- A hard error with remediation when neither runner exists (previously the poetry fallback would just die with poetry's own error).

## Tests

No package code changed; the suite runner table in `check.sh` is untouched, so no test suites are affected. Verification is of the lint paths themselves, on the machine that reproduces the bug (bare pre-commit present, no `jupyter` on PATH):

- **Bug repro (pre-fix behavior):** `pre-commit run jupyter-notebook-clear-output --all-files` with the bare runner → exit 1, `Executable \`jupyter\` not found`.
- **Preferred path:** full `--all-files` run through the root poetry env's pre-commit (3.8.0) → **17/17 hooks green**, including the jupyter hook actually executing (Passed, not skipped); `jupyter` resolves inside that env.
- **Fallback path:** `./scripts/check.sh --lint` in a poetry-env-less worktree → picks bare pre-commit, prints the skip note, `== OK (lint)`; bare `--all-files` with the SKIP → all other hooks Passed, jupyter hook Skipped.
- **commit.sh:** this PR's own commit was made with the new `commit.sh` on the fallback path — succeeded first try, skip note printed, commit-time hook run covered by the exported SKIP.
- Guard negative cases: `poetry env info --path` in this worktree exits non-zero (no env) → falls through as designed.

## Hardware verification

Not applicable — repo tooling only; nothing touches scan execution or devices.

### DIFF
--- a/scripts/check.sh
+++ b/scripts/check.sh
@@ -151,11 +151,32 @@ if [ "$DRY_RUN" -eq 1 ]; then
     exit 0
 fi
 
-# --- Lint (pre-commit; same fallback as scripts/commit.sh) ---------------------
-if command -v pre-commit >/dev/null 2>&1; then
-    PC=(pre-commit)
-else
+# --- Lint (pre-commit; same runner preference as scripts/commit.sh) ------------
+# Prefer the root Poetry env's pre-commit over one on PATH. The config's
+# `language: system` hooks (jupyter-notebook-clear-output) resolve their
+# executable from the invoking environment: the root env has jupyter, a
+# system pre-commit typically does not — which turns into a phantom
+# "Executable `jupyter` not found" lint failure. The guard is self-verifying
+# (env exists AND pre-commit runs in it), so environments without a usable
+# root env fall back to a PATH pre-commit as before.
+if command -v poetry >/dev/null 2>&1 \
+    && poetry env info --path >/dev/null 2>&1 \
+    && poetry run pre-commit --version >/dev/null 2>&1; then
     PC=(poetry run pre-commit)
+elif command -v pre-commit >/dev/null 2>&1; then
+    PC=(pre-commit)
+    # A PATH pre-commit runs `language: system` hooks in its own environment.
+    # If that environment cannot find jupyter, the notebook-clearing hook dies
+    # with "Executable `jupyter` not found" on every .ipynb — a phantom
+    # failure unrelated to the change. Skip that one hook loudly instead;
+    # CI's pre-commit workflow still enforces it.
+    if ! command -v jupyter >/dev/null 2>&1; then
+        export SKIP="${SKIP:+$SKIP,}jupyter-notebook-clear-output"
+        echo "   note : no 'jupyter' on PATH — skipping jupyter-notebook-clear-output (CI still runs it)"
+    fi
+else
+    echo "check.sh: no pre-commit found (neither the root poetry env nor PATH) — run 'poetry install' at the repo root" >&2
+    exit 2
 fi
 
 LINT_OK=1
--- a/scripts/check.sh
+++ b/scripts/check.sh
@@ -156,34 +156,45 @@ fi
 # `language: system` hooks (jupyter-notebook-clear-output) resolve their
 # executable from the invoking environment: the root env has jupyter, a
 # system pre-commit typically does not — which turns into a phantom
-# "Executable `jupyter` not found" lint failure. The guard is self-verifying
-# (env exists AND pre-commit runs in it), so environments without a usable
-# root env fall back to a PATH pre-commit as before.
-if command -v poetry >/dev/null 2>&1 \
-    && poetry env info --path >/dev/null 2>&1 \
-    && poetry run pre-commit --version >/dev/null 2>&1; then
-    PC=(poetry run pre-commit)
-elif command -v pre-commit >/dev/null 2>&1; then
-    PC=(pre-commit)
-    # A PATH pre-commit runs `language: system` hooks in its own environment.
-    # If that environment cannot find jupyter, the notebook-clearing hook dies
-    # with "Executable `jupyter` not found" on every .ipynb — a phantom
-    # failure unrelated to the change. Skip that one hook loudly instead;
-    # CI's pre-commit workflow still enforces it.
-    if ! command -v jupyter >/dev/null 2>&1; then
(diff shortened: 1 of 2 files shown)


skbarber:

Reviewer brief: refute the diff via correctness / redundancy / placement, whole-repo context, report only findings surviving self-refutation.

### Finding 1 — HIGH (correctness): the "self-verifying" guard mis-verifies — `poetry run` falls through to caller PATH for binaries the env lacks
`poetry run pre-commit --version` succeeds even when the env has no pre-commit (verified empirically: `GeecsCAGateway/.venv` has neither pre-commit nor jupyter, yet the probe returned `pre-commit 4.5.1` via PATH fall-through). Combined with commit.sh's unpinned cwd, invoking `../scripts/commit.sh` from a package dir with an in-project venv selected the poetry branch with the bare runner underneath and no SKIP — reintroducing the exact phantom failure this PR kills, in a place the repo's own release flow sends people (`cd <pkg> && poetry version ...`). Same mis-verification at the root for a `--without dev` env.

**Disposition: FIXED** in 38e28b48 — guard now requires the env's own `bin/pre-commit` (`[ -x "$root_env/bin/pre-commit" ]`), and the jupyter check is hoisted so the SKIP applies to whichever runner actually won (poetry branch accepts jupyter from env bin OR caller PATH, matching `poetry run`'s PATH composition). **Reviewer re-verified: CONFIRMED FIXED** — new predicate rejects the package env in the same shell state where the old probe still passes; `--without dev` corner now degrades to bare+SKIP instead of a phantom failure.

### Finding 2 — MEDIUM (correctness / internal contradiction): commit.sh never pinned its cwd
The "same runner preference" cross-comments asserted an equivalence the code didn't have: check.sh cd's to the repo root, commit.sh didn't, so its poetry branch resolved whatever env the caller's cwd implied — and (pre-existing, latent) the re-staging loop silently re-staged nothing from a subdir, because `git diff --cached --name-only` emits root-relative paths.

**Disposition: FIXED** in 38e28b48 — commit.sh now cd's to the repo root before the staged-file capture; the comment documents the one behavior change (pathspec args to `git commit` are interpreted from the repo root). Verified live: the fix commit itself was made by invoking `../scripts/commit.sh` from `docs/`. **Reviewer re-verified: CONFIRMED FIXED** (including the "nothing staged" path from a subdir).

### Finding 3 — MEDIUM-LOW (correctness regression): check.sh hard-exited before modes that never lint
The new `exit 2` ran unconditionally before the explicit-packages / nothing-to-lint branches, so a machine with no pre-commit anywhere (e.g. fresh Linux setup with only a package env) failed `./scripts/check.sh <pkg>` at a lint runner that mode never invokes — the old code only *assigned* the fallback and never executed it there.

**Disposition: FIXED** in 38e28b48 — selection is now `select_lint_runner()`, called only from the two branches that actually lint. **Reviewer re-verified: CONFIRMED FIXED** — `env PATH="/usr/bin:/bin" ./scripts/check.sh GeecsBluesky` proceeds to the suite stage (failing there honestly with the /env-doctor pointer); `--lint` on the same PATH exits 2 with remediation; all-deletions changesets never probe.

### Note — LOW (redundancy): ~20-line selection block duplicated across the two scripts; a sourced `scripts/lib/` helper is the natural home
**Disposition: WAIVED** — borderline at exactly two consumers; after the cwd pin the two copies are semantically identical (the divergence that motivated the note is gone), and the remaining differences (note to stdout vs stderr, exit 2 vs 1) follow each script's own conventions. Reviewer had no objection to the waiver. Revisit if a third consumer appears.

### Reviewer non-findings (attempted and refuted)
`${SKIP:+$SKIP,}` safe under `set -euo pipefail`; `pre-commit run` locates the git toplevel from subdirs; no other repo script runs pre-commit (no duplication); no `SKIP` consumers elsewhere, so the export leaking into the pytest phase is harmless; `.github/workflows/pre-commit.yml` (pip-installed `pre-commit nbconvert`) keeps enforcing the jupyter hook, so the "CI still runs it" claim is accurate; public-repo hygiene clean; docs/skills (env-doctor, /check, getting_started) consistent with the new preference. Residual note: on Windows/Git-Bash a poetry venv uses `Scripts/` not `bin/`, so the poetry branch never wins there — it degrades to the PATH-pre-commit+SKIP fallback rather than misbehaving.

**Reviewer bottom line after re-verification:** all three findings closed; fallback and preferred paths both re-verified on the reproducing machine; no new issues introduced by the fixes.

PR #624, opened 2026-08-20
GEECS-Console 0.20.1: R6 log tail dedupes consecutive narration lines

skbarber:
## What + why

Field report from the 2026-08-19 test scans (Scan004): the R6 log tail showed **"scan running" twice** and the **final step line twice**. Root causes, both narration-side:

- The bridge deliberately re-emits RUNNING once the scan number is claimed (that's how the "Scan NNN" label works) — the narrator rendered both identically.
- The engine legitimately re-emits the final step event; identical rendered text, two lines.

Fix, confined to `events_adapter.py`: the narration path now suppresses **exact consecutive repeats** (a `_narrate` helper all log-tail emissions route through), and the claimed-number RUNNING re-emission renders as `scan running (Scan004)` — new information instead of a duplicate. **Data signals never dedupe** — state pill, progress bar, scan-number label see every event exactly as before; only the human-readable tail changed. Scan004's ticker becomes: `scan initializing` / `scan running` / `scan running (Scan004)` / one line per shot / `scan done`.

~30 LOC + 4 pinning tests (dedupe, number narration, data-signals-not-deduped, alternating-lines-not-suppressed). Console CLAUDE.md R6 bullet updated in the same PR (the narration contract lives there).

## Tests

GEECS-Console: **791 passed, 2 skipped** (787 + 4 new), offscreen.

## Hardware verification

Not required — no engine or device surface touched (pure GUI narration); the event stream itself is unchanged and covered by the adapter's hermetic fakes. The next routine console scan will show the deduped tail.

### DIFF
--- a/GEECS-Console/CHANGELOG.md
+++ b/GEECS-Console/CHANGELOG.md
@@ -9,12 +9,17 @@ semantic.
 ### Fixed
 
 - R6 log tail no longer prints duplicate lines (field report, 2026-08-19
-  Scan004: "scan running" twice and the final step line twice): the
-  events adapter's narration now suppresses exact consecutive repeats,
-  and the bridge's deliberate RUNNING re-emission (once the scan number
-  is claimed) narrates as new information — "scan running (Scan004)".
-  Data signals (state pill, progress bar, scan number) are never
-  deduped, only the human-readable tail.
+  Scan004: "scan running" twice and the final step line twice).
+  Consecutive-duplicate suppression lives at the tail's one convergence
+  point — `NowPanelController.append_log` — so it measures against the
+  actually visible tail (the adapter's narration AND the window's direct
+  status lines both flow through it; PR #624 review finding 1), and the
+  cache re-arms when a new scan announces its totals so scan N's last
+  line can never swallow scan N+1's first (finding 3). Separately, every
+  post-claim lifecycle line now carries the scan number — "scan running
+  (Scan004)", "scan done (Scan004)" — which renders the bridge's
+  deliberate RUNNING re-emission as new information. Data signals (state
+  pill, progress bar, scan number) are never deduped.
 
 ## [0.20.0] - 2026-08-20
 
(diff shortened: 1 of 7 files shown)


skbarber:

Three findings, all dispositioned below; thread-safety, redundancy, data-signal purity, and test strength were attacked and cleared (the reviewer verified the new tests fail against both the pre-PR code and a naive dedupe-everything implementation).

**1. Medium (correctness + placement) — dedupe at the wrong layer.** The tail has two writers: the adapter's narration AND ~10 direct `MainWindow.append_log` sites ("stop requested", "operator: continue", …) that bypassed the adapter-side cache, so "consecutive" was measured against the wrong sequence. Concrete live scenario: a re-asked operator question after an interleaved "operator: continue" would be falsely suppressed (while its modal still pops), AND the two "operator: continue" answers would render as exactly the consecutive duplicate this PR fixes. Better home: the tail's convergence point.

**2. Low (docs/contract) — the (ScanNNN) suffix actually applies to every post-claim lifecycle line** (the bridge stamps the number on each emission), but every document described it as the RUNNING re-emission only, and no test pinned a non-RUNNING state.

**3. Low (robustness) — cross-scan cache persistence** was safe only via bridge invariants two packages away (the finally-emitted terminal line); a bridge change could silently swallow the next scan's first line.

## Disposition (all re-verified by the same reviewer against `9c4d86d8`, who independently reran the suites: 795 passed, 2 skipped reproduced)

| # | Disposition |
|---|---|
| 1 | **FIXED** — cache moved to `NowPanelController.append_log` (verified as the tail's sole `appendPlainText` point); adapter reverted to unconditional emission and *pinned* not to dedupe; the false-suppression scenario pinned verbatim in `test_interleaved_writers_measure_against_the_tail` (fails against the pre-fix implementation); direct-writer duplicates now covered too. CONFIRMED FIXED. |
| 2 | **FIXED** — behavior kept wide, contract text widened in both homes (CLAUDE.md R6 + CHANGELOG), `scan done (Scan004)` pinned so a doc-motivated narrowing would fail a test. CONFIRMED FIXED. |
| 3 | **FIXED** — cache re-arms in `set_totals` (the same new-scan re-arm site as the beep counter), pinned. Reviewer notes the zero-total-scan non-re-arm is covered by the bridge's terminal-line invariant — defense-in-depth as requested, no further change warranted. CONFIRMED FIXED. |

Hardware verification: not required (pure GUI narration; event stream unchanged) — the next routine console scan shows the deduped tail.

PR #625, opened 2026-08-20
GeecsBluesky 0.52.1: root BAX algorithm_results_file into the claimed scan folder at bind

skbarber:
## What + why

Xopt's `BaxGenerator` opens `<algorithm_results_file>_<n>.pkl` as-is on every `generate` call, and the config default is a bare relative `bax_probe_results`. `BaseOptimizer.from_config` only roots that path when a `scan_data_manager` provides a scan folder — but no live caller passes one (the block is dormant, per its own breadcrumb). So on the delegated console path the BAX generator wrote its probe-results pickles into the **process working directory**: live-observed 2026-08-20, a run launched from the GEECS-Console package dir dropped `bax_probe_results_1.pkl` into the repo checkout and it nearly got committed. Under the future queueserver worker the same default would litter the service's cwd.

**Fix:** `SessionOptimizationBridge.bind` — the first moment the run's real folder is known on the delegated path (the runner claims the scan pre-bind and passes `scan_folder`) — now roots a relative `algorithm_results_file` into the claimed scan folder. When the claim failed (`scan_folder=None`), it falls back to a fresh temp directory so cwd is never the destination. Absolute paths are explicit intent and pass through untouched; generators without the attribute (random, UCB, …) are a no-op.

**Scan-folder invariant:** the rebase only writes *into* the already-claimed folder (`claim_scan_number` created it, pre-bind) — no `mkdir` anywhere in this change.

## Changes (~130 LOC, one concern)

- `session_bridge.py` (+50): `_root_algorithm_results_file()` + call from `bind`, docstring update
- `base_optimizer.py` (+2): dormant-block breadcrumb now points at the live rooting site
- `tests/optimization/test_session_bridge.py` (+66): `_make_bax_optimizer` helper + 4 pinning tests (relative→scan folder, absolute passthrough, no-folder→temp dir never cwd, attribute-less generator no-op)
- version bump + CHANGELOG

## Tests

- `./scripts/check.sh`: lint green, GeecsBluesky **644 passed, 3 deselected** (CI-style env, `ca tiled` extras — optimization layer skips without the `optimize` extra, as in CI)
- Optimization layer (env with xopt 3.1.1): **128 passed** in `tests/optimization/`, including the 4 new pinning tests; the pre-fix code fails the new relative-default test

## Hardware verification

OWED: one console/delegated BAX optimize scan — expect `bax_probe_results_<n>.pkl` files inside `scans/ScanNNN/` (next to `xopt_dump.yaml`) and **no** pickles in the launch directory. (Non-BAX optimize scans are unaffected — their generators have no `algorithm_results_file`.)

### DIFF
--- a/GeecsBluesky/CHANGELOG.md
+++ b/GeecsBluesky/CHANGELOG.md
@@ -4,6 +4,25 @@ All notable changes to `geecs-bluesky` are documented here.
 
 Format follows [Keep a Changelog](https://keepachangelog.com/en/1.1.0/).
 
+## [0.52.1] - 2026-08-20
+
+### Fixed
+
+- **BAX algorithm-results pickles no longer land in the process working
+  directory.** Xopt's `BaxGenerator` opens
+  `<algorithm_results_file>_<n>.pkl` as-is on every generate call, and the
+  config default is a bare relative `bax_probe_results` — on the delegated
+  console path (where no `scan_data_manager` exists to root it) the pickles
+  dropped into whatever directory the process ran from (live-observed
+  2026-08-20: into the repo checkout, nearly committed; under the future
+  queueserver worker it would litter the service's cwd).
+  `SessionOptimizationBridge.bind` now roots a relative
+  `algorithm_results_file` into the claimed scan folder (writing into the
+  existing folder only — the runner claims pre-bind; scan-folder invariant
+  respected), falling back to a fresh temp directory when the claim failed.
+  Absolute paths pass through untouched; generators without the attribute
+  are a no-op.
+
 ## [0.52.0] - 2026-08-20
 
 ### Added
--- a/GeecsBluesky/CHANGELOG.md
+++ b/GeecsBluesky/CHANGELOG.md
@@ -20,8 +20,10 @@ Format follows [Keep a Changelog](https://keepachangelog.com/en/1.1.0/).
   `algorithm_results_file` into the claimed scan folder (writing into the
   existing folder only — the runner claims pre-bind; scan-folder invariant
   respected), falling back to a fresh temp directory when the claim failed.
-  Absolute paths pass through untouched; generators without the attribute
-  are a no-op.
+  Absolute paths pass through untouched; a relative value with directory
+  components is flattened to its basename (nothing creates those
+  directories, and `..` must not escape the scan folder); generators
+  without the attribute are a no-op.
 
 ## [0.52.0] - 2026-08-20
 
--- a/GeecsBluesky/geecs_bluesky/optimization/base_optimizer.py
+++ b/GeecsBluesky/geecs_bluesky/optimization/base_optimizer.py
@@ -578,7 +578,9 @@ class BaseOptimizer:
                 # scan_data_manager (scan_folder stays None) — a future
                 # caller on a scans/ScanNNN path must go through
                 # claim_scan_number instead (cross-package scan-folder
-                # invariant, root CLAUDE.md).
+                # invariant, root CLAUDE.md).  On the live delegated path
+                # the equivalent rooting happens post-claim, in
+                # SessionOptimizationBridge._root_algorithm_results_file.
                 scan_folder.mkdir(parents=True, exist_ok=True)
 
                 for key, block in list(overrides.items()):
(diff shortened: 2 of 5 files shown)


skbarber:

**Result: 2 low-severity findings, no high/medium findings survived. Both dispositioned below (separate comment).**

Verification performed before writing findings: full diff plus `session_bridge.py`, `base_optimizer.py` (from_config rooting block), `scan_request_runner.py` (claim/bind ordering), `scanner_bridge/bluesky_scanner.py` (bridge lifecycle), the BAX factory/config, the installed xopt 3.1.1 `BaxGenerator` source, and the test module + conftest skip guard. The four new tests were run against the real xopt classes: all pass.

### Surviving findings

**1. [Low — correctness edge]** A relative `algorithm_results_file` containing a directory component crashes the optimization mid-run at the first generator call. Scenario: `algorithm_results_file: "diagnostics/bax_probe"` → bind roots it to `<scan_folder>/diagnostics/bax_probe`; xopt 3.1.1's `BaxGenerator._get_acquisition` writes with a bare `open(...)` (no mkdir, exception propagates), so the run dies with FileNotFoundError at iteration 3 — after hardware has already moved. Not a regression (pre-fix the same value crashed unless cwd happened to contain the directory), but the new documented contract invites this shape and the failure is late and expensive. Same latent hole exists in the mkdtemp fallback and (pre-existing) in the dormant `from_config` rooting block.

**2. [Low — redundancy/placement]** This is the second implementation of the "root a relative algorithm_results_file" policy; the dormant scanner-side copy in `BaseOptimizer.from_config` differs (injects a `<key>_algo_results` default when the value is omitted; `.resolve()`s the result). Today they compose safely (config-time rooting produces an absolute path bind passes through, and the diff cross-links them), and the divergence is unreachable for BAX (`MultipointProbeConfig` defaults to `"bax_probe_results"`, never None) — but when the scanner-side path wakes up, someone must reconcile two drifted copies. Natural consolidation: one `BaseOptimizer` helper both call sites use, keeping generator-attribute knowledge in the optimizer module.

### Attacks attempted and refuted (recorded so they aren't re-raised)

- **Rebind staleness** (second scan's pickles into the first scan's folder): refuted — `bluesky_scanner.py` builds a fresh bridge (fresh optimizer/Xopt/generator) per scan request; bind runs exactly once per bridge on the live path.
- **Generator called before bind** (pickles into cwd during seeding): refuted — `seed_from_dumps` only calls `add_data`; `generate` is reached solely via `bridge.suggest`, post-bind.
- **Scan-folder invariant**: respected — the runner claims via `claim_scan` immediately before invoking the binder, bind performs no mkdir on any scan path, and the fallback mkdtemp is not a scan path.
- **Pydantic assignment onto the generator**: `algorithm_results_file` is a declared `Optional[str]` field on xopt's `BaxGenerator`; the new tests exercise assignment against the real class.
- **Dump/seed round-trip poisoning** (absolute temp path serialized into `xopt_dump.yaml` then re-imported by warm-start): refuted — `seed_from_dumps`/`load_xopt_dump` consume only VOCS + data rows, never generator config.
- **Test discrimination**: the new tests fail against plausible wrong implementations (no-op → `is_absolute()` fails; cwd-rooting → `parent != Path.cwd()` fails); the directory-level conftest skips cleanly without the `optimize` extra.
- **Process/policy**: patch bump correct for a bug fix, CHANGELOG accurate, no public-repo hygiene issues.

Bottom line per the reviewer: "the fix is correct on the live path and well-tested for what it claims. The two findings above are hardening/maintenance items, not blockers."

skbarber:
## Review disposition

- **Finding 1 (directory components → late FileNotFoundError): FIXED** in 18a069c8 — a relative `algorithm_results_file` with directory components is flattened to its basename at bind time with a WARNING (also closes the `..`-escape shape the same path allowed), pinned by `test_relative_path_with_directories_is_flattened`. **Reviewer verdict on the fix commit: CONFIRMED** ("verified against the actual diff and by running the test class, 5/5 pass … no failure scenario survives" — including the degenerate bare-`..` value, which becomes a harmless `.._<n>.pkl` inside the scan folder, not traversal).
- **Finding 2 (second copy of the rooting policy vs the dormant `from_config` block): WAIVED** — the scanner-side copy is dormant with no live caller, the two sites cross-reference each other in comments, and today they compose safely (config-time rooting yields an absolute path that bind passes through). Consolidating into one `BaseOptimizer` helper is the right move *when the scanner-side path wakes up* — doing it now is speculative cleanup of a dormant path (per the repo's deferred-debt doctrine). Reviewer agrees the waiver is consistent with that doctrine. Cheap to veto: say the word and I'll consolidate in a follow-up.

Optimization-layer test count after the fix commit: **129 passed** (xopt 3.1.1 env), CI-style suite unchanged.

PR #626, opened 2026-08-20
GeecsCAGateway 0.18.0: bounded DB connects + opt-in shot number in TCP push frames

skbarber:
## What + why

PR 1 of the **geecs-core arc** (declutter/replace GEECS-PythonAPI with a thin `GeecsDevice` client over the access layer; extraction of `transport/` + `db/` into a `GEECS-Core` package follows in PR 2). Two small behavior tweaks land here while the code still lives in one place, so the later extraction PR stays purely mechanical:

1. **`GeecsDb` MySQL connects are time-bounded** — `connection_timeout` via module constant `CONNECT_TIMEOUT_S = 10.0`. Off-network callers (a laptop off VPN, CI) fail in seconds instead of blocking ~75 s on the OS TCP-connect default. On the lab network connects complete in milliseconds, so this is a no-op there. This is also what makes self-skipping DB integration tests practical in geecs-core.

2. **`GeecsTcpSubscriber.subscribe(include_shot=True)`** — opt-in delivery of the push frame's shot counter to callbacks under the reserved key `"shot number"` (an `int`). Attached only when at least one subscribed variable matched, so callback gating is unchanged. Default `False`: the CA gateway and PVA gateway consumers are byte-for-byte unaffected. First consumer: the planned `GeecsDevice` client, which mirrors the legacy `state["shot number"]` behavior scripts rely on.

## Per-concern LOC

| Concern | Files | LOC |
|---|---|---|
| DB connect timeout | `db/geecs_db.py` | +14/−1 |
| include_shot opt-in | `transport/tcp_subscriber.py` | +34/−7 |
| Tests | `tests/test_geecs_db.py`, `tests/test_transport.py` | +76 |
| Version + CHANGELOG | `pyproject.toml`, `CHANGELOG.md` | +19 |

## Tests

`./scripts/check.sh`: lint green, **GeecsCAGateway 197 passed** (191 existing + 6 new: 1 DB-timeout kwarg pin, 1 fake-server end-to-end include_shot, 4 parser-contract pins incl. non-integer shot field and empty-frame gating).

## Hardware verification

None owed. Both changes are offline-verifiable: the timeout is a connector kwarg pinned by unit test (lab-network behavior unchanged — millisecond connects never hit a 10 s bound); `include_shot` is inert until a consumer passes it (none in this PR). PV-visible gateway behavior is untouched, so no `PV_CONTRACT.md` change.

- Reserved-key name `"shot number"` (legacy GEECS-PythonAPI `state` key) rather than a new identifier — chosen for familiarity to existing script users; documented as reserved in the docstring.
- `CONNECT_TIMEOUT_S` as a module-level constant (tunable by callers) rather than a per-call parameter — keeps the `GeecsDb` classmethod signatures untouched.

### DIFF
--- a/GeecsCAGateway/CHANGELOG.md
+++ b/GeecsCAGateway/CHANGELOG.md
@@ -3,6 +3,24 @@
 All notable changes to `geecs-ca-gateway` are documented here, following
 [Keep a Changelog](https://keepachangelog.com/en/1.1.0/) and semantic versioning.
 
+## [0.18.0] - 2026-08-20
+
+### Added
+
+- **`GeecsTcpSubscriber.subscribe(include_shot=True)`** — opt-in delivery of
+  the push frame's shot counter to callbacks under the reserved key
+  `"shot number"` (an `int`). Attached only to frames where at least one
+  subscribed variable matched, so callback gating and all existing consumers
+  (default `False`) are unchanged. First consumer: the planned `GeecsDevice`
+  client (geecs-core arc), which mirrors the legacy `state["shot number"]`.
+
+### Changed
+
+- **`GeecsDb` MySQL connects are now time-bounded** (`connection_timeout`,
+  module constant `CONNECT_TIMEOUT_S = 10.0` in `db/geecs_db.py`). Off-network
+  callers fail in seconds instead of blocking ~75 s on the OS TCP-connect
+  default. On the lab network (millisecond connects) this is a no-op.
+
 ## [0.17.1] - 2026-08-18
 
 ### Changed
--- a/GeecsCAGateway/geecs_ca_gateway/db/geecs_db.py
+++ b/GeecsCAGateway/geecs_ca_gateway/db/geecs_db.py
@@ -82,14 +82,27 @@ def _find_credentials() -> dict:
     return _credentials
 
 
+# Bound on the TCP connect to the DB host.  On the lab network connections
+# complete in milliseconds; off-network the OS default lets a connect block
+# for ~75 s per query.  Module-level so callers with unusual links can tune it.
+CONNECT_TIMEOUT_S: float = 10.0
+
+
 def _connect_mysql(mysql_connector):
     """Open a MySQL connection using the pure-Python connector implementation.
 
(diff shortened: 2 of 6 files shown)


skbarber:

The reviewer attempted to refute both changes against the full transport source, the legacy GEECS-PythonAPI parser this mirrors, the installed mysql-connector-python 9.7.0 source, and all in-repo callers. **Two low-severity findings survived; both fixed in 161fd321 and confirmed by the same reviewer.**

### Findings + disposition

1. **(Low) Per-frame warn spam on a malformed shot field** — a non-integer shot field is a device-configuration property, so the parser-level warning would fire on every ~5 Hz frame. **FIXED** (161fd321): warning moved to `_listen_loop` behind a per-subscription `warned_bad_shot` flag, matching the file's `_warned_missing_variables` convention; `_parse_subscription` stays a silent pure function. Reviewer **confirmed**, noting the detection condition is exact (with fix 2 in place, a non-empty parse missing the key can only be the `int()` failure path).

2. **(Low) Reserved-key collision documented but not guarded** — `include_shot=True` plus a subscribed variable literally named `"shot number"` would silently overwrite the variable's value. **FIXED** (161fd321): `subscribe()` raises `ValueError` on the combination, placed before the connected check; new test `test_reserved_name_in_variables_raises` pins both the guard and its ordering. Reviewer **confirmed**.

### Verified-and-refuted concerns (recorded for posterity)

- `connection_timeout` does **not** become a hidden query/read timeout: verified against connector 9.7.0 source — the socket timeout is cleared after auth (`set_connection_timeout(None)`), so the bound covers connect+handshake only.
- Shot-field extraction is byte-identical to the legacy parser's, including the matched-variables-only gating.
- Existing consumers (CA gateway, PVA gateway) pass no new kwargs; default-path parse is byte-identical — no `PV_CONTRACT.md` change needed.

### Reviewer nits on the fixes (waived)

- Warn-once behavior itself has no pinning test — **waived**: logging-only behavior; the key-omission contract it accompanies *is* pinned.
- Warning message dropped the raw frame excerpt — **waived**: reviewer's own assessment is host:port suffices to identify the device.

Suite after fixes: **198 passed**.

PR #627, opened 2026-08-20
GEECS-Core 0.1.0: extract the GEECS access library out of GeecsCAGateway

skbarber:
## What + why

PR 2 of the **geecs-core arc**. The GEECS access-layer library parts — UDP/TCP transport, experiment DB, PV naming, the exception tree, and the fake device server — move out of GeecsCAGateway into a new **GEECS-Core** package (`geecs_core`). Rationale: a fourth library consumer arrives next (the thin `GeecsDevice` client replacing GEECS-PythonAPI), and "library parts inside a service package" meant every consumer dragged caproto along and the supervise/naming/db surface had no owner of record. The gateway's own DESIGN.md has carried this extraction as a known gap since the transport first moved there.

**Purely mechanical — no behavior change.** Files move verbatim (`git mv`, rename-detected); the only content edits are import paths, two logger-name strings inside moved tests, and doc truth-ups.

## What moved where

- `geecs_ca_gateway/{transport/,db/,pv_naming.py,exceptions.py,testing/}` → `geecs_core/*`, with their test suites (`test_transport`, `test_udp_reply_correlation`, `test_geecs_db`, `test_coerce`).
- `alarms.py` → `geecs_core/db/alarms.py` **whole** (judgment call, cheap to veto: the audit suggested splitting it, but it's one cohesive pydantic unit modeling the `ca_alarm_limits` table `GeecsDb` reads — a split would sever `AlarmLimits` from its own field types).
- `naming.py` (thin re-export shim) **deleted**; its users import `geecs_core.pv_naming` directly. `test_naming.py` split: naming-policy primitives → `GEECS-Core/tests/test_pv_naming.py`, DeviceSpec assembly tests stay in the gateway.
- New: `GEECS-Core/DESIGN.md` (the three layering rules: one-way deps, single sync/async bridge point, the admission rule), `CLAUDE.md`, `README.md`, lazy package `__init__` (`import geecs_core.transport` stays stdlib-only).

## Dependency rewiring

- **GeecsCAGateway 0.19.0** — path-deps geecs-core; drops `mysql-connector-python` (rides in with core).
- **GeecsBluesky 0.52.2** — `geecs-ca-gateway` path dep **dropped entirely**; the CA gateway is now consumed purely as a service (its PVs), matching the long-stated doctrine. Imports come from geecs-core.
- **GeecsPvaGateway 0.4.3** — adds geecs-core; keeps a gateway dep only for `config.effective_vartype`.
[...]

PR #628, opened 2026-08-20
GEECS-Core 0.2.0: GeecsDevice — the entry-level synchronous client

skbarber:
## What + why

PR 3 of the **geecs-core arc** — the payload the extraction was for: `geecs_core.client.GeecsDevice`, the thin synchronous get/set/subscribe client that succeeds the legacy GEECS-PythonAPI device object (which is deleted in PR 4). Familiar shape for existing script users (`GeecsDevice("U_Hexapod")`, `.get`, `.set`, `.subscribe`, `.state`, `.close`), new semantics where the old ones were bugs.

## Design (per the DESIGN.md rules)

- **`client/_loop.py`** — one lazily-started daemon-thread asyncio loop per process; the package's single sync/async bridge (rule 2). No per-device threads, no cross-device command lock (the legacy class-global `threads_lock` serialized *every* device in the process — pinned gone by a two-device concurrency test).
- **`client/geecs_device.py`** (~370 lines incl. docs) — composes the existing, hardware-proven `GeecsUdpClient` + `GeecsTcpSubscriber` verbatim; zero transport logic of its own.
- **Deliberate semantic changes from legacy**: errors raise `geecs_core.exceptions` types (never `None`-on-failure); get/set return the device's parsed exe response (not a state-cache readback); no aliases, no `exp_info` global, no composites (GeecsBluesky pseudo-variables are the successor), no liveness probe at construction (one DB SELECT, no sockets).
- **Subscription**: frames feed `state` with reserved keys `"shot number"` (the PR-1 `include_shot` opt-in, always on here) and `"connected"`; optional `on_update` callback (loop thread; exceptions logged, stream continues); **auto-reconnect supervisor** with the gateways' 0.5→30 s backoff (opt-out `reconnect=False`); initial connect failure raises synchronously at the call site.
- **`close()`** idempotent, releases **both** UDP sockets — the legacy exe-socket leak (WinError 10048 on rapid reconnect) is pinned dead by a 5-cycle open/get/close test.
- Lazy `__init__` export preserved: `from geecs_core import GeecsDevice` works; `import geecs_core.transport` verified to still load zero pydantic/mysql modules.

## Tests

`./scripts/check.sh` green: **GEECS-Core 85 passed, 5 deselected** (68 existing + 17 new client tests). New coverage: get/set typing + state feed, error raising, two-device concurrency, subscription frames/shot-number/callback ordering, unsubscribe semantics, double-subscribe refusal, loud bad-endpoint failure, **supervisor reconnect across a real server restart on the same port**, the rapid-reconnect socket pin, idempotent close + fail-loud after close, DB-lookup construction paths.

**Live-lab tiers** (both deselected by default, self-skipping off-network thanks to PR 1's bounded DB connect): `-m integration` (real MySQL endpoint + metadata shape), `-m hardware` (real device get + subscribe, `GEECS_HW_DEVICE`/`GEECS_HW_VAR` env vars, default U_S1H/Current — same convention as GeecsBluesky). A set test exists but is double-gated behind `GEECS_HW_ALLOW_SET=1` and only writes back the value just read.

- `subscribe(variables=None)` subscribes the device's full DB-declared variable set with **no automatic text-variable inference** — string-typed values that look numeric get coerced unless the caller passes `text_variables`. Deriving text-ness needs the DB→dtype rules (`effective_vartype`), which live gateway-side; per the reviewer note on #627, if this bites, that's the moment `effective_vartype` moves down into geecs-core.
- `set` default timeout 30 s (GEECS sets block until convergence — the gateway's put budget), `get` 10 s; both per-call overridable.
- The supervisor is the third copy of the reconnect pattern in the family (CA gateway, PVA gateway, here) — hoisting a shared `supervise()` into `transport/` is deliberately deferred until a consumer forces the refactor.

## Hardware verification

**OWED**: the `-m hardware` tier against U_S1H `Current` from the lab network — expected: `get` returns a float, subscription delivers frames with an advancing integer `"shot number"` and `connected: True`. Offline coverage is complete (the fake server speaks the real wire protocol), so this is a confirmation run, not a functionality gate.


skbarber:

The reviewer ran the suite plus **empirical probe scripts** (deadlock reproduction, deterministic-schedule race probes, test-sabotage falsifiability checks, a 60 s-watchdog stress run). Seven findings survived; all fixed in b646afd8 + the final residuals commit, each **confirmed by the same reviewer re-running their probes**.

### Findings + disposition

1. **(HIGH) `run_sync` from inside `on_update` deadlocked the whole client layer** — empirically reproduced (loop wedged, every device dead, no diagnostic). **FIXED**: `run_sync` raises `RuntimeError` on the loop thread, naming the callback rule; fork caveat documented; pinned by test. Reviewer re-ran the deadlock probe: raises loudly, loop stays alive.
2. **(MED) `close()` racing `get()`/`subscribe()` leaked live sockets on a closed device** — empirically reproduced. **FIXED**: one lifecycle lock guards every transition; closed-check moved inside it; subscribe/unsubscribe/close serialized. Reviewer re-ran the race probe: getter now raises, zero leaks. Their explicit hold-lock-across-run_sync deadlock hunt (stress run, watchdog) found nothing wedgeable — every under-lock coroutine is bounded, long exchanges run outside the lock.
3. **(MED) The rapid-reconnect "socket pin" test couldn't fail** (fresh ephemeral ports masked leaks — verified by no-op-close sabotage passing). **FIXED**: asserts `is_closing()` on transports captured before `close()`. Sabotage now fails the test.
4. **(LOW-MED) Concurrency test was sequential.** **FIXED**: barrier-threaded gets + a structural pin that the two transport clients' locks are distinct objects (the exact legacy-regression shape).
5. **(LOW-MED) Reconnect path dropped half-established subscribers unclosed + narrow except could kill the supervisor silently** — real drift from both gateway supervisors. **FIXED**: `_connect_subscriber` owns the socket until handoff (`except BaseException: close; raise`); retry catches broad `Exception` after re-raising cancellation. Reviewer's swallowed-cancellation refutation attempt found no reachable path.
6. **(LOW) `timeout=0` coalesced to the default.** **FIXED** in the client (identity check), and the reviewer's re-check caught the same pre-existing `or` one layer down in `GeecsUdpClient.get/set` — also fixed (final commit).
7. **(LOW) `"connected"` reserved key unguarded.** **FIXED**: `subscribe` raises `ValueError` for both reserved keys; get/set skip reserved-key state writes. Noted behavior: a device DB-declaring a variable literally named `connected` gets a loud raise on `subscribe()` rather than silent clobbering — loud is the intended default.

### Waived (reviewer-assessed non-blocking)

- **Worst-case lock-hold ~5 s** when subscribing to a black-holing endpoint stalls other threads' get/set on the same device — inherent to the design, fine for a script-tier client.
- **Cosmetic message edge**: an in-flight get racing close() can surface the transport's "not connected" RuntimeError instead of "device is closed" — loud, no leak, no hang.
- **Reconnect churn on an accept-then-drop flapper** (~2 Hz retries) — shared family behavior with both gateway supervisors, not a divergence.

Suite after all fixes: **87 passed, 5 deselected**. Reviewer's verdict: merge-ready.

skbarber:
## Hardware verification — DONE ✅ (2026-08-20, lab network via VPN, rtt ~19 ms)

The OWED confirmation run passed on first attempt:

```
tests/test_live_lab.py::TestLiveDb::test_find_device_returns_endpoint PASSED
tests/test_live_lab.py::TestLiveDb::test_device_variables_have_metadata_shape PASSED
tests/test_live_lab.py::TestLiveDevice::test_get_returns_numeric PASSED
tests/test_live_lab.py::TestLiveDevice::test_subscribe_delivers_frames_with_shot_number PASSED
tests/test_live_lab.py::TestLiveDevice::test_set_back_current_value SKIPPED (GEECS_HW_ALLOW_SET not set)
4 passed, 1 skipped in 3.71s
```

Against the real U_S1H over the live DB + device: `get("Current")` returned numeric, the TCP subscription delivered frames with `connected: True` and an advancing integer `"shot number"`. The set-back test remains deliberately gated (commands hardware). Nothing further owed on this PR.

skbarber:
**Set path also hardware-verified** (2026-08-20, owner-authorized `GEECS_HW_ALLOW_SET=1`): `test_set_back_current_value` PASSED against U_S1H `Current` — write-back of the just-read value, device readback matched. All three hardware-tier tests green in 3.0 s; every public `GeecsDevice` method (get / set / subscribe / state / close) is now verified against real hardware.

Issue #634, opened 2026-08-21
qs/W4: s-file export as a stop-document callback (queueserver round 1)


skbarber:
**Campaign**: queueserver migration, Round 1. **Worker**: Codex. **Base + PR target**: `feat/queueserver` (never master, never dev). Repo context: read `AGENTS.md` → the root and `GeecsBluesky/CLAUDE.md`. The coordinator reviews and merges; do not self-merge.

## Task (self-contained)

Today the legacy s-file export runs as a **post-scan method call**: `GeecsSession._export_scalar_files(scan_number)` in `GeecsBluesky/geecs_bluesky/session.py` (defined near the bottom of the class; called from two post-`RE()` sites). Under the bluesky-queueserver migration (`Planning/cutover_strategy/02_queueserver_migration.md`), it must become a **RunEngine document-stream callback** driven by the stop document, because the worker has no post-call site.

Deliverable: a **new module** `GeecsBluesky/geecs_bluesky/sfile_callback.py`:

- A callback suitable for `RE.subscribe(...)` (a callable taking `(name, doc)`, or a class with that `__call__`) that: captures the run's `start` document (it carries the scan metadata — scan number under the run metadata the engine already emits; see `plans/run_wrapper.py` for what `claim_scan_number` puts in the start doc), and on the matching `stop` document performs the same best-effort export `_export_scalar_files` performs today (Tiled read-back by uid → scalar s-file write via the existing `geecs_data_utils` export helpers that method uses — reuse the same functions it calls, do not reimplement the export itself).
- Same failure posture as the method: **best-effort** — any failure logs a warning and never raises into the RunEngine (a callback exception must not kill a scan).
- Multi-run safe: keyed by run uid (start/stop matched by uid), tolerant of runs whose start doc lacks a scan number (log-and-skip).

**Purely additive**: do NOT modify `session.py` or any existing module — the wiring into worker startup is a separate task. No bluesky-queueserver imports.

## Tests

Hermetic pytest beside the existing GeecsBluesky tests (`tests/test_sfile_callback.py`): feed synthetic start/stop document pairs; assert the export helper is invoked with the right (uid, scan number) via monkeypatching; assert a raising export is swallowed with a logged warning; assert a scan-number-less run is skipped. No hardware, no Tiled server — mock the read-back seam.

## Conventions (binding)

- Python >=3.11, NumPy docstrings, type hints, Pydantic v2 idioms (root `CLAUDE.md`).
[...]

PR #635, opened 2026-08-21
qs/W4: export s-files from stop documents

skbarber:
## Summary
- Add `SFileExportCallback` for RunEngine document streams, keyed by run UID and driven by matching stop documents.
- Reuse `geecs_data_utils.write_scalar_files_from_tiled(uid)` behind a best-effort export helper that logs warnings instead of raising into the RunEngine.
- Add hermetic callback tests and bump `geecs-bluesky` to 0.52.3 with a changelog entry.

## Tests
- `cd GeecsBluesky && poetry install --with dev` passed.
- `cd GeecsBluesky && poetry run pytest tests -m 'not integration and not fake_server' -q` -> `434 passed, 26 skipped, 2 deselected, 7 warnings in 6.87s`.

Refs #634

### DIFF
--- a/GeecsBluesky/CHANGELOG.md
+++ b/GeecsBluesky/CHANGELOG.md
@@ -4,6 +4,15 @@ All notable changes to `geecs-bluesky` are documented here.
 
 Format follows [Keep a Changelog](https://keepachangelog.com/en/1.1.0/).
 
+## [0.52.3] - 2026-08-20
+
+### Added
+
+- Legacy scalar s-file export can now run as a RunEngine stop-document
+  callback (`SFileExportCallback`), matching starts/stops by run UID,
+  reading the scan number from start metadata, and preserving the existing
+  best-effort Tiled export posture for queueserver worker wiring.
+
 ## [0.52.2] - 2026-08-20
 
 ### Changed
--- a/GeecsBluesky/CHANGELOG.md
+++ b/GeecsBluesky/CHANGELOG.md
@@ -11,7 +11,9 @@ Format follows [Keep a Changelog](https://keepachangelog.com/en/1.1.0/).
 - Legacy scalar s-file export can now run as a RunEngine stop-document
   callback (`SFileExportCallback`), matching starts/stops by run UID,
   reading the scan number from start metadata, and preserving the existing
-  best-effort Tiled export posture for queueserver worker wiring.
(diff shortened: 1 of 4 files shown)


skbarber:
## Adversarial review (coordinator-run — the PR author is an external agent, per the campaign protocol in #634)

The fresh-context reviewer verified the module against issue #634's contract and `GeecsSession._export_scalar_files`'s actual behavior, ran the full suite on the PR head, and **sabotage-tested the new tests** (deliberate bugs must make them fail). Contract compliance was clean across the board: exact file set, correct base, scan number read from where `claim_scan_number` actually puts it, stop→start matched on the correct event-model key, the export delegated to the same `geecs_data_utils` helper (no reimplementation), scan-folder invariant honored, version + CHANGELOG + exact counts all present.

### Findings + disposition (coordinator-applied fixes, confirmed by the same reviewer)

1. **(Med) Exported s-files for aborted/failed runs** — the method this replaces is only *called* after clean completion, so today an aborted scan writes no s-file; the callback fired on every stop document. **FIXED** (26eeb3d0): success-only gate on `exit_status`, with the parity rationale in a comment; new test pins abort+fail+state-hygiene. Reviewer confirmed via gate-removal sabotage (test fails without the gate). Fail-closed on a missing `exit_status` — reviewer verified that matches both event-model requirements and the parity rationale. **Maintainer veto point**: if partial s-files from aborted runs are ever *wanted*, that's a one-line deliberate change.
2. **(Low) uid-matching test couldn't fail** — an implementation ignoring uids and exporting "the latest start" passed it. **FIXED** (26eeb3d0): stops the earlier run first, asserts both pairings. Reviewer re-ran the sabotage: now fails as required.
3. **(Note) Full start-document retention** — a lost stop (realistic under a future 0MQ RemoteDispatcher subscription) would leak a whole start-doc copy. **FIXED** (26eeb3d0): `_starts` keeps `uid → scan_number` only.
4. **(Nit) CHANGELOG didn't mention the success-only gate.** **FIXED** (67524acb).

Suite after fixes: **648 passed, 3 deselected** (full-extras env); pre-commit clean. The author's PR-body counts (434/26 skipped) reflect a no-extras env — the known collapse documented in GeecsBluesky/CLAUDE.md; process obligation met.

Merging into `feat/queueserver` (integration branch — coordinator merge per campaign protocol; master promotion stays maintainer-only).

Issue #636, opened 2026-08-21
qs/W2a: queueserver launch assets — permissions, launcher, ops notes (queueserver round 2)


skbarber:
**Campaign**: queueserver migration. **Worker**: Codex. **Base + PR target**: `feat/queueserver` (never master). Read `AGENTS.md` → root + `GeecsBluesky/CLAUDE.md` first. Coordinator reviews and merges; do not self-merge.

## Task (self-contained)

Create the operational launch assets for running a bluesky-queueserver RE Manager for GEECS, as a **new directory** `GeecsBluesky/qserver/` (nothing outside it except the version bump + CHANGELOG):

1. **`user_group_permissions.yaml`** — the manager refuses every queue submission with `success: False` if launched without one (empirically confirmed on the interim host; the failure is silent at the CLI). Provide: `root` (all null = everything), `admin` (regex-all plans/devices/functions), and a restricted `operator` group allowing only plans matching `:geecs_.*` (the campaign's plan-naming convention; `geecs_scan_request_plan` is coming in a parallel task). Comment each group.
2. **`launch_re_manager.sh`** — a user-level launcher (no sudo, no systemd — that comes at deployment): starts a local redis-server if 127.0.0.1:6379 isn't answering (path to the binary via `QS_REDIS_SERVER` env var, default `redis-server`), then `exec`s `start-re-manager` with `--startup-dir` (env `QS_STARTUP_DIR`, default `./startup`), the permissions file above, and `--zmq-publish-console ON`. Fail loud with a clear message if start-re-manager isn't on PATH. Shell-check-clean bash.
3. **`README.md`** in that directory — how to launch, how to verify (`qserver status`, `qserver environment open`), the permissions-file gotcha (silent add-rejection without it), and the explicit note that the startup profile content lands in a separate task (the plan preamble work) — this directory ships launch *mechanics* only.
4. A placeholder `startup/` dir with a README line saying the profile arrives with the plan-preamble task — no Python.

## Constraints

- Do NOT touch any existing file except `GeecsBluesky/pyproject.toml` (poetry version patch) and `GeecsBluesky/CHANGELOG.md` (entry under the new version).
- No hostnames, usernames, or lab account names anywhere (public repo; internal 192.168.6.x IPs are acceptable but none are needed here).
- bash with `set -euo pipefail`; NumPy-style isn't applicable to shell — comment the non-obvious lines only.

## Acceptance

[...]

PR #638, opened 2026-08-21
qs/W2a: add queueserver launch assets

skbarber:
## Summary
- Add `GeecsBluesky/qserver/user_group_permissions.yaml` with root, admin, and restricted operator groups for RE Manager startup.
- Add `launch_re_manager.sh` to check `start-re-manager`, start local Redis when `127.0.0.1:6379` is not answering, and exec RE Manager with the permissions file and console publishing enabled.
- Add qserver operator notes plus a placeholder `startup/` README for the separate plan-preamble task.
- Bump `geecs-bluesky` to 0.52.4 and add a changelog entry.

## Validation
- `bash -n GeecsBluesky/qserver/launch_re_manager.sh` -> no output, exit 0.
- `python3.11 -c "import yaml,sys; yaml.safe_load(open('GeecsBluesky/qserver/user_group_permissions.yaml'))"` -> no output, exit 0.
- `cd GeecsBluesky && poetry run python -c "import yaml,sys; yaml.safe_load(open('qserver/user_group_permissions.yaml'))"` -> no output, exit 0.
- `git diff --check` -> no output, exit 0.
- `scripts/commit.sh -m "Add queueserver launch assets"` -> hooks passed, including `check yaml` and `check toml`; jupyter clear-output skipped because `jupyter` is not installed locally.

Note: the literal `python -c ...` form from the issue cannot run on this machine because there is no bare `python` executable on `PATH`; the YAML was validated with both `python3.11` and Poetry's package Python.

Refs #636

### DIFF
--- a/GeecsBluesky/CHANGELOG.md
+++ b/GeecsBluesky/CHANGELOG.md
@@ -4,6 +4,15 @@ All notable changes to `geecs-bluesky` are documented here.
 
 Format follows [Keep a Changelog](https://keepachangelog.com/en/1.1.0/).
 
+## [0.52.4] - 2026-08-21
+
+### Added
+
+- Added `qserver/` launch mechanics for the queueserver migration: a
+  user-group permissions file, a user-level RE Manager launcher that brings up
+  local Redis when needed, operator README notes, and a placeholder startup
+  directory for the separate plan-preamble task.
+
 ## [0.52.3] - 2026-08-20
 
 ### Added
--- a/GeecsBluesky/CHANGELOG.md
+++ b/GeecsBluesky/CHANGELOG.md
@@ -4,7 +4,7 @@ All notable changes to `geecs-bluesky` are documented here.
 
 Format follows [Keep a Changelog](https://keepachangelog.com/en/1.1.0/).
 
-## [0.52.4] - 2026-08-21
+## [0.52.4] - 2026-08-20
 
 ### Added
 
--- a/GeecsBluesky/pyproject.toml
+++ b/GeecsBluesky/pyproject.toml
@@ -1,6 +1,6 @@
 [tool.poetry]
 name = "geecs-bluesky"
-version = "0.52.3"
+version = "0.52.4"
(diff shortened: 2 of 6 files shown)


skbarber:
## Adversarial review (coordinator-run — the PR author is an external agent, per campaign protocol)

The reviewer validated the deliverables directly (bash -n, yaml.safe_load in the package env, hygiene greps, file-set audit, demo-format comparison for the permission regexes) and checked the PR against issue #636 **including its comments**. Body-level compliance was complete and the shell/YAML quality genuinely good — but **all three comment-borne empirical requirements were missing** (F1 `--keep-re`, F2 the `primary` group, F3 the RE-requirement + function-idle docs). Root cause is a coordination lesson, not worker sloppiness: the comments landed *after* the worker snapshotted the issue. Campaign protocol updated — brief amendments go into the issue body with an explicit re-ping.

### Disposition (coordinator-applied fixes, confirmed by the same reviewer)

- **F1 (blocker)** launcher lacked `--keep-re` → every `queue start` would silently bounce. **FIXED** (000cafd4), with the failure mode documented above the exec. Confirmed.
- **F2 (blocker)** no `primary` group → every CLI submission rejected — the shipped YAML reproduced the exact gotcha its own README warns about. **FIXED** (admin-equivalent + production-tightening note). Confirmed; groups now admin/operator/primary/root.
- **F3 (major)** READMEs missing the `RE = RunEngine({})` requirement and the function-execute-requires-idle note. **FIXED**: startup/README states the RE pairing + symptom; main README gains a Troubleshooting section covering all three empirical findings. Confirmed.
- **Nit** CHANGELOG dated tomorrow. **FIXED** (2026-08-20).

### Recorded advisories (accepted as-is, per the issue body's own dictates)

Unanchored `":geecs_.*"` operator pattern (re.search semantics); empty operator allowed_devices/functions lists; `primary` omitting the explicit `forbidden_*: - null` keys (semantically identical).

Merging into `feat/queueserver` (coordinator merge per campaign protocol).

PR #639, opened 2026-08-21
GeecsBluesky 0.53.0: geecs_scan_request_plan — run a ScanRequest as one plan (qs/W1, #633)

skbarber:
Executes #633 (queueserver campaign, round 1, worker W1). **Base = `feat/queueserver`; the coordinator merges — do not merge from this side.**

## What + why

The queueserver migration's one structural unit (`Planning/cutover_strategy/02_queueserver_migration.md` § The one structural task): `geecs_scan_request_plan(request)` — a plan generator meaning "run this ScanRequest". The `run_scan_request()` prologue relocates into the plan **preamble**, executed worker-side inside the RunEngine, so every bound method/closure (ShotController stubs, per-step action callables) is constructed from the JSON request and never crosses a process boundary:

1. authoritative fail-fast validation (`validate_scan_request` — amendment 2: no separate validation stub; clients re-run it pre-submit),
2. name resolution + worker-side construction (shot controller, action signals, detectors, axis movables — connects deferred),
3. in-plan connects (`ensure_connected` batches for the strict tier, soft gather for Tier-2 telemetry, shot-control reachability) — still pre-claim, so failures burn no scan number,
4. scan-number claim + ScanInfo + per-scan `scan.log`,
5. the same inner plan as today (`build_step_scan_plan`), with a finalize disconnecting everything the plan created.

No bluesky-queueserver import — an ordinary plan callable; W2 registers it with the manager. `set_plan_session()` installs the worker-wide default session so `RE(geecs_scan_request_plan(request))` needs only the JSON request.

This is **relocation, not rewrite**: the preamble calls the runner's own module-level prologue functions. The one extraction: `run_scan_request`'s inline md/ScanInfo/grid assembly became the pure `build_step_scan_spec()` (+ `StepScanSpec`), consumed verbatim by both entry points so they cannot drift. `run_scan_request` and the GUI-bridge path are otherwise unchanged (the whole existing suite pins them).

## Per-concern LOC breakdown

| Concern | Files | ~LOC |
|---|---|---|
| Runner extraction (`StepScanSpec` + `build_step_scan_spec`, tail rewired) | `scan_request_runner.py` | +145 / −90 (net logic unchanged) |
| The plan callable + facade + in-plan connect/cleanup stubs | `plans/scan_request_plan.py` (new) | +560 |
| Hermetic pins + document-parity tests | `tests/test_scan_request_plan.py` (new) | +500 |
[...]

PR #644, opened 2026-08-21
GeecsBluesky 0.54.0: RE Manager startup profile + optimize loader seam (#640)

### LINKED ISSUE #640: qs/W2: worker startup profile — session wiring, plan registration, optimize loader (queueserver round 2)
**Campaign**: queueserver migration, Round 2. **Worker**: Claude Code terminal session. **Base + PR target**: `feat/queueserver`. Coordinator merges; report ready-to-merge with your adversarial review + disposition posted on the PR first.

## Task

Build the RE Manager **startup profile** that turns the merged pieces into a runnable worker: `GeecsBluesky/qserver/startup/startup.py` (+ helpers as needed under `qserver/`). The launcher (#638) and the plan (#639, `geecs_scan_request_plan` + `set_plan_session`) exist; this task wires them.

The profile must, in this order:

1. `import geecs_bluesky` **first** — its `__init__` applies `[epics] ca_addr_list` from `~/.config/geecs_python_api/config.ini` before any aioca import (`epics_env.py` contract; see GeecsBluesky/CLAUDE.md). Document in a comment why this import leads. (Design decision on record: EPICS_CA_ADDR_LIST stays config-file/systemd-env sourced, never DB-sourced — bootstrap circularity + import-time-network hazard; restate briefly in the profile docstring.)
2. Build the `GeecsSession` — experiment name from env `QS_EXPERIMENT`, falling back to config.ini's `[Experiment]` section if present; fail loud at startup if neither yields a name.
3. `set_plan_session(session)` (the #639 seam).
4. `RE = RunEngine({})` — REQUIRED (launcher passes `--keep-re`; see qserver/startup/README.md for the silent-bounce failure mode).
5. Subscribe: TiledWriter (relocate the session's `[tiled]` config mechanism — best-effort, skip-with-log if unreachable, same posture as today), `SFileExportCallback` from #635, and the scan-log root capture per `scan_log.py`'s worker-relocation note in the Planning doc amendments.
6. **Register the optimization loader** (closes the deferred cross-vendor P1 from #639): when `request.optimization` is present the plan preamble must invoke a config→Xopt-stack factory built from `geecs_bluesky.optimization` (decision 5 in Planning/cutover_strategy/02_queueserver_migration.md — the loader CALL moves worker-side, code stays put). You own the seam change in `plans/scan_request_plan.py` that replaces the validated-then-refused optimize path with loader invocation when a loader is registered (keep the loud refusal when none is registered — headless installs without the optimize extra). Model the loader-registration API on how GEECS-Console's `services/optimization.py` builds its loader, but import nothing from GEECS-Console.
7. Expose `geecs_scan_request_plan` in the manager's plan list.

Also:
- Add an optional extra to GeecsBluesky: `qserver = ["bluesky-queueserver"]`, pinned `^0.0.25` (the sandbox-validated version).
- Update `qserver/README.md` + `qserver/startup/README.md` to reflect the profile's existence (drop the placeholder language).

## File ownership (collision contract)

You own: `GeecsBluesky/qserver/**`, `geecs_bluesky/plans/scan_request_plan.py` (optimize-seam only), new tests, `pyproject.toml`, `CHANGELOG.md`. Do NOT touch: `session.py`, `scanner_bridge/`, `plans/orchestration.py`/`step_scan.py`/`free_run_step_scan.py` (a parallel task owns pause semantics there), GEECS-Console, geecs-core.

## Tests (hermetic — no lab, no redis)

- Profile importable headless and defines RE + the plan (import it with a monkeypatched config chain).
- The optimize seam: with a fake loader registered, an optimize ScanRequest reaches the loader and threads its binder into the run (mock backends); with no loader, the loud refusal stands (existing test updated, not deleted).
- bluesky-queueserver's `gen_list_of_plans_and_devices` (or equivalent profile-load API) succeeds on the startup dir — pins that the manager can actually consume the profile.

## Acceptance

All suites green (report exact counts); version minor bump + CHANGELOG; the coordinator's live checkpoint (real `qserver` submit → manager → plan on the interim host) is explicitly OWED in the PR body, not claimed.

skbarber:
## Summary

Closes #640 (queueserver migration, round 2): builds the RE Manager startup
profile and closes the deferred cross-vendor optimize-mode seam.

- **`qserver/startup/startup.py`** (new): imports `geecs_bluesky` first
  (before any `aioca` import — sets `EPICS_CA_ADDR_LIST`/`EPICS_CA_AUTO_ADDR_LIST`
  from `config.ini` before libca's CA context is created), resolves the
  experiment from `QS_EXPERIMENT` (falling back to `config.ini`'s
  `[Experiment] expt`, failing loud at import time if neither yields a
  name), builds `GeecsSession(experiment, tiled=True)`, exposes `session.RE`
  as the module-level `RE` the manager keeps alive across queue items
  (`--keep-re` requires this — see the startup README's failure-mode note),
  subscribes `SFileExportCallback` (#635) and the session's own best-effort
  Tiled subscription, and registers `geecs_scan_request_plan`.
- **`geecs_bluesky/plans/scan_request_plan.py`** (optimize seam only):
  `set_optimization_loader()` installs a worker-wide loader; the OPTIMIZE
  branch now invokes it in-plan (worker-side deferred-connect construction,
  unserved-vars preflight merged with the loader's `device_requirements`,
  the claim boundary, then `geecs_adaptive_scan` wrapped the same way
  `GeecsSession.optimize` wraps it) instead of the old `NotImplementedError`
  refusal. No loader registered still refuses loudly — never silent
  degradation.
- **`geecs_bluesky/optimization/worker_loader.py`** (new): the loader
  implementation (`OptimizationSpec` → `SessionOptimizationBridge` via
  `BaseOptimizer.from_config`), modeled on GEECS-Console's
  `services/optimization.py` but independently implemented — imports
  nothing from GEECS-Console. `optimization_available()` gates registration
  on `xopt`/`scan_analysis` being importable.
- New optional extra `qserver = ["bluesky-queueserver"]` pinned `^0.0.25`;
  `poetry.lock` refreshed to match (purely additive — new transitive deps
  from `bluesky-queueserver` plus marker updates on already-shared deps;
  no existing package version changed, confirmed via diff).
- `qserver/README.md` / `qserver/startup/README.md` updated, placeholder
  language dropped.

### Per-concern LOC breakdown

- Startup profile + its README + worker_loader.py: new files, ~230 LOC
- Optimize seam in `scan_request_plan.py`: ~350 LOC (the loader hookup +
  the in-plan ask/move/bin/objective/tell loop body, mirroring
  `GeecsSession.optimize`'s existing shape)
- Tests: `test_qserver_startup.py` (new, 3 tests) + `test_scan_request_plan.py`
  additions (2 tests) — ~215 LOC
- `pyproject.toml` + `poetry.lock` + `CHANGELOG.md`: mechanical

## File-ownership contract (per #640)

Touched only: `qserver/**`, `geecs_bluesky/plans/scan_request_plan.py`
(optimize-seam only), new tests, `pyproject.toml`, `poetry.lock`,
`CHANGELOG.md`. Did not touch `session.py`, `scanner_bridge/`,
`plans/orchestration.py`/`step_scan.py`/`free_run_step_scan.py`,
GEECS-Console, or geecs-core.

## Tests (hermetic — no lab, no redis)

`GeecsBluesky`: **661 passed, 1 skipped (pre-existing), 3 deselected**
(`poetry run pytest -q`, with the `qserver` extra installed so all 3
`test_qserver_startup.py` cases execute rather than skip). Lint
(`./scripts/check.sh --lint`) clean.

New coverage:
- `test_qserver_startup.py` (new): startup profile defines `RE`/
  `geecs_scan_request_plan` headlessly with `QS_EXPERIMENT` set and a
  stubbed Tiled subscription; fails loud (`RuntimeError`) with neither
  `QS_EXPERIMENT` nor a resolvable `config.ini` experiment; `bluesky_queueserver`'s
  own `gen_list_of_plans_and_devices` succeeds against `qserver/startup/`
  — this last case is skipped when the `qserver` extra isn't installed
  (CI's default `poetry install --with dev` does not install it).
- `test_scan_request_plan.py`: renamed/updated
  `test_optimize_mode_without_a_loader_is_refused_loudly_after_validation`
  (loud refusal still stands, existing test updated not deleted) plus new
  `test_optimize_mode_reaches_a_registered_loader_and_runs_the_bins` — a
  fake loader/bridge drives a scripted suggester through real bins with
  mock CA backends. Notable wrinkle: `geecs_adaptive_scan`'s t0 sync runs
  *before* any staging message (unlike the step/noscan path, which sits
  inside `bpp.stage_wrapper`), so the existing `_install_stage_pacer`
  "stage"-triggered test helper fires too late for this path. The fix
  seeds/paces mock `acq_timestamp` from inside the fake bridge's
  `bind(devices=...)` call instead — devices are already connected by
  that point (bind() runs after in-plan connects, before the adaptive-scan
  generator starts), which is the last hook available ahead of the t0 sync
  read.

## Hardware verification

**OWED** — this PR is code-complete and hermetically tested only. Owed
before/at the coordinator's live checkpoint:
- `start-re-manager --startup-dir qserver/startup/` actually boots against
  a real `QS_EXPERIMENT` and gateway, `--keep-re` holds across two queued
  plans (no "Run Engine is not found" bounce)
- A real `qserver queue add plan 'geecs_scan_request_plan' ...` /
  `queue start` round-trip against live devices (step, noscan, and — if
  the `optimize` extra is installed on that worker — optimize mode)
- SFileExportCallback and Tiled actually receive documents from a
  manager-run scan (this PR only exercises the subscription wiring, not a
  live write)

## Not merging this myself

Per repo convention, `feat/queueserver` merges are coordinator-handled.
Reporting as ready-to-merge once the adversarial review (below) is
dispositioned and CI is green.

### DIFF
--- a/GeecsBluesky/CHANGELOG.md
+++ b/GeecsBluesky/CHANGELOG.md
@@ -4,6 +4,55 @@ All notable changes to `geecs-bluesky` are documented here.
 
 Format follows [Keep a Changelog](https://keepachangelog.com/en/1.1.0/).
 
+## [0.54.0] - 2026-08-20
+
+### Added
+
+- **RE Manager startup profile — the worker is now runnable**
+  (`qserver/startup/startup.py`, issue #640, queueserver migration round 2).
+  Imports `geecs_bluesky` first (before any `aioca` import, so
+  `EPICS_CA_ADDR_LIST`/`EPICS_CA_AUTO_ADDR_LIST` are set from
+  `~/.config/geecs_python_api/config.ini` before libca's CA context is
+  created — config-file/systemd-env sourcing is deliberate, never
+  DB-sourced, to avoid an import-time network round trip and a bootstrap
+  circularity), resolves the experiment from `QS_EXPERIMENT` (falling back
+  to `config.ini`'s `[Experiment] expt`, failing loud at startup if
+  neither yields a name), builds the worker's `GeecsSession`, exposes
+  `session.RE` as the module-level `RE` the manager keeps alive
+  (`--keep-re`), subscribes `SFileExportCallback` (#635) and the session's
+  own best-effort Tiled subscription, and registers
+  `geecs_scan_request_plan` in the manager's plan list.
+- **The optimization loader seam is closed** (the deferred cross-vendor P1
+  from #639, decision 5 in
+  `Planning/cutover_strategy/02_queueserver_migration.md`):
+  `plans/scan_request_plan.py` gains `set_optimization_loader()`, and the
+  optimize-mode body of the preamble now invokes the registered loader
+  in-plan (worker-side construction with deferred connects, unserved-vars
+  pre-flight over the save-set devices plus the loader's
+  `device_requirements`, the claim boundary, then
+  `plans.optimize.geecs_adaptive_scan` wrapped exactly as
+  `GeecsSession.optimize` wraps it) instead of refusing loudly. A worker
+  without the `optimize` extra still refuses loudly — no loader registered
+  means no silent degradation. `geecs_bluesky/optimization/worker_loader.py`
+  is the loader implementation (`OptimizationSpec` → `SessionOptimizationBridge`
+  via `BaseOptimizer.from_config`), modeled on GEECS-Console's
+  `services/optimization.py` but independently implemented (GeecsBluesky
+  imports nothing from GEECS-Console).
+- New optional extra `qserver = ["bluesky-queueserver"]`, pinned `^0.0.25`
+  (the sandbox-validated release); `poetry.lock` refreshed to match.
+- `tests/test_qserver_startup.py` (hermetic — no lab, no redis): the
+  startup profile defines `RE`/`geecs_scan_request_plan` headlessly with
+  `QS_EXPERIMENT` set, fails loud with neither `QS_EXPERIMENT` nor
+  `config.ini`'s `[Experiment] expt`, and (when the `qserver` extra is
+  installed) `bluesky_queueserver`'s own
+  `gen_list_of_plans_and_devices` succeeds against `qserver/startup/` —
+  skipped by default since the CI job does not install the `qserver`
+  extra. `tests/test_scan_request_plan.py` gains the optimize-loader-seam
+  coverage: no loader registered still refuses loudly post-validation,
+  and a fake loader/bridge drives a scripted suggester through real bins
+  (mock acq-timestamp pacing seeded from the bridge's `bind(devices=...)`,
+  since the adaptive scan's t0 sync runs before any staging message).
+
 ## [0.53.1] - 2026-08-20
 
 ### Added
--- a/GeecsBluesky/CHANGELOG.md
+++ b/GeecsBluesky/CHANGELOG.md
@@ -52,6 +52,18 @@ Format follows [Keep a Changelog](https://keepachangelog.com/en/1.1.0/).
   and a fake loader/bridge drives a scripted suggester through real bins
   (mock acq-timestamp pacing seeded from the bridge's `bind(devices=...)`,
   since the adaptive scan's t0 sync runs before any staging message).
+- PR #644 review fix wave: `worker_loader.py` gains
+  `warm_up_optimization_stack()`, which imports the heavy optimize-extra
+  modules (`xopt`, `scan_analysis`, the loader's own submodules) on a
+  daemon thread right after a loader is registered, so the first real
+  `geecs_scan_request_plan` optimize call in a worker session doesn't pay
+  that import cost in-plan; `startup.py` calls it immediately after
+  `set_optimization_loader()`, a no-op when the `optimize` extra isn't
+  installed. `tests/test_qserver_startup.py`'s import-order test is now a
+  `_qserver_startup_probe.py` subprocess run in a fresh interpreter — the
+  prior in-process `runpy.run_path` version could never actually observe
+  first-import order since `geecs_bluesky`/`aioca` are already cached in
+  `sys.modules` by the time any in-process test runs.
 
 ## [0.53.1] - 2026-08-20
 
--- a/GeecsBluesky/CLAUDE.md
+++ b/GeecsBluesky/CLAUDE.md
@@ -107,9 +107,13 @@ geecs_bluesky/
                             #   the plan preamble (validate → resolve →
                             #   construct+connect in-plan → claim → the same
                             #   inner plan); set_plan_session installs the
-                            #   worker default session; optimize refused
-                            #   loudly (later round); s-file export is NOT
-                            #   here (worker stop-doc callback seam)
+                            #   worker default session; optimize-mode
+                            #   requests run in-plan too (set_optimization_
+                            #   loader registers the worker's loader — see
+                            #   optimization/worker_loader.py — refused
+                            #   loudly, not mid-scan, when unregistered);
+                            #   s-file export is NOT here (worker stop-doc
+                            #   callback seam)
   devices/
     ca/                     # THE device family: CA-backed via GeecsCAGateway PVs (`ca` extra)
       triggerable.py        # CaAcqTimestampReadable (persistent CA monitor) + CaTriggerable
@@ -161,6 +165,13 @@ geecs_bluesky/
                             #   — the legacy engine re-imports them via shims).
                             #   Heavy deps ride the `optimize` extra; tests
                             #   skip whole without it
+    worker_loader.py         # the queueserver worker's optimization_loader:
+                            #   OptimizationSpec -> SessionOptimizationBridge
+                            #   (worker-side twin of GEECS-Console's
+                            #   geecs_console.services.optimization); also
+                            #   warm_up_optimization_stack(), called once at
+                            #   worker startup (qserver/startup/startup.py)
+                            #   to pre-import the heavy stack off-thread
   tiled_integration.py      # subscribe_tiled + descriptor patch + safe callback
   data_paths.py             # local ↔ device-server path mapping, asset roots
   scanner_configs.py        # configs-repo resolution + shot-control YAML loading
--- a/GeecsBluesky/geecs_bluesky/optimization/worker_loader.py
+++ b/GeecsBluesky/geecs_bluesky/optimization/worker_loader.py
@@ -1,0 +1,155 @@
+"""The queueserver worker's ``optimization_loader``: OptimizationSpec → bridge.
+
+Decision 5 (``Planning/cutover_strategy/02_queueserver_migration.md``):
+optimization stays in-worker — the stack's code
+(``geecs_bluesky.optimization``) does not move, only the *call site* that
+invokes it moves from the GUI process into the queueserver worker's plan
+preamble (``geecs_bluesky.plans.scan_request_plan``,
+:func:`~geecs_bluesky.plans.scan_request_plan.set_optimization_loader`).
+
+This module is the worker-side twin of GEECS-Console's
+``geecs_console.services.optimization`` — same three functions, same
+shapes, independently implemented so GeecsBluesky imports nothing from
+GEECS-Console (the dependency graph runs the other way: Console depends on
+GeecsBluesky, never the reverse). Keep the two in sync by hand if the
+``OptimizationSpec``/``BaseOptimizerConfig`` mapping changes.
+"""
+
+from __future__ import annotations
+
+import importlib.util
+import logging
+from typing import Any, Callable, Optional
+
+logger = logging.getLogger(__name__)
+
+
+def optimizer_config_from_spec(spec: Any) -> dict:
+    """Map an ``OptimizationSpec`` onto the ``BaseOptimizerConfig`` dict shape.
+
+    Pure and import-light (duck-typed off the spec's attributes), so the
+    mapping is testable without the ``optimize`` extra installed. The exact
+    inverse of ``geecs_schemas.convert.convert_optimizer_config``
+    (``generator.options`` <-> ``xopt_config_overrides[generator.name]``).
+    ``max_iterations`` is deliberately not mapped — the plan consumes it
+    from the spec directly.
+
+    Parameters
+    ----------
+    spec : geecs_schemas.OptimizationSpec
+        The request's resolved optimization block.
+
+    Returns
+    -------
+    dict
+        The parsed-YAML shape ``BaseOptimizer.from_config`` validates:
+        ``vocs`` (variables/objectives/observables/constraints),
+        ``evaluator`` (module/class/kwargs), ``generator`` (name),
+        ``xopt_config_overrides`` keyed by the generator name (only when
+        the spec carries generator options), ``seed_dump_files`` and
+        ``move_to_best_on_finish``.
+    """
+    config: dict[str, Any] = {
+        "vocs": {
+            "variables": {
+                name: list(bounds) for name, bounds in spec.variables.items()
+            },
+            "objectives": dict(spec.objectives),
+            "observables": list(spec.observables),
+            "constraints": {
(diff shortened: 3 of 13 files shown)


skbarber:
## Adversarial review (fresh-context subagent, three lenses: correctness / redundancy / placement)

**Correctness:** No surviving findings. `claim_scan`/`ScanTag` handling, `finally`-block `opt_bridge.finish()`, NaN-objective fallback on a failing objective, `device_requirements` merge defensiveness, and the validate→loader-check→bind→preflight→claim→run ordering all check out. `_observe_previous(len(history) + 1)` correctly scores the final bin. No scan-folder-creation invariant violation, no public-repo hygiene issue.

**Redundancy:** One finding — `worker_loader.py`'s `optimizer_config_from_spec` duplicates GEECS-Console's `services/optimization.py` version (currently identical), with no CI check that the two stay in sync if `OptimizationSpec` changes.

**Placement:** No surviving findings. `worker_loader.py` in `geecs_bluesky/optimization/` and `startup.py` outside the package in `qserver/startup/` both match the dependency graph's boundaries.

## Disposition

**Redundancy finding — WAIVED.** This duplication is the issue's explicit design ("model API on GEECS-Console's `services/optimization.py` but import nothing from GEECS-Console" — #640), for the same reason `geecs-core`/the gateways don't import each other's server code: GeecsBluesky must not depend on GEECS-Console (dependency graph, root CLAUDE.md). The module docstring already states the by-hand-sync obligation. Adding a cross-package parity test would need to import both packages together, which is itself outside either package's own boundary and outside this issue's scope — noting it here as a fair follow-up if `OptimizationSpec` mapping logic grows nontrivial, rather than blocking this PR.

No findings required a code change; nothing else to fix.

skbarber:
## Adversarial review (coordinator-run) — disposition

Verified on a clean clone of the PR head: suites green in both extras mixes (662/0/3 full), `poetry check --lock` clean, three sabotage runs (objective-threading and RE-pairing caught by tests; one tripwire found dead — finding 4). The spec's ordered list is fully delivered: import order verified to the source, claim-then-bind matches `_run_optimize_request`, `worker_loader.py` imports nothing from Console and keeps xopt call-time (proven by running the seam tests with no xopt installed), gen-list test genuinely enumerates the plan, and the out-of-grant `optimization/worker_loader.py` placement is **endorsed** (startup/ isn't importable; plans/ would be a layering smell). `session.RE` over a literal `RunEngine({})` is a justified spec improvement.

### Findings → worker fix wave (rows 1–7)

1. **(P2) `move_to_best_on_finish: true` silently ignored** — the runner path moves to the best-observed point; the plan drops it with only a docstring note. This is the operator's *point* of optimizing. Fix: implement (a `bps.mv` to best inputs at plan end) or, minimum, `logger.warning` + an md key when the flag is set + CHANGELOG disclosure (the current "wrapped exactly" claim overstates).
2. **(P2) `optimization.json` silently dropped** — the per-iteration history is built and never persisted; nothing else produces this artifact. Fix: the same `json.dumps` block `session.optimize` uses, when `scan_folder` is set.
3. **(P3) Post-preflight empty device set → post-claim `IndexError`** (burned scan number). Fix: the step body's own pre-claim `GeecsConfigurationError` guard, after `_preflight_unserved`.
4. **(P3) Dead tripwire**: the no-loader refusal test patches `claim_scan_number` but the optimize path claims via `claim_scan` — sabotage passed. Fix the patch target.
5. **(P3) `finish()` runs on failure/abort** — the Console consumer deliberately runs it only on success; a failed run's partial `xopt_dump.yaml` can poison warm-start seeding. Mirror the success-only rule (or document loudly with the other drops — mirroring preferred).
6. **(P3) No optimization-stack warm-up in the worker** — the first queued optimize item eats the tens-of-seconds torch/xopt cold import inside the plan preamble, looking hung. Fix: warm at startup when the loader is available (Console's `warm_up_optimization_stack` precedent).
7. **(P3, grant extension) CLAUDE.md staleness**: the layout tree's "optimize refused loudly (later round)" line is now false, and `optimization/` doesn't list `worker_loader.py`. The coordinator **extends the file grant to GeecsBluesky/CLAUDE.md for exactly these lines** (no collision — the sibling PR doesn't touch it). Also thread `applied_defaults` into the optimize body's md while in there.

### Noted, no action in this PR

Version 0.54.0 collision with #645 (coordinator resolves at merge); geecs-core lock-metadata refresh (benign re-lock side effect); the worker_loader ↔ Console twin needs a reciprocal keep-in-sync pointer on the Console side — follow-up chip; gen-list test only runs with the qserver extra (disclosed).

skbarber:
## Cross-vendor review addendum (Codex, relayed by the maintainer; reviewed head ad44f4f7, pre-fix-wave)

Findings 1, 2, 4 **independently confirm** disposition rows 1, 2, 6 (move_to_best ignored; optimization.json dropped; cold-import in the preamble) — treble-confirmed operator-facing regressions. Two extensions, both added to the worker's wave:

- **Row 1 is widened to the full `on_finish` ladder**: Codex notes the legacy path not only moves-to-best on success but **restores initial values on abort/failure** (session.py's on_finish semantics). The wave implements the ladder as `session.optimize` does, not just the success branch — end-of-run machine state must match the legacy path in all three outcomes.
- **Row 7 gains `db_scan_runtime`**: the optimize start-doc provenance must carry both `applied_defaults` and `db_scan_runtime` (legacy records both; the parity tests don't assert metadata on the optimize path — add that assertion).
- **NEW row 8 (P3, test strength): the startup import-order test is vacuous** — its autouse fixture imports `geecs_bluesky.session` before `runpy` executes the profile, so the load-bearing ordering contract can't fail in-process. Fix: assert the ordering in a **subprocess** (fresh interpreter, e.g. `python -c` running the profile with a sentinel/monkeypatched aioca probe), so a regression that imports the session above `import geecs_bluesky` actually fails.

Current wave for the worker: rows 1 (full ladder), 2, 3, 4, 5, 6, 7 (+db_scan_runtime, + optimize-md parity assertion), 8.

skbarber:
Review fix wave pushed (commit 9618def8).

**Rows addressed this pass:**
- Rows 1/2/3/5/7: test coverage additions/fixes on `test_scan_request_plan.py` (19 passed).
- Row 6: worker warm-up — `worker_loader.warm_up_optimization_stack()` pre-imports the heavy optimize-extra modules (`xopt`, `scan_analysis`, loader submodules) on a daemon thread as soon as a loader is registered, so the first real optimize call in a worker session doesn't eat that import cost in-plan. Wired into `startup.py` right after `set_optimization_loader()`, gated so it's a no-op without the `optimize` extra. New coverage in `tests/test_optimization_warmup.py` (4 passed), mirroring GEECS-Console's equivalent test.
- Row 7 (CLAUDE.md): fixed a stale "optimize refused loudly" line that no longer matched the implemented behavior, and added the new `worker_loader.py` to the `optimization/` directory listing.
- Row 8: rebuilt the qserver-startup import-order test as a subprocess probe (`tests/_qserver_startup_probe.py`). The prior in-process `runpy.run_path` version asserted the right thing but could never actually catch a regression — by the time any test function runs, `geecs_bluesky` and `aioca` are already cached in this test session's `sys.modules` from earlier collection, so "import order" was untestable in-process. The subprocess starts with an empty `sys.modules` and uses a `sys.meta_path` finder to record real first-import order plus the `EPICS_CA_ADDR_LIST` env state at the moment `aioca` first imports. Other two tests in that file are unchanged.

CHANGELOG.md amended under the existing `[0.54.0]` entry (this is still the same open PR, not a new release bump) with a bullet summarizing the above; no further `poetry version` bump.

Full `GeecsBluesky` suite: **673 passed, 1 skipped (qserver extra), 3 deselected**.

**New finding, not fixed here (out of scope for this branch):** `apply_experiment_defaults` in `scan_request_runner.py` builds `applied_defaults` dict keys like `"actions.setup"` (dotted). `event_model.compose_run`'s start-doc schema rejects dots not just in top-level keys but also in *nested* dict keys under a top-level key — so any real scan that actually applies an actions-based default (both the noscan/step path via `build_step_scan_spec` and the optimize path via `_optimize_request_body` funnel through the same helper) will raise `jsonschema.exceptions.ValidationError: 'actions.setup' does not match any of the regexes...` the moment it's recorded into run metadata. I hit this validating row-7 optimize-metadata behavior against a real `RunEngine` and worked around it test-side (spy-and-sanitize the dotted value out before it reaches `compose_run`, while still asserting on the real value) rather than fixing the shared helper, since that file is outside this branch's scope. Flagging for a coordinator call on where the fix belongs.

skbarber:
## Final disposition — all 8 rows confirmed, merging

The findings author re-verified the wave (commit 9618def8) on a clean clone with sabotage re-runs: the on_finish ladder matches session.py's three outcomes line-by-line (incl. the no-finite-objective fallback and halt semantics), optimization.json byte-shape-identical, the claim tripwire and the subprocess import-order probe both **fail under the sabotages that previously passed** — the probe correctly catches the realistic regression (aioca-importing module above `import geecs_bluesky`) even in no-extras environments. Suite 674/0/3, lint clean.

**Dotted-key `applied_defaults` bug (worker-discovered, pre-existing)**: confirmed at source — `apply_experiment_defaults` emits `actions.<slot>` keys that `event_model.compose_run` rejects, killing the run **post-claim** for any experiment that defines default bracket actions (none does yet — why production hasn't hit it). **Correctly deferred** (fix spans scan_request_runner's producer + two readers + md consumers, outside this grant); filed as its own issue with the reviewer's analysis.

Residual notes accepted: `_json_safe` private reach-in (consistent with the fork-bar's established pattern), two cosmetic nits, optimize+pseudo+move-to-best untested (matches the standing deferred-smoke posture), Console twin pointer already chipped.

Merging into `feat/queueserver` with the version untangle (this PR's 0.54.0 → **0.55.0**; #645 holds 0.54.0). Live checkpoint remains OWED and is next.

skbarber:
**Live integration checkpoint — 2026-08-21 (VERIFIED — closes this PR's owed live rows: startup profile, optimize loader registration + seam, warm-up)**

- **Startup profile live:** `environment open` loads `startup.py` with **zero warnings/errors** once the host config.ini is complete; `RE` kept via `--keep-re`, `geecs_scan_request_plan` in the allowed list, Tiled/s-file/scan-log callbacks all proven by the runs on #639.
- **Optimize seam live (Scan004):** optimize-mode ScanRequest (S1H ∈ [0.95, 1.05] A, MaxCountsEvaluator on UC_TopView, `random` generator, 3 iterations, 2 shots/iter, strict, HTU-NoGas): `exit: completed`. **`optimization.json` written** — 3 iterations with random inputs inside the range and per-iteration `n_rows: 2` — plus `xopt_dump.yaml`; the evaluator's auto-generated device requirements pulled UC_TopView into the Amp4In device set (both cameras' folders + s-file columns present). Objective −0.0 throughout (TopView dark under the no-gas profile — the machinery, not the physics, was under test). `move_to_best_on_finish` stayed off (deferred-untested posture unchanged).
- **Registration gotcha (deployment fact):** the loader registers only if the optimize extra is importable **in the manager's process tree** — a manager started *before* `poetry install -E optimize` kept refusing optimize requests (worker inherits stale import state) even after `environment close/open`. A full manager restart fixed it. Filed into `qserver/deploy/DEPLOYMENT.md` on the branch together with a required-config.ini-keys reference (`GEECS_DATA_LOCAL_BASE_PATH`, `[tiled] uri`, `scan_analysis_configs_path` were each discovered missing the hard way; the optimizer path additionally *requires* `scan_analysis_configs_path`).
- **Pre-existing configs-repo issue (not this PR):** `scan_analysis_configs/analyzers/UNCLASSIFIED/UC_Amp4_IR_input.yaml` fails current ImageAnalysis config validation (7 `extra_forbidden` errors, e.g. `thresholding`) — optimize requests naming that diagnostic fail pre-claim with a clear message. UC_TopView's config validates fine. Needs a configs-repo migration pass, tracked separately.

Issue #641, opened 2026-08-21
qs/W3: pause semantics — quiesce-on-pause, checkpoint placement, failed-move retry (queueserver round 2)


skbarber:
**Campaign**: queueserver migration, Round 2. **Worker**: Claude Code terminal session (parallel-safe with the W2 startup task — disjoint file ownership below). **Base + PR target**: `feat/queueserver`. Coordinator merges; post your adversarial review + disposition on the PR before reporting ready.

## Context (binding reading, in order)

1. `Planning/cutover_strategy/02_queueserver_migration.md` — decisions 1 and 4, **and the 2026-08-20 sandbox-spike amendment with its scoping note**: deferred pause waits out blocking puts and replays nothing; **hard pause (`bps.pause()`, `re pause immediate`) resumes by replaying from the last checkpoint — a replayed GEECS set re-executes and hardware re-moves**. Checkpoint placement governs latency for deferred pause but *what replays* for hard pause.
2. `GeecsBluesky/CLAUDE.md` — the acquisition-modes architecture, the shot-control bracketing (arm/disarm/quiesce), and the documented pause-latency floor.

## Task

Implement decisions 1 + 4 in the plan layer:

1. **Quiesce-on-pause**: when a scan pauses (either verb), the trigger must stop. Today's engine quiesces via the pause supervisor (dying at Round 3); the queue world needs it plan-level. Evaluate the two candidate mechanisms and implement the better one, documenting why: (a) a `SupplementalData`/suspender-style pause hook if bluesky offers a clean seam, or (b) structuring per-step bracketing so the paused state is always trigger-quiescent (the existing arm→shots→disarm bracketing already means pause-at-checkpoint lands in a disarmed window for free-run — verify and pin that claim rather than assuming it; strict mode is already quiescent between shots by construction).
2. **Checkpoint placement audit for hard-pause replay idempotence**: walk `build_step_scan_plan` / `geecs_step_scan` / `geecs_free_run_step_scan` checkpoint sites; for each, answer "if an immediate pause lands after this checkpoint and resume replays from it, what re-executes?" Moves re-executing to the SAME absolute target = acceptable; re-fired shots duplicating event rows or double-executed relative/stateful segments = not. Fix placements that fail the test; document the audit table in the PR body.
[...]

PR #645, opened 2026-08-21
GeecsBluesky 0.54.0: qs/W3 pause semantics — quiesce-on-pause, hard-pause replay audit, failed-move retry (#641)

skbarber:
Implements #641 (qs/W3: pause semantics — quiesce-on-pause, checkpoint placement, failed-move retry). Base and PR target: `feat/queueserver`. Per the Planning doc's task split, this PR owns `orchestration.py`, `step_scan.py`, `free_run_step_scan.py`, the new `plans/pause_semantics.py` helper, their tests, `pyproject.toml`, `CHANGELOG.md` — it does **not** touch `plans/scan_request_plan.py` or `qserver/**` (parallel task).

Designed from the Planning doc's 2026-08-20 sandbox-spike amendment and verified directly against installed bluesky 1.15.0 source: **deferred pause** waits for the next checkpoint and replays nothing; **hard/immediate pause** replays from the last checkpoint, and a replayed GEECS `set` re-executes — hardware re-moves. That corollary drives both halves of this PR.

## 1. Quiesce-on-pause

`ShotControlPauseQuiescer` (`plans/pause_semantics.py`) implements bluesky's `Pausable` protocol (`async pause()` / `async resume()`). The RunEngine awaits `Pausable.pause()` on every object it has "seen" (via any message referencing it) while entering the paused state, and awaits `Pausable.resume()` on all of them *before* replaying cached messages on resume — so this is the correct seam to force shots off during a pause, independent of where in a plan the pause lands.

- `register()` yields a no-op `Msg("null", self)` so the RE sees the quiescer; wired into `build_step_scan_plan` in `orchestration.py` via a `_with_pause_quiescer` wrapper, registered before the wrapped plan runs.
- `QUIESCE_FROM = {SCAN, STANDBY}`: on pause, if `controller.last_state` is one of these, drive to `OFF` and stash the prior state; on resume, re-assert the stashed state. `ARMED` (strict-shot-control's baseline) is a no-op — already quiescent by construction.
- If `OFF` isn't defined for the controller, log a warning and skip quiescing rather than raising — `pause()`/`resume()` never raise, since a `Pausable` exception mid-pause would leave the RE in an inconsistent state.
- **Transitional coexistence with the console's `PauseSupervisor`** (dies at Round 3): while both are live on the bridge/console path, a pause double-quiesces (redundant `OFF`/restore writes on both sides) and the supervisor's restore is no longer a safety net for a plan-driven pause. End state is provably correct either way; documented in `pause_semantics.py` rather than fixed, since it's inherently temporary (cross-vendor review, #645).

## 2. Checkpoint-placement audit

Every `bps.checkpoint()` call was audited against "what would a hard pause replay from here" by reimplementing the RE's uncacheable/rewind cache logic as a test helper (`_assert_no_replayable_completed_row` in the new test file) and running it over the full message stream of both step plans.

| Segment | Replays on hard pause | Verdict |
|---|---|---|
| Plan setup (stage/quiescer-register) | n/a — before first checkpoint | OK, no data yet |
| Pre-move (existing) | The upcoming move | OK — re-commanding a position is idempotent |
| Post-move (NEW) | Nothing (already at checkpoint) | Closes a window: previously a pause after move-before-shots would replay the move on resume |
| **Post-per_step (NEW, #645 cross-vendor addendum, P1)** | Nothing (already at checkpoint) | Closes a window: previously a pause after `per_step()`'s compiled ActionPlan writes (`bps.abs_set` calls — not guaranteed idempotent like an absolute move) but before arm would replay those completed writes |
| Pre-row (existing) | The upcoming row's shots | OK — shots haven't fired yet |
| Post-rows / pre-disarm (NEW) | Nothing | Closes a window: previously a pause after the last row but before disarm (strict) / disarm+advance (free-run) would replay completed shot events |
| Post-flush (NEW, free-run only) | Nothing | Closes a window after the tail create/read/save flush |
| Strict-mode end | n/a — final checkpoint before unstage | OK |

Three residuals, all irreducible at checkpoint-placement granularity and documented in code comments, out of scope for this PR:
- **Mid-shot bounded refire (strict mode):** a hard pause landing between the pre-row checkpoint and the next checkpoint — i.e. during a shot's own create/read/save — replays from the pre-row checkpoint, re-firing that one shot. Bounded to a single row, but it is a real re-fire, not a no-op.
- **The save→checkpoint instant:** the two-message gap between a row's `save` completing and the following checkpoint executing is not itself checkpointed (bluesky has no finer granularity). A hard pause requested in that exact window resumes from the *previous* checkpoint, which still holds the just-completed row — so it **can** replay a completed row and duplicate its event, not merely re-arm with nothing at stake. Width is two messages; not closable without a checkpoint between every message pair, which the RE doesn't support.
- **Mid-action residual (per_step):** a hard pause landing *during* `per_step()` itself (mid-action, before it yields back to the loop) still replays from the post-move checkpoint through the partial `per_step` execution — same class as the two residuals above, not closable by checkpoint placement alone (would need per-action idempotence in the ActionPlan compiler, out of this module's scope).

## 3. Failed-move → pause → retry

`move_with_failed_move_pause` (`plans/step_scan.py`, kept in the owned file rather than `pause_semantics.py` per module ownership) wraps `move_changed_axes`: on `FailedStatus`, logs `FAILED MOVE - pausing for operator` with the failing axes/targets and the exception cause (bluesky never attaches the failing device to the status object itself, so the cause — e.g. `GeecsMotorTimeoutError.device_name`/`variable` — is the reliable attribution point when multiple axes moved concurrently), then `bps.pause()`s. Resuming replays the move from the last checkpoint (decision 4). If the replayed move also fails, the loop catches the re-thrown `FailedStatus` and pauses again — indefinitely, until the move succeeds or the operator stops the scan.

**`failed_move_policy: Literal["raise", "pause"] = "raise"`** gates this path, threaded from `build_step_scan_plan` down through both step plans. Default `"raise"` preserves exact pre-#641 behavior (propagate `FailedStatus`, no pause) for any caller that doesn't opt in — notably `scan_request_plan.py` (the bridge/console path, untouched by this PR), which never enters the new pause path. This closes a coordinator-review finding: the console's `PauseSupervisor` auto-resumes a pause it doesn't recognize as operator-intentional, which would otherwise turn a failed move into an unbounded silent retry loop (and, per a cross-vendor follow-up, also let a Stop issued mid-pause bypass the finalize chain via the supervisor's park/resume logic). With the default `raise`, the bridge path is unaffected by either issue; wiring `'pause'` into `scan_request_plan.py` for the queueserver path is a follow-up owned by the coordinator once the sibling PR merges.

## Tests

- `tests/test_pause_semantics.py`: replay-cache audit over both plans' full message streams; `ShotControlPauseQuiescer` behavior (hard/deferred pause parity, ARMED no-op, missing-OFF warning, registration order); flagship `test_hard_pause_mid_move_resume_replays_only_the_move` (threads a real `RE.request_pause(defer=False)` mid-move, asserts the move re-fires exactly once and event/seq_num integrity holds); failed-move retry tests (single failure, double failure, stop-ends-gracefully); `test_failed_move_policy_defaults_to_raise_like_legacy` (pins the default — `FailedStatus` propagates, single attempt, no pause); message-level pins for the new post-per_step checkpoint in both the strict and free-run plans (`test_step_plan_checkpoints_immediately_after_per_step`, `test_free_run_plan_checkpoints_immediately_after_per_step`).
- `tests/test_pause_checkpoints.py` (updated counts for the checkpoints added by this PR, including the post-per_step checkpoint).
- `tests/test_grid_plans_message_level.py` (fixed a placement assertion that needed to skip checkpoints, not just find the nearest preceding message; already generic enough to tolerate the post-per_step checkpoint without further changes).

`./scripts/check.sh GeecsBluesky`: **800 passed, 1 skipped, 3 deselected**.

## LOC breakdown

- `plans/pause_semantics.py`: +190 (new + F3/P1 docstring additions)
- `plans/step_scan.py`: retry helper + `failed_move_policy` flag + 3 checkpoints + comments, ~+65
- `plans/free_run_step_scan.py`: switched to retry helper + `failed_move_policy` flag + 4 checkpoints + comments, ~+45
- `plans/orchestration.py`: quiescer wiring + `failed_move_policy` passthrough, ~+15
- `tests/test_pause_semantics.py`: +~750 (new file + F1/P1 pinning tests)
- `tests/test_pause_checkpoints.py`, `tests/test_grid_plans_message_level.py`: assertion fixes only

## Judgment calls

- `QUIESCE_FROM` includes `STANDBY` as well as `SCAN` — `STANDBY` is a shots-armed-but-idle state on some controllers; quiescing from it on pause errs toward safety.
- Quiescer failures are logged and swallowed, never raised — a broken quiesce shouldn't prevent the RE's own pause/resume machinery from completing.
- Failed-move retries are unbounded (operator-driven), matching decision 4's intent — no retry-count cap was requested by the issue.
- `failed_move_policy` defaults to `"raise"` rather than `"pause"` — the new pause/retry behavior only activates where a caller explicitly opts in, so every existing caller (in-grant tests aside) keeps its pre-#641 failure behavior unless and until it's deliberately switched over.
- `Pausable` is satisfied for both `pause()` and `resume()`; there's no separate `stop()` hook to implement — plan-level cleanup on stop is unaffected by this change.

Not yet exercised against live hardware. Owed at the next campaign integration checkpoint:
- Hard pause mid-move on a real GEECS motor: confirm the move re-issues on resume and the device actually re-settles at the commanded position.
- Deferred pause: confirm no replay occurs and the scan resumes exactly where it left off.
- Quiesce-on-pause: confirm `SCAN`→`OFF`→`SCAN` round-trip on a real `ShotController`-backed device during a live pause/resume, and that no stray shots fire while paused.
- Failed-move retry: force a real move failure (e.g. out-of-range setpoint) and confirm the operator sees the `FAILED MOVE` log line and that resume retries correctly, with `failed_move_policy="pause"` explicitly selected.

Per the issue, the coordinator merges this PR — not opening it pre-merged.

### DIFF
--- a/GeecsBluesky/CHANGELOG.md
+++ b/GeecsBluesky/CHANGELOG.md
@@ -32,20 +32,32 @@ Format follows [Keep a Changelog](https://keepachangelog.com/en/1.1.0/).
     pre-move checkpoint** — the retry, at the same absolute target
     (`pause` is uncacheable, so the pause itself never replays); a retry
     that fails pauses again; stop ends the run gracefully through the
-    finalize chain.
+    finalize chain.  Gated behind a new `failed_move_policy: "raise" |
+    "pause"` flag on `geecs_step_scan` / `geecs_free_run_step_scan` /
+    `build_step_scan_plan`, **default `"raise"`** — exact pre-#641
+    behavior, so the bridge/console path (which does not pass the flag)
+    is unaffected, sidestepping both the coexisting pause supervisor's
+    auto-resume-on-failed-move loop and the related stop-from-paused
+    bypass (cross-vendor review on #645; the queueserver-side opt-in wires
+    up separately once `plans/scan_request_plan.py`'s parallel PR merges).
 
 ### Changed
 
 - **Checkpoint placement is now a hard-pause replay contract** (the #641
   audit): both step plans add a post-move checkpoint (a replayed per-step
-  action prefix no longer re-runs the move) and a post-rows checkpoint (a
-  bin's last completed row can no longer replay into the disarm window —
-  previously a hard pause there duplicated the event row and re-fired the
-  shot); the free-run plan adds a post-flush checkpoint (the tail-flush
-  event can no longer replay into the finalize chain).  A rewind-cache
-  reconstruction walk over both plans' message streams pins the invariant
+  action prefix no longer re-runs the move), a post-per_step checkpoint
+  (a replayed arm can no longer drag a completed ActionPlan write along
+  with it — cross-vendor review on #645, item P1; the residual mid-action
+  pause is documented as irreducible, same class as strict's bounded
+  refire), and a post-rows checkpoint (a bin's last completed row can no
+  longer replay into the disarm window — previously a hard pause there
+  duplicated the event row and re-fired the shot); the free-run plan adds
+  a post-flush checkpoint (the tail-flush event can no longer replay into
+  the finalize chain).  A rewind-cache reconstruction walk over both
+  plans' message streams pins the invariant
   (`tests/test_pause_semantics.py`); deferred-pause behavior is unchanged
   (still lands at the next checkpoint, still replays nothing).
+
 ## [0.53.1] - 2026-08-20
 
 ### Added
--- a/GeecsBluesky/CHANGELOG.md
+++ b/GeecsBluesky/CHANGELOG.md
@@ -4,6 +4,48 @@ All notable changes to `geecs-bluesky` are documented here.
 
 Format follows [Keep a Changelog](https://keepachangelog.com/en/1.1.0/).
 
+## [0.54.0] - 2026-08-20
+
+### Added
+
+- **Plan-level pause semantics for the queueserver world** (issue #641;
+  decisions 1 + 4 of `Planning/cutover_strategy/02_queueserver_migration.md`
+  — the pause supervisor's successors, ahead of its Round-3 deletion):
+  - **Quiesce-on-pause** (`plans/pause_semantics.py`):
+    `ShotControlPauseQuiescer` rides the RunEngine's `Pausable` device
+    notification — the one seam covering both operator pause verbs *and*
+    the in-plan failed-move pause, with `RunEngine.resume()` awaiting the
+    re-assert **before** the rewind replay runs.  Pausing from a
+    free-running standing state (`SCAN`/`STANDBY`) drives the shot
+    controller's `OFF` writes (profile order, each completing first);
+    resume re-asserts the interrupted state.  `ARMED` (strict) is
+    deliberately skipped — already quiescent by construction (pinned);
+    a trigger profile without `OFF` writes draws a loud warning instead
+    of a silent no-quiesce.  `build_step_scan_plan` registers the
+    quiescer as the composed plan's first message whenever it has a
+    controller — both front doors get it for free.
+  - **Failed-move → pause** (`move_with_failed_move_pause`, wired at both
+    step plans' move sites): a failed move status records its reason as
+    the documented `FAILED MOVE - pausing for operator: ...` ERROR line
+    (scan.log captures it via the root-logger handler) and issues a hard
+    `bps.pause()`.  Resume **replays the failed `set`/`wait` from the
+    pre-move checkpoint** — the retry, at the same absolute target
+    (`pause` is uncacheable, so the pause itself never replays); a retry
+    that fails pauses again; stop ends the run gracefully through the
+    finalize chain.
+
+### Changed
+
+- **Checkpoint placement is now a hard-pause replay contract** (the #641
+  audit): both step plans add a post-move checkpoint (a replayed per-step
+  action prefix no longer re-runs the move) and a post-rows checkpoint (a
+  bin's last completed row can no longer replay into the disarm window —
+  previously a hard pause there duplicated the event row and re-fired the
+  shot); the free-run plan adds a post-flush checkpoint (the tail-flush
+  event can no longer replay into the finalize chain).  A rewind-cache
+  reconstruction walk over both plans' message streams pins the invariant
+  (`tests/test_pause_semantics.py`); deferred-pause behavior is unchanged
+  (still lands at the next checkpoint, still replays nothing).
 ## [0.53.1] - 2026-08-20
 
 ### Added
--- a/GeecsBluesky/geecs_bluesky/plans/free_run_step_scan.py
+++ b/GeecsBluesky/geecs_bluesky/plans/free_run_step_scan.py
@@ -51,7 +51,7 @@ from geecs_bluesky.devices.scan_context import ScanContext
 from geecs_bluesky.devices.shot_id import ShotIdSupport
 from geecs_bluesky.plans.step_scan import (
     motor_md,
-    move_changed_axes,
+    move_with_failed_move_pause,
     normalize_motors,
 )
 from geecs_bluesky.plans.t0_sync import geecs_t0_sync
@@ -256,14 +256,20 @@ def geecs_free_run_step_scan(
         scan_event_index = 0
         previous: tuple | None = None
         for bin_number, pos in enumerate(_positions, start=1):
-            # Deferred-pause boundary (issue #552): a checkpoint before the
-            # move and before every row means request_pause(defer=True)
-            # lands with an empty rewind cache — resume replays nothing.
-            # A between-rows pause sits inside a SCAN window; the pause
-            # supervisor owns driving the trigger to a safe state there.
+            # Checkpoint placement is a pause contract (issues #552/#641) —
+            # the full rationale lives in geecs_step_scan (same layout).
+            # Free-run specifics: the pre-move checkpoint sits in the
+            # disarmed STANDBY window, but the per-row checkpoints sit
+            # INSIDE the SCAN window — a pause landing there (either verb)
+            # relies on the ShotControlPauseQuiescer (pause_semantics) to
+            # stop the trigger; plan structure alone cannot make those
+            # windows quiescent without whole-bin pause latency.
             yield from bps.checkpoint()
             if _motors and pos is not None:
-                previous = yield from move_changed_axes(_motors, pos, previous)
+                previous = yield from move_with_failed_move_pause(
+                    _motors, pos, previous
+                )
+            yield from bps.checkpoint()
             if per_step is not None:
                 # After the move, before arming: per-step actions run with
                 # the shot controller disarmed (outside the SCAN window).
@@ -289,6 +295,11 @@ def geecs_free_run_step_scan(
                     )
                 if scan_event_index == 1:
                     _t0_seed_check(detectors)
+            # Post-rows checkpoint: the bin's last completed row must never
+            # sit in a hard-pause replay landing in the disarm window
+            # (re-saving it would duplicate the event row) — see
+            # geecs_step_scan's placement rationale.
+            yield from bps.checkpoint()
             if disarm_trigger is not None:
                 yield from disarm_trigger()
         # End of scan: stop the trigger BEFORE the tail machinery — STANDBY
@@ -303,6 +314,10 @@ def geecs_free_run_step_scan(
             for dev in _read_devices:
                 yield from bps.read(dev)
             yield from bps.save()
+            # Post-flush checkpoint: a hard pause during the finalize chain
+            # must not replay the completed flush event (or the end quiesce)
+            # into the cleanup path.
+            yield from bps.checkpoint()
 
     _end_quiesced = {"done": False}
 
--- a/GeecsBluesky/geecs_bluesky/plans/free_run_step_scan.py
+++ b/GeecsBluesky/geecs_bluesky/plans/free_run_step_scan.py
@@ -41,7 +41,7 @@ from __future__ import annotations
 
 import logging
 import time
-from typing import Any, Callable, Iterable, Sequence
+from typing import Any, Callable, Iterable, Literal, Sequence
 
 import bluesky.plan_stubs as bps
 import bluesky.preprocessors as bpp
@@ -51,6 +51,7 @@ from geecs_bluesky.devices.scan_context import ScanContext
 from geecs_bluesky.devices.shot_id import ShotIdSupport
 from geecs_bluesky.plans.step_scan import (
     motor_md,
+    move_changed_axes,
     move_with_failed_move_pause,
     normalize_motors,
 )
@@ -98,6 +99,7 @@ def geecs_free_run_step_scan(
     enable_saving: Callable | None = None,
     t0_sync_window_s: float = 0.2,
     tail_flush: bool = True,
+    failed_move_policy: Literal["raise", "pause"] = "raise",
     md: dict[str, Any] | None = None,
 ):
     """Free-run step scan: rows paced by *reference*, contributors never block.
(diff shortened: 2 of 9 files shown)


skbarber:
## Adversarial review (fresh-context, three lenses: correctness / redundancy / placement)

**No blocking findings.**

### Correctness — verified
- **Pausable ordering**: confirmed against installed bluesky 1.15.0 source (`run_engine.py`) — RE awaits `pause()` on all `Pausable`s while entering the paused state, then awaits `resume()` on all of them **before** replaying cached messages. This is the ordering the quiescer design depends on, and it holds.
- **Checkpoint placement**: re-derived the RE's rewind-cache semantics independently and ran it over both plans' full message streams — no window exists where a hard pause could replay a completed event row. Every checkpoint (pre-move, post-move, pre-row, post-rows, post-flush) has a distinct, non-redundant purpose.
- **Failed-move retry loop**: confirmed the `FailedStatus` re-thrown into the plan generator at the `yield from bps.pause()` point is caught correctly on a second failed replay, and the loop continues to pause indefinitely rather than mishandling the second exception.
- **QUIESCE_FROM logic**: `{SCAN, STANDBY}` correctly excludes `ARMED` (strict mode, quiescent by construction — pinned by test); capture/restore on resume is correct and doesn't double-apply.

### Redundancy — none found
No duplicated logic between the quiescer and `ShotController`; no adjacent checkpoints with nothing replayable between them; no test duplicating existing coverage.

### Placement — sound, one alternative noted
`move_with_failed_move_pause` living in `step_scan.py` rather than `pause_semantics.py` was flagged as a legitimate alternative (colocate in `pause_semantics.py`, import into both step plans) but the reviewer agrees the chosen placement — co-located with `move_changed_axes`, respecting the file-ownership split from the issue — is sound and introduces no circular-dependency risk.

### Non-blocking observations (not correctness bugs)
1. **Missing-OFF misconfiguration** — if a trigger profile has no `OFF` state defined, the quiescer warns and continues without quiescing; the scan pauses with the trigger still running (unbounded orphan frames). The warning is loud but doesn't prevent the misconfiguration from reaching a live pause.
2. **Silent quiescer exception handling** — `pause()`/`resume()` swallow all exceptions by design (never raise, to avoid destabilizing RE's own pause/resume machinery); a hard failure (e.g. a CA timeout) surfaces later as a downstream trigger-wait timeout rather than immediately at the pause point.

### Disposition
- (1) and (2) are accepted as intentional design tradeoffs, not defects — both are already called out in this PR's description ("Judgment calls" section) and covered by the "Hardware verification — OWED" section for live validation. **Waived**, no code change.
- No other findings to disposition.

Recommendation from review: ready to merge (coordinator merges per issue #641 — not merging from this branch).

skbarber:
## Adversarial review (coordinator-run) — disposition

The reviewer verified the mechanism against the installed bluesky source (Pausable seam ordering, replay-cache contents at each pause site, retry-by-replay, stop-from-paused), independently re-derived the audit table via message-level traces for strict and free-run shapes, and sabotage-tested three ways — **all caught** (checkpoint removal, reason-line deletion, replay-scope change). File grant exact, no Round-3 dependencies, all existing pins intact (pause-latency floor, STANDBY window, record_interruptions). Suite: 798 passed on the PR head.

### Findings + disposition

**F1 (Med/High — coexistence): the console's PauseSupervisor auto-resumes the failed-move pause → unbounded silent retry loop** on any bridge scan while Round 2 coexists with the live console (reviewer reproduced: 5 automatic resumes, zero operator interaction; pre-PR a failed move failed loudly). **Disposition: gate via policy flag** — `build_step_scan_plan` (and the move wiring) gains `failed_move_policy: 'raise' | 'pause'`, **default `'raise'`** (exact legacy behavior — console scans unchanged), `'pause'` enabling the new path. The worker fixes this in-grant; the one-line `'pause'` wiring in `scan_request_plan.py` belongs to the coordinator **after both open PRs merge** (that file is under the sibling PR's active fix wave — no cross-PR file sharing). Tests: existing failed-move tests exercise `'pause'`; add one pinning that the default raises like legacy.

**F2 (Low): PR-body audit-table wording** understates two irreducible residual windows (the save→checkpoint instant that can replay a completed row; mid-row hard pause = bounded refire in strict). Code handles both honestly — fix is body wording to match the code's own comments.

**F3 (Low, transitional): double-quiesce + poisoned `entry_state` in the supervisor interplay** — end state provably correct, ~3 redundant gateway writes per console pause, supervisor restore no longer a safety net. **Disposition: comment in `pause_semantics.py` noting the interplay + a line in the Round-3 deletion notes**; no code change (dies with the supervisor).

**Nits**: CHANGELOG blank line; failed-move `detail` zips all grid axes — clarify the failing-axis attribution or note the cause-text disambiguation.

Worker wave = F1 flag + default-raise pin test, F2 body wording, F3 comment, nits. Confirm pass by the same reviewer; coordinator merges both PRs afterwards (0.54.0/0.54.1 version untangle at merge) and then wires the queueserver-side `'pause'` flag as a follow-up commit with its own confirm.

skbarber:
## Cross-vendor review addendum (Codex, relayed by the maintainer; reviewed head 4f0e9d0a, pre-fix-wave)

**P0 — independently confirms F1 and extends it.** Same PauseSupervisor auto-resume loop, plus a new consequence: **Stop-from-paused can be bypassed** on the bridge path (`stop_scanning_thread` deliberately skips `RE.abort()` when paused and relies on supervisor parking; the empty-window branch resumes instead, so an operator's Stop during a failed-move pause can replay the move rather than reach the finalize chain). *Disposition: covered by the F1 fix already commissioned* — with `failed_move_policy` defaulting to `'raise'`, bridge scans never enter the pause path, eliminating both the retry loop and the stop bypass for the entire transition window; on the queueserver path there is no supervisor and stop goes through the RE Manager. The stop-bypass mechanism is recorded here so the Round-3 deletion of the supervisor re-checks it.

**P1 — NEW (critical class): `per_step` action writes sit inside the hard-pause replay window.** The post-move checkpoint correctly bounds motor replay, but a hard pause landing between `per_step()` (compiled ActionPlan writes via `bps.abs_set` — potentially non-idempotent) and the pre-row checkpoint replays those completed action writes. The PR body scoped ActionPlan replay out; the issue's audit test explicitly puts non-idempotent device writes in the critical class. *Disposition: fix in the wave* — add a checkpoint **after `per_step()`** (both step and free-run plans) so the replay window excludes completed actions; the residual mid-action window (pause landing during `per_step` itself replays the in-flight action sequence) is irreducible checkpoint-granularity, same class as the documented bounded-refire — document it in the audit table and the module docstring, and pin the new placement with a message-level test.

Both added to the worker's wave (rows F1-ext, P1).

skbarber:
## Fix-wave confirm pass (same reviewer, head ea89e018)

Re-reviewed the fix wave against the coordinator's disposition and the cross-vendor addendum. **CONFIRMED — no remaining gaps.**

- **F1/P0**: `failed_move_policy: Literal["raise", "pause"] = "raise"` on `build_step_scan_plan` → threaded through both step plans, selecting `move_with_failed_move_pause` vs. `move_changed_axes`. Default `"raise"` verified to reproduce exact legacy behavior (untouched `scan_request_plan.py` never enters the pause path). Pinned by `test_failed_move_policy_defaults_to_raise_like_legacy`.
- **F2**: PR body's audit table and residuals section rewritten to honestly reflect the save→checkpoint two-message gap (can replay a completed row, not a no-op as previously worded) and the mid-shot bounded-refire window, plus the new P1 residual — matches the code's own comments.
- **F3**: transitional double-quiesce / poisoned `entry_state` coexistence with the console's `PauseSupervisor` documented in `pause_semantics.py`'s docstring, tagged to die at Round 3. Docstring-only, as scoped.
- **Nits**: CHANGELOG blank-line fixed; failed-move log line now attributes via `exc.__cause__` rather than the ambiguous per-axis `detail` string — confirmed correct, since bluesky's `FailedStatus` never attaches `.obj` to the failing device.
- **P1**: `bps.checkpoint()` added immediately after `per_step()` in both `step_scan.py` and `free_run_step_scan.py` (correctly ordered: checkpoint → move → checkpoint → per_step → checkpoint → arm_trigger). Irreducible mid-action residual documented in both module docstrings and the PR body's audit table. Placement pinned by `test_step_plan_checkpoints_immediately_after_per_step` and `test_free_run_plan_checkpoints_immediately_after_per_step`.

Suite: 801 passed, 3 deselected. Lint clean.

Ready for merge (coordinator merges per issue #641).

skbarber:
## Final disposition — all streams closed, merging

Three review streams on this PR: the worker's own reviewer, the coordinator's primary reviewer (findings author), and the cross-vendor lens. Consolidated outcome:

- **F1 + cross-vendor P0** (supervisor auto-resume loop + Stop-from-paused bypass): closed by `failed_move_policy` defaulting to `'raise'` — bridge/console scans byte-identical to pre-#641 behavior. Default pinned at BOTH layers (RE-path test with the kwarg omitted + a builder-signature assert); reviewer sabotage-verified each layer independently (flipping only the inner plan's default is caught by the RE path; flipping the builder's is caught by the signature assert).
- **Cross-vendor P1** (per_step action writes in the hard-pause replay window): closed by the post-per_step checkpoint in both plans; placement pinned by falsifiable tests (checkpoint-removal sabotage now fails the strict test on its own); the irreducible mid-per_step residual documented in the audit table + module docstring.
- **F2/F3/nits/N1-N3**: all applied and confirmed; the emitted failed-move log line verified verbatim against the documented contract during sabotage runs.

Suite: **801 passed, 3 deselected** on the merge head.

**Standing coordination item (owed, coordinator's):** decision 4 ships dormant — the queueserver opt-in (`failed_move_policy='pause'` in `scan_request_plan.py`'s wiring) lands as a coordinator commit after #644 merges, with its own confirm. Live pause/resume/stop exercise owed to the integration checkpoint.

skbarber:
**Live integration checkpoint — 2026-08-21 (VERIFIED — closes this PR's owed live rows: quiesce-on-pause, checkpoint placement, stop-from-paused)**

1D step scan (Scan002: S1H 0.95→1.05 A, 5 positions × 3 shots, strict, HTU-NoGas), `qserver re pause deferred` issued mid-scan:

- **Deferred pause landed exactly per the sandbox-spike model:** the request arrived during the move to 1.0 A; the RE stayed `running` until the move completed, then `pausing → paused` at the **post-per_step checkpoint** (this PR's addition). scan.log shows move → pause with no shot in between.
- **Trigger quiescent while paused:** strict mode is plan-owned firing, so quiescence is structural — zero `SINGLESHOT` commands after `paused` in scan.log, and the image count stayed frozen (6 → 6 across the pause window).
- **Resume replayed nothing:** scan completed with **exactly 15 shots / 15 images** (header+15 rows in the s-file) — no duplicated set, no re-fired shot, `exit: completed`, clean STANDBY finalize.
- **Stop-from-paused (Scan003):** same scan paused at the same point, then `qserver re stop`: `exit_status: stopped`, **partial data preserved** (6 shots / 6 images), s-file pair written on stop (graceful stop keeps its data through the #635 gate), shot controller → STANDBY, manager straight back to idle. This is the current baseline for the Stop-from-paused path to re-check during W5 deletions (the cross-vendor P0 note on this PR).

Not exercised: hard/immediate pause (operator surface is the deferred verb; the hard-pause replay corollary stays documented in the Planning doc) and a pause during an optimize-mode scan (the 0.55.2 optimize-path quiescer ran only unpaused here). The `failed_move_policy='pause'` path itself wasn't triggered — no move failed live; its coverage remains the hermetic suite.

Issue #648, opened 2026-08-21
W6 — Console queueserver client: manager-API submission, pre-submit preflight, paused-state rendering, document-stream progress


skbarber:
## Context (read first, in this order)

1. `Planning/cutover_strategy/02_queueserver_migration.md` — decisions 3/4 and the **Console (client refactor)** scope section; the 2026-08-21 phase-complete amendment lists what is already live-verified on the worker side.
2. `GeecsBluesky/qserver/README.md` (esp. Troubleshooting) and `qserver/startup/startup.py` — the worker you are a client of, exactly as deployed.
3. `GEECS-Console/CLAUDE.md` — controller checklist (no Qt parent + dispose, contract travels, pin every fix) and existing service patterns.
4. PR #639/#644/#645 checkpoint comments — live-verified behavior your UI renders.

The engine side is DONE and live-verified (GeecsBluesky 0.55.3, feat/queueserver): `geecs_scan_request_plan` runs noscan/step/optimize through a qserver RE Manager. This issue is the console becoming a peer client of that manager.

## Scope

1. **Pre-submit preflight as plain GUI code** (decision 3): run `validate_scan_request` client-side at submit for immediate feedback (the worker re-runs it authoritatively — duplication is by design, see the scope doc's reinitialize amendment); unserved/liveness/staleness questions become ordinary synchronous dialogs *before* queueing; stamp outcomes into request metadata, not scan.log.
2. **Submission via the manager API** — use `bluesky-queueserver-api` (Python dicts in, no serialization games). Do NOT shell out to the `qserver` CLI: it parses payloads as Python literals (JSON null/true/false rejected) and its `history get` truncates with a literal `'...'` — both documented in qserver/README.md Troubleshooting.
3. **Queue semantics the client MUST handle:** on plan failure the manager returns the failed item to the **front** of the queue (default `ignore_failures: false`). A client that add-and-starts after a failure re-runs the failed item. Surface the failed item and clear/remove it explicitly on resubmit.
4. **Paused-state rendering** (decision 4): render `re_state: paused` with the recorded failed-move reason (the queueserver path runs `failed_move_policy='pause'` — see #645; do not change that asymmetry). The device-error Continue/Abort dialog becomes a rendering of paused + the stock verbs: resume (retry from checkpoint) / stop (graceful end, partial data preserved — live-verified Scan003 2026-08-21).
[...]

PR #650, opened 2026-08-21
GEECS-Schemas 0.10.0: ScanRequest.submission provenance record (#648)

skbarber:
## What

Adds the optional `submission` field to `ScanRequest` — the vehicle for W6 (#648) decision-3 provenance. `SubmissionRecord` carries which client queued the request and when (free-text client id + ISO-8601 timestamp), plus a list of `PreflightOutcome`s (`check`, `result`: `passed`/`continued`/`skipped`, `detail` in the operator's words). All fields default; the engine never acts on the record — a request without one runs exactly the same. Saved presets leave it unset.

## Why

Under the queue, preflight dialogs move client-side pre-submit (migration doc decision 3). The record of "operator was asked and chose to continue" must travel with the queued item into run metadata — but `ScanRequest` is `extra="forbid"` and the queue-facing plan signature is a bare `request: dict`, so provenance needs a real schema field. Maintainer approved the schema-field approach 2026-08-21.

- An aborted submission is never queued, so `aborted` is deliberately not a `PreflightCheckResult` value.
- The worker-side `md["submission"]` recording lands in the next PR (GeecsBluesky worker enablers), which depends on this one.

## Scope / LOC

One concern. ~90 LOC of models + descriptions in `scan_request.py`, exports, ~50 LOC tests, regenerated golden (`focuscan_scan_request.json` gains `"submission": null`) and generated `docs/geecs_schemas/schema_reference.md` (+22).

## Tests

`./scripts/check.sh`: lint OK; GEECS-Schemas suite passed (261 passed, corpus-integration test deselected as in CI — its pre-existing stale count floor is noted separately); GeecsBluesky 454 passed, 30 skipped, 2 deselected.

## Hardware verification

Not applicable — schema-only, no scan-execution or device surface. The field's end-to-end flow (client stamp → run metadata) gets exercised in the W6 console-client live check (maintainer's, post-merge, per #648).

## Deployment sequencing (review finding 1)

The interim-host worker (running geecs-schemas 0.9.1) must be **pulled + manager-restarted before any client starts stamping `submission`**: `SchemaModel` is `extra="forbid"` and schema validation happens inside the plan at execution, so a stamped request against an old worker dequeues, fails validation, and returns to the **front** of the queue (the #648 item-3 re-run trap, triggered by validation instead of a plan failure). Reverse skew (new worker, old client) is safe — the field is optional. An environment reopen is NOT enough; restart the manager process (same empirical rule as the extras install, see qserver/README.md).

### DIFF
--- a/GEECS-Schemas/CHANGELOG.md
+++ b/GEECS-Schemas/CHANGELOG.md
@@ -5,6 +5,19 @@ All notable changes to GEECS-Schemas are documented here.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [0.10.0] - 2026-08-21
+
+### Added
+
+- `ScanRequest.submission` — optional submission-provenance record
+  (queueserver migration, issue #648 decision 3): `SubmissionRecord`
+  carries which client queued the request and when, plus the pre-submit
+  preflight outcomes (`PreflightOutcome` /
+  `PreflightCheckResult`: passed / continued / skipped). Filled in by the
+  submitting client at queue time, copied into run metadata by the
+  engine, never acted on — a request without one runs exactly the same.
+  Saved presets leave it unset.
+
 ## [0.9.1] - 2026-08-20
 
 ### Changed
--- a/GEECS-Schemas/geecs_schemas/__init__.py
+++ b/GEECS-Schemas/geecs_schemas/__init__.py
@@ -36,9 +36,12 @@ from geecs_schemas.scan_request import (
     PositionList,
     PositionRange,
     Positions,
+    PreflightCheckResult,
+    PreflightOutcome,
     ScanAxis,
     ScanRequest,
     ScanRequestMode,
+    SubmissionRecord,
 )
 from geecs_schemas.scan_variables import (
     CompositeMode,
@@ -70,6 +73,9 @@ __all__ = [
     "OptimizationSpec",
     "EvaluatorSpec",
     "GeneratorSpec",
+    "SubmissionRecord",
+    "PreflightOutcome",
+    "PreflightCheckResult",
     # derived_channels
     "DerivedChannels",
     "DerivedChannel",
--- a/GEECS-Schemas/geecs_schemas/scan_request.py
+++ b/GEECS-Schemas/geecs_schemas/scan_request.py
@@ -360,6 +360,89 @@ class OptimizationSpec(SchemaModel):
         return self
 
 
+class PreflightCheckResult(str, Enum):
+    """How one pre-submit check ended.
+
+    Attributes
+    ----------
+    PASSED : str
+        The check found nothing to ask about.
+    CONTINUED : str
+        The check raised a question and the operator chose to continue
+        anyway.
+    SKIPPED : str
+        The check could not run (for example the database was unreachable)
+        and submission went ahead without it.
+    """
+
+    PASSED = "passed"
+    CONTINUED = "continued"
+    SKIPPED = "skipped"
+
+
+class PreflightOutcome(SchemaModel):
+    """One pre-submit check and how it ended, kept for the scan's record.
+
+    A check that would have stopped the submission never produces one of
+    these — an aborted submission is never queued, so there is nothing to
+    record.
+    """
+
+    check: str = Field(
+        description=(
+            "Name of the pre-submit check, e.g. 'unserved_variables', "
+            "'gateway_liveness', 'free_run_staleness'."
+        )
+    )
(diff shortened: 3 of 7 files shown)


skbarber:

Verified independently before writing findings: re-ran the GEECS-Schemas suite in the worktree (261 passed; the single failure is the pre-existing `test_corpus_integration` count-floor the PR body already discloses, deselected in CI). The golden change is forced by a full `model_dump(mode="json")` comparison, and `docs/geecs_schemas/schema_reference.md` is guarded by the no-drift CI test (`tests/test_schema_reference.py`), so both regenerated files are mechanically honest. `schema_version` staying at 1 matches the established precedent for additive optional fields (axes, save-sets union, `optimization` were all added without a bump). Version bump minor + CHANGELOG ride together per root policy; intra-repo consumers use unconstrained path deps, so 0.10.0 breaks no resolution. Placement and the decision-3 alignment check out against `Planning/cutover_strategy/02_queueserver_migration.md` and issue #648.

### Findings (ranked)

1. **Low, operational sequencing — the running interim worker will bounce the first stamped request, and it fails in the worst spot.** `SchemaModel` is `extra="forbid"`, and the queue-facing plan signature is a bare `request: dict`, so schema validation happens inside the plan at execution, not at `queue_item_add`. Concrete scenario: the interim-host manager is still running at 98780aaa (geecs-schemas 0.9.1); a W6 console stamping `submission` submits fine, the item dequeues, `model_validate` rejects it with "extra inputs not permitted", and the manager returns the failed item to the **front** of the queue — the exact re-run trap issue #648 item 3 warns clients about, now triggered by a validation error rather than a plan failure. Nothing in the PR body or #648 says "pull + restart the worker before the console starts stamping." One sentence in the PR body or the #648 live-check notes closes this. (Reverse skew — new worker, old client — is safe: the field is optional.)

2. **Low — `submitted_at`'s documented contract ("ISO 8601 with timezone") is enforced nowhere.** The most likely client bug is precisely `datetime.now().isoformat()` — Python emits a *naive* timestamp with no offset — and it validates silently, so cross-machine time comparisons over run metadata would be quietly wrong in the way the description explicitly tries to forbid. `Optional[AwareDatetime]` would catch this at the client's build step (where loud failure is free) and still round-trips through `model_dump(mode="json")` as the ISO string the tests pin. The counter-argument — a plain `str` can never abort a queued item at the worker's authoritative re-validation over an informational field — is real; but that argues for a lenient worker, not for an unenforceable client contract. Flagging so the trade-off is chosen deliberately rather than by default.

3. **Nit — present-tense overpromise in shipped docs during the interim.** The field description, CHANGELOG, and the published `schema_reference.md` all state the engine "copies it into the run metadata," but that copying lands only in the next PR (declared in the PR body). Between the two merges, anyone reading the published reference and looking for `md["submission"]` in a run finds nothing. Fine if the enabler PR follows promptly; worth remembering if it stalls.

**Redundancy — no surviving findings.** Checked: `GeecsBluesky/geecs_bluesky/preflight.py` (the engine's `Passed`/`Aborted`/`Ask` dataclasses are control-flow outcomes, not serializable records — no overlap, and its check roster matches the example names in the new descriptions, so the vocabulary is grounded); `scan_request_runner.py`'s applied-defaults/save-set provenance (separate mechanism, separate keys); `operator_channel.py` (`ANSWER_*` constants are dialog transport, deleted by decision 3); repo-wide grep for `SubmissionRecord`/`PreflightOutcome`/`submission`/`provenance` in GeecsBluesky and GEECS-Console found no prior model. The enum's `continued`/`skipped` values have no engine equivalent to drift from.

**Placement — no surviving findings.** GEECS-Schemas is the only package an arbitrary queue client (script, agent) can import without dragging in GeecsBluesky, and since the queue item is a bare validated dict, a schema field is the only vehicle that travels with the item — the alternative homes (a GeecsBluesky preflight record, an out-of-band md channel) fail one of those two constraints.

skbarber:
## Review disposition

1. **Interim-worker sequencing trap — FIXED**: "Deployment sequencing" section appended to the PR body (pull + manager-restart before any client stamps `submission`; env reopen insufficient; reverse skew safe). Confirmed by a fresh-context verifier given only the finding text and the fix (the original reviewer's session had ended — substitution per /land).
2. **`submitted_at` unenforced tz contract — WAIVED**: deliberate choice for the lenient side of the reviewer's stated trade-off — an informational provenance field must never make the worker's authoritative re-validation refuse a queued scan, and one shared model can't be strict client-side but lenient worker-side. The console (the client we own) will stamp aware timestamps and pin that in its own tests (W6 client PR).
3. **Present-tense overpromise during the interim — WAIVED**: the enabler PR (worker `md["submission"]` recording) is already built and lands immediately after this one, same session.

No undispositioned findings.

PR #651, opened 2026-08-21
GEECS-Schemas 0.9.2: drift-tolerant corpus count in save-element test

skbarber:
## What + why

`TestFullCorpus::test_every_save_element_converts` pinned a hard floor on the legacy save_devices file count (`converted >= 70`, "71 legacy files at the time of writing"). The corpus is a **live** sibling GEECS-Plugins-Configs checkout, and legacy files are rewritten in place to the new schema as they migrate — three flipped around 2026-08-20 (see GEECS-Plugins-Configs#13), leaving 68 legacy files, so the test fails on every branch with no code change, and any new floor would just fail again as migration continues.

Fix: make the count assertion drift-tolerant — `assert converted > 0` (pins that the legacy-conversion path was actually exercised, i.e. the corpus glob found files and the loop ran; the failure message names both possible causes: broken glob, or migration complete). The **per-entry legacy-pin assertions are unchanged** (`db_scalars is False`, empty `at_scan_start`/`at_scan_end` on every converted entry) — they remain the real guard.

## Second concern (flagged for cheap veto): GeecsBluesky ruff-format fix

The pre-commit CI job failed on this PR's first push for changes that are **not in this PR's diff**: `scan_request_runner.py` and `tests/test_pause_semantics.py` were committed on feat/queueserver (817550da) unformatted, so every PR targeting the branch goes red. Included here as a formatting-only fix (ruff-format output verbatim, 2 files / 9 lines) with a GeecsBluesky 0.55.3 → 0.55.4 patch bump per convention. Veto = drop commit fd2ff692 and this PR stays red on pre-commit through no fault of its own.

## Scope (LOC per concern)

- **Corpus drift fix**: `GEECS-Schemas/tests/test_corpus_integration.py` — 1 assertion → drift-tolerant assert + comment (~9 lines); `pyproject.toml` + `CHANGELOG.md` bump 0.9.1 → 0.10.1 (re-versioned after merging the base forward: #650 landed GEECS-Schemas 0.10.0 mid-flight)
- **Inherited format fix**: `GeecsBluesky/geecs_bluesky/scan_request_runner.py` + `GeecsBluesky/tests/test_pause_semantics.py` — 9 lines mechanical ruff-format; `pyproject.toml` + `CHANGELOG.md` bump 0.55.3 → 0.55.4

## Tests

- `GEECS-Schemas` corpus integration suite against the live sibling checkout: **10 passed**
- `./scripts/check.sh --base origin/feat/queueserver` (GEECS-Schemas + dependent GeecsBluesky): **819 passed, 1 skipped, 3 deselected**
- Lint (pre-commit full hook chain): passed

## Hardware verification

Not applicable — test-only + formatting-only change, no scan-execution or device surface.

### DIFF
--- a/GEECS-Schemas/CHANGELOG.md
+++ b/GEECS-Schemas/CHANGELOG.md
@@ -11,9 +11,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - Test-only: `TestFullCorpus::test_every_save_element_converts` no longer
   pins a hard floor on the legacy save_devices file count (`>= 70`). The
-  corpus is a live GEECS-Plugins-Configs checkout and legacy files
-  disappear as they migrate to the new schema (68 remain as of this
-  change), so the count assertion is now drift-tolerant
+  corpus is a live GEECS-Plugins-Configs checkout and legacy files are
+  rewritten in place to the new schema as they migrate (68 remain as of
+  this change), so the count assertion is now drift-tolerant
   (`converted > 0`); the per-entry legacy-pin assertions are unchanged
   and remain the real guard.
 
--- a/GEECS-Schemas/CHANGELOG.md
+++ b/GEECS-Schemas/CHANGELOG.md
@@ -5,6 +5,18 @@ All notable changes to GEECS-Schemas are documented here.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [0.9.2] - 2026-08-21
+
+### Fixed
+
+- Test-only: `TestFullCorpus::test_every_save_element_converts` no longer
+  pins a hard floor on the legacy save_devices file count (`>= 70`). The
+  corpus is a live GEECS-Plugins-Configs checkout and legacy files
+  disappear as they migrate to the new schema (68 remain as of this
+  change), so the count assertion is now drift-tolerant
+  (`converted > 0`); the per-entry legacy-pin assertions are unchanged
+  and remain the real guard.
+
 ## [0.9.1] - 2026-08-20
 
 ### Changed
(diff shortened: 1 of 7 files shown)


skbarber:

Reviewed per the /land ritual against the live sibling corpus (GEECS-Plugins-Configs: 68 legacy + 4 new-schema files under `*/save_devices/`), the full test file, and repo versioning/scope policy.

**Reviewer verdict:** "the fix is correct and minimal … nothing blocks merge."

### Findings and disposition

1. **LOW — tripwire message conflated two failure causes** (broken glob vs. migration complete), and the comment/CHANGELOG said legacy files "disappear" when they are actually rewritten in place. → **Fixed** in 8d2feebb; **reviewer re-confirmed** (message now names both causes; wording matches the observed migration mechanism).
2. **NIT — PR-body arithmetic**: 71 → 68 is three migrated files, not two. → **Fixed** in the PR body; **reviewer re-confirmed** against the live corpus count.
3. **INFORMATIONAL — no action (deliberate)**: four sibling floors in the same class (`>= 2`, `>= 8`, `>= 12`, `>= 11`) share the pattern but are not currently drifting (the scan-variables migration keeps legacy pairs in place; no new-schema files in the other folders). Left alone per minimal-scope policy; noted so the pattern is recognized if one fails later.

### Attack lines attempted and refuted (reviewer's own refutations)

- "Test lost its teeth" (floor 70 → 1): partial-discovery regressions are still caught by sibling cross-experiment floors and a second walk of the same glob; dispatch regressions by the hermetic `TestSaveSetLoadDispatch` and `assert save_set.entries, path`.
- Redundancy: no other pin of the corpus count exists in the repo; no duplicate corpus-walk test.
- Placement/versioning: change lives in the only file owning the assertion; patch bump + "Fixed" category correct.

### Added commit fd2ff692 (GeecsBluesky format fix) — reviewer assessment

AST-compared before/after: **both files AST-identical, zero semantic change**; confirmed `ruff format --check` fails on the base branch's copies (so pre-commit CI is red for every PR targeting feat/queueserver) and passes after. Patch bump + CHANGELOG per policy. Advisory note: bundling means other queueserver PRs stay red until this merges — fine if this merges promptly.

No undispositioned findings.

PR #652, opened 2026-08-21
GeecsBluesky 0.56.0: worker-side enablers for the console queueserver client (#648)

skbarber:
## What

The worker half of W6 (#648), so the console client PR that follows is pure GUI code. Four concerns, each small:

1. **`md["submission"]` recording** (~15 LOC): the geecs-schemas 0.10.0 `ScanRequest.submission` provenance record is copied verbatim into run metadata at all three emission sites (runner step/optimize + plan optimize; the shared step-path builder covers both entry points). One helper, `metadata_submission()`. Closes the "engine copies it into run metadata" promise from #650.
2. **`geecs_run_action_plan(name)`** (~90 LOC incl. docstring): on-demand ActionPlan execution as an ordinary queue item — decision 2's replacement for `session.run_action`'s direct `RE(...)` call, which cannot run under a manager-owned RE. Reuses the in-scan action machinery (`compile_action_slot`, `prefetch_action_signals`, `_DeferredConnectFactories`, in-plan connects, finalize disconnect). No run opened, matching today's `run_action`. Manager signature validation pinned (the 0.55.3 bare-namespace lesson).
3. **ZMQ document publisher** (~25 LOC startup + ~20 LOC launcher): the startup profile publishes bluesky documents to a `bluesky-0MQ-proxy` the launcher starts alongside Redis (5567 in / 5568 out; `QS_DOC_PROXY*`/`QS_DOC_PUBLISH_ADDR` overrides; best-effort with a loud warning). **Brief correction, verified against installed bluesky-queueserver 0.0.25:** the #648 brief's "launcher already passes `--zmq-publish-console ON`" conflated two streams — that flag publishes captured stdout/stderr *text* (port 60625), never documents; 0.0.25 has no built-in document publishing. This publisher/proxy pair is the migration doc's "ZMQ document publisher" worker build item, previously unbuilt. README + DEPLOYMENT.md now spell out both streams and the port matrix.
4. **Manual-verb functions** (~50 LOC): `geecs_move_variable` / `geecs_describe_action` in the startup namespace for the manager's idle-only `function_execute`; `operator` group gains `allowed_functions: [":geecs_.*"]`. Maintainer decision 2026-08-21: the console's Actions menu and Movable panel re-home to the worker. `move_variable` is deliberately a *function*, not a plan — it moves outside the RE by design (manual-move lock, RE stays idle). Describe is served worker-side so previews reflect the worker's configs checkout, not a possibly-drifted client one; the trade (no mid-scan preview — functions are idle-only) is flagged in the README.

Judgment-call additions, cheap to veto: the `operator`-group `allowed_functions` widening; serving `describe_action` worker-side at all.

Also carries the one-hunk ruff-format cleanup of `scan_request_runner.py` from direct commit 817550da (committed without hooks); the sibling test-file hunk is its own preceding format-only commit.

## Tests

`./scripts/check.sh`: lint OK; GEECS-Schemas + GeecsBluesky suites — **696 passed, 3 deselected** (was 673+skips pre-extras; run here with `ca tiled qserver` extras all installed). New: submission-record start-doc pins on both entry points (+ absent-key case), `geecs_run_action_plan` execute/unknown-name/manager-signature pins, startup subprocess probe extended to the new namespace surface.

## Hardware verification

**OWED (maintainer, on the interim host, next live session):**
- Pull + restart the manager (env reopen insufficient — README rule), then verify a stamped `submission` request lands its record in the run start doc (Tiled or scan.log).
- `qserver queue add plan '{"name": "geecs_run_action_plan", "args": ["<some action>"], "item_type": "plan"}'` → the sets execute, queue history records the item.
- Console-output + document streams: `bluesky-0MQ-proxy` running, a scan produces documents on port 5568 (RemoteDispatcher smoke), beeps/progress deferred to the console PR's live check.
- `function_execute` of `geecs_move_variable` on a safe variable (e.g. U_S1H Current small delta and back) while idle; confirm refusal while a scan runs.

Deployment sequencing from #650 applies: worker updates before any client stamps `submission`.

### DIFF
--- a/GeecsBluesky/CHANGELOG.md
+++ b/GeecsBluesky/CHANGELOG.md
@@ -4,6 +4,32 @@ All notable changes to `geecs-bluesky` are documented here.
 
 Format follows [Keep a Changelog](https://keepachangelog.com/en/1.1.0/).
 
+## [0.56.0] - 2026-08-21
+
+### Added
+
+- Worker-side enablers for the W6 console queueserver client (#648):
+  - **`md["submission"]` recording** — a `ScanRequest.submission`
+    provenance record (geecs-schemas 0.10.0) is copied verbatim into run
+    metadata at all three emission sites (runner step/optimize, plan
+    optimize; the shared step-path builder covers both entry points).
+    `metadata_submission()` in `scan_request_runner` is the one helper.
+  - **`geecs_run_action_plan(name)`** — on-demand ActionPlan execution as
+    an ordinary queue item (decision 2: actions are queue items). Same
+    resolution/flattening/prefetch code paths as the in-scan action
+    slots, in-plan connects, finalize disconnect, no run opened. Manager
+    signature validation pinned (the 0.55.3 bare-namespace lesson).
+  - **ZMQ document publisher** — the startup profile publishes bluesky
+    documents to a `bluesky-0MQ-proxy` (launcher-started, 5567 in /
+    5568 out, `QS_DOC_PROXY*` / `QS_DOC_PUBLISH_ADDR` overrides,
+    best-effort with a loud warning). This is the GUI progress stream;
+    the manager's `--zmq-publish-console` stream is text, not documents.
+  - **Manual-verb functions** — `geecs_move_variable` /
+    `geecs_describe_action` in the startup namespace for the manager's
+    idle-only `function_execute` (the queueserver enforcement of the
+    session's "scan in progress" refusals); `operator` group gains
+    `:geecs_.*` allowed_functions.
+
 ## [0.55.3] - 2026-08-21
 
 ### Fixed
--- a/GeecsBluesky/CHANGELOG.md
+++ b/GeecsBluesky/CHANGELOG.md
@@ -22,13 +22,18 @@ Format follows [Keep a Changelog](https://keepachangelog.com/en/1.1.0/).
   - **ZMQ document publisher** — the startup profile publishes bluesky
     documents to a `bluesky-0MQ-proxy` (launcher-started, 5567 in /
     5568 out, `QS_DOC_PROXY*` / `QS_DOC_PUBLISH_ADDR` overrides,
-    best-effort with a loud warning). This is the GUI progress stream;
-    the manager's `--zmq-publish-console` stream is text, not documents.
+    best-effort; a startup TCP probe warns when no proxy answers, since
+    a zmq PUB connect succeeds silently). This is the GUI progress
+    stream; the manager's `--zmq-publish-console` stream is text, not
+    documents.
   - **Manual-verb functions** — `geecs_move_variable` /
     `geecs_describe_action` in the startup namespace for the manager's
-    idle-only `function_execute` (the queueserver enforcement of the
-    session's "scan in progress" refusals); `operator` group gains
-    `:geecs_.*` allowed_functions.
+    `function_execute` (foreground calls require an idle manager — the
+    queueserver enforcement of the session's "scan in progress"
+    refusals); `operator` group allows exactly those two functions.
+    Both queue plans now check the session's manual-move lock before
+    starting (the PR #597 mutual exclusion, carried to the queue path —
+    background-executed moves leave the manager IDLE while converging).
 
 ## [0.55.4] - 2026-08-21
 
--- a/GeecsBluesky/geecs_bluesky/plans/scan_request_plan.py
+++ b/GeecsBluesky/geecs_bluesky/plans/scan_request_plan.py
@@ -88,6 +88,7 @@ from geecs_bluesky.scan_request_runner import (
     compile_action_slot,
     make_scalar_policy,
     metadata_applied_defaults,
+    metadata_submission,
     merge_optimizer_device_requirements,
     prefetch_action_signals,
     resolve_movable_target,
@@ -103,7 +104,12 @@ from geecs_schemas import AcquisitionMode, ScanRequest, ScanRequestMode
 
 logger = logging.getLogger(__name__)
 
-__all__ = ["geecs_scan_request_plan", "set_plan_session", "set_optimization_loader"]
+__all__ = [
+    "geecs_scan_request_plan",
+    "geecs_run_action_plan",
+    "set_plan_session",
+    "set_optimization_loader",
+]
 
 #: Per-connect timeout for the in-plan strict batches.  Near-parity, not
 #: parity, with ``GeecsSession._connect``: there 20 s is the outer wall on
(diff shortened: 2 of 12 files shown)


skbarber:

Verified against: installed bluesky 1.15.0 / bluesky-queueserver 0.0.25 sources, the repo's session/runner/plan modules, and a full suite run (696 passed, 3 deselected — matches the PR claim). Findings ranked; everything below survived my own refutation attempts. Refuted concerns are listed at the end so they don't get re-litigated.

**1. MEDIUM (correctness / invariant): the manual-move mutual exclusion does not travel to the queue-plan path, and this PR is what makes the gap reachable.**
`GeecsSession` pins a deliberate contract (PR #597 review, `session.py:1418`): the manual-move lock "is what keeps a scan, an action, or a second move from starting mid-move" — enforced by `session.scan` (:750), `optimize` (:953), and `run_action` (:1332). Neither `geecs_scan_request_plan` nor the new `geecs_run_action_plan` checks `_refuse_if_manual_move`. Pre-PR that was moot: nothing in the worker process ever took the lock. This PR installs `geecs_move_variable` in the worker, so the collision now exists. Concrete scenario: a client calls `function_execute` on `geecs_move_variable` with `run_in_background=True` (permitted — the yaml cannot forbid the flag, and I verified in bluesky_queueserver 0.0.25 that only foreground execution requires `MState.IDLE`: `manager.py:1373`, `worker.py:692`; background functions run in a thread while the env is busy or idle). The manager stays IDLE during a 20–30 s motor move; `queue start` launches a scan or action immediately; the scan claims a number and starts driving hardware while the manual move is still converging on the same RE loop. Note the built-in guards cover only the *other* direction (`move_variable`'s `RE.state != "idle"` check refuses a move during a scan; foreground moves set `EXECUTING_TASK`, blocking queue start). Cheap fix: both plans call `session._refuse_if_manual_move(...)` at the top (the session is in hand at that point). Related docs bug in the same hunk: the README ("Function execution requires a fully **idle** manager") and both verb docstrings state the idle-only enforcement unconditionally — it holds only for `run_in_background=False`.

**2. LOW-MEDIUM (correctness of the stated failure posture): the promised "loud warning" for a missing document stream can never fire from the worker side, and a dead proxy is silent and unsupervised.**
`Publisher.__init__` (bluesky 1.15 `callbacks/zmq.py`) only creates a PUB socket and calls `connect()` — zmq connect is asynchronous and succeeds with nothing listening. So the `try/except` in `startup.py` can only catch import/address-parse errors; when the proxy is absent or dead, the code logs the *success* line and no warning ever appears. Concrete silent-death scenario: another process holds out-port 5568 while in-port 5567 is free → the launcher's guard (which probes 5567 only) passes → `bluesky-0MQ-proxy` starts, fails to bind 5568, and dies with its stderr discarded → zero diagnostics anywhere, GUI progress permanently empty. Also, DEPLOYMENT.md's "dying and restarting with the manager is correct behavior" oversells: systemd's `Restart=` supervises only the main process, so a proxy that dies alone is not restarted until the next manager restart. Suggested fixes: let the proxy's stderr go to the journal, and either soften the "loud warning" claims or make the warning real with a cheap TCP probe of the publish address at startup. Scans are genuinely unaffected either way — PUB sockets drop on mute/HWM, verified — so this is an observability gap, not a scan hazard.

**3. LOW (permissions scope): `allowed_functions: [":geecs_.*"]` matches more than the two verbs.**
The pattern also admits `geecs_scan_request_plan` and `geecs_run_action_plan` as `function_execute` targets (calling a generator function returns an unexecuted generator — no hardware effect, just a failed/garbage task result), plus any future `geecs_`-named callable in the namespace by default. Since the PR itself flags this widening as a cheap-to-veto judgment call: pinning to `":geecs_move_variable$"` / `":geecs_describe_action$"` costs nothing and keeps the security-boundary file explicit. (The `primary` group's admin-equivalent `":.*"` is pre-existing, out of scope.)

**4. NIT: the two OFF switches disagree on case.**
`startup.py` accepts any case (`.upper() != "OFF"`); the launcher's `[[ "${QS_DOC_PROXY:-ON}" != "OFF" ]]` is exact-match. `QS_DOC_PROXY=off` + `QS_DOC_PUBLISH_ADDR=off` leaves an idle proxy running that nothing publishes to. Harmless residue; one-line fix if touched.

**Redundancy — no surviving findings.** Checked: no prior document publisher anywhere (grepped `zmq`/`Publisher`/`RemoteDispatcher` across GeecsBluesky and GEECS-Console); no prior submission-md recording (all `md[...]` sites audited — the claimed three-site coverage is real and complete). `geecs_run_action_plan` vs `session.run_action` and `plans/action_direct.py`: three deliberate dispatches over the *shared* compile/flatten/prefetch machinery — the new plan reuses `compile_action_slot`, `prefetch_action_signals`, `_DeferredConnectFactories`, and `_disconnect_plan` rather than duplicating them; the failure-path teardown is complete.

**Placement — one informational note.** `geecs_run_action_plan` lives in `plans/scan_request_plan.py`, a module named and documented for the ScanRequest plan. Justified today by its reuse of that module's private helpers, but if actions-as-queue-items grows (args, per-step options, a run for provenance), a `plans/queue_actions.py` split will need those helpers made non-private. The manual verbs as startup-profile code (not package code) are fine — 2-line delegations, covered by the subprocess probe.

**Concerns raised and refuted by the reviewer** (recorded so the next reviewer doesn't repeat them): `bluesky-0MQ-proxy` is the real console-script name in bluesky 1.15 and the positional-port invocation works (deprecated, FutureWarning only); `Proxy(int)` binds all interfaces, so DEPLOYMENT.md's reachability claims are correct; port 60625 for `--zmq-publish-console` confirmed; the backgrounded proxy stays in the unit's cgroup across the launcher's `exec` and dies with the unit; an absent/slow proxy cannot stall or break scans (PUB drops when mute/full); `md["submission"]` is not duplicative; the new tests would fail if the code were wrong; version bump (minor) and CHANGELOG are correct; no public-repo-hygiene violations.

skbarber:
## Review disposition

All four findings **FIXED** in 48fcded4 and confirmed by a fresh-context verifier given only the findings and the fix commit (the original reviewer's session had ended — substitution per /land):

1. **Manual-move mutual exclusion (medium)** — both queue plans now call `session._refuse_if_manual_move(...)` at start; pinned by `test_queue_plans_refuse_while_a_manual_move_is_in_flight` (real lock, real refusal, the queue's exact call shape). Docs scope the manager's idle gate to foreground function execution.
2. **Silent missing proxy (low-med)** — startup TCP-probes the publish address (warning is now real; verifier exercised up/down/malformed paths — cannot break startup), launcher keeps proxy stderr attached, DEPLOYMENT.md states the dies-alone supervision gap.
3. **Over-broad allowed_functions (low)** — pinned to exactly the two verbs; verifier validated the regexes against bluesky-queueserver's own permission matcher (superstrings rejected).
4. **OFF-switch case (nit)** — launcher upper-cases before comparing.

The placement note (`geecs_run_action_plan` in `scan_request_plan.py`) is accepted as-is per the reviewer's own justification (reuses that module's private helpers); the future `plans/queue_actions.py` split threshold is recorded in the review comment above.

Also in this branch since review start: sync-merge of feat/queueserver (#651 landed the same ruff-format fixups independently — the duplicates fell out of the diff; version resolved to 0.56.0 with 0.55.4 kept in the CHANGELOG).

Suite after fixes: **697 passed, 3 deselected**. No undispositioned findings.

PR #653, opened 2026-08-21
GEECS-Console 0.21.0: queueserver client services layer (#648, W6 part 1)

skbarber:
## What

The console half of W6, part 1 of 2: the **services layer** for the queueserver client — built, tested, and deliberately **not yet wired into the window** (part 2 is the window switch: submission path, paused rendering, progress/beeps, bridge import removal). Splitting keeps each review tractable.

Three new modules, each following the console's standing seam conventions (protocol + real + stub, lazy imports import-safe offline, never-raises pollers, #534 controller checklist):

1. **`services/queue_client.py`** (~450 LOC with docstrings): the one module that speaks `bluesky-queueserver-api`. `ZmqQueueClient` — submission with the **failed-item-at-front guard** (#648 item 3: pending queue items are surfaced, cleared only on explicit `clear_pending=True`), deferred pause/resume, graceful **stop sequencing** (from paused → `re_stop` directly, the live-verified Scan003 path; from running → `queue_stop` + deferred pause + bounded wait + `re_stop`), `function_execute` manual verbs with task polling, never-raises `status()`. Config: new `[qserver]` section (`host` + optional `control_addr`/`info_addr`/`doc_addr`) in the shared config file, path via the blessed `ops_paths.USER_CONFIG_PATH` literal.
2. **`services/submit_preflight.py`** (~330 LOC): decision-3 client-side pre-submit — the engine's own `validate_scan_request` as the hard gate, the engine's `UnservedVariablesCheck` **reused not reimplemented** (no drift; DB-unknown reads as skipped, never blocking), `CONNECTED` liveness reads (fail-open — only the exact `"Disconnected"` reading counts), free-run staleness sample on the reference device. `stamp_submission` writes the geecs-schemas 0.10.0 `SubmissionRecord` with an **aware** timestamp (the #650 review's naive-timestamp bug, pinned by test). Pure check layer — dialogs stay in the window (part 2).
3. **`app/scan_monitor.py`** (~280 LOC): `ScanMonitorController` — manager status polling on the `HealthPoller` pattern, the document stream (`RemoteDispatcher` on a daemon thread, worker-owned signals), and the manager console-output stream with **failed-move reason extraction** (prefix pinned against the engine's `FAILED_MOVE_LOG_PREFIX` — the paused pill's "why" in part 2).

Dependency: `bluesky-queueserver-api ^0.0.13` (pulls pyzmq). Rider (#648 item 6): reciprocal keep-in-sync cross-reference on `services/optimization.py` (its worker twin already pointed here).

**W5 heads-up:** the preflight module is a live consumer of `geecs_bluesky.preflight` (`UnservedVariablesCheck`, `PreflightContext`, `Ask`/`Passed`) — W5's deletion pass must keep that module importable (`Ask.question`'s `OperatorQuestion` type will need a new home when `operator_channel.py` dies).

Judgment calls, cheap to veto: the `[qserver]` config section shape (host + derived standard ports); the 2 s staleness sample window (a submit click on free-run costs at most that); worker-side `describe_action` (idle-only) as the actions-preview source in part 2.

## Tests

`./scripts/check.sh`: lint OK; **GEECS-Console 836 passed** (39 new across `test_queue_client.py` / `test_submit_preflight.py` / `test_scan_monitor.py` — config reader, pending-guard, stop sequencing, task polling incl. failure/timeout, every preflight check's pass/question/skip paths, aware-timestamp pin, stream parsing, controller lifecycle); GEECS-Schemas + GeecsBluesky 697 passed, 3 deselected.

## Hardware verification

Nothing here executes against hardware yet (unwired services). The live check rides part 2's OWED list (submission end-to-end against the interim worker). Note the #650 deployment sequencing still applies: interim worker must be at ≥0.56.0 (pull + manager restart) before this client's stamped submissions run.

### DIFF
--- a/GEECS-Console/CHANGELOG.md
+++ b/GEECS-Console/CHANGELOG.md
@@ -4,6 +4,39 @@ All notable changes to GEECS-Console are documented here.  Format follows
 [Keep a Changelog](https://keepachangelog.com/en/1.1.0/); versioning is
 semantic.
 
+## [0.21.0] - 2026-08-21
+
+### Added
+
+- The queueserver client services layer (#648, W6 part 1 — built and
+  tested, not yet wired into the window; the window switch is the next
+  PR):
+  - `services/queue_client.py` — the RE Manager client seam
+    (protocol + `ZmqQueueClient` over `bluesky-queueserver-api` +
+    offline `StubQueueClient`): submission with the failed-item-at-front
+    guard (surface pending items, clear only on explicit confirm),
+    deferred pause / resume, graceful stop (from paused directly; from
+    running via pause-then-stop sequencing), `function_execute` manual
+    verbs with task polling, and a never-raises `status()` for polling.
+    Config: the new `[qserver]` section (`host` + optional address
+    overrides) of the shared config file.
+  - `services/submit_preflight.py` — client-side pre-submit checks
+    (decision 3): the engine's own `validate_scan_request` (hard gate),
+    the engine's `UnservedVariablesCheck` (reused, not reimplemented),
+    gateway `CONNECTED` liveness reads (fail-open), and a free-run
+    trigger-staleness sample; plus `stamp_submission`, which stamps the
+    geecs-schemas 0.10.0 `SubmissionRecord` (aware timestamp pinned by
+    test) onto the outgoing request.
+  - `app/scan_monitor.py` — `ScanMonitorController` (#534 controller
+    shape): manager status polling on the `HealthPoller` pattern, the
+    document stream (`RemoteDispatcher`, per-shot progress source), and
+    the manager console-output stream with failed-move reason
+    extraction (pinned against the engine's `FAILED_MOVE_LOG_PREFIX`).
+- New dependency: `bluesky-queueserver-api` (pulls pyzmq).
+- Cross-reference rider (#648 item 6): `services/optimization.py` now
+  names its worker-side twin (`geecs_bluesky.optimization.worker_loader`)
+  with the keep-in-sync rule; the twin already pointed here.
+
 ## [0.20.2] - 2026-08-20
 
 ### Changed
--- a/GEECS-Console/geecs_console/app/scan_monitor.py
+++ b/GEECS-Console/geecs_console/app/scan_monitor.py
@@ -54,19 +54,25 @@ def _failed_move_prefix() -> str:
 class DocumentStreamWorker(QObject):
     """Daemon-thread consumer of the worker's bluesky document stream.
 
-    Emits ``document(name, doc)`` for every document received.  The signal
-    is worker-owned; consumers connect it ``QueuedConnection``.  ``start()``
-    spawns the thread; :meth:`stop` makes further emissions impossible and
-    best-effort closes the dispatcher (never joins — the thread is daemon).
+    Emits ``document(name, doc)`` for every document received and
+    ``stream_failed(str)`` if the stream cannot be set up (bad address,
+    missing import) — post-setup, the zmq SUB socket reconnects on its own
+    and needs no supervision.  Signals are worker-owned; consumers connect
+    them ``QueuedConnection``.  ``start()`` spawns the thread; :meth:`stop`
+    only gates emission and **abandons** the daemon thread: the
+    dispatcher's zmq socket/loop must never be touched from another thread
+    (pyzmq sockets are not thread-safe — a cross-thread close can trip a
+    libzmq assertion, which *aborts* the process rather than raising; #653
+    review finding 3).
     """
 
     document = Signal(str, object)
+    stream_failed = Signal(str)
 
     def __init__(self, doc_addr: str) -> None:
         super().__init__()
         self._addr = doc_addr
         self._stopped = False
-        self._dispatcher: Any = None
         self._thread: Optional[threading.Thread] = None
 
     def start(self) -> None:
@@ -83,7 +89,6 @@ class DocumentStreamWorker(QObject):
             from bluesky.callbacks.zmq import RemoteDispatcher
 
             dispatcher = RemoteDispatcher(self._addr)
-            self._dispatcher = dispatcher
 
             def _forward(name: str, doc: dict) -> None:
                 if not self._stopped:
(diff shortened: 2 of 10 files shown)


skbarber:

Verified against: bluesky-queueserver-api 0.0.13 as installed (all `REManagerAPI` signatures the client calls — `item_add(item, user=...)`, `function_execute(item, user=...)`, `re_pause(option=...)`, `task_result(task_uid)`, constructor kwargs — match; `task_result` payload keys confirmed against the manager/worker sources; `ConsoleMonitor_ZMQ_Threads` constructor and `next_msg` timeout behavior match), the real engine APIs (`validate_scan_request`, `resolve_save_sets_and_rituals`, `save_set_to_devices_config`, `UnservedVariablesCheck`/`PreflightContext`/`Ask`/`Passed`, `GeecsDbServedSetProvider.served_by_device` incl. its caching, `ca_pv`/`bare_pv`, `FAILED_MOVE_LOG_PREFIX`), geecs-schemas 0.10.1, and the worker profile (plan/function names, permissions yaml, DEPLOYMENT.md ports). Shared-`REManagerAPI` thread-safety across the status poller and worker calls is fine: the sync ZMQ API rides `ZMQCommSendThreads`, which serializes requests behind `_lock_zmq`. The PR's 39 new tests pass locally; PR CI is green.

Findings that survived refutation, by severity:

**1. HIGH (correctness): the gateway-liveness check can never fire in production — `_read_pv` reads the CONNECTED enum as its integer index, not the choice string.**
`submit_preflight.py` calls `aioca.caget(pv, timeout=...)` with no `datatype`. `datatype=None` fetches the channel's *native* type; the gateway serves `CONNECTED` as a caproto `ChannelEnum` (DBR_ENUM), so the read returns 0/1 and `str(reading) == "Disconnected"` compares `"0"`/`"1"` against the string: always False. Every save-set device down → the check records `passed` and asks nothing; the fail-open doctrine masks the bug completely, and the tests mask it too (the patched fake returns strings). The engine's own reader avoids this deliberately (`epics_signal_r(str, ...)` in `triggerable.py`). Fix: `caget(pv, datatype=str)` for the CONNECTED read (leave `acq_timestamp` native).

**2. MEDIUM (correctness): `_submit_item` reports failure after the item is already queued — the "failed" scan or action then starts later on its own.**
One try/except wraps `queue_get → item_add → queue_start`. If `item_add` succeeds but `queue_start` raises (a refusal while an item is executing; a transient recv timeout), the operator is told the submission failed while the item sits queued and auto-runs later. For `submit_action` this is an *armed action executing after the operator was told it didn't*. Suggested shape: separate the add and start outcomes; best-effort `item_remove` the uid before reporting failure, or name the stuck item explicitly.

**3. MEDIUM-LOW (Qt/threading): `DocumentStreamWorker.stop()` tears down zmq objects from the wrong thread — the failure mode is an uncatchable abort, not an exception.**
`stop()` calls `dispatcher.stop()` from the GUI thread while the daemon thread is blocked inside `RemoteDispatcher.start()`: `_task.cancel()` cross-thread, `_socket.close()` + `context.destroy()` under a peer thread's pending recv (libzmq assertion = process abort, uncatchable), `loop.close()` on a running loop. Safer: don't touch the dispatcher cross-thread at all — abandon it (the thread is daemon and `_stopped` already gates emission).

**4. LOW (robustness): both stream workers are one-shot — a setup-time failure kills the stream for the console's lifetime with no UI-visible signal and no way to restart.** Worth at least a worker-owned `failed` signal the window can surface.

**5. LOW (thread-safety): `ZmqQueueClient._manager()` lazy init is unlocked** — the poller's first `status()` and an early worker-thread verb can race it and leak the loser's `REManagerAPI` zmq receive thread. One `threading.Lock` closes it.

**6. LOW (resource pattern, shared with existing code): `asyncio.run`-per-CA-read leaks aioca channel caches.** aioca keys its `ChannelCache` by event loop and never purges closed-loop entries; every `_read_pv` call leaks a cache entry + CA channel. This is the pattern `services/health.py` already established (every 5 s poll), so inherited debt — but the PR is the second copy. One shared helper (or the device panel's persistent-loop pattern) would serve both.

**7. LOW (docs contract): the new `[qserver]` config section is documented nowhere but the CHANGELOG.** The canonical config.ini reference (`docs/tutorials/getting_started.md`) and the qserver deploy docs don't mention the section operators must add. Fine to defer to the window-switch PR, but it should be owed explicitly there.

**Placement (flag, cheap veto — no change requested):** `queue_client.py` and `submit_preflight.py` are zero-Qt, pure manager-facing logic; the planned OSPREY GEECS-side scan MCP is exactly a second RE-Manager client that would need both verbatim and cannot import a GUI package — its existence would force these modules down into GeecsBluesky (e.g. `geecs_bluesky.qs_client`, with the api dep as an extra there). Console-side placement is defensible today; the move is queued behind the MCP. `scan_monitor.py` in `app/` matches the #534 controller precedent — right home.

**Redundancy — checked, clean otherwise:** no prior queueserver client anywhere in the repo; the engine checks are genuinely reused; the `_FAILED_MOVE_PREFIX_FALLBACK` twin is pinned by test. One minor overlap noted: `ConsoleStreamWorker` hand-builds a `ConsoleMonitor_ZMQ_Threads` although `REManagerAPI` owns an identically-configured `.console_monitor` — independent construction is defensible (separate socket, no coupling to the control client).

Verified non-findings recorded to save the next reviewer the trip: `validate_scan_request`'s discarded return is fine; defaults never add save sets; `served_by_device()` is cached; the stop sequencing and `queue_stop`-first ordering are correct; `PreflightOutcome` strings match the schema enum; the aware-timestamp test genuinely pins the tz contract; each address format matches its consumer.

skbarber:
## Review disposition

Findings 1–5 **FIXED** in 7d7c52d8, all confirmed by a fresh-context verifier given only the findings and the fix commit (the original reviewer's session had ended — substitution per /land):

1. **CONNECTED enum read (HIGH)** — `_read_pv` gained `datatype`; the liveness read passes `datatype=str`; the test fake is enum-faithful and the verifier traced that dropping the fix now fails the disconnected-device test.
2. **Item left queued on start failure (medium)** — add/start outcomes separated; best-effort `item_remove` with honest messages either way; two new tests.
3. **Cross-thread zmq teardown (med-low)** — `stop()` only gates emission; the dispatcher is abandoned with its daemon thread, rationale in docstrings.
4. **Silent stream death (low)** — both workers emit `stream_failed(str)` on setup failure.
5. **Unlocked lazy init (low)** — `threading.Lock` around `_manager()`.

6. **aioca loop-cache leak (low, shared with health.py)** — **WAIVED here, chip filed**: pre-existing pattern inherited from the health probe; a background-task chip proposes the shared leak-free CA-read helper covering both call sites (the fix belongs in one place, not a third copy of the pattern).
7. **\[qserver\] docs (low)** — **DEFERRED as owed**: the getting-started config.ini reference gains the `[qserver]` section in the window-switch PR (recorded in that PR's scope).

Placement flag (client modules may move to GeecsBluesky when the OSPREY scan MCP becomes a second manager client) — **accepted as informational**, recorded for that future work.

Suite after fixes: **838 passed**. No undispositioned findings.

Issue #649, opened 2026-08-21
W5 — GeecsBluesky: delete the bridge machinery (BlueskyScanner, action_direct, OperatorChannel, pause_supervisor remnants)


skbarber:
## Context (read first, in this order)

1. `Planning/cutover_strategy/02_queueserver_migration.md` — the **Delete** scope section and decisions 2/3/4 (what replaced each deleted seam), plus the 2026-08-21 phase-complete amendment.
2. `GeecsBluesky/CLAUDE.md` — the **do-not-fix pins** (e.g. the pause-latency floor documentation). Deleting code must not 'fix' a pinned behavior or delete its documentation.
3. PR #645's cross-vendor P0 note and its 2026-08-21 checkpoint comment — the Stop-from-paused baseline this issue must preserve.

**Ordering: this issue lands only after W6 (#648) merges** — the console must stop importing the bridge before the bridge dies.

## Scope (deletions)

- `geecs_bluesky/scanner_bridge/bluesky_scanner.py` — BlueskyScanner thread/state-flags/progress machinery + its tests (they pin GUI-thread semantics and die with the bridge). The RE Manager's queue/status API *is* this functionality now.
- `geecs_bluesky/plans/action_direct.py` and the three-way `ActionDecisionRequest` dialog (`events.py`) (decision 2: pause-window actions dropped; actions are ordinary queue items).
- `OperatorChannel` (`geecs_bluesky/operator_channel.py`) and `DialogRequest` (`geecs_bluesky/events.py`) — the cross-thread dialog transport (decision 3: preflight moved client-side pre-submit; no cross-thread dialog transport remains).
- Most of `geecs_bluesky/pause_supervisor.py` (decision 1: pause means quiesce, handled at plan level since 0.54.0/0.55.x). Whatever survives must have a live consumer — justify each retained line in the PR body.

## Musts beyond the deletions

1. **Re-check the Stop-from-paused path.** The supervisor parking the console Stop button relied on machinery that dies here. The required end state (live-verified 2026-08-21, Scan003): stop from paused = graceful finalize, `exit_status: stopped`, partial s-file + data preserved, shot controller → STANDBY. Add/keep a hermetic test pinning that path post-deletion.
2. **Extract the `session.scan` tail-helper** (PR #639 disposition row 5): the post-`RE()` tail shared between `session.scan` and the plan path becomes one helper instead of two copies.
[...]

PR #654, opened 2026-08-21
GEECS-Console 0.22.0: the window switch — console is a queueserver client end to end (#648)

skbarber:
## What

W6 part 2 of 2: the window stops using the in-process bridge and runs entirely on the [#653](https://github.com/GEECS-BELLA/GEECS-Plugins/pull/653) services layer. After this, GEECS-Console no longer imports `BlueskyScanner` anywhere — the precondition for W5 (#649).

Per #648's scope items:

1. **Pre-submit preflight as plain GUI code** — Start runs a three-phase pipeline: check phase (`run_submit_preflight`) on the submit worker, questions as ordinary synchronous modals (`_ask_binary`; a render failure reads as abort), `stamp_submission` writes the provenance record, then the queue submit on the worker. Both worker callables capture their own exceptions into refusal/failure results — `BackgroundResult` swallows raises without emitting, which would otherwise strand the pipeline in-flight with Start disabled forever (found while porting the optimize tests; pinned by `test_submitter_exception_is_surfaced_and_releases_start`).
2. **Submission via the manager API** — `submission.py` rewritten around `QueueSubmitter` (`make_queue_submitter`); the `Submitter` protocol is now the queue shape (submit / stop_scan / pause / resume / run_action / describe_action / move_variable / status). No CLI anywhere.
3. **Failed-item-at-front handling** — a non-empty queue surfaces a "Remove && submit?" modal; the stamped payload is held and resubmitted verbatim with `clear_pending=True` (no re-stamp). Cancel leaves the queue untouched.
4. **Paused-state rendering** — the status poll asserts running/paused on the pill (authoritative, including scans other clients started) and falls back to idle when the stream is down; "unknown" when the manager is unreachable. The paused pill's *why* is the console-output stream's failed-move reason line (`_on_pause_reason` — status bar + log). Stop sequences pause-then-stop on the stop worker with a result slot: a **failed** sequencing (pause never landed) releases the "Stopping…" hold for a retry, a successful one waits for the terminal state. `_TERMINAL_SCAN_STATES` gains `idle` (the poll's stream-down fallback).
5. **Progress from the document stream** — start doc → totals (`num_points × shots_per_step`), scan number, "scan running (Scan NNN)"; primary-stream event `seq_num` → progress bar + per-shot beeps (descriptor-name filtered); stop doc → done/aborted by `exit_status` with the reason in the log tail.
6. *(Rider landed in #653.)*

**Dropped feature (maintainer decisions 2026-08-21):** the G-actions v2 pause-window flow — `request_action_during_scan`, the three-way decision modal, and the mid-run operator-dialog transport (`_on_operator_dialog`/`_unblock_engine`) are deleted; the actions dialog's Run disables during scans ("actions are idle-only queue items"). The Movable panel's catalog moves now run the worker's `geecs_move_variable`; raw `device:variable` direct gateway puts are unchanged.

Also: `events_adapter.py` + `tests/fake_events.py` deleted; `main.py` drops the GUI-process optimization warm-up; **`services/optimization.py` is now consumer-less** — deliberately NOT deleted here (kept the diff reviewable; the W5 sweep deletes it with its tests and the `optimization` extra, and un-twins the worker loader's docstring). `docs/tutorials/getting_started.md` gains the `[qserver]` section (the #653 review's owed item). Console CLAUDE.md truthed up (R5/R6, seams, actions, optimization, movable).

Judgment calls, cheap to veto: Start stays enabled while the manager is unreachable (clicking reports the config/connection error — discoverability over pre-blocking); the submitter is built eagerly at construction so the status poll runs from startup; `_TERMINAL_SCAN_STATES` including `idle`.

## Tests

`./scripts/check.sh`: lint OK; **GEECS-Console 820 passed** (TestSubmission rewritten for the pipeline — stamped-record content, refusal, question-continue/abort, pending-items clear/cancel, stop sequencing incl. failure release, pause/resume, pause reason, disconnected pill; document-driven progress/beeps/scan-number/stop tests; actions idle-only gating); GEECS-Schemas + GeecsBluesky 697 passed, 3 deselected.

## Hardware verification

**OWED (maintainer, interim host at ≥ GeecsBluesky 0.56.0 — pull + manager restart first, per #650/#652):**
- Console submits a noscan end to end: preflight modals behave, run start doc carries `submission` (check Tiled/scan.log), pill runs/completes, progress bar + beeps track shots, scan number renders.
- Failed-item trap: submit a request that fails validation worker-side, then resubmit — the "Remove && submit?" modal must appear and clearing must work.
- Pause/Resume from the button; Stop from running (sequenced) and from paused; paused pill shows a failed-move reason if one occurs.
- Actions menu: preview + Run (queued item) while idle; Run disabled during a scan. Movable panel catalog move via the worker; raw device:variable direct put unchanged.


skbarber:

Verified before writing findings: start-document keys exist worker-side (`md["scan_number"]` in `geecs_bluesky/plans/run_wrapper.py:207`; `num_points`/`shots_per_step` in `step_scan.py` and `free_run_step_scan.py` — noscan is `positions=[None]` so totals work there too); `geecs_run_action_plan` opens no run, so action items cannot fake a "scan running" pill; `FAILED_MOVE_LOG_PREFIX` fallback string matches `pause_semantics.py:98` exactly; `md["submission"]` is consumed worker-side; repo-wide grep finds no dangling references to `events_adapter`, `ScanEventsAdapter`, `request_action_during_scan`, `is_scanning_active`, `make_bluesky_submitter`, `_on_operator_dialog`, or `fake_events` outside deliberately-kept `services/optimization.py`/its tests and two dead fake attributes. The double-click guard, the pending-items retry, modal reentrancy, and the offline-first import discipline held up under attack. Console suite: 820 passed at review time.

### Findings (ranked)

**1. Medium (correctness) — the status poll's fallback cannot recover from a worker-environment death mid-scan; the pill sticks at RUNNING with contradictory gating.** `_on_queue_status` handles exactly running/paused (assert), `"idle"` (fallback), else no-op. If the worker process/environment dies mid-scan, the manager stays reachable and reports `re_state=None` — the else branch forever: no stop document will ever come, the pill stays RUNNING/PAUSED indefinitely, while `_scanning()` returns False — Start *enabled* under a RUNNING pill, Stop disabled, action dialogs stuck Run-disabled. The same black hole covers `pausing`/`stopping`/`halting`, permanent only for the crash case.

**2. Medium (correctness/UX) — `stream_failed` is emitted by both stream workers and connected nowhere; degraded mode is silent.** A typo'd `doc_addr` or missing import → scans run, but scan number, totals, progress, and beeps silently never appear, with nothing in the GUI telling the operator why. The composed operator-facing messages are dead code; the new CLAUDE.md text reads as if they were surfaced.

**3. Low (correctness) — a stale status snapshot can clobber the terminal pill after the stop document.** A snapshot *taken* while running but *delivered* after the stop document flips a just-rendered DONE/ABORTED pill back to RUNNING for ~1 s, then lands on IDLE — the DONE indication is lost. The one place the "stream narrates, poll asserts" split contradicts itself; no test delivers a post-stop snapshot.

**4. Low (robustness) — the stop worker callable is the one worker dispatch not exception-wrapped.** The PR's own rule is that worker callables capture their own exceptions (BackgroundResult swallows raises without emitting). `_on_stop_clicked` dispatches `submitter.stop_scan` raw — safe today only because both implementations catch internally; any future implementation that raises leaves "Stopping…" stuck.

**5. Low (test rigor) — the state-machine branches this PR added are the ones the rewritten tests skip.** The idle-fallback branch is never exercised at window level (the one test drives idle only *after* "done", so the branch is dead in the test); no stream-down scan-end or unknown→idle recovery test. If `"idle"` fell out of `_TERMINAL_SCAN_STATES`, 820 tests would still pass.

**Redundancy (note)** — `Submitter` and `QueueClient` are near-mirror protocols with `QueueSubmitter` a rename shim; defensible as seam stability, but the duplicated docstring contracts will drift — worth remembering at the W5 sweep. The preflight correctly *reuses* engine checks — no drift.

**Placement (note)** — `_on_scan_document` moves the document→GUI-vocabulary translation into the window, where the deleted `ScanEventsAdapter` deliberately kept it out. `scan_monitor.py` is the natural home for a pure reducer emitting old-shaped signals; a second consumer would force the move, and the md-key contract with `EVENT_SCHEMA.md` would be pinned in one testable place. Cheap now, annoying later.

**Nits:** dead `request_calls`/`request_error` in the actions-menu fake; dead `is_scanning_active` in the movable-panel fake.

Attacked and refuted (recorded so the next reviewer skips the trip): double-click/experiment-switch mid-pipeline; pending-items infinite loop (each iteration needs an operator click); the 1 s Start-re-enable gap after submit (a double submit lands in the queue-start-refused → item-removed path with an honest message); beep/progress re-arm across runs; Qt threading (worker-owned signals, queued connections, modals GUI-thread only, closeEvent teardown); the enum-`str` CONNECTED read.

skbarber:
## Review disposition

Findings 1–5 **FIXED** in db6e0fe8, all confirmed by a fresh-context verifier given only the findings and the fix commit (original reviewer session ended — substitution per /land):

1. **Worker-death honesty (medium)** — `re_state None` reads UNKNOWN with a one-time status-bar line (verifier checked it cannot repeat per poll); transitional states render as-is; `_scanning()` counts every live state.
2. **stream_failed surfaced (medium)** — both workers wired to `_report`; pinned by test.
3. **Terminal-pill grace window (low)** — 2.5 s suppression of live-state asserts after a terminal document; verifier confirmed a genuinely new run's start document re-enables asserts immediately, and the test pins both suppression and expiry.
4. **Stop dispatch wrapped (low)** — the submit-pipeline exception rule now uniform.
5. **State-branch tests (low)** — `TestStateModel` covers all of the above; verifier checked each test fails under its fix's reversion.

Nits (dead fake attributes) fixed. Verifier's non-blocking observation (gating refresh rides the previous poll during the grace window) accepted — self-healing within one poll.

**Redundancy note (Submitter/QueueClient near-mirror protocols)** — accepted, recorded for the W5 sweep.
**Placement note (document→GUI translation in the window vs a reducer in scan_monitor)** — WAIVED for this PR with reason: single consumer today, and the extraction is mechanical when a second consumer (e.g. a browser live view) appears; recorded as a W5-adjacent candidate so the md-key contract gets one testable home then.

Suite after fixes: **826 passed**. No undispositioned findings.

PR #658, opened 2026-08-21
GEECS-Console 0.23.1: refresh gating on every status snapshot (live-session finding)

skbarber:
## What

Live-session finding (2026-08-21, Scan008): with a scan running, **Stop was disabled and Start still enabled**. Root cause: the document stream narrates RUNNING (start doc) before the first running status poll; when that poll arrives agreeing with the pill, `_on_queue_status` took the no-transition path and never called `_refresh_submit_enabled` — but `_scanning()` reads the *stored snapshot*, which the same handler had just updated. Gating stayed frozen on the pre-scan snapshot until some transition (a pause, or a failed double-submit's cleanup — both observed live) forced a refresh.

Fix: `_on_queue_status` stores the snapshot, applies state transitions (extracted to `_apply_status_state`, logic unchanged), then **always** refreshes gating. Refresh is idempotent widget-enable computation — no cost concern at 1 Hz.

Pinned by `test_equal_state_poll_still_refreshes_gating`, which reproduces the exact live sequence (start doc first → equal-state poll → Stop must be enabled, Start disabled).

Bonus live evidence from the same minutes (no code change needed): the double-submit-while-busy path fired exactly as #653's finding-2 fix intended — `queue_start` refused ("RE Manager is busy"), the item removed itself, honest message shown; and the sequenced stop-from-running ran end to end ("stop requested (paused, then stopped)", graceful, partial data).

## Tests

`./scripts/check.sh`: lint OK; GEECS-Console **131 window tests / 817 total passed**; GeecsBluesky 559; GEECS-Schemas 255.

## Hardware verification

This PR exists *because of* the live session; re-verification is the next ladder step (Stop enabled immediately once a scan runs, no workaround).

### DIFF
--- a/GEECS-Console/CHANGELOG.md
+++ b/GEECS-Console/CHANGELOG.md
@@ -4,6 +4,18 @@ All notable changes to GEECS-Console are documented here.  Format follows
 [Keep a Changelog](https://keepachangelog.com/en/1.1.0/); versioning is
 semantic.
 
+## [0.23.1] - 2026-08-21
+
+### Fixed
+
+- Live-session finding (Scan008, 2026-08-21): Start/Stop gating froze on
+  the pre-scan snapshot for the whole scan when the document stream
+  narrated RUNNING before the first running status poll — the
+  equal-state poll skipped the refresh, leaving Stop disabled and Start
+  enabled mid-scan. `_on_queue_status` now refreshes gating on every
+  snapshot, transition or not (pinned by
+  `test_equal_state_poll_still_refreshes_gating`).
+
 ## [0.23.0] - 2026-08-21
 
 ### Removed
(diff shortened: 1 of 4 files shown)


skbarber:

**No surviving findings.** All four attack vectors traced and refuted: (1) transition logic byte-identical to the pre-fix inline body — the removed inner else-refresh is strictly redundant under the unconditional refresh (single caller verified); (2) no new mis-gating paths — the terminal-grace race leaves buttons in byte-for-byte pre-fix states, and `_refresh_submit_enabled` already folds submit/stop-in-flight into its formula and never touches button text; (3) the pin test genuinely fails pre-fix (traced: equal-state poll returned without refresh → both assertions fail); (4) the transition-path double refresh is a pure idempotent recompute with correct ordering. Housekeeping (patch bump, CHANGELOG, test placement) correct. Suite: 817 passed. Clear to land.

## Disposition
No findings to disposition.

PR #659, opened 2026-08-22
GEECS-Console 0.23.2: optimize progress totals from max_iterations (live-session finding)

skbarber:
## What

Live finding (Scan013): the optimize progress bar hit 100% at iteration 3 of 5. Stream-verified root cause: `geecs_adaptive_scan`'s start doc records `max_iterations` (5) and `shots_per_step` (5) but **no `num_points`** — the console skipped its totals update and the bar kept the previous 1D scan's 15-shot total; a 25-shot optimization filled it at exactly shot 15 = iteration 3.

Fix (console-only — the iteration bound is already in the start doc): totals fall back to `max_iterations × shots_per_step` (an upper bound, same contract as the old bridge's `on_scan_start` — an early suggester stop finishes under 100%, honestly), and a start doc with *no* computable totals **resets** the bar instead of inheriting stale totals. Two pin tests (the max_iterations sizing, and the stale-bar reset incl. an event leaving an unknown-totals bar untouched).

## Tests

`./scripts/check.sh`: lint OK; GEECS-Console **132 window / 819 total passed**; GeecsBluesky 559; GEECS-Schemas 255.

## Hardware verification

Found live, re-verify live: rerun the 5-iteration optimize after pulling — the bar should size to 25 and track all 5 iterations.

### DIFF
--- a/GEECS-Console/CHANGELOG.md
+++ b/GEECS-Console/CHANGELOG.md
@@ -4,6 +4,19 @@ All notable changes to GEECS-Console are documented here.  Format follows
 [Keep a Changelog](https://keepachangelog.com/en/1.1.0/); versioning is
 semantic.
 
+## [0.23.2] - 2026-08-21
+
+### Fixed
+
+- Live-session finding (Scan013, 2026-08-21): the optimize progress bar
+  hit 100% at iteration 3 of 5 — `geecs_adaptive_scan` records
+  `max_iterations`, not `num_points`, so the console skipped the totals
+  update and inherited the previous scan's bar. Start-document totals
+  now fall back to `max_iterations × shots_per_step` (an upper bound —
+  early suggester stops finish under 100%, honestly), and a start doc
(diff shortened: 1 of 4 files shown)


skbarber:

**No surviving findings.** All attack lines dispositioned: (1) `_on_totals_known(0)` is sane end to end — honestly-empty 0/1 bar, zero-totals events no-op, the beep/dedupe re-arm is the cleaner version of the pre-existing compensation (and a small correctness improvement for the tail dedupe); (2) no in-repo plan lacks both totals keys (step/free-run record num_points, optimize records max_iterations + shots_per_step — key names verified against the real md block); third-party plans degrade to the honest reset, strictly better than inheriting; (3) both pins verified to fail empirically on the reverted code; (4) the only consumer treats 0 as a designed input (`max(1, ·)` guards the Qt marquee edge). Patch bump + CHANGELOG accurate incl. the upper-bound caveat. Suite: 819 passed.

## Disposition
No findings to disposition.

Issue #664, opened 2026-08-22
Free-run: dead reference device passes t0 sync (stale-cache seed) and surfaces only as a trigger timeout


skbarber:
## What happened (live, 2026-08-22, Undulator Scans 001–002 of 26_0822)

The camera server at 192.168.6.100 rebooted uncleanly (Windows updates), killing its LabVIEW GEECS devices. Two queued free-run noscans with `UC_Amp4_IR_input` as the reference then failed with:

```
GeecsTriggerTimeoutError('UC_Amp4_IR_input: no shot within 3.0s')
```

after `scan.log` showed a *successful* `t0 sync complete: 1 devices seeded, spread 0.000s`. The DG645 and the rest of the chain were verified fine (another camera triggered 50/50 in internal mode). Diagnosis cost real time because the failure signature pointed at triggering, not at the dead device.

## Why t0 sync passed

`plans/t0_sync.py` seeds from each device's **cached** `last_acq_timestamp` under a quiesced trigger and validates only the **cross-device spread**. With a single sync device the spread is trivially `0.000s`, so a dead device's hours-old cached timestamp seeds "successfully". The module docstring already names the hazard ("a dead/off device serves a stale cached timestamp forever") but the laggard detection only works relative to other devices — a dead *reference* (or an all-dead set) sails through.

## Why nothing else caught it

The CONNECTED liveness check is **client-side pre-submit only** since W5 (#649) — the engine-side `GatewayLivenessCheck` died with the bridge, and the console's `submit_preflight` is its successor. That leaves two uncovered paths, both real:

- a client that skips/shortcuts preflight (this incident: a smoke script that stamped provenance without running the checks);
- a device that dies **between submission and queue-front execution** — the long submission-to-execution gap is exactly the scenario the migration doc worries about, and the worker re-runs validation + unserved-variables at execution but *not* liveness.

## Candidate remedies (pick at design time, not all)

1. **t0-sync freshness bound**: `acq_timestamp` is LabVIEW-epoch wall time; a cached value older than N seconds (N ≫ clock skew, e.g. 30 s) at seed time means the device is not producing frames — refuse with the device's name instead of seeding. Cheap, catches the dead-reference case regardless of client behavior, single-device case included.
2. **Worker-side CONNECTED re-check pre-claim** (the execution-time analogue of the console's preflight read, fail-open on unreadable per the liveness doctrine, refuse only on the exact "Disconnected").
[...]

PR #665, opened 2026-08-22
GeecsBluesky 0.59.0: worker-side CONNECTED re-check + t0-sync liveness gate (#664)

skbarber:
## What + why

Live incident 2026-08-22 (evidence on #664): the camera server at 192.168.6.100 rebooted uncleanly for Windows updates, and queued free-run noscans with the now-dead `UC_Amp4_IR_input` as reference failed with a cryptic mid-scan `GeecsTriggerTimeoutError` — t0-sync had "seeded" from the dead device's stale cache (single-device spread is trivially zero), and the CONNECTED liveness check existed only client-side pre-submit, leaving the submission-to-execution gap (and preflight-skipping clients) uncovered. Diagnosis cost real time because nothing named the dead device.

Two remedies from #664, both execution-side (the client-side pre-submit check stays):

- **`_preflight_connected`** (~80 LOC + wiring): pre-claim CONNECTED re-check on all four execution paths (runner + queue plan, scan + optimize) — the execution-time analogue of the console's preflight read. A Disconnected **load-bearing** device refuses with `GeecsDeviceDownError` naming it, before any scan number is burned — free-run: the reference (rows are paced by it); strict: any synchronous device (rows await all). Any other Disconnected device warns and continues, recorded in run metadata as `disconnected_devices`. Fail-open per the liveness doctrine (unreadable ≠ down); reads via the `oneshot` helper, `datatype=str` (the DBR_ENUM contract).
- **t0-sync liveness gate** (~45 LOC): `geecs_t0_sync` refuses to seed when a device's `connected_status` reads the exact `Disconnected` choice string (in-plan `bps.rd` — the strict refire gate's read; fail-open on missing attribute/failed read). **Deliberately NOT a timestamp-freshness bound**: #664's remedy 1 as originally sketched is unsound in this facility — at rest with the trigger parked (laser-off operation) a *healthy* cache is legitimately hours old (Scan014 of 26_0821 passed t0-sync in exactly that state and ran clean), so freshness cannot distinguish dead from quiescent; `connected_status` can.

Design note flagged for cheap veto: the severity split (refuse load-bearing / warn-and-continue others) is my judgment call — a queued overnight scan shouldn't die for a snapshot device, but a scan that cannot produce a single row should refuse with the device named rather than burn a scan number and time out.

Test amendments that travel with the behavior: the read-path staging and telemetry read-count pins now allow the gate's **bounded once-per-scan** `connected_status` reads (deliberately outside the staged cache — a liveness probe must not be served from a possibly-dead monitor); per-row costs remain pinned at zero/4-per-row.

## Tests

GeecsBluesky: **609 passed, 3 deselected** (`check.sh` green end to end). New pins: `tests/test_preflight_connected.py` (8 tests — refusal by role/mode, warn-and-continue, fail-open, enum-string contract, no-read short-circuits) and `TestT0SyncLivenessGate` in `tests/test_shot_id.py` (RunEngine-driven: Disconnected refuses before seeding, Connected seeds, statusless device is fail-open).

## Hardware verification

- The incident itself is the negative case: two live failures (26_0822 Scans 001–002) with the exact signature this PR eliminates; `scan.log` + manager history preserved.
- `_preflight_connected` executed live against the real Undulator gateway from this branch (2026-08-22): with the restored `UC_Amp4_IR_input` it reads the DBR_ENUM choice string `Connected` and passes clean.
- The refusal path is pinned hermetically; a live dead-device refusal requires deliberately killing a device and is **OWED at the next natural opportunity** (e.g. next host update or device restart: submit a queued noscan against a down device and confirm the pre-claim `GeecsDeviceDownError` names it, no scan folder created).

### DIFF
--- a/GeecsBluesky/CHANGELOG.md
+++ b/GeecsBluesky/CHANGELOG.md
@@ -13,23 +13,39 @@ Format follows [Keep a Changelog](https://keepachangelog.com/en/1.1.0/).
   sat queued, and the dead free-run reference surfaced only as a cryptic
   mid-scan `GeecsTriggerTimeoutError`).  `_preflight_connected` runs
   pre-claim on all four execution paths (runner + queue plan, scan and
-  optimize): a Disconnected **load-bearing** device — free-run's
-  reference, or any synchronous device in strict mode — refuses with
-  `GeecsDeviceDownError` naming it (no scan number burned); any other
-  Disconnected device warns and continues, recorded in run metadata as
-  `disconnected_devices`.  Fail-open per the liveness doctrine: an
-  unreadable PV, no experiment, or missing CA support is never a
-  verdict.  The client-side pre-submit read stays — this covers the
-  submission-to-execution gap and clients that skip preflight.
+  optimize): a Disconnected **synchronous** device refuses with
+  `GeecsDeviceDownError` naming it (no scan number burned) in both
+  modes — strict rows await every sync device, and a dead free-run sync
+  device would fail t0 sync post-claim anyway (its stale cached
+  timestamp blows the spread window), so a softer disposition would
+  just defer the same death past the claim (review finding on the PR).
+  A Disconnected asynchronous (snapshot) device warns and continues,
+  recorded in run metadata as `disconnected_devices`.  Fail-open per
+  the liveness doctrine: an unreadable PV, no experiment, or missing CA
+  support is never a verdict.  The client-side pre-submit read stays —
+  this covers the submission-to-execution gap and clients that skip
+  preflight.
 - **t0-sync liveness gate**: `geecs_t0_sync` refuses to seed from a
   device whose `connected_status` reads the exact `Disconnected` choice
-  string (in-plan `bps.rd`, the strict refire gate's read; fail-open
-  otherwise).  The seed comes from the *cached* `acq_timestamp` under a
-  quiescent trigger, and a dead device serves its stale cache forever —
-  with one sync device the spread check is trivially satisfied and the
-  failure deferred to the mid-scan timeout above.  Timestamp freshness
+  string (in-plan, the strict refire gate's read; fail-open otherwise).
+  The seed comes from the *cached* `acq_timestamp` under a quiescent
+  trigger, and a dead device serves its stale cache forever — with one
+  sync device the spread check is trivially satisfied and the failure
+  deferred to the mid-scan timeout above.  Timestamp freshness
   deliberately cannot be the guard: at rest with the trigger parked
   (laser-off operation) a *healthy* cache is legitimately hours old.
+- **The liveness idioms consolidated into two shared homes** (review
+  finding — four hand-rolled copies risked silent drift of the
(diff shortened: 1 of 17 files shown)


skbarber:

Per the /land ritual; each fix went back to the same reviewer for confirm/deny.

**1 (HIGH, internal contradiction): the free-run warn-and-continue for dead sync contributors was unreachable** — the t0 gate (and even the pre-existing spread check) killed them *post-claim*, strictly worse than refusing pre-claim, making the PR's severity split a fiction for everything but snapshots. — **FIXED** in 53406b85 by converging on the reviewer's option (a): **any dead synchronous device refuses pre-claim in both modes** (`GeecsDeviceDownError`, named, no scan number burned); warn-and-continue is real only for asynchronous/snapshot devices, which the t0 gate never reads. The `free_run` parameter is deleted; docs/CHANGELOG/tests tell the converged story. Reviewer: CONFIRMED (original repro now unreachable).

**2 (MEDIUM, contract violation + RE-loop blocking): the probe used `oneshot` in-plan against its own docstring, sequentially — up to ~2s × N devices of RE-loop stall with CA unreachable.** — **FIXED** in 53406b85: one concurrent batch via the new `oneshot.try_caget_many` (worst case ≈ one 2s budget + 2s grace **regardless of device count**, pinned by a one-batch assertion), and `oneshot.py`'s contract text now names the pre-claim preamble probe as its one sanctioned in-plan use (the devices don't exist yet, so there is no signal to read). Reviewer: CONFIRMED (blocking math verified).

**3 (redundancy): 3rd and 4th hand-rolled copies of the CONNECTED idiom risked silent drift of the fail-open convention.** — **FIXED** in 53406b85: consolidated into two shared homes — `devices/ca/liveness.py::probe_disconnected` (out-of-plan; now used by BOTH the client-side preflight and the worker re-check; the DBR_ENUM `datatype=str` subtlety lives there once, pinned) and `plans/liveness.py::rd_confirmed_down` (in-plan; now used by BOTH the strict refire gate and the t0 seed gate). Reviewer: CONFIRMED.

**Reviewer-caught regression on the fix commit itself (must-fix): `try_caget_many`'s unguarded aioca import broke fail-open** — a worker without the `ca` extra would refuse every scan with `ModuleNotFoundError`, and the old guard in `_preflight_connected` had become dead code. — **FIXED** in a8641743: the import is inside the fail-open (all-`None` batch), the guard wraps the probe *call* (with the `GeecsDeviceDownError` raise deliberately outside it, verified not swallowed), and `test_missing_aioca_is_fail_open` pins the no-aioca path through the real probe chain. Reviewer: CONFIRMED.

**Accepted trade-off** (reviewer's non-blocking note): a no-`ca`-extra install now skips the check silently rather than logging "no CA support" — matching `try_caget_once`'s semantics; flagged here for the record.

No undispositioned findings remain.

skbarber:
The OWED live verification from the PR body is **closed** — dead-device refusal and healthy-scan bookend both verified on hardware 2026-08-22; evidence on #664 (now closed).

Issue #668, opened 2026-08-22
Engine splits: scan_request_runner.py and session.py (gated on the scan-MCP work)


skbarber:
Deferred-by-design from the post-queueserver organizational assessment (2026-08-21), recorded here so the trigger condition isn't hostage to session memory:

- `GeecsBluesky/geecs_bluesky/scan_request_runner.py` (~1,950 lines) → a `request/` subpackage (resolution / preflight / metadata assembly / detector building are already distinct function families).
- `GeecsBluesky/geecs_bluesky/session.py` (~1,750 lines) → split along the session-facade vs plan-tail seam.

**The forcing function is the GEECS scan MCP** — now homed in this repo (owner decision 2026-08-22) — whose build will touch these files' import surfaces. Execute the splits when that work starts, not before (repo doctrine: no speculative refactors; both files are heavily pinned by the runner/plan suites, so the mechanical risk is low when the time comes).

skbarber:
Sequencing confirmed by a second review (codex, 2026-08-22): build the scanning MCP first; split only when it creates concrete pressure. Two disciplines to hold during the MCP build: (1) every time the MCP wants a private helper from `scan_request_runner.py`/`session.py`, promote that helper into a small public module instead of letting the MCP couple to internals — the splits then emerge from real seams instead of guessed ones; (2) when splitting, keep compatibility re-exports so tests/notebooks/console imports don't all churn at once.

skbarber:
Closing per the 2026-09-01 issue audit: the forcing function this was gated on (the scan MCP build, #670–#690) came and went with zero imports from scan_request_runner.py or session.py, so no seam ever formed. The promote-don't-couple discipline is recorded in GEECS-MCP/CLAUDE.md. Re-file if a real seam appears (the runner is ~2,240 lines today; the risk noted here is unchanged, just untriggered).

Issue #666, opened 2026-08-22
Extract Qt-free config listings for non-GUI clients (scan-MCP prerequisite)


skbarber:
**Context (decision record, 2026-08-22):** the GEECS scan MCP server will live **in GEECS-Plugins**, not the osprey repo (owner decision, reversing the earlier work-split; the osprey-side agent holds the plan document). The MCP's v0 read-only tier needs config listings — which save sets, trigger profiles, and presets exist for an experiment — and today the only listing implementation is Qt-adjacent console code (`GEECS-Console/geecs_console/` — `NamedConfigStore.list_names()` and friends), which a headless MCP/notebook client can't import.

**The task:** extract listing (names + basic metadata, resolved from the configs repo the same way `ConfigsRepoResolver` resolves individual configs) down to GeecsBluesky — natural home: beside `config_resolver.py`, or as resolver methods (`scan_variable_catalog()` is the precedent for promoting a listing surface). The console then consumes the shared implementation, same shape as the qs_client extraction (#663).

Scope discipline: listing only — no editing/writing surface moves.

PR #670, opened 2026-08-22
GEECS-Scan-MCP 0.1.0: the scan MCP server, v0 read-only tools

skbarber:
## What + why

First PR onto the `feat/scan-mcp` integration branch: the **GEECS scan MCP server package** (homed in this repo by owner decision 2026-08-22), built to the scan-MCP planning document's v0 tier. Five read-only tools, zero write risk:

| Tool | Backing | Returns |
|---|---|---|
| `scan_status` | `qs_client` `status()` + `queue_items()` | manager snapshot + pending queue items; never fails (disconnected → `connected: false` + detail) |
| `scan_history(limit)` | `history_items()` | recent items with exit status + the operator-facing error line (last traceback line), field-tolerant |
| `get_scan_result(scan_number\|uid, day?)` | `geecs_data_utils.tiled_catalog` | run metadata incl. `submission` provenance, column names, capped per-column mean/std — never the full event table |
| `list_scan_configs(kind)` | the resolver's #666 listing surface + `scan_variable_catalog()`/`action_plan_registry()` | 6 kinds; scan-variable rows carry kind/target(s)/confirm — never limits (hardware truth, not schema data; corrected on review finding 2) |
| `validate_scan_request(request)` | `ScanRequest.model_validate` + `run_submit_preflight` | `{valid, refusal?, warnings, outcomes}` — nothing submitted |

**Architecture** (osprey house pattern, package `CLAUDE.md` is the contract): FastMCP stdio server (`python -m geecs_scan_mcp`), self-registering tool modules, `tool_names.py` constants leaf so osprey `profile.yml` permission lists import symbols, structured JSON envelopes (`errors.py` taxonomy — tools never raise to the agent, engine text preserved verbatim), lazy cached runtime singletons from the standard `config.ini` with honest degradation everywhere. **A client, never an engine**: only `qs_client` + resolver + `tiled_catalog`; engine internals are off-limits by CLAUDE.md rule (the #668 promote-don't-couple discipline).

**Flagged judgment calls (cheap to veto):** `tiled_unreachable` added to the plan's error-kind taxonomy (archive failures are neither the manager's nor the request's fault); the client identity is `geecs-scan-mcp <version>` (the plan's `osprey-htu-assistant` string belongs to the *deployment*, which can override when v1 stamps submissions); stats capped at 40 numeric columns.

## Tests

GEECS-Scan-MCP: **17 passed** (hermetic — fakes on the `runtime` patch seam; every tool's envelope pinned incl. all error kinds, defensive history mapping, the registration test). CI wired: new job steps in `unit-tests.yml`, `check.sh` own-env mapping.

## Hardware verification (live, v0 checklist from the planning doc)

Run against the real Undulator stack from this branch (2026-08-22): `scan_status` agrees with the idle manager; `scan_history` surfaces this morning's dead-device refusal as its `error` line (`GeecsDeviceDownError … UC_Amp4_IR_input`); `list_scan_configs` matches the console's dropdown folders for all kinds; `get_scan_result(4)` resolves Scan004 from Tiled with the `submission` provenance and capped stats; `validate_scan_request` passes a good request (correctly raising the free-run staleness warning — the trigger is parked) and refuses a bad save-set name with the engine's message verbatim. Nothing owed.

## Review process note

Per the maintainer's process amendment (2026-08-22): after the standard fresh-context adversarial review is dispositioned, this PR **waits for a follow-up codex review** before merging.


skbarber:

Per the /land ritual; each fix went back to the same reviewer for confirm/deny. All findings **FIXED** in 238482ef:

**1 (HIGH): `get_scan_result` emitted invalid JSON whenever a stat was NaN** — `json.dumps` writes bare `NaN` tokens (`default=str` never consults floats), and NaN stats are *routine*: a one-row run's ddof-1 std, or an all-NaN column from a device dead for the whole scan (the event schema's designed null cell). Strict consumers (a TS MCP client's `JSON.parse`) reject the whole envelope. — FIXED: `_finite()` maps non-finite to `null`, pinned with a one-row + all-NaN-column frame asserting no bare `NaN` token. The corollary (envelope assembly outside the try) is closed globally: every tool wrapper routes through `_run_guarded`, turning any impl bug into an `internal_error` envelope — the tools-never-raise contract is now enforced, not aspirational. Reviewer: CONFIRMED (old code fails both new pins; `Exception` not `BaseException`, so anyio cancellation still propagates).

**2 (HIGH): scan-variable rows probed fields the schema deliberately does not carry** (`min`/`max`/`units`/`device` — the schema module itself says limits are hardware truth), and the test pinned that invented shape with a `SimpleNamespace` no resolver can produce — the test passed while real behavior degraded to `{name, kind}`, and an agent told to respect bounds would find none and might infer "unbounded". — FIXED: rows rebuilt from the real models (`target`/`kind`/`confirm`; `targets`/`mode` for pseudo), the bounds claim corrected in docstring/CHANGELOG/PR body with an explicit "absence here does NOT mean unbounded" warning, and the test now builds real `ScanVariable`/`PseudoScanVariable` models and asserts limits never appear. Reviewer: CONFIRMED (re-checked every emitted field against the schema).

**3 (MEDIUM): an unknown uid read as `tiled_unreachable` with message `"'uid-x'"`** — the catalog's unknown-uid contract is `KeyError`; a typo'd uid (e.g. copied from `scan_history`) misdiagnosed as an outage. — FIXED: `KeyError → not_found` before the generic handler; `day` parses before any catalog I/O so catalog `ValueError`s can't masquerade as `invalid_request`. Reviewer: CONFIRMED.

**4 (LOW/MED): the runtime singleton first-use race reproduced the exact zmq-receive-thread leak qs_client's #653 lock prevents one level down** (concurrent first-use under FastMCP's parallel tool dispatch). — FIXED: all cached builds go through `_cached()` under a `threading.Lock`; deadlock-checked (the queue-client build reads config directly rather than re-entering the lock). Reviewer: CONFIRMED.

**5 (LOW): listing freshness was inconsistent** — four kinds fresh per call, two (scan_variables/actions) frozen in a forever-cached resolver, so mid-session console edits never appeared. — FIXED: `get_resolver` returns a fresh resolver per call (lazy, cheap) with the rationale in its docstring. Reviewer: CONFIRMED.

**Minors fixed:** `anyio` declared as a direct dependency; day-timezone sentence in the `get_scan_result` docstring. **Accepted as-is** (reviewer's own suggestion): a stub-client history failure reads as `manager_unreachable` with the honest missing-config message.

Reviewer's redundancy/placement lenses came back clean: the seams are consumed, not reimplemented (the promote-don't-couple pattern working); no engine-internal imports; CI/check.sh wiring matches the own-env pattern exactly. One noted future candidate: this is the repo's third independent `[Experiment]` config reader — a shared helper on the next config touch, not this PR.

No undispositioned findings remain. Suites after fixes: 20 passed.

---

**Per the maintainer's process amendment: this PR now waits for the follow-up codex review before merging.**

skbarber:
## Codex follow-up review + disposition (second reviewer, per the new process)

Codex reviewed after the first disposition; both findings **FIXED** in db4da57a:

**P2: `make_ok` could still emit invalid JSON for non-finite floats** — the first-round fix sanitized the known stats path, but the envelope contract is global; any future payload field could regress it. — FIXED exactly as codex prescribed: the serializer itself now enforces strict JSON (`_json_safe` recursively normalizes non-finite floats to `null` through dicts/lists/tuples) with `allow_nan=False` as the raising backstop (a leak becomes an `internal_error` envelope via the dispatch guard, never invalid output). The per-site sanitizer is deleted — one enforcement point. Pinned by a direct `make_ok(x=float("nan"), nested=...)` test asserting nulls and no bare tokens.

**P3: `get_scan_result` constructed the catalog before validating the selector** — pure argument validation depended on archive setup, and the test's catalog monkeypatch was the tell. — FIXED: the no-selector check runs first; the test dropped its patch.

Codex's not-findings recorded for the audit trail: package boundary clean (no engine-internal imports; `validate_scan_request` reaches the engine only through the `qs_client` seam), and the first-round fixes verified real.

Both reviews (fresh-context Claude reviewer + codex) are now fully dispositioned. Suites: 21 passed; CI running on db4da57a.

PR #680, opened 2026-08-24
GEECS-LogTriage 0.3.0: classify NotConnectedError + expected soft-drop down-ranking (#621)

### LINKED ISSUE #621: LogTriage: classify NotConnectedError tracebacks (now recurring in scan.log after #620)
PR #620 (GeecsBluesky 0.51.0) made scan.log the complete per-scan record: tolerated soft-telemetry drops now land in the file **with full `NotConnectedError` tracebacks** on every scan where a ghost/soft device exists (a standing condition given the known DB orphan-row situation — e.g. `U_GhostLowPowerWFS` on every Undulator scan today).

`NotConnectedError` is absent from `GEECS-LogTriage`'s `classifier.CLASSIFICATION_MAP` and no `_MESSAGE_HINTS` entry matches, so every such scan now produces an `unknown`-classified fingerprint in `/triage` output.

From the PR #620 adversarial review (finding 4). Suggested:
- add `"NotConnectedError": HARDWARE_ISSUE` to the map
- decide whether an *expected* soft-drop (the engine's own "Dropping background-telemetry device … (soft tier — never aborts the scan)" WARNING precedes the traceback) should be down-ranked rather than surfaced as a hardware issue per scan

skbarber:
## What + why

Closes #621 (both halves of its ask, including the down-ranking decision it left open):

- **`NotConnectedError` → `hardware_issue`** in `CLASSIFICATION_MAP` — since GeecsBluesky 0.51.0 made scan.log the complete per-scan record (#620), every scan with a ghost soft-telemetry device (the standing DB orphan-row situation, e.g. `U_GhostLowPowerWFS`) produced an `unknown` fingerprint.
- **The down-rank, decided and implemented**: a new `EXPECTED_CONDITION` classification for conditions the engine tolerates *by design*. The guard keys on the engine's own tolerated-drop phrase (`"soft tier — never aborts the scan"`, the exact string `GeecsBluesky` `session.telemetry` logs) and runs **before** the exception-type map — so a standing ghost device groups into a trailing "Expected conditions (tolerated by design)" report section instead of filing as a fresh per-scan hardware issue, while a genuine `NotConnectedError` anywhere else stays `hardware_issue`. The enum docstring pins the rule that Stage-2 issue filing must not act on this category.

Minor bump (new report category = behavior change). Render section order/labels extended; `/triage` skill unchanged (it wraps the CLI).

## Tests

**56 passed** (3 new pins: bare `NotConnectedError` → hardware; the exact engine phrase → expected, with and without the exception type present; an unrelated hardware message carrying the same exception type is NOT down-ranked). `check.sh GEECS-LogTriage` green.

## Hardware verification

Not applicable — pure log classification, no scan/device surface. The classifier change is exercised against the exact production strings (the engine phrase verified against `session.py`, the ghost-device scenario from the #621 report).

## Review process

Fresh-context adversarial review next; then the codex gate before the maintainer merges.

### DIFF
--- a/.claude/skills/triage/SKILL.md
+++ b/.claude/skills/triage/SKILL.md
@@ -147,7 +147,7 @@ show the draft body so the user can file manually.
 
 ## Notes
 
-- Only file for `bug_candidate`. `hardware_issue`, `config_issue`, and
+- Only file for `bug_candidate`. `hardware_issue`, `config_issue`,
   `operator_error`, and `expected_condition` (engine-tolerated
   soft-telemetry drops — only present on `--level warning` runs; the
   default min level is ERROR) are noted in the summary but do not
--- a/.claude/skills/triage/SKILL.md
+++ b/.claude/skills/triage/SKILL.md
@@ -148,7 +148,10 @@ show the draft body so the user can file manually.
 ## Notes
 
 - Only file for `bug_candidate`. `hardware_issue`, `config_issue`, and
-  `operator_error` are noted in the summary but do not produce issues unless the
+  `operator_error`, and `expected_condition` (engine-tolerated
+  soft-telemetry drops — only present on `--level warning` runs; the
+  default min level is ERROR) are noted in the summary but do not
+  produce issues unless the
   user explicitly asks.
 - Deduplicate: if the same fingerprint hash already appears in an open GitHub
   issue (check with `gh issue list --search "<fingerprint_hash>"`), skip it and
--- a/GEECS-LogTriage/CHANGELOG.md
+++ b/GEECS-LogTriage/CHANGELOG.md
@@ -20,7 +20,22 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   ghost device (the DB orphan-row situation) is grouped and rendered
   in its own trailing report section instead of filing as a fresh
   per-scan hardware issue.  Stage-2 issue filing must not act on this
-  category.
+  category.  Note: the report's Expected-conditions section only
+  populates on `--level warning` runs (the harvester's default min
+  level is ERROR).
+
+### Fixed (review findings on the PR)
+
+- The exception-type map lookup normalizes to the last dotted segment —
+  real tracebacks print non-builtins fully qualified
+  (`ophyd_async.core._utils.NotConnectedError`), so the exact-key match
+  was dead code for every record the harvester can produce; the
+  normalization also rescues the pre-existing dotted-class entries.
+  Pinned end-to-end through the fingerprint extractor.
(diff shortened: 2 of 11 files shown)


skbarber:

Per the /land ritual; fixes verified by the same reviewer. This review was exemplary — the reviewer built a byte-identical scan.log replica and ran the *entire* pipeline (handler → parser → fingerprint extractor → classifier) rather than reading the diff.

**1 (HIGH): the `NotConnectedError` map entry never fired on real logs** — real tracebacks print non-builtins fully qualified (`ophyd_async.core._utils.NotConnectedError`) and the lookup was an exact-key match, so the PR's headline entry was dead code for every record the harvester can produce; the original test hand-fed the bare name (a self-confirming pin). — **FIXED** in 8a4f5be8: `classify()` normalizes to the last dotted segment for the map lookup only (fingerprint signatures keep the full dotted form — dedup stability preserved, the reviewer's stated constraint), which also rescues the pre-existing dotted-class entries. Pinned with the reviewer's prescribed honest end-to-end test through `_extract_traceback_signature`. Reviewer: CONFIRMED, re-verified empirically (genuine ERROR record now `hardware_issue`, tolerated WARNING still `expected_condition`).

**2 (MEDIUM): the phrase coupling was documented on one side only** — a GeecsBluesky wording pass would silently revert every ghost device to per-scan noise with no test failing in either package. — **FIXED**: GeecsBluesky 0.61.1 rides this PR — `session.SOFT_TELEMETRY_DROP_MSG` at both drop sites (reviewer AST-verified byte-identity with both pre-fix strings — true no-behavior-change) with the reword-only-together warning per the `FAILED_MOVE_LOG_PREFIX` pattern; plus a triage-side bidirectional pin that greps the sibling source when present. Reviewer: CONFIRMED.

**3 (LOW): three taxonomy enumerations not extended** — FIXED across the classifier docstring, LogTriage CLAUDE.md, and the /triage skill (including the reviewer's observation that the Expected-conditions section only populates on `--level warning` runs — the harvester default is ERROR). The reviewer caught that my CLAUDE.md fix itself mangled the `operator_error` bullet (inserted mid-bullet) plus a doubled "and" in the skill — both repaired in the follow-up commit, cosmetic-only per the reviewer.

**Also verified by the reviewer, for the record**: the load-bearing question of the whole PR — the WARNING + traceback reach the classifier as ONE LogEntry with the phrase intact (UTF-8 em-dash survives the handler→parser round trip), at both drop sites.

No undispositioned findings remain. Suites: LogTriage 59, GeecsBluesky green.

---

**Per the process: this PR now waits for the codex follow-up review before the maintainer merges.**

PR #681, opened 2026-08-24
GEECS-MCP 0.4.0: the analysis domain — get_scan_analysis + get_scan_figure (#675)

### LINKED ISSUE #675: GEECS-MCP: analysis/figure verbs (get_scan_analysis, get_scan_figure)
**Priority 1 of the post-promotion MCP roadmap** (owner design sessions, 2026-08-22) — the top ask: closes the "scan → analyze → present" loop, and the analyzers mostly exist already.

Route existing **ScanAnalysis outputs** per scan through two new read verbs in the `scans/` (or a new `analysis/`) domain of GEECS-MCP:

- `get_scan_analysis(scan_number|uid, day?)` — the analysis results for a completed scan (the `analysis/ScanNNN/` output tree + task-queue status), context-sized (summaries/metadata, never bulk arrays).
- `get_scan_figure(scan_number, ...)` — fetch a rendered summary figure. Note MCP result-size constraints: likely return a path/reference plus enough metadata for the agent to describe it, or a downscaled image if osprey's client renders images.

Design notes carried from the arc: read-only (auto-allow tier); envelope conventions per `geecs_mcp.errors`; the scan-folder invariant applies with full force (analysis reads are consumers — never create `scans/ScanNNN/` paths); pure-numpy analysis is cross-platform so this fits the Linux central server — anything needing Windows-only SDKs goes to the future Windows satellite server instead (GEECS-MCP/CLAUDE.md records the pattern).

skbarber:
## What + why

Closes #675 — the top post-promotion ask, closing "scan → analyze → present". The first non-scans domain on the general server (`geecs_mcp/analysis/`), proving the domains-as-subpackages design:

- **`get_scan_analysis(scan_number, day?)`** — per-analyzer task statuses from `analysis_status/*.yaml` (parsed tolerantly — the `TaskStatus` schema is ScanAnalysis-owned; torn/partial YAMLs read as `unreadable`, never crash) plus the capped analysis output tree.
- **`get_scan_figure(scan_number, name?, day?)`** — a rendered summary figure as **actual MCP image content** (pillow downscale, ≤1024 px longest edge). Routing: `display_files` from completed tasks first, then tree images; no name + one candidate returns it; ambiguous returns the candidate list to pick from by substring.

**The load-bearing discipline: strictly read-only over the data share.** Only `ScanPaths`' *pure static* path builders are used — the instance `get_analysis_folder()` silently `os.makedirs` a missing folder and is explicitly banned here (documented in the module + CLAUDE.md), pinned by an mtime-snapshot test asserting a miss creates *nothing* on the share.

**Live-run finding fixed en route:** `ScanPaths.paths_config` starts `None` and nothing initializes it at import — an unconfigured host raised `AttributeError` instead of degrading. The impl now does a one-time `reload_paths_config()` and refuses honestly (naming the `[Paths] geecs_data` key) when it stays `None`.

Both tools are read-only/auto-allow; README + DEPLOYMENT.md allow-lists extended; `pillow` declared (direct import — the anyio lesson).

## Tests

**50 passed** (10 new: tree+status parsing incl. the queued/done shapes, torn-YAML tolerance, the read-only nothing-created pin, unconfigured refusals, downscale format/size, candidate listing + substring pick + no-match naming the available, registration). `check.sh GEECS-MCP` green.

## Hardware verification

- The **unconfigured-host refusal path verified live** on this machine (real `GeecsPathsConfig` resolution failing → the honest envelope).
- The happy path against the real netapp tree is **OWED at the qserver-box deployment** (the serving host mounts the share natively; the lab network is unreachable from the dev machine right now): run `get_scan_analysis`/`get_scan_figure` against a date with real ScanAnalysis outputs and confirm statuses/figures match the analysis folder.

## Review process

Fresh-context adversarial review next; then the codex gate before the maintainer merges.

### DIFF
--- a/GEECS-MCP/CHANGELOG.md
+++ b/GEECS-MCP/CHANGELOG.md
@@ -4,6 +4,27 @@ All notable changes to `geecs-mcp` are documented here.
 
 Format follows [Keep a Changelog](https://keepachangelog.com/en/1.1.0/).
 
+## [0.4.0] - 2026-08-22
+
+### Added
+
+- **The analysis domain** (#675 — the top post-promotion ask, closing
+  "scan → analyze → present"): `get_scan_analysis` (per-analyzer task
+  statuses from `analysis_status/*.yaml` — tolerantly parsed, the
+  schema is ScanAnalysis-owned — plus the capped analysis output tree)
+  and `get_scan_figure` (a rendered summary figure as actual MCP image
+  content, ≤1024 px longest edge via pillow; `display_files` routed
+  first, then tree images; ambiguous → the candidate list).  Both
+  read-only/auto-allow.  **Strictly read-only over the data share**:
+  only ScanPaths' pure static path builders (the instance
+  `get_analysis_folder()` silently `os.makedirs` and is banned here),
+  pinned by a nothing-created-on-miss test.  Requires `[Paths]
+  geecs_data` + the mounted share on the serving host; an unconfigured
+  host refuses honestly (live-verified) after a one-time
+  `reload_paths_config()` init (live-run finding: the class attribute
+  starts `None` and raised instead of degrading).
+- `pillow` declared (figure downscaling).
+
 ## [0.3.0] - 2026-08-22
 
 ### Changed
--- a/GEECS-MCP/CHANGELOG.md
+++ b/GEECS-MCP/CHANGELOG.md
@@ -39,11 +39,14 @@ Format follows [Keep a Changelog](https://keepachangelog.com/en/1.1.0/).
 - Field coercions sit inside the per-file guard and `display_files`
   entries are type-checked — one odd YAML on the writable share
   degrades that entry to `unreadable`, never the whole tool.
-- **Figure candidates are bounded to the share root** (resolve +
-  `is_relative_to`): a `display_files` entry pointing outside the share
-  (absolute or `../`) is dropped with a warning, closing the
+- **Figure candidates are bounded to the scan's own analysis folder**
+  (resolve + `is_relative_to`): a `display_files` entry pointing
+  anywhere else — outside the share, at another scan's outputs, or into
+  the raw `scans/` tree — is dropped with a warning, closing the
   confused-deputy path where share-writers could make the MCP serve
-  arbitrary host-readable files.
+  other host-readable files.  (The first review pass bounded to the
+  share root; the codex pass tightened it to the scan's analysis folder,
+  which is where the writer puts every legitimate entry.)
 - A 64 MP decode cap refuses giant share-resident images before the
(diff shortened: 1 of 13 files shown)


skbarber:

Per the /land ritual; fixes verified by the same reviewer. **6/6 CONFIRMED FIXED** in 4b8ad5e9:

**1 (CRITICAL): the status parser read a schema the writer never writes.** Written from ScanAnalysis's *stale CLAUDE.md* (`status`/float `heartbeat`) while `TaskStatus.to_dict()` actually writes `state`/`error`/`claimed_by`/`last_heartbeat` (ISO-8601) — against production data every task would have read null and failed tasks would never show why; the original fixtures pinned the bug, not the contract. — FIXED: the parser reads the real shape (ISO heartbeat → age with naive-UTC tolerance matching `task_queue._parse_ts`; `error` surfaced; `no_data` in the documented vocabulary), the authoritative-shape comment names the writer, **the stale ScanAnalysis doc is corrected at its source** with a read-the-writer warning, and the fixtures now write the writer's schema with the ISO parse asserted. Reviewer: CONFIRMED. This is exactly what the deployment-time live check would have caught — better here.

**2 (MED): the tolerant parse had intolerant tails** (unguarded coercions; a scalar `display_files` iterating characters). — FIXED: all coercions inside the per-file guard, `display_files` list/str-filtered; pinned with odd-typed fields. Reviewer: CONFIRMED.

**3 (MED): share-resident `display_files` could point `get_scan_figure` at any host-readable image** (absolute or `../` entries from YAML on a writable share — a confused-deputy path). — FIXED: candidates `resolve()` and must be `is_relative_to` the share root; escapes warn and drop; pinned (an outside absolute entry is not served). Reviewer: CONFIRMED (incl. symlink-consistent comparison).

**4 (LOW):** `pyyaml` declared — the anyio lesson applied to *both* direct imports this time. **5 (LOW):** a 64 MP decode cap fires from header dimensions before any full decode (Pillow's own ceiling is ~178 MP ≈ 700 MB RAM on a long-lived server); pinned. **6 (INFO):** the `paths_config` comment states the true mechanism. All CONFIRMED.

**Placement:** the reviewer's structural suggestion (promote a shared tolerant reader into geecs_data_utils) is deliberately deferred until a second `analysis_status/` consumer exists — now tracked as its own issue per the reviewer's follow-up note, so the trigger condition doesn't evaporate.

**Reviewer-verified clean:** the read-only discipline holds across every filesystem touch (the pure builders are side-effect-free; no config-write path exists; the mtime pin is real protection); the fastmcp Image return path is guard-covered on all raise paths; day/timezone semantics match the scans domain.

No undispositioned findings remain. Suites: GEECS-MCP 53, ScanAnalysis green.

---

**Per the process: this PR now waits for the codex follow-up review before the maintainer merges.**

PR #684, opened 2026-08-24
GEECS-MCP 0.5.1: first-deployment live fixes — Windows display_files + nested outputs

skbarber:
Two live findings from today's qserver-box HTTP deployment, caught the first time the analysis verbs met the real netapp tree (May 1 Scan001):

1. **Windows `display_files` crashed the figure tool.** Production task statuses are written by the Windows analysis machines, so `display_files` carry `Z:\data\...\analysis\Scan<NNN>\...` paths. On the Linux service host that string is not absolute → joined onto the analysis folder as one giant backslash component → the stat raised `OSError` (Errno 22 on the NFS mount) outside the per-candidate guard, aborting candidate gathering entirely — even the healthy tree-scan fallback never ran. Fix: `_localize_display_entry` re-roots a Windows-style entry by its tail after `analysis\Scan<NNN>\` onto the local analysis folder (so those figures actually **serve**, which is the point of `display_files`-first routing); unmatchable entries skip with a warning; the whole per-candidate filesystem path is guarded. The scan's-own-analysis-folder security bound (codex, #681) is unchanged.

2. **`get_scan_analysis` read every device as empty.** The real layout nests analyzer subdirs (`Scan<NNN>/<device>/<Analyzer>/files`); the one-level listing reported `n_files: 0` for every device. Outputs now walk the device subtree (relative names, listing still capped).

Live verification against the deployed service rides this PR (the raw `bellaappserver`-claimed production status must parse, localize, and serve the actual figure over HTTP) — results will be posted here. 91 tests pass (3 new pins).

### DIFF
--- a/GEECS-MCP/CHANGELOG.md
+++ b/GEECS-MCP/CHANGELOG.md
@@ -4,6 +4,30 @@ All notable changes to `geecs-mcp` are documented here.
 
 Format follows [Keep a Changelog](https://keepachangelog.com/en/1.1.0/).
 
+## [0.5.1] - 2026-08-24
+
+First-deployment live findings (qserver-box HTTP service, real netapp
+tree — both invisible to the hermetic fixtures):
+
+### Fixed
+
+- **Windows-written `display_files` entries localize instead of
+  crashing**: production statuses come from the Windows analysis
+  machines (`Z:\data\...\analysis\Scan<NNN>\...`); on the Linux
+  service host such an entry is not absolute, was joined onto the
+  analysis folder as one giant backslash component, and the stat's
+  `OSError` killed the whole figure tool — including the healthy
+  tree-scan fallback.  Now: a Windows-style entry re-roots by its tail
+  after `analysis\Scan<NNN>\` onto the local analysis folder (served
+  like any candidate), an unmatchable one is skipped with a warning,
+  and every per-candidate filesystem touch is guarded so one bad entry
+  can never take down the tool.
+- **`get_scan_analysis` outputs walk the nested analyzer tree**: the
+  production layout is `Scan<NNN>/<device>/<Analyzer>/files`, and the
+  one-level listing read every device as `n_files: 0`.  Files are
+  counted through the whole device subtree (names relative, listing
+  still capped).
+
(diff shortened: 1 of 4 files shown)


skbarber:
**Live-verified against the deployed service (2026-08-24).** 0.5.1 installed into the qserver-box venv from this branch; over HTTP from a client machine, May 1 Scan001:

- `get_scan_analysis`: the raw production statuses (`bellaappserver`-claimed, Windows `Z:\...` display_files) parse cleanly, and `outputs` now reports the nested tree (e.g. `UC_Amp2_IR_input: n_files 3` with `Array2DScanAnalyzer/...` names — was `n_files: 0` for every device on 0.5.0).
- `get_scan_figure`: no-name call returns the candidate list (13 figures, display_files localized + tree scan); a named call served the actual `U_BCaveMagSpec-interpSpec_summary_waterfall.png` as image content (253 KB PNG, ≤1024 px). On 0.5.0 the same call crashed with `internal_error [Errno 22]`.

skbarber:
**Adversarial review: complete (2 rounds), merge-ready.** Round 1 verified the #681 security bound survives localization (no escape found under `..` tails, UNC prefixes, mixed separators, or repeated `analysis` parts) and raised two fix-first items; both landed in e19d48f7 and were re-verified:

1. `ValueError` (embedded null byte — legal YAML from the writable share) now joins the per-candidate guard alongside `OSError`.
2. The tests now actually pin the live failure — **empirically confirmed**: `test_pathological_entries_hit_the_guard_not_the_tool` fails on 0.5.0 via the real mechanism (a >NAME_MAX joined component makes the unguarded stat raise and kill the tool) and passes on this branch; `test_localize_display_entry_unit_contract` asserts the localization directly instead of being maskable by the tree fallback.

Accepted as-is per review: the Windows-heuristic trade (two contrived-but-legal shapes now drop with a warning; the tree fallback covers real files), and the rglob walk cost (same class as the pre-existing figure gatherer; symlink loops not followed on 3.11).

Live verification already posted above (deployed service, May 1 Scan001). 93 tests pass.

PR #687, opened 2026-08-24
GEECS-MCP 0.6.0: figure references + /figures/ fetch route (payload discipline)

skbarber:
**Stacked on #684** (0.5.1 live fixes) — merge that first; this contains it.

Integration finding from the first osprey web-UI test (reported by the osprey-side session, verified against the live service): `get_scan_figure`'s inline PNG return (247 KB → 337k base64 chars) forced context compaction on the deployment's haiku-tier agent. Direction per the osprey side: references + artifacts, not smaller inline images.

## Changes (GEECS-MCP 0.6.0)

- **`get_scan_figure` default = a reference, not pixels**: figure label, dimensions (header read, no decode), byte size, share-relative POSIX path (never a `Z:\` or `/mnt/` primary handle), and `figure_url`. Candidate/ambiguity listings carry the same reference entries.
- **`thumbnail=true`** opts into bounded inline image content: ≤768 px longest edge, JPEG q80 (~30–60 KB for a matplotlib summary) — the only decode path, still behind the 64 MP cap.
- **New `GET /figures/{day}/{scan_number}/{label}`** custom route on the same FastMCP HTTP app: streams the ORIGINAL file bytes (bypasses model context — the osprey bridge fetches and saves as a local artifact). Bounded by exactly the tool's candidate set (the scan's-own-analysis-folder security bound from the #681 codex review, exact label match) + a 50 MB byte cap; clean 4xx on bad inputs, never a 500 for malformed paths; unreachable over stdio by construction. `figure_url` is **server-relative by design** — clients resolve against the MCP base URL they already hold, so no advertised host is baked into results and the planned service re-homing stays client-config-only (osprey side confirmed this works for their bridge).
- **Payload budgets on `get_scan_analysis`**: 20 `display_files` per task, 40 output dirs, explicit `*_truncated` flags — truncation never silent.

95 tests pass (new pins: reference shape incl. share-relative path + URL, bounded JPEG thumbnail, route serves original bytes / 404 non-candidates / 4xx malformed, truncation flags). Live verification on the deployed service + the osprey bridge test ride this PR — the osprey-side session adds its fetch-and-save skill once the route is live.

### DIFF
--- a/GEECS-MCP/CHANGELOG.md
+++ b/GEECS-MCP/CHANGELOG.md
@@ -4,6 +4,30 @@ All notable changes to `geecs-mcp` are documented here.
 
 Format follows [Keep a Changelog](https://keepachangelog.com/en/1.1.0/).
 
+## [0.5.1] - 2026-08-24
+
+First-deployment live findings (qserver-box HTTP service, real netapp
+tree — both invisible to the hermetic fixtures):
+
+### Fixed
+
+- **Windows-written `display_files` entries localize instead of
+  crashing**: production statuses come from the Windows analysis
+  machines (`Z:\data\...\analysis\Scan<NNN>\...`); on the Linux
+  service host such an entry is not absolute, was joined onto the
+  analysis folder as one giant backslash component, and the stat's
+  `OSError` killed the whole figure tool — including the healthy
+  tree-scan fallback.  Now: a Windows-style entry re-roots by its tail
+  after `analysis\Scan<NNN>\` onto the local analysis folder (served
+  like any candidate), an unmatchable one is skipped with a warning,
+  and every per-candidate filesystem touch is guarded so one bad entry
+  can never take down the tool.
+- **`get_scan_analysis` outputs walk the nested analyzer tree**: the
+  production layout is `Scan<NNN>/<device>/<Analyzer>/files`, and the
+  one-level listing read every device as `n_files: 0`.  Files are
+  counted through the whole device subtree (names relative, listing
+  still capped).
+
 ## [0.5.0] - 2026-08-23
 
 The v2 verbs (issue #676) — actions, manual moves, pause/resume, and the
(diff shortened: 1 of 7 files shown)


skbarber:
**Live-verified on the deployed service (2026-08-24, May 1 Scan001):** reference default returns ~700 chars of metadata (`share_relative_path` = `Undulator/Y2026/...` POSIX, dims 1400×1184, 155,789 bytes, server-relative `figure_url`) — versus the 337k base64 chars that blew the osprey agent's context on 0.5.x. `thumbnail=true` → 67 KB JPEG ≤768 px. `GET /figures/...` → byte-identical original (155,789 == 155,789). Osprey side notified; their fetch-and-save bridge test runs against this deployment.

skbarber:
**Adversarial review: complete (2 rounds), merge-ready — all nine dispositions CONFIRMED, zero denied.** Round 1's fix-first items landed in fad6cdc8 and were re-verified against the diff:

1. **HIGH**: the `/figures/` route's share lookup now runs off the event loop (`_locate_figure` via `anyio.to_thread`) — a hung NFS stat was one request away from stalling every verb including the halt family. Reviewer confirmed nothing left in the handler path touches the share on the loop (starlette's `FileResponse` threads its own I/O).
2. **MEDIUM**: labels relativize against the resolved root — a symlinked base can no longer collapse labels into collisions that would have let the route serve wrong bytes under the right name.
3. **MEDIUM**: the payload caps are now actually pinned (both caps, both truncation flags, no-flag-when-untruncated), and the commit record states its test counts honestly.

Also landed: the comparison-only traversal property pinned directly on `_locate_figure` (`../`, `/etc/passwd` → 404; the earlier TestClient probe was neutered by httpx URL normalization), DEPLOYMENT.md's posture paragraph covers the route, dead `_MAX_FIGURE_EDGE_PX` deleted, `_figure_metadata` stat guarded. Accepted per review: per-request gather cost (lab-internal, now threaded), the gather-time TOCTOU residue, the untested 413 branch.

**External contract confirmed unchanged** (figure_url rule, metadata default, thumbnail opt-in — the osprey bridge's dependencies); only widening: `bytes` is nullable on an unreadable file, like `width`/`height`.

Deployed at fad6cdc8 on the qserver box; route re-verified byte-exact. 97 tests pass. Merge after #684.

PR #689, opened 2026-08-25
GEECS-Core 0.3.0 + GeecsCAGateway 0.20.0: close the half-open-socket blind spot (#611)

### LINKED ISSUE #611: CA gateway: ungracefully-dropped device connections freeze readbacks without alarming (half-open socket blind spot)
## Symptom

A device's readback PVs freeze at their last values — **no INVALID alarm, no CONNECTED MAJOR** — while `:SP` writes still work and the hardware responds. From an operator's seat it looks like "Phoebus stopped updating", and nothing on any screen says otherwise.

## Live incident (2026-08-17, u_s1h)

Gateway on 192.168.6.14 (0.16.0, up 24 days, 111 devices connected). `undulator:u_s1h:current` readback frozen; simultaneous `camonitor` of readback + `:SP` while toggling the setpoint 0→1→0→1 A showed every SP publish arriving and **zero readback events** across a real current excursion. Writing `undulator:cagateway:restart` (exit 86 → systemd relaunch → subscriptions rebuilt) fixed it immediately.

## Root cause

`GeecsCaGateway._supervise` (gateway.py) detects a drop only when the subscriber's listener task ends, i.e. when the socket actually closes. That's deliberate — GEECS devices legitimately idle, so *silence is not a drop*. But it means an **ungraceful** death of the peer — host crash/power cycle, a Windows restart (e.g. domain-forced update reboot), a network partition that eats the FIN/RST — leaves a half-open socket the gateway believes is healthy. It waits forever on a dead line; readbacks stay valid-stale.

Graceful device-app restarts are already handled correctly (socket closes → INVALID + CONNECTED MAJOR while down → auto-reconnect with 0.5→30 s backoff → cache cleared so first frame posts). This issue is only about the ungraceful path.

Adjacent gap in the same loop, worth fixing together: reconnects retry the endpoint resolved at startup — if a device app re-registers on a different port, the supervisor retries the stale address until a gateway restart (restart is currently the DB-resync mechanism by design).

## Current operator workaround

Write `Restart` to `[experiment:]cagateway:restart` — few seconds of CA disconnect fleet-wide, monitors reconnect automatically. Documented rule of thumb: frozen-at-plausible-values + working `:SP` + no alarm ⇒ half-open socket ⇒ bounce the gateway.

## Proposed fix (contained to `_supervise` / `GeecsTcpSubscriber`)

1. **Staleness detection**: enable TCP keepalive on the subscription socket (cheap, catches dead peers at OS level), and/or a per-device last-push-age check — a device that has pushed nothing for a generous multiple of its known cadence gets its subscription proactively cycled. Must not regress the "silence is not a drop" doctrine for legitimately idle devices, so any age-based probe should re-issue the subscribe (or a benign UDP get) rather than assume death.
2. **Endpoint re-resolve on reconnect**: inside the backoff loop, re-query the device endpoint from the DB before dialing, so a port change heals without a gateway restart.
3. Optional observability: per-device `last_push_age` (or similar) diagnostic PV so a frozen subscription is *visible* on a screen instead of inferred.

Per the device-error doctrine (build when a real case forces it): this is real case #1. One occurrence in 24 days with a one-PV workaround — not urgent, but the failure mode is silent, and silent is the bad kind.

skbarber:
Closes #611 — the silent failure mode from the 2026-08-17 live incident (u_s1h readbacks frozen at valid-stale values, `:SP` writes working, no alarm, for 24 days of gateway uptime until a manual `cagateway:restart`).

## What + why

**1. TCP keepalive on the subscription socket — GEECS-Core 0.3.0, ~60 LOC (`transport/tcp_subscriber.py`):**
The subscription is read-only after the `Wait>>` command, so the OS never sends on it — an *ungracefully* dead peer (host crash, power cycle, a partition that eats the FIN/RST) left a half-open socket the listener waited on forever. `GeecsTcpSubscriber` now enables `SO_KEEPALIVE` by default with platform-guarded tuning (`TCP_KEEPIDLE`/`TCP_KEEPALIVE`, `TCP_KEEPINTVL`, `TCP_KEEPCNT`; bare `SO_KEEPALIVE` + OS defaults where unexposed): dead-peer detection in roughly a minute (30 s idle + 3 × 10 s probes), surfacing as the ordinary drop path — INVALID readbacks, `CONNECTED` MAJOR, supervised reconnect. **The "silence is not a drop" doctrine is preserved**: keepalive probes are answered by a live peer's TCP stack even when its application is silent, so quiet devices are never disturbed. This was the fix the supervisor's own docstring and PV_CONTRACT §5 already named as the intended future refinement.

**2. Endpoint re-resolve at the backoff ceiling — GeecsCAGateway 0.20.0, ~110 LOC (`gateway.py`, `__main__.py`):**
The issue's adjacent gap: reconnects retried the endpoint resolved at startup forever. Once a device's backoff hits its cap (the outage is not a blip), the supervisor re-queries the device's `(host, port)` via a new injectable `endpoint_resolver` (production: `GeecsDb.find_device`, wired in `__main__`) — off the event loop with a 10 s budget, so an unreachable DB mid-partition cannot stall the supervisor. On a move, the UDP set client is rebuilt too (`_rebuild_udp`; the setter closures resolve `self._udp` at call time, so sets heal along with readbacks). Any resolve failure keeps the last known endpoint — exactly the old behavior. Not on every retry by design: during a partition the DB is down too, and fast retries must not be paced by DB timeouts.

**Contract/docs travel with the behavior** (per the gateway's ground rule): PV_CONTRACT §5 (the known-gap bullet becomes the closed-gap statement with detection-latency honesty), §7 (re-resolve semantics; restart remains the resync mechanism for added/removed devices/variables), the pinned-by test map, DEPLOYMENT.md's restart section, and the gateway class docstring.

## Deliberately not built (from the issue's proposal list — cheap to veto)

- **App-level last-push-age probing**: keepalive detects dead peers at the transport level without interpreting application silence; an age-based prober would re-introduce the "slow vs dead" conflation the doctrine exists to avoid.
- **Per-device `last_push_age` diagnostic PV**: with self-healing subscriptions the frozen state it would surface no longer persists. Add later if a real case asks.

## Tests

- GEECS-Core: **89 passed** (2 new: keepalive on by default / can be disabled, socket-option asserted against a live fake-server connection).
- GeecsCAGateway: **136 passed** (3 new: keepalive present on the gateway's live subscription socket; endpoint re-resolve heals a device that moved ports end-to-end incl. the UDP client following; a raising resolver degrades to the old retry-same-endpoint behavior and still recovers).
- Lint clean via `scripts/check.sh` (plan: GEECS-Core + GeecsCAGateway).

## Hardware verification

**OWED** (this is the production gateway path): after merge + deploy to the gateway host, two checks —
1. *Keepalive is active in production*: `ss -o state established '( sport = :* )' | grep -A1 <device-ip>` on the gateway host shows a `timer:(keepalive,…)` on device subscription sockets. Expected: present on every device connection.
2. *The incident scenario heals*: hard-drop one expendable device (pull its network / kill the host, not a graceful app exit) and confirm within ~2 minutes: readbacks go INVALID, `CONNECTED` goes MAJOR, and recovery is automatic when the device returns — no `cagateway:restart` needed. The endpoint-move case can ride the next natural device re-registration rather than a staged test.

Note the deployed host runs 0.19.x; this PR does not change wire behavior toward devices (same subscribe command, same UDP protocol), so the upgrade is the routine `git pull && poetry install && systemctl restart`.

### DIFF
--- a/GEECS-Core/CHANGELOG.md
+++ b/GEECS-Core/CHANGELOG.md
@@ -3,6 +3,27 @@
 All notable changes to `geecs-core` are documented here, following
 [Keep a Changelog](https://keepachangelog.com/en/1.1.0/) and semantic versioning.
 
+## [0.3.0] - 2026-08-24
+
+### Added
+
+- **TCP keepalive on `GeecsTcpSubscriber`, on by default** (issue #611 —
+  the half-open-socket blind spot; live incident 2026-08-17, u_s1h): the
+  subscription is read-only after the `Wait>>` command, so an ungracefully
+  dead peer (host crash, power cycle, a partition that eats the FIN/RST)
+  used to leave the listener waiting forever on a half-open socket, with
+  readbacks frozen at valid-stale values and no alarm.  With keepalive the
+  OS probes the idle peer and a dead one resets the connection, ending the
+  listener — the supervised-reconnect path.  Probes are answered by a live
+  peer's TCP stack even when its application is silent, so the "silence is
+  not a drop" doctrine is preserved.  New constructor knobs `keepalive`
+  (default True), `keepalive_idle_s` (30), `keepalive_interval_s` (10),
+  `keepalive_count` (3) — dead-peer detection in roughly a minute where
+  the platform exposes the tuning (Linux/macOS; a bare `SO_KEEPALIVE`
+  fallback with OS defaults elsewhere).  Tuning failures are logged and
+  never fail the connect.  Pinned by `test_keepalive_enabled_by_default` /
+  `test_keepalive_can_be_disabled`.
+
 ## [0.2.0] - 2026-08-20
 
 ### Added
--- a/GEECS-Core/geecs_core/transport/tcp_subscriber.py
+++ b/GEECS-Core/geecs_core/transport/tcp_subscriber.py
@@ -338,6 +338,12 @@ class GeecsTcpSubscriber:
             logger.debug("TCP connection closed by server")
         except asyncio.CancelledError:
             pass
+        except OSError as exc:
+            # The ungraceful peer-death path — e.g. ConnectionResetError from
+            # a keepalive-detected dead peer (#611) or an RST from a rebooted
+            # host.  Ordinary connection loss, not "unexpected": the caller's
+            # supervisor owns the once-per-episode down logging.
+            logger.info("TCP connection to %s:%s lost: %s", self._host, self._port, exc)
         except Exception:
             logger.exception("unexpected error in TCP listener")
 
--- a/GEECS-Core/geecs_core/transport/tcp_subscriber.py
+++ b/GEECS-Core/geecs_core/transport/tcp_subscriber.py
@@ -24,6 +24,7 @@ from __future__ import annotations
 import asyncio
 import logging
 import re
+import socket
 import struct
 from typing import Any, Callable, Awaitable, Collection, Sequence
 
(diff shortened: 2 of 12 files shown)


skbarber:

The reviewer verified the new tests are stable (5/5 repeated runs of the timing-sensitive endpoint tests), found no redundancy (no other keepalive or re-resolve logic in the repo; the setter closures were already rebuild-tolerant), and confirmed placement (keepalive in the transport layer). It also attempted and refuted: `wait_for`+`to_thread` cancellation during `close()`, the `_udp` swap racing an in-flight caput, keepalive disturbing quiet-but-live devices, platform-guard safety on Windows, contract/CHANGELOG number consistency, and the committed device-name hygiene question.

### Findings and dispositions

**1. (Medium) The keepalive-detected death — the very path this PR designs in — logged as `ERROR "unexpected error in TCP listener"` with a traceback** (a `ConnectionResetError` fell through to the generic handler; empirically reproduced by the reviewer). **FIXED** in d58db86d: `_listen_loop` catches `OSError` as ordinary connection loss, one INFO line, supervisor keeps the once-per-episode down logging. **Reviewer re-verified: CONFIRMED** — re-ran the RST reproduction against the fixed code; single INFO line, clean drop path, and handler ordering is safe (`IncompleteReadError` subclasses `EOFError`, not `OSError`).

**2. (Medium) The production resolver wiring was captured but never asserted** — deleting `endpoint_resolver=_db_endpoint_resolver` from `__main__` would silently disable re-resolve in production with every suite green. **FIXED** in d58db86d: `test_entrypoint` asserts the kwarg `is _db_endpoint_resolver`. **Reviewer re-verified: CONFIRMED** (dropping the kwarg now fails the test).

**3. (Low) Steady-state DB churn: re-resolve ran on every ceiling cycle, one fresh MySQL connection each** — 30 devices off overnight ≈ a query per second, forever. **FIXED** in d58db86d: a per-supervisor holdoff counter throttles ceiling resolves to every 10th cycle (~one query per ~5 min per down device); an endpoint change resets backoff + holdoff. **Reviewer re-verified: CONFIRMED**, and its one residue — holdoff not reset on successful reconnect, delaying episode-2 move discovery by minutes — **FIXED** in 583ca2ee (reset in the successful-subscribe branch). **Reviewer re-verified: CONFIRMED, no remaining findings.**

**4. (Low) `LAST_SET_ERROR`'s "bind failed at startup" message became wrong in the new rebind-failed/in-flight states.** **FIXED** in d58db86d: reworded to name both origins; the existing substring test pin still passes. **Reviewer re-verified: CONFIRMED.**

**5. (Low, placement) The endpoint re-resolve lives in `GeecsCaGateway._supervise`, so GeecsPvaGateway can't reuse it** (it does inherit the keepalive for free via the shared subscriber default — the actual #611 blind-spot fix reaches both gateways). **WAIVED**: extraction to a shared home is deferred until the PVA gateway actually needs it — the repo's promote-on-demand discipline. **Reviewer accepted the waiver** ("no pushback").

Final state: GEECS-Core **89 passed**, GeecsCAGateway **136 passed** (reviewer independently re-ran both). No undispositioned findings. Hardware verification remains **OWED** as stated in the PR body (keepalive timer visible via `ss -o` on the deployed host; one staged ungraceful device drop heals without `cagateway:restart`).

Issue #682, opened 2026-08-24
Promote a shared tolerant analysis_status reader into geecs_data_utils when a second consumer appears


skbarber:
Deferred-by-design from the #681 review (GEECS-MCP analysis domain): the tolerant `analysis_status/*.yaml` reader currently lives in `geecs_mcp/analysis/read_tools.py` with `TaskStatus.to_dict()` (ScanAnalysis `task_queue.py`) named as the authoritative shape. **Trigger: the moment a second consumer of `analysis_status/` appears**, promote a schema-light tolerant reader into geecs_data_utils (the scan-folder layout owner, already a dependency of both) and pin ScanAnalysis's writer against it with a cross-package test — the #679-style drift class. Until then, no code owed; the read-the-writer warnings in both CLAUDE.mds carry the contract.

PR #688, opened 2026-08-24
GEECS-Console 0.24.1: adopt shared resolver listings + schema shot-count derivation (#679)

skbarber:
Closes #679 — the two console-touching cleanups deferred by scope discipline during the MCP arc (the deferred halves of #666 and #671). Both are behavior-identical swaps; they ride together because the issue filed them as one bundle.

## What + why

**1. Listings rewire (#666's deferred half) — ~70 LOC across `services/`:**
- `ConsoleConfigs.listing()` and `NamedConfigStore.list_names()` now delegate to `ConfigsRepoResolver.list_save_sets()/list_trigger_profiles()/list_presets()/list_optimizer_configs()` — the one folder-scan implementation every queue client shares.
- The console's duplicated folder constants (`SAVE_SET_FOLDER`, `TRIGGER_FOLDER`, `OPTIMIZATION_FOLDER`) and the local `yaml_stems` helper are **deleted**; folder names come from the resolver's class constants (`ConfigsRepoResolver.OPTIMIZER_FOLDER` — per the issue's rename trap, no third name introduced).
- Listing semantics are identical (sorted stems, both suffixes, missing → empty). One deliberate improvement: an I/O failure mid-scan (SMB blip on a mounted configs share) now reads as an empty listing instead of raising — the resolver's never-raise contract, which `listing()`'s docstring always promised but the old `yaml_stems` path didn't deliver.
- The listing resolver is constructed over the **same experiments root the console resolved** (`experiments_root=base` / the store's `_root()`), so the `_configs_base` monkeypatch seam and `GEECS_SCANNER_CONFIG_DIR` stay coherent — a production-resolution resolver would silently scan a different tree than the one tests patch.

**2. Shot-count consolidation (#671's follow-up) — ~25 LOC in `request_builder.py`:**
- `_position_count` is deleted; `estimate_total_shots` counts axis positions via the schema's `Positions.n_positions()` (GEECS-Schemas 0.11.0) — the same never-materializing arithmetic behind `ScanRequest.planned_shots()`, retiring the third parallel derivation. No safety exposure: the console's counter was already arithmetic.

## Judgment calls (cheap to veto)

[...]

PR #690, opened 2026-08-25
GEECS-MCP 0.7.0: analysis execution slice — run_scan_analysis + listings (#686)

### LINKED ISSUE #686: GEECS-MCP: analysis-domain execution verb — run ScanAnalyzers on demand (owner request)
Owner request (Sam, 2026-08-24): the MCP server should be able to *execute* ScanAnalysis analyzers, not just read their outputs (#675 built the read slice: `get_scan_analysis`/`get_scan_figure`).

Sharpening observation from the same day's live checks: the data share has **no `analysis_status/` folders anywhere in Jul/Aug 2026** — the LiveTaskRunner fleet has not been running post-migration, so nothing currently produces fresh analysis at all. An MCP `run_analysis` verb would not just expose existing capability to agents; it would restore a currently-dormant capability through the new architecture.

Design considerations to settle before building:

- **Verb shape**: `run_analysis(scan_number, day?, analyzer|group)` — submit-and-poll like scans (analysis of a real scan takes minutes; never a blocking tool call), with `get_scan_analysis` as the natural poll (task statuses + outputs appearing).
- **Execution mechanism**, two candidates: (a) in-process on the serving host — load the group config (`GEECS-Plugins-Configs` `groups/`/`analyzers/`), `create_scan_analyzer(...)`, `run_analysis(tag)` in a worker thread/process; (b) enqueue into the ScanAnalysis task queue (`init_status_for_scan`) and let a runner claim it — requires a runner fleet that doesn't currently exist. (a) is the self-contained start.
- **Platform split**: pure-Python analyzers (numpy/matplotlib pipelines) run fine on the Linux qserver box; analyzers needing Windows-only SDKs are the documented *satellite-server* case (GEECS-MCP CLAUDE.md domain roadmap) — the verb should refuse those honestly by config inspection, not crash.
- **Safety class**: Q (ask + headless `write_tools`) — it writes to the share (`analysis/Scan<NNN>/` tree + `analysis_status/` inside the existing scan folder) and burns CPU on a box shared with the production manager (the systemd unit's resource caps matter here). The cross-package scan-folder invariant applies in full: never create `scans/Scan<NNN>/`; a missing scan folder refuses.
- **Dependencies**: pulls ScanAnalysis/ImageAnalysis into the MCP install (heavy) — likely an extra, mirroring GeecsBluesky's `analysis`/`optimize` pattern.

Slots after the current approved queue (#679, then #611) unless reprioritized.

skbarber:
Closes #686 (owner request). The MCP server can now *execute* ScanAnalysis, not just read its outputs — restoring a capability that has been dormant since the queueserver migration (no `analysis_status/` files on the share since ~May: the LiveTaskRunner fleet stopped running). Design settled with the owner 2026-08-24: **ScanAnalysis as-is is the backend; the Tiled-based analysis stack is the recorded long-term direction**, so the verb surface is backend-neutral (names a diagnostic/group + scan + day) and `analysis_status/` is the progress contract — a future Tiled backend slots in behind the same tools with `get_scan_analysis` polling unchanged.

## What (per-concern LOC)

**1. `run_scan_analysis` (Q-class) — `analysis/run_tools.py` + `analysis/run_worker.py`, ~330 LOC:** submit-and-poll, never blocking. The tool validates *everything refusable before any side effect*:
- exactly one of `analyzer`/`group` (names from the new listings);
- the configs root resolves (`SCAN_ANALYSIS_CONFIG_DIR` / `[Paths] scan_analysis_configs_path` — the shared `config_roots` resolution every ScanAnalysis consumer uses);
- **the scan folder exists** — the cross-package invariant in full: analysis never creates `scans/ScanNNN/`; a missing folder is a `not_found` refusal, pinned by a nothing-created tree-snapshot test;
- **the analyzers construct on this host** — a diagnostic whose image-analyzer class needs a Windows-only SDK (HASO WaveKit, Grenouille FROG.dll — both verified Windows-bound at analysis time) fails its import at build and is refused up front with a message naming the future Windows satellite server, instead of half-running.

Then it initializes the status YAMLs server-side (`init_status_for_scan`, idempotent — a worker that dies pre-claim leaves *visible* queued rows, never silence) and spawns a detached worker subprocess that drives ScanAnalysis's own task queue (claim / 30 s heartbeat / stale-reclaim). `rerun_failed`/`rerun_completed` ride `build_worklist`'s native flags. **Google-Doc upload stays hard-off** (`gdoc_enabled` default) — publishing to the experiment log would need its own explicitly gated verb.

**2. `list_analyzers` / `list_analysis_groups` (R-class) — ~60 LOC:** discovery over the configs repo's `analyzers/`/`groups/` trees so agents name diagnostics instead of guessing; payload-capped with explicit `truncated` flags per the 0.6.0 payload doctrine.

**3. Packaging + gating + docs:** new optional `analysis-run` extra (ScanAnalysis path dep — scipy/opencv ride in via ImageAnalysis, hence an extra mirroring GeecsBluesky's pattern; without it all three tools refuse naming the extra and the server starts normally). `run_scan_analysis` joins `QUEUE_TOOLS` (native `ask` + headless `write_tools` — DEPLOYMENT.md's gating section updated with the new entry and the host requirements). CI installs the extra so the new suite actually runs there. CLAUDE.md layout/roadmap truth-up.

## Deliberately out of scope (cheap to veto)

- **Day-sweep** (`scan_number` omitted = whole day): dropped for scope — one scan per call; an agent loops. Add later with a scans-per-call cap if usage asks.
- **A `requires_windows` marker on the diagnostic YAML schema**: not needed — the construct-to-validate check refuses unrunnable diagnostics with zero schema changes across three packages. Revisit only if construction cost ever becomes a problem.
- **Issue #682** (shared tolerant `analysis_status` reader in geecs-data-utils): considered and not tripped — this PR adds a status *writer* consumer (ScanAnalysis's own `task_queue`), not a second tolerant reader; the MCP read tools remain the only independent reader.

## Tests

GEECS-MCP: **109 passed** (12 new in `test_analysis_run_tools.py`: selector/root/unknown-name refusals, the missing-scan-folder refusal with a tree-snapshot nothing-created pin, the unimportable-diagnostic refusal, happy paths for analyzer and group with server-side queued statuses + captured worker payload, rerun-flag passthrough, both listings, and a worker test driving the real `build_worklist` against a tmp share). The suite `importorskip`s without the extra; CI now installs it. Lint clean via `scripts/check.sh`.

## Hardware verification

**OWED (live, needs the share + configs repo on the serving host — no scan hardware involved):** after deploy with the extra,
1. `list_analyzers` returns the real diagnostic corpus;
2. `run_scan_analysis` on a known archived scan (e.g. a May Scan with camera data) produces `analysis_status/` rows transitioning queued → claimed → done, figures land in the analysis tree, and `get_scan_figure` serves them;
3. the refusal path on a nonexistent scan number returns `not_found` and creates nothing (spot-check the share).
Also owed osprey-side: add `run_scan_analysis` to the profile's `write_tools` and re-render (`osprey build`).

### DIFF
--- a/.github/workflows/unit-tests.yml
+++ b/.github/workflows/unit-tests.yml
@@ -70,7 +70,9 @@ jobs:
 
       - name: Install GEECS-MCP dependencies
         working-directory: GEECS-MCP
-        run: poetry install --with dev
+        # analysis-run: the #686 execution-slice tests importorskip
+        # without it — install it so CI actually runs them.
+        run: poetry install --with dev -E analysis-run
 
       - name: Run GEECS-MCP tests
         working-directory: GEECS-MCP
--- a/GEECS-MCP/CHANGELOG.md
+++ b/GEECS-MCP/CHANGELOG.md
@@ -4,6 +4,41 @@ All notable changes to `geecs-mcp` are documented here.
 
 Format follows [Keep a Changelog](https://keepachangelog.com/en/1.1.0/).
 
+## [0.7.0] - 2026-08-24
+
+### Added
+
+- **The analysis-domain execution slice** (issue #686, owner request —
+  restores the post-migration-dormant analysis capability through the
+  new architecture; ScanAnalysis as-is is the backend by owner decision
+  2026-08-24, with the verb surface kept backend-neutral and
+  `analysis_status/` as the progress contract so the Tiled-based stack
+  can slot in behind the same tools later):
+  - `run_scan_analysis(scan_number, day?, analyzer|group, rerun_failed,
+    rerun_completed)` — Q-class (native `ask` + `write_tools`),
+    submit-and-poll: validates everything refusable *before* side
+    effects (exactly-one selector, configs root, the scan folder EXISTS
+    — the cross-package invariant, pinned by a nothing-created test —
+    and the analyzers construct on this host, so a Windows-only-SDK
+    diagnostic is refused up front instead of half-running), initializes
+    the status YAMLs server-side (a dead worker leaves visible queued
+    rows, never silence), then spawns a detached worker subprocess
+    (`analysis/run_worker.py`) driving ScanAnalysis's own task queue
+    (claim/heartbeat/stale-reclaim).  Poll with the existing
+    `get_scan_analysis`; figures via `get_scan_figure`.  Google-Doc
+    upload stays hard-off.
+  - `list_analyzers` / `list_analysis_groups` — R-class listings over
+    the configs repo's `analyzers/`/`groups/` trees
+    (`discover_analyzers`/`discover_groups`), so agents name
+    diagnostics instead of guessing (payload-capped, truncation
+    flagged).
+  - New optional `analysis-run` extra (ScanAnalysis path dep — heavy
+    via ImageAnalysis, hence an extra mirroring GeecsBluesky's
+    pattern); without it the three tools refuse naming the extra and
+    the server starts normally.  New consumed config:
+    `SCAN_ANALYSIS_CONFIG_DIR` / `[Paths] scan_analysis_configs_path`
+    (see `deploy/DEPLOYMENT.md`).
+
 ## [0.6.0] - 2026-08-24
 
 Payload discipline for figures (osprey-side integration finding: an
--- a/GEECS-MCP/CHANGELOG.md
+++ b/GEECS-MCP/CHANGELOG.md
@@ -38,6 +38,19 @@ Format follows [Keep a Changelog](https://keepachangelog.com/en/1.1.0/).
     the server starts normally.  New consumed config:
     `SCAN_ANALYSIS_CONFIG_DIR` / `[Paths] scan_analysis_configs_path`
     (see `deploy/DEPLOYMENT.md`).
+  - Review-hardened status semantics (adversarial findings 1–2): the
+    rerun flags reset `failed`/`done` (and stale-claimed) rows to
+    queued *server-side* (`reset_status_for_scan`), so the dead-worker
+    visibility contract holds on every path; a task another runner is
+    actively working (fresh heartbeat — ScanAnalysis 1.16.0's
+    `claim_is_active`) refuses the call (`policy_refusal`) instead of
+    double-running into the same output files; the envelope reports
+    `tasks` (what this call runs) vs `skipped` (done/failed without
+    their rerun flag), and an all-skipped call is an honest
+    `started: false` no-op with no worker spawned.  Task ids come from
+    ScanAnalysis's exported `analyzer_task_id` (no mirrored
+    derivation), and the server-side validation builds analyzers
+    through the same `run_worker.build_analyzers` the worker executes.
 
 ## [0.6.0] - 2026-08-24
 
--- a/GEECS-MCP/CHANGELOG.md
+++ b/GEECS-MCP/CHANGELOG.md
@@ -38,6 +38,14 @@ Format follows [Keep a Changelog](https://keepachangelog.com/en/1.1.0/).
     the server starts normally.  New consumed config:
     `SCAN_ANALYSIS_CONFIG_DIR` / `[Paths] scan_analysis_configs_path`
     (see `deploy/DEPLOYMENT.md`).
+  - **The pre-claim double-start window is closed** (Codex P1): the
+    realistic vector — a second call into this server while the first
+    worker is still importing its stack — refuses via an in-process
+    dispatch ledger (side-effect-free, before any status write; entries
+    expire when the pid dies, the tasks leave `queued`, or after the
+    queue's own 180 s staleness bound), and the cross-process backstop
+    is ScanAnalysis 1.16.0's atomic claim gate in `run_worklist` (a
+    losing worker skips the task instead of double-running it).
   - Review-hardened status semantics (adversarial findings 1–2): the
     rerun flags reset `failed`/`done` (and stale-claimed) rows to
     queued *server-side* (`reset_status_for_scan`), so the dead-worker
--- a/GEECS-MCP/CLAUDE.md
+++ b/GEECS-MCP/CLAUDE.md
@@ -11,8 +11,11 @@ Domain roadmap (add as they earn their keep, never speculatively):
 `scans/` (built: v0 read + v1 control), then candidates in rough value
 order — `health/` (gateway/Tiled/DB probes, read-only), `db/`
 (device-variable metadata lookups), `logs/` (the /triage analysis as a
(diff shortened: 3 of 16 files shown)


skbarber:

The reviewer verified up front: the scan-folder invariant is honored on every path (refusal pre-side-effect, `_require_scan_folder` guards, `analysis_status/` via `mkdir(exist_ok=True)` only, tree-snapshot pin would fail on any violation); validation ordering, gdoc hard-off, registration/gating consistency, config-root resolution names, lazy-import extra behavior, CI `-E analysis-run`, and the loader exception mapping all check out. Six findings, all dispositioned across two fix commits (67fe99f2, aa730c12), each re-verified by the same reviewer:

**1. (Medium) The rerun path broke the dead-worker visibility contract** — `init_status_for_scan` is skip-if-exists, so `rerun_*` calls left old `done`/`failed` rows untouched until the worker claimed, and the envelope note misdirected. **FIXED**: the tool now resets `failed`/`done` (+ stale-claimed) rows to queued server-side via `reset_status_for_scan`. Digging in surfaced a **real latent ScanAnalysis bug**: reset routed through `update_status`, whose keep-on-None semantics silently preserved stale `error`/claim/heartbeat fields (the old error would survive onto the rerun's *done* row). Fixed in **ScanAnalysis 1.16.0** (fresh queued record written directly; `TestResetClearsStaleFields` pin; deliberate consequences documented in its CHANGELOG). **CONFIRMED.**

**2. (Medium-low) Two near-simultaneous calls could double-run one task into the same output files.** **FIXED** (the recommended partial fix): the tool refuses `policy_refusal` on any actively-claimed requested task via ScanAnalysis 1.16.0's new `claim_is_active` export (no re-derived timestamp semantics); stale claims stay runnable; the envelope now reports `tasks` vs `skipped`, and an all-skipped call is an honest `started: false` no-op with no spawn. The reviewer's follow-up (refusal ran *after* the reset, so a refused call had side effects) is **FIXED** in aa730c12 — refusal now precedes any status write, pinned by a byte-exact file-unchanged assertion. Acknowledged residual, by construction: two calls inside the worker-startup import window can still race — the ScanAnalysis task queue has no cross-process claim lock; this closes the realistic slow-second-call case. **CONFIRMED.**

**3. (Low-medium) Mirrored ScanAnalysis logic** (the analyzer-id derivation ×5, the analyzer-construction block ×2). **FIXED**: ScanAnalysis 1.16.0 exports `analyzer_task_id()` and all four internal `task_queue` sites now call it (zero inlined copies remain); the MCP uses the export; `build_analyzers` lives once in `run_worker.py` and both the tool's validation and the worker call it — validation can no longer drift from execution. **CONFIRMED** (reviewer's taste note about the shared builder living in the worker module: accepted as-is — it guarantees validation-equals-execution).

**4. (Low) systemd restarts kill mid-run workers** (cgroup vs session detachment) without the docs saying so. **FIXED**: DEPLOYMENT.md states the kill behavior, the 180 s stale-claim recovery, and both death signatures; the envelope note gained the claimed-with-growing-`heartbeat_age_s` mode. **CONFIRMED.**

**5. (Low) `list_analysis_groups` `count` double-counted** (each group indexed under two name forms). **FIXED**: `count` = distinct config files; the reviewer's deny on the missing test pin addressed in aa730c12 (`count == 1`/`== 2` assertions). **CONFIRMED.**

**6. (Low, pre-existing) `poetry.lock` re-commits a home-directory worktree path.** **WAIVED**: pattern predates this PR, inert for installs (deployment installs from the checkout, not the lock), and regenerating from this worktree would re-embed the same path — spun off as a background-task chip for the owner (audit + regenerate all package locks from the main checkout). **Waiver accepted by the reviewer.**

Final state: GEECS-MCP **113 passed**, ScanAnalysis **175 passed** (reviewer re-ran both independently); CI green on every commit. Reviewer's bottom line: *"No surviving findings remain against the branch."* Hardware/live verification remains **OWED** as stated in the PR body.

skbarber:
## Codex P1 (pre-claim double-start) — fixed, adversarially verified over rounds 3–5

Codex's finding: two `run_scan_analysis` calls inside the worker-startup window both see `queued` rows and spawn two workers; `run_worklist` claims unconditionally, so both run the same analyzer into the same output files. Closed at both layers (commits 126c9b3e, 260ceebc, d44ac291, 87051a70), each round independently attacked and re-verified by the fresh-context reviewer:

**Layer 1 — ScanAnalysis 1.16.0: atomic per-task claim gate in `run_worklist`** (the cross-process fix, the "future multi-app can add lock files" extension the module docstring reserved). `try_acquire_claim`/`release_claim` (public): an `O_CREAT|O_EXCL` lock at `analysis_status/<id>.claim`; exactly one runner claims per task, a loser skips with the status row untouched (pinned end-to-end). Liveness stays in the status heartbeat (`claim_is_active`); a dead holder's lock is broken under a **single-breaker sentinel** with a sentinel-held age re-check (round 4 closed the reviewer's two-breaker TOCTOU), and the sentinel's own stale-cleanup is **rename-atomic** (round 5). Lock files are invisible to status readers (`*.yaml` glob); the gate never creates a missing scan folder (invariant pinned).

**Layer 2 — GEECS-MCP: per-task dispatch ledger with reserve-at-check** (the in-process fix, including the exact test shape Codex requested: two calls with the spawn captured — second refuses `policy_refusal`, exactly one worker, tree unchanged). Entries are keyed `(scan, task)`, reserved under the same lock hold that checked them (two truly concurrent calls cannot both pass), converted to the real pid after spawn, released on every non-spawn path, and evicted only when dead/expired. The reviewer's two reproduced regressions along the way — a live-entry eviction on unrelated-task dispatch (round 4) and rerun/reclaim dispatches escaping the window (round 5) — are both fixed with their reproductions converted into in-suite pins.

**Accepted residual, documented in code rather than engineered against** (reviewer's explicit recommendation): the syscall-granularity sentinel-cascade floor of path-based locking — reachable only via a breaker killed inside a milliseconds-wide window, plus 180 s elapsing, plus two independent syscall-granularity straddles in cascade, and even then final admission is the main-lock `O_EXCL`. POSIX/SMB offer no compare-and-remove primitive, so no path-name scheme reaches zero; `_break_stale_lock`'s docstring names this floor and the sibling unfenced-lease class.

**Reviewer's final bottom line** (5 rounds, 3 independent reproductions all converted to pins, both suites re-run each round): *"No practically reachable findings survive. From the adversarial-review side, d44ac291 is clean to land."* Final counts: GEECS-MCP **119 passed**, ScanAnalysis **183 passed**; CI green on every commit.

(Codex reported no findings on #688/#689 — no action there.)

PR #695, opened 2026-08-27
capture writer: shuffle+gzip-1 frame compression (0.64.1)

skbarber:
Flips the v0 'no compression' default after the owner spotted the 3x size inversion vs LV PNGs. Empirical basis (real Scan003 frames, benchmark in session): shuffle+gzip-1 = **2.04 MB vs 7.92 raw vs ~2.5 PNG**, 3.9 ms/frame write cost (negligible at 1–5 Hz). Built-in HDF5 filters only — every reader decodes them, files stay self-describing, schema unchanged (geecs-capture/1). Test asserts the filters on the dataset; capture suite 18/18. The running daemon picks this up on its next restart; existing raw stacks remain readable.

### DIFF
--- a/GeecsBluesky/CHANGELOG.md
+++ b/GeecsBluesky/CHANGELOG.md
@@ -4,7 +4,7 @@ All notable changes to `geecs-bluesky` are documented here.
 
 Format follows [Keep a Changelog](https://keepachangelog.com/en/1.1.0/).
 
-## [0.64.1] - 2026-08-27
+## [0.65.0] - 2026-08-27
 
 ### Changed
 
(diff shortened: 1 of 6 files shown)


skbarber:
Adversarial review: clean verdict, 5 findings dispositioned in `9ed6a35a` — minor bump 0.64.1→**0.65.0** (byte-level behavior change per the repo's semver table; cheap-veto if you disagree), `compression_opts==1` pinned in the test, frame-size cost caveat added to CHANGELOG (reviewer measured ~25 ms/frame at TopView size — trivial at 1 Hz fleet-wide; a many-large-camera 5 Hz scan would saturate the single writer thread and degrade to counted queue_drops; per-device writer threads = future remedy), scope-doc 'light/no compression' → 'light compression' truth-up. Reviewer confirmed the important part: whole-chunk-aligned appends mean compression does not interact with the flush-per-append crash-window design at all. Ready for your codex gate.

PR #696, opened 2026-08-27
Phase 4 read side: capture-stack consumption (data-utils 0.14.0, ImageAnalysis 1.12.0, ScanAnalysis 1.17.0)

skbarber:
## What

The read side of the capture arc (scope doc Phase 4): the analysis stack can now consume the per-device HDF5 frame stacks the capture daemon dual-writes — **opt-in per diagnostic, with unconditional fallback to per-shot files**, so the old basis keeps working for every scan, config, and facility that never adopts stacks.

Per-package (LOC ≈ 170 / 15 / 130 + tests 100 / 40 / 150):

- **geecs-data-utils 0.14.0** — `io/scan_stack.py`: `find_stack_file`/`is_stack_file` (dispatch on the `schema` attribute, never extension), `read_stack_timestamps` (LabVIEW-epoch conversion via new `LABVIEW_EPOCH_OFFSET`), `read_shot` (one chunk read), and `ShotRef` — a `Path` subclass carrying a frame index that travels the per-shot pipeline unchanged and **pickles correctly** (process-pool workers). Read-only by design. CLAUDE.md truth-up (io/ was undocumented).
- **ImageAnalysis 1.12.0** — `ImageAnalyzer.load_image` resolves `ShotRef`s to single frames; list loading and subclass format overrides unaffected.
- **ScanAnalysis 1.17.0** — `data_format: device_hdf5` on a diagnostic's `scan:` block, threaded config → factory → Array1D/Array2D → `SingleDeviceScanAnalyzer`. `_map_shots_from_stack` mirrors the existing canonical-millisecond `acq_timestamp` join (valid-column skip included). Fallback on every failure shape: no stack, wrong schema, no timestamp column, zero joins — each pinned by a test. Default (`data_format` unset) is byte-for-byte today's behavior; instances built without `__init__` (test-harness/notebook pattern) default safely via `getattr`.

## Verification

- geecs-data-utils: scan_stack 5/5; suite 166 passed (5 collection errors + 1 failure are pre-existing missing-extras gaps: sklearn/scipy — unrelated)
- ImageAnalysis: shotref tests 2/2
- ScanAnalysis: **full suite 190 passed, 12 skipped**; both mapping suites 25/25
- **End-to-end against the real Scan003 daemon stack** (today's live acceptance artifact): all 10 shots mapped to `ShotRef`s — the pre-save-window extra frame correctly excluded by the timestamp join — and `ImageAnalyzer.load_image` returned the (600,600) uint16 frame (the one proven pixel-identical to its LV PNG).

## Deliberately deferred (recorded, scope discipline)

Background pre-pass stack enumeration, `infer_device_ext`/expected-asset-column guards, per-bin whole-stack bulk reads — all only matter in the post-deprecation stack-only world (scope doc Phase 4 notes); dual-write scans always have the PNGs those paths use.

## Review


skbarber:
**Adversarial review: 8 findings, all dispositioned — fix wave in `4af84e8d`.**

The substantive three: **F1** (the reviewer *proved* a malformed stack — valid schema, missing `/acq_timestamp` — failed the task instead of falling back, defeating the PR's central guarantee): stack reads now guarded (`OSError`/`KeyError` → warn + fall back), pinned by a corrupt-stack test. **F2** (proved `ShotRef` breaks at construction on Python 3.12 — masked today by the repo's <3.12 pin): 3.12-proofed now while cheap (`__init__` path-forwarding + `with_segments` deriving plain Paths). **F3** (MagSpec/LineStitcher-class analyzers derive per-shot output names from `file_path.stem` — under a stack every shot shares one path, so opted-in magspec diagnostics would silently overwrite their calibrated outputs): loud WARNING in the `data_format` field docs; a runtime guard is a candidate follow-up if opt-in spreads.

Also fixed: derived-ShotRef refusal is a clean TypeError (F4, + test), class docstring documents the equality/hash and derived-path limits (F5), the claimed-but-missing no-timestamp-column test added (F6), three-strategies docstring (F7), numpy param style + keep-first comment (F8).

**Reviewer affirmations worth keeping** (all empirically probed): fallback atomicity is structurally sound (a partial stack map can never mix with file strategies); the epoch round-trip arithmetic is sound to ~0.5 µs against the ±1 ms candidate window; the writer's frame-before-timestamp append order means timestamp-derived indices are always readable even on crash-truncated stacks; ShotRef construction/pickling/derived-ops all verified working on 3.11 including a real spawn-mode process pool; per-task exception handling means a bad read degrades to a skipped shot, never a crashed run; and a stack-only device dir already passes the data-presence check (no post-deprecation deferral needed there). **Today's end-to-end Scan003 claim explicitly not invalidated.**

Suites after fixes: data-utils scan_stack 6/6, ScanAnalysis mapping 27/27, ImageAnalysis 2/2. Ready for the codex gate.

PR #698, opened 2026-08-27
pva gateway: timestamp plausibility post-conversion (0.4.4) — capture-arc chip

skbarber:
The audit-found bug (chip): `_frame_timestamp` checked `value > 0` BEFORE subtracting the LabVIEW epoch offset, so a value in `(0, offset]` (device counting from boot, zeroed channel) published a **negative Unix timestamp** instead of falling back to receive time — poisoning any consumer that keys frames on the PVA timestamp: the capture daemon's dedupe/stale filter and the analysis-side `acq_timestamp` join. Now checked on the converted value (CA gateway PV_CONTRACT ladder parity), with 3 tests (valid conversion, small-positive fallback, ladder preference). Patch bump 0.4.4 + CHANGELOG. Fleet pickup: pull-on-restart launcher — reaches camera servers on next service restart; no urgency (no device currently emits implausible stamps — Phase-0 evidence shows healthy ladders).

### DIFF
--- a/GeecsPvaGateway/CHANGELOG.md
+++ b/GeecsPvaGateway/CHANGELOG.md
@@ -3,7 +3,7 @@
 All notable changes to this package will be documented here.
 Format follows [Keep a Changelog](https://keepachangelog.com/en/1.1.0/).
 
-## [0.4.4] - 2026-08-27
+## [0.4.4] — 2026-08-27
 
 ### Fixed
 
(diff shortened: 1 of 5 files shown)


skbarber:
Adversarial review: **clean approve** — exact semantic parity with the CA gateway's PV_CONTRACT ladder confirmed against its implementation (post-conversion >0 plausibility, per-rung fall-through, receive-time terminal), and the capture-daemon blast radius traced end-to-end: every documented stale-filter invariant survives (the (1,1) placeholder never passes through this function; cached pre-scan frames still filter), while broken-ladder mid-scan frames flip from silently-all-dropped to saved-with-wall-clock timestamps — the better failure mode. One informational residue in the record: a device with BOTH ladder rungs implausible loses idle-repush dedupe (fresh receive time per push) — edge-of-edge and saving-duplicates beats losing-everything. Two cosmetic nits fixed in the follow-up commit (tests moved off the inaccurate fake_server marker into tests/test_timestamp.py; CHANGELOG dash style). 17/17. Ready for the codex gate.

PR #697, opened 2026-08-27
native_image_save: the devicetype-scoped PNG-deprecation toggle (schemas 0.12.0, GeecsBluesky 0.66.0)

skbarber:
## What

The Phases 3+6 redesign (owner decision, scope doc): **one devicetype-scoped toggle** decides whether capture-eligible cameras (Point Grey — the capture registry) write native per-shot files, or the capture daemon's frame stacks are their only image record. Replaces the former per-save-set-entry `capture_mode` design entirely — the LV save infrastructure is permanent for proprietary devices (HASO, scopes), so image-saving ownership needed one switch, not N fields.

- **GEECS-Schemas 0.12.0** — `ExperimentDefaults.native_image_save` (default **ON**; flipping it is the deprecation step) + `ScanRequest.native_image_save` tri-state override (the `background_telemetry` pattern). Docgen reference + golden fixtures regenerated; 273 tests green.
- **GeecsBluesky 0.66.0** —
  - `db_runtime.GeecsDbDeviceTypes`: fail-open batch devicetype provider (DB blip → nothing eligible → native saving untouched; hermetic sessions without `.experiment` likewise).
  - `resolve_native_image_save` / `select_capture_devices` / `apply_native_image_save_off` in the runner, wired **identically in both execution paths** (headless + queue plan). With the toggle off, captured cameras remain full scan participants (scalars, shot-id columns, strict rows); only the native save + PNG asset docs are suppressed (`save_nonscalar_data=False` → the existing machinery gates itself).
  - Run picture md: `capture_devices` + `native_image_save` (EVENT_SCHEMA.md truth-up).
  - `geecs_run_wrapper`: engine-side `makedirs` of capture device dirs **before the start doc** — the daemon still never mkdirs, and this closes the lazy-writer race window entirely for captured devices.
  - Daemon: prefers the explicit `capture_devices` start-doc key (paths composed as `scan_folder/<device>`) over inferring from `nonscalar_save_paths`; an explicit empty list means "nothing eligible"; absent key falls back to today's inference (older-engine compatible).
  - Optimize-mode scans keep native saving regardless (v0, documented).
- Zero signature changes anywhere — the whole thread rides `extra_md`/md.

## Verification

- GeecsBluesky **full suite 646 passed** (was 618+28 before the hermetic-session fail-open guard); toggle suite 5, daemon suite +2 (explicit-list preference, empty-list semantics), run_wrapper +2 (pre-start-doc mkdir pins)
[...]

PR #699, opened 2026-08-27
active save-off for capture-owned cameras (0.67.0) — codex C1 follow-up

skbarber:
Closes the C1 HIGH from #697's codex gate: with `native_image_save` off, a camera whose LV save flag was left on out-of-band could keep writing PNGs to its stale previous path — the scan wrote no save commands to captured cameras at all.

Now: `apply_native_image_save_off` marks captured devices `save_control_only`; the device classes build **only the `save` control child** (no `localsavingpath`, no save-path column, no asset docs — `save_nonscalar_data` stays False and wins if both flags are ever set); `geecs_run_wrapper` finds marked devices in its existing `devices` param (zero signature changes) and **eagerly commands `save="off"`** at scan start — turning off needs no trigger windowing, unlike save-on. FORMAT.md's documented assumption is replaced by the active contract.

Tests: device-shape (save child present, localsavingpath absent, both-flags precedence), run_wrapper off-write (marked camera gets `off`; native saver's on/off bracket untouched; unmarked devices untouched), wiring assertion in the toggle suite. **Full suite 655 passed.**

Live verification note: the next toggle-off scan will show the off-write in the gateway/device traffic; the behavior is also exercised implicitly on every toggle-off scan from now on.

### DIFF
--- a/GeecsBluesky/CHANGELOG.md
+++ b/GeecsBluesky/CHANGELOG.md
@@ -17,7 +17,9 @@ Format follows [Keep a Changelog](https://keepachangelog.com/en/1.1.0/).
   out-of-band can no longer write native files to a stale path during
   a toggle-off scan. Threaded config → `_build_request_detectors` →
   `session.detector`/`contributor` → the device classes;
-  `apply_native_image_save_off` sets the flag automatically.
+  `apply_native_image_save_off` sets the flag automatically — on ALL
+  three role branches including snapshot (codex P2: an async
+  capture-owned camera previously dropped the flag silently).
 
 ## [0.66.0] - 2026-08-27
 
--- a/GeecsBluesky/CHANGELOG.md
+++ b/GeecsBluesky/CHANGELOG.md
@@ -4,6 +4,21 @@ All notable changes to `geecs-bluesky` are documented here.
 
 Format follows [Keep a Changelog](https://keepachangelog.com/en/1.1.0/).
 
+## [0.67.0] - 2026-08-27
+
+### Added
+
+- **Active save-off for capture-owned cameras** (codex finding C1 on
+  PR #697): with `native_image_save` off, captured cameras are built
+  `save_control_only` — only the `save` control child exists (no
+  `localsavingpath`, no save-path column, no asset docs) — and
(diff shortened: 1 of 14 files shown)


skbarber:
Adversarial review: **approve, no blocking findings** — disposition of the 5 items in the follow-up commit:

- The central attack was **empirically disproven on the installed ophyd-async**: a late-set `save` child registers with the parent (connects, gets named) but never enters describe()/read() — no event-document leakage; every mixin column independently gates on `_save_nonscalar_data`.
- Off-write timing verified: capture-dir mkdir → `save=off` mv → open_run, on both front doors, independent of `defer_save_on`; target liveness covered by the existing CONNECTED re-check (captured cameras remain sync detectors); duck-typed matching cannot false-positive on anything in the tree; the queue plan inherits the new kwarg for free through the class-function-binding facade; optimize paths branch away before the toggle on both doors, consistently.
- **Fixed (LOW×2)**: `save_control_only` numpydoc entries on both device classes; a free-run **contributor-branch wiring test** added (was strict/detector-branch only). Suites: 143 across the touched files.
- **Recorded (notes)**: the async-capture-eligible corner (config-marked but snapshot branch drops the flag — pre-existing shape, matters only if async PG configs ever exist) and the new connect-time requirement (gateway must serve `device:save` for captured cameras; unserved dies loudly pre-claim, which UnservedVariablesCheck can't pre-warn about) — both in the scope doc.
- The no-off-write smoke test's admitted weakness (NIT 5) accepted as-is — the hasattr guard is the protection, not the test.

Ready for the codex gate.

skbarber:
**Codex P2 fixed in the follow-up commit**: the snapshot branch of `_build_request_detectors` now threads `save_control_only` like the other two roles — `CaSnapshotReadable` grows the same save-control child (created outside `add_children_as_readables`, so never in event rows) and `session.snapshot` forwards it. The direction chosen keeps async capture-eligible cameras SELECTED (the daemon captures regardless of role — excluding them would orphan their images under toggle-off) and closes the stale-path hazard for that shape. Pinned by a device-shape test + an async-branch wiring test; full suite **658 passed**. Per the standing process, no re-confirm loop — merging on your go, or immediately if the prior codex verdict stands as the gate.

PR #700, opened 2026-08-28
capture production trio: liveness preflight, systemd kit, dual-write diff CLI (0.68.0)

skbarber:
The three in-repo promotion prerequisites from the arc-end definition, bundled with per-concern breakdown (LOC ≈ 95 / 100 / 160 + tests 65 / 0 / 110):

**1. Daemon-liveness preflight** — `capture/heartbeat.py`: the daemon touches a JSON heartbeat every 10 s from a side thread (atomic replace; path = `~/.local/state/geecs-capture/heartbeat.json`, override `[capture] heartbeat_path`); the engine's `preflight_capture_liveness` **refuses a toggle-off scan pre-claim, fail-closed**, when the heartbeat is missing or >30 s old — wired identically in both execution paths, inert on the default dual-write path. Closes the HARD Phase-6 precondition from #697's review. Honest scope documented in the module: proves *process* liveness; subscription health stays covered by reconciliation counters. The file transport piggybacks the existing shared-filesystem deployment constraint (upgrades to a PV only if worker/daemon ever split hosts).

**2. systemd deployment kit** — `capture/deploy/geecs-capture.service` + `DEPLOYMENT.md` (the qserver template conventions: placeholder paths, absolute-poetry gotcha, restart policy). Worker co-location stated as a *requirement* (save-path view + heartbeat), which also makes the future services-box migration a copy-unit-and-install operation.

**3. Dual-write diff CLI** — `geecs-capture-diff`: per scan/device, joins the stack's timestamps against native PNG filenames on the same canonical-millisecond keys the analysis join uses, pixel-compares every matched pair (IMAQ-decoded), attributes extras (`stack_only` = pre-window frames; `png_only` = **the bad bucket**), verdicts `pass`/`mismatch`/`capture_only`/`no_stack`, appends JSONL evidence via `--log`, exits 1 on mismatch. Weeks of clean log = the Phase-6 gate, greppable.

## Verification
- capture suite 32/32 incl. 11 new tests; **full suite 668 passed**
- Live deployment of the unit + a heartbeat-gated toggle-off scan + a diff run over Scans 003–005 planned immediately post-merge (lab access active)

## Review

### DIFF
--- a/GeecsBluesky/CHANGELOG.md
+++ b/GeecsBluesky/CHANGELOG.md
@@ -4,6 +4,29 @@ All notable changes to `geecs-bluesky` are documented here.
 
 Format follows [Keep a Changelog](https://keepachangelog.com/en/1.1.0/).
 
+## [0.68.0] - 2026-08-28
+
+### Added
+
+- **The arc-end production trio** (promotion prerequisites,
+  `Planning/data_capture/01_central_pva_capture_scope.md`):
+  1. **Daemon-liveness preflight** — `capture/heartbeat.py` (daemon
+     touches a JSON heartbeat every 10 s from a side thread; engine's
+     `preflight_capture_liveness` REFUSES a toggle-off scan pre-claim,
+     fail-closed, when the heartbeat is missing/stale — with native
+     saving off, a dead daemon means images exist nowhere). Wired in
+     both execution paths; the heartbeat file inherits the existing
+     shared-filesystem deployment constraint. Proves process liveness
+     only; subscription health stays covered by reconciliation counters.
+  2. **systemd deployment kit** — `capture/deploy/geecs-capture.service`
+     + `DEPLOYMENT.md` (the qserver pattern; co-location with the worker
+     is a stated requirement; migration = copy unit + install).
+  3. **Dual-write diff CLI** — `geecs-capture-diff` (`capture/diff.py`):
+     per scan/device, canonical-ms timestamp join of stack vs native
+     PNGs with per-pair IMAQ-decoded pixel comparison; verdicts
+     pass/mismatch/capture_only/no_stack; JSONL evidence log via
+     `--log`; exit 1 on mismatch. Weeks of clean log = the Phase-6 gate.
+
 ## [0.67.0] - 2026-08-27
 
 ### Added
--- a/GeecsBluesky/CHANGELOG.md
+++ b/GeecsBluesky/CHANGELOG.md
@@ -21,8 +21,14 @@ Format follows [Keep a Changelog](https://keepachangelog.com/en/1.1.0/).
      devices the daemon isn't covering (a daemon started before a
      camera joined the DB); the daemon **removes the heartbeat on
      clean shutdown** so `systemctl stop` refuses immediately instead
-     of after the 30 s stale window; the beat thread survives any
-     write exception; a relative `[capture] heartbeat_path` override
+     of after the 30 s stale window — including under SIGTERM, which
+     the daemon converts to `SystemExit` so the cleanup actually runs
+     (codex gate P1: Python's default SIGTERM action skips `finally`,
+     so systemd's stop signal would have left the heartbeat behind); a
+     fresh heartbeat **without a device roster is refused** as corrupt
+     state, not tolerated (codex gate P2 — coverage cannot be
+     verified); the beat thread survives any write exception; a
+     relative `[capture] heartbeat_path` override
      is rejected (it would split writer from reader by CWD). The check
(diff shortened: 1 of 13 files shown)


skbarber:
## Adversarial review — dispositioned (fix wave: 6eed0a7c)

Fresh-context reviewer, 10 findings. All addressed in-PR except one deliberate accept:

|---|-----|---------|-------------|
| 1 | HIGH | Diff join lacked the analysis join's ±1 ms `timestamp_key_candidates` tolerance — reviewer Monte-Carlo'd the epoch/wire float round-trip: ~2×10⁻⁴ of frames land one canonical-ms key away, each a false `png_only`+`stack_only` mismatch pair poisoning the Phase-6 evidence log | **FIXED** — candidate pairing after the exact intersection; docstring claim made honest; pinned by a boundary-key test |
| 2 | MED | `png.astype(frame.dtype)` wraps out-of-range values → a real pixel difference could read "identical" | **FIXED** — cast dropped, direct comparison under numpy promotion; pinned (257→uint8 wrap test) |
| 3a | MED | Clean daemon stop (`systemctl stop`) left a ≤30 s window where toggle-off preflight still passes | **FIXED** — daemon removes the heartbeat in its shutdown path; stop → immediate refusal. (SIGKILL/power-loss still takes the stale window — inherent, documented) |
| 3b | MED | Heartbeat carried only a target *count*: a daemon predating a roster addition passed preflight while not monitoring that camera | **FIXED** — payload now carries the device-name roster; preflight refuses toggle-off scans with uncovered devices, naming them |
| 4 | MED | Runbook said mid-scan restart "loses capture only, native files untouched" — false on toggle-off scans | **FIXED** — restart guidance states the toggle-off exception |
| 5 | LOW | "Shared filesystem" claim inaccurate (file is host+user-local); off-host headless callers get a misdiagnosing refusal | **FIXED** — docstring corrected; refusal message states the check is host-local; payload read once (no age/verdict race) |
| 6 | LOW | Beat thread died permanently on non-OSError; relative `heartbeat_path` override splits writer/reader by CWD | **FIXED** — `except Exception`; relative overrides rejected with a warning |
| 7 | LOW | CLI conflated operational failure with mismatch; one bad path aborted the sweep with no log rows | **FIXED** — exit 2 for operational errors, per-scan try/except, incremental JSONL writes; pinned |
| 8 | LOW | Duplicate-key policy diverged from the analysis join (last vs first) | **FIXED** — `setdefault` (keep-first) |
| 9 | LOW | Nothing pinned the queue-plan preflight wiring | **FIXED** — `test_toggle_off_without_daemon_heartbeat_refuses_before_claim` (refusal + no claim burned) |
| 10 | INFO | systemd `${VAR}`-in-ExecStart question | **CONFIRMED CORRECT** by the reviewer (systemd expands `Environment=` vars as one unsplit token). Nits taken: runbook notes the heartbeat lives in the *service user's* home, and adds the recurring diff-sweep cron line so the Phase-6 gate is operational, not aspirational |

Accepted residual (documented, not fixed): a daemon killed without cleanup keeps its heartbeat up to 30 s; mid-scan daemon death on a toggle-off scan is detected only by reconciliation — both stated in module docs and runbook.

Tests: **676 passed** (7 new).

## Review
- [x] Adversarial review — dispositioned above

skbarber:
## Codex gate — both findings fixed (812e0e23)

**P1 (SIGTERM skips the tombstone) — confirmed and fixed.** Codex is exactly right: Python's default SIGTERM action terminates without running `finally`, so under `systemctl stop` the clean-stop tombstone never fired — on precisely the deployment path it exists for. The daemon now installs a SIGTERM handler converting it to `SystemExit`, so the heartbeat removal (and `daemon.shutdown()`) run. Chose the code-side fix over `KillSignal=SIGINT` in the unit: it also covers non-systemd invocations (the current nohup interim daemon) and keeps the unit file stock. Pinned by a test that delivers a real SIGTERM to the process and verifies the `finally` cleanup ran.

**P2 (rosterless heartbeat tolerated) — agreed and flipped.** My "older daemon" tolerance case was vestigial — this PR's writer always emits the roster, so a fresh payload without one is corrupt/foreign state and coverage cannot be verified. Now a refusal with its own message; the tolerate-test became a refuse-test.

**On the verification env**: the `[Errno 17] File exists` venv collision is the known worktree-poetry footgun — reviewers should run against the package's existing env (`poetry env use` on the main checkout's interpreter) rather than letting poetry create a fresh one. Full suite here: **677 passed**.

## Review
- [x] Adversarial review — dispositioned (10 findings)
- [x] Codex gate — P1 + P2 fixed in 812e0e23

Issue #412, opened 2026-05-28
Move scalar-metric prefix/suffix from ImageAnalysis to ScanAnalysis


skbarber:
## Summary

The string used to prefix every scalar metric an analyzer emits (e.g. `UC_GaiaMode_centroid_x`) currently lives inside ImageAnalysis as `CameraConfig.name` / `Line1DConfig.name`. Promote that concern to ScanAnalysis, where the device-name context already lives, and let ImageAnalysis emit bare scalar keys.

Companion to issue #400 (the PR-E loader-API refactor). #400 cleaned up the YAML schema and the load path; this issue cleans up the analyze-and-name path.

## Motivation

`camera_config.name` (and `line_config.name`) plays three roles today, only one of which is load-bearing:

1. **Metric key prefix** — every scalar in `ImageAnalyzerResult.scalars` is emitted as `{name}_centroid_x`, `{name}_size`, etc. This is the only role that affects downstream data.
2. **Result metadata** — passively embedded in `ImageAnalyzerResult.metadata` as `camera_name` / `line_name`. Nothing currently reads these.
3. **Log identifier** — cosmetic `logger.info` lines.

The metric-prefix role doesn't belong in ImageAnalysis. ScanAnalysis is the layer that:

- Knows the `device_name` (it has to, to find the data folder)
- Knows the per-analyzer-instance `metric_prefix` override from the YAML
- Actually writes the scalars to the s-file

Pushing prefix-naming into ImageAnalysis means analyzers carry naming state that scientifically belongs to the scan context. It also forces the awkward `StandardAnalyzer.name_suffix` pattern — which **mutates `camera_config.name` in place** to make variant analyses (same `BeamAnalyzer`, two ROIs) emit distinct scalar columns. Variant naming is a scan-level concern, not an analyzer-internal one.

## Current architecture

```
ImageAnalysis: analyze_image() -> ImageAnalyzerResult
                                  scalars = {
                                    "UC_GaiaMode_centroid_x": 512.3,
                                    "UC_GaiaMode_size": 45.0,
                                    ...
                                  }
                                          |
                                          v
ScanAnalysis:  aggregate results -> s-file row with pre-named columns
```

ImageAnalysis is doing the naming. ScanAnalysis is a passthrough for column names. The prefix is baked in at the analyzer-instance level, and `StandardAnalyzer.name_suffix` is the only way to vary it — by mutating the config.

## Proposed architecture

```
[...]

PR #701, opened 2026-08-28
toggle-off scans keep the acq_timestamp s-file column (0.68.1) — stack join key

skbarber:
## What

Two-line gate fix in the sync device roles: the `<Device> acq_timestamp` s-file column — the key ScanAnalysis's `device_hdf5` strategy joins stack frames to scan rows by — now survives the `native_image_save` toggle. The `localsavingpath`/`save` native-save children still ride `save_nonscalar_data` only (the gate is split; both directions pinned).

## Why (found live, not by review)

First real ScanAnalysis run over a capture stack — today's PNG-free Scan001 — mapped **zero shots**: the toggle-off s-file had no device columns at all, because `save_control_only` devices fell into the pure-scalar column rule. With no PNGs, the fallback couldn't mask it: the scan's images were simply orphaned from analysis. Dual-write Scan003's s-file has the column; toggle-off Scan004/Scan001 didn't.

Notably, **three review passes (adversarial + codex, PRs #697/#699) accepted this shape** — the prior test even asserted `acq_timestamp` absent as correct for capture-owned devices. The event data itself is in Tiled (only the s-file column mapping was gated), so yesterday's toggle-off scans are recoverable by re-export if ever needed.

Known residual (unchanged, pre-existing): async (snapshot) capture-owned cameras have no acq column — the snapshot role never had one.

## Verification

- 678 passed (the flipped pin + a contributor twin + save-children gate pinned both ways)
- Live re-verification planned immediately: pull on the worker, manager restart, fresh toggle-off scan, ScanAnalysis stack run expecting 10/10 mapped ShotRefs

### DIFF
--- a/GeecsBluesky/CHANGELOG.md
+++ b/GeecsBluesky/CHANGELOG.md
@@ -4,6 +4,26 @@ All notable changes to `geecs-bluesky` are documented here.
 
 Format follows [Keep a Changelog](https://keepachangelog.com/en/1.1.0/).
 
+## [0.68.1] - 2026-08-28
+
+### Fixed
+
+- **Toggle-off scans keep the `acq_timestamp` s-file column** — the
+  stack join key. `save_control_only` devices (capture-owned cameras)
+  reused the pure-scalar column rule, so the s-file carried no
+  `<Device> acq_timestamp` column and ScanAnalysis's `device_hdf5`
+  strategy had nothing to join on: the whole scan's images were
+  orphaned from analysis. Found LIVE on the first real
+  stack-analysis run (26_0828 Scan001 — PNG-free, so the fallback
+  couldn't mask it; the prior test suite had pinned the buggy shape
+  as correct). A capture-owned camera is a file-producing device:
+  the column now survives the toggle on both sync roles
+  (`CaGenericDetector`, `CaTimestampedReadable`); the
+  `localsavingpath`/`save` native-save children still ride
+  `save_nonscalar_data` only. Async (snapshot) capture-owned cameras
+  still have no acq column — a pre-existing property of the snapshot
+  role, noted, not changed here.
+
 ## [0.68.0] - 2026-08-28
 
 ### Added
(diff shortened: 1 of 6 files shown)


skbarber:
## Live verification — the arc-end loop closed end to end (26_0828 Scan002)

Worker pointed at this branch (3e680338), manager restarted, fresh heartbeat-gated toggle-off scan:

- Scan002 ran PNG-free (folder holds only `UC_Amp4_IR_input.h5`); daemon reconciliation clean.
- s-file now carries `UC_Amp4_IR_input acq_timestamp` (the fix; Scan001's, pre-fix, has no device columns).
- **Real ScanAnalysis run**: `Array2DScanAnalyzer(device_name=..., image_analyzer=StandardAnalyzer(CameraConfig(bit_depth=16)), data_format='device_hdf5')` → `Mapped 10 shot(s) into capture stack`, all 10 as `ShotRef`s, per-shot pipeline executed (ProcessPoolExecutor — ShotRef pickling proven in anger), average-image HDF5 + rendered visual written to `analysis/Scan002/UC_Amp4_IR_input/Array2DScanAnalyzer/`.

This is the arc-end definition demonstrated live: HDF5-only save + analysis handles it.

Incidental (pre-existing, NOT this arc): the configs-repo camera YAMLs (e.g. `UC_Amp4_IR_input`) fail `load_camera_config` post-#412 (`extra=forbid` rejects the dropped `name:` field + stray keys like `location`) — same error visible in an old failed-optimize queue item. The configs-repo modernization is its own chore; the planned first-opt-in `device_hdf5` diagnostic PR should fix that camera's YAML while it's in there.

## Review

skbarber:
## Adversarial review — dispositioned (polish: 0d6a93dc)

**Verdict: no blocking defects** — all five attack angles clean (four-combo gate enumeration byte-identical where it must be; no duplicate `save` child in any combo; every `_column_headers` consumer traced — none infers native saving from the column; snapshot-branch leakage impossible, routing is strictly on `synchronous`).

|---|-----|---------|-------------|
| F1 | LOW | `EVENT_SCHEMA.md` still claimed `-acq_timestamp` is "excluded by construction" — false since the original native-saver gate, more visibly so now | **FIXED in-PR** — the file-producing-device exception is now documented (natives + capture-owned) |
| F2 | LOW | Pre-existing (#697/#699): an async capture-eligible camera with an empty `variable_list` lands in `capture_devices` but no device is built, so the eager `save=\"off\"` never reaches it — the stale-path hazard | **FILED** as #702 (not this PR's regression; sync roles unaffected) |
| F3 | NIT | New contributor test duplicates the existing shape test | **ACCEPTED as-is** — the duplication reads as intent documentation |
| F4 | NIT | Both-flags combo didn't assert the column | **FIXED in-PR** — pinned via either operand of the `or` |

48/48 device+toggle tests pass. With the live end-to-end evidence above (Scan002: 10/10 mapped, analysis outputs written), ready for the Codex gate.

## Review
- [x] Adversarial review — dispositioned above

Issue #702, opened 2026-08-28
Async capture-eligible camera with empty variable_list escapes the active save-off


skbarber:
Found during the adversarial review of PR #701 (pre-existing — introduced neither by #701 nor worsened by it; the #697/#699 toggle machinery).

`_build_request_detectors` (`GeecsBluesky/geecs_bluesky/scan_request_runner.py` ~1210) skips asynchronous devices with no scalar variables ("no scalars to record") before any device object is built — but `select_capture_devices` selects on `save_nonscalar_data` + devicetype only. An async capture-eligible camera with an empty `variable_list` therefore:

- lands in `capture_devices` / the toggle start-doc metadata,
- gets `save_control_only=True` in the config copy,
- yet no device object exists — so `run_wrapper`'s eager `save=\"off\"` loop never commands it off.

That's exactly the stale-path hazard the active off-write surface was built against (a save flag left on out-of-band writes native files to a stale path during a toggle-off scan).

Likely shapes: build the save-control device even for scalar-less async cameras when they're capture-owned, or drop them from `capture_devices` with a loud warning. Sync roles (the production case) are unaffected.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

skbarber:
Extending this issue with a sibling async-role gap found by the #693 pre-promotion review (engine-wiring dimension): a snapshot-role (asynchronous) capture-owned camera with a NON-empty `variable_list` gets the `save` control child and the eager off-write, but — unlike the sync roles since 0.68.1 — surfaces no `acq_timestamp` s-file column (`CaSnapshotReadable` has no acq machinery at all), so its capture stack cannot be row-joined unless the save set lists `acq_timestamp` explicitly. Same orphaning class 0.68.1 fixed for sync roles. EVENT_SCHEMA.md now documents the exception and points here (0.68.2). Both shapes in this issue share the same alternative remedy: either build full capture support for async cameras (save child + join column), or drop async devices from `capture_devices` with a loud warning.

PR #693, opened 2026-08-27
Central PVA image capture arc (feat/pva-capture → master)

skbarber:
# Central PVA image capture arc — promotion to master

The complete arc from `Planning/data_capture/01_central_pva_capture_scope.md`: a central capture daemon subscribes to the PVA image PVs of a scan's Point Grey cameras (gated by the Bluesky document stream) and writes **one HDF5 frame stack per device per scan** beside the LabVIEW per-shot PNG save — attacking the measured NAS pathology (21k files/scan, ~40 MB/s path cap) at the file-count root. Dual-write is the default and the PNGs remain the system of record; the deprecation path is the `native_image_save` toggle, hard-gated on daemon liveness.

**Every constituent PR below carried its own review ritual, and the integrated whole was additionally re-reviewed at the owner's request**: a six-dimension fresh-context adversarial pass over the full 73-file diff plus a cloud ultra review — no blockers from either; every finding fixed in #703 or explicitly dispositioned (see the review comments below). This PR is the integration merge.

## Constituent PRs (all merged into `feat/pva-capture`)

| PR | What | Review |
|----|------|--------|
| #694 | Capture daemon v0: `geecs_bluesky.capture` — scan-gated doc-stream consumer, deep-queue p4p monitors, acq_timestamp dedupe + stale filter, per-device HDF5 stacks (`FORMAT.md`, schema `geecs-capture/1`), reconciliation counter identity | adversarial (1 HIGH fixed: lazy writer creation under defer_save_on) + codex (2 P1 fixed: pyzmq extra, append_failures bucket) |
| #695 | Writer compression: shuffle+gzip-1 chunked (1,H,W) — stack now ~3× smaller than the PNG sum | both clear |
| #696 | Phase 4 read side: `ShotRef`/`find_stack_file`/`read_shot`/`read_stack_timestamps` (data-utils `io/scan_stack.py`), `load_image` ShotRef branch, ScanAnalysis opt-in `data_format: device_hdf5` with canonical-ms timestamp join + unconditional per-shot-file fallback | both clear |
| #697 | The `native_image_save` toggle (schemas 0.12.0): ExperimentDefaults bool + ScanRequest tri-state, devicetype-scoped capture eligibility (fail-open DB provider), `capture_devices` start-doc key, engine mkdirs capture dirs pre-start-doc | adversarial + codex (C2 fixed: fallback false-alarm scoping; C1 split to #699) |
[...]

PR #703, opened 2026-08-29
capture #693 pre-promotion review fix wave (0.68.2 / 1.17.1 / 0.12.1)

skbarber:
The fix wave from the six-dimension adversarial review of promotion PR #693 (owner-requested whole-diff re-review). **No blockers were found**; this lands every worthwhile finding. Per-concern breakdown:

## Code fixes
- **Counter-identity close race** (daemon dimension, MED-HIGH): a p4p in-flight frame delivered after `close()` began landed in NO reconciliation bucket — the one provable break of the identity the evidence engine leans on. Closing flag now set under the lock before unsubscribe; `_on_frame` check-and-divert to `late_frames` is atomic. Pinned by a new test (frame injected during source close → identity closes in the finalized file).
- **Partial-subscribe source leak** (MED): an exception on monitor *k* left monitors 0..k-1 + the p4p context alive forever, holding the gateway's subscription gate open. Failure path now closes the source. Pinned.
- **Tombstone resurrection race** (LOW): a beat mid-write during shutdown could atomically replace the heartbeat after the unlink. Beat thread now has a stop event and is joined before `clear_heartbeat()`.
- **`finalize()` handle leak** on failed attr stamping — closes regardless (file is un-finalized, already tolerated).
- **`geecs-capture-diff --log`** creates its parent dir (fresh-host cron sweep no longer tracebacks as a fake mismatch).
- **systemd `RestartPreventExitStatus=2`** — permanent misconfig no longer restart-loops.
- **ScanAnalysis fallback hardening** (read-path dimension): stack-read guard broadened to `Exception` — the reviewer constructed a malformed-but-schema-valid stack whose `TypeError` escaped `(OSError, KeyError)` and failed the task instead of falling back to the PNGs beside it.
- **Optimize mode WARNs** when discarding an explicit `native_image_save=false` (both paths; was silent — matters once the experiment default ever flips).

## Docs / provenance
- FORMAT.md + daemon docstring truth-ups (pre-start-doc dir creation replaced the stale claims; first-append-failed file = valid-but-empty; conditional `scan_number`; dedupe-on-accepted semantics + the 0.4.4 receive-time-fallback signature).
- heartbeat docstring names the **wedged-writer residual gap** (h5py global lock — beat stays green while later scans' writes block); scope doc gains the pre-Phase-6-flip hardening list (writer health in the beat, fleet ordering, `background_source` on PNG-free scans, #702, optimize provenance).
- DEPLOYMENT.md: **upgrade the PVA fleet to ≥0.4.4 before any toggle-off scan** (a 0.4.3 box stale-filters 100% of a broken-ladder camera's frames — invisible to the liveness preflight).
- EVENT_SCHEMA: the acq-column exception is sync-roles-only (async snapshot gap → #702, extended).
- Schema field description states the refusal/inert behaviors (reference regenerated); GeecsBluesky relock (sibling path-dep versions were one bump stale — the scope reviewer's F1).

## Explicitly accepted (no change)
Requested-vs-effective metadata convention; the extra per-scan DB query (fail-open, pinned by test); partial-stack-join no-fallback (parity doctrine); post-claim eager save-off abort semantics (matches existing save puts).

Tests: GeecsBluesky **680 passed** (2 new pins), ScanAnalysis 192, GEECS-Schemas 273 (docgen no-drift included).


skbarber:
## Ultra review of #693 — both findings folded in (2f904d74)

The cloud multi-agent review returned **2 findings, both quality nits, zero blockers** — independent confirmation of the six-dimension verdict.

1. **Toggle-off block duplicated across both execution paths** → extracted `resolve_and_apply_capture_toggle`, one module-level prologue helper shared by the runner and the queue plan (the pattern every other prologue step already follows — the two paths are now lockstep by construction). Plan-path refusal test retargeted and green.
2. **`read_shot` reopens the stack per shot** (the open-handle cache was in the Phase-4 design list but neither shipped nor marked deferred) → recorded as an **explicit deferral** in the scope doc with the per-process-cache design sketch (each ProcessPool worker amortizes its own share; h5py thread-safety noted). Deferred until an opt-in `device_hdf5` diagnostic measurably hurts — correctness is untouched.

Full suite: **680 passed**.

## Review
- [x] Six-dimension adversarial review — fix wave (this PR)
- [x] Ultra review (cloud) — 2 nits, both dispositioned above

PR #709, opened 2026-08-29
Portal 0.2.0 + Data-Utils 0.17.0: resource viewer — tiered image gallery (portal arc phase 4)

skbarber:
## What

Phase 4 of the data-portal arc: the image gallery — the piece that answers "can I see my beam images in a browser." Live-verified against real data this session (Scan 012, 2026-08-21: real `UC_Amp4_IR_input` shots served).

Per-concern breakdown:

- **`geecs_portal/resources.py` (new, ~250 LOC)**: (scan folder, device, shot) → displayable image per the scope doc's tiering — capture HDF5 stacks via `scan_stack` (Tier A, frame = shot−1), native files (Tier B) via `ScanPaths.build_asset_path`/`infer_device_ext` + `read_imaq_image`, vendor formats (`.himg`/`.has`) as a path card (Tier C). Percentile-windowed (1–99.7) 16-bit→8-bit rendering. Device names validated against the folder's real subfolders (traversal guard). Strictly read-only, tree-untouched pinned for hits *and* misses.
- **The Bluesky-native filename join (the live discovery)**: production scans write `<device>_<labview_seconds>.png`, not the legacy `ScanNNN_device_shot` names (the open compatibility question in TILED_SETUP.md). The viewer joins those by the event row's device `acq_timestamp` — epoch-agnostic (event rows carry LabVIEW epoch, verified live; Unix sources get the offset), ±1 s tolerance, **404 on mismatch rather than serving the wrong shot**, ordinal order as the no-metadata fallback.
- **Routes/UI**: gallery section in `/run/{uid}` (`?device=&shot=`, prev/next + shot input, tier-degraded vendor card), `/run/{uid}/image.png`.
- **GEECS-Data-Utils 0.17.0 (behavior change, minor)**: `resolve_scan_folder` no longer gives up when the recorded `scan_folder` path doesn't exist on *this* host — it falls through to the local daily-path fallback. The recorded path is host-specific (worker records `/mnt/...`; clients mount `/Volumes/...` or `Z:`), so the old short-circuit made every cross-host consumer fail spuriously (found live: the gallery was empty on macOS for every worker-recorded run). Console's Open button inherits the same improvement; affected tests made hermetic (the fallback must never touch the real data root in tests).

## Tests

`check.sh --base origin/feat/data-portal`: **Portal 40 passed** (13 new: tier routing, timestamp join incl. wrong-timestamp-404s, ordinal fallback, traversal guard, read-only pins, vendor card, gallery routes), **Data-Utils 218 passed / 3 skipped** (incl. the new foreign-host re-basing test), **Console 788 passed**, lint green.

## Hardware verification

Live-verified this session over VPN against the real catalog + share: gallery lists real devices, real shot served as PNG (~2.6 s over VPN), timestamp join confirmed against actual file stamps. OWED (phase 5): the same check served *from* the worker host, plus a capture-arc HDF5-stack scan once one exists on disk.


skbarber:
## Adversarial review (fresh-context subagent) — report & dispositions

The strongest review of the arc: mutation probes proved the original tests never asserted *which* frame a request returns, and three of four substantive findings shared one root cause — the PR re-implemented joins the repo already owns.

### Findings & dispositions (all FIXED in 4e80a63d + 8cd3eaae, all reviewer-confirmed with mutations re-run)

**1. HIGH — stack tier joined by ordinal `shot−1`, contradicting the stack's own contract** (FORMAT.md: `/acq_timestamp` is "the universal join key"; a pre-scan extra frame shifts every ordinal; ScanAnalysis's `_map_shots_from_stack` joins by timestamp). Mutation-verified blind: serving the wrong frame passed all tests.
→ Canonical join: `read_stack_timestamps(labview_epoch=True)` + `native_files` millisecond keys; no-match 404s; ordinal only without metadata. Fixture now carries a **leading pre-scan extra frame** and per-index pixel markers; the reviewer re-ran its mutation — **now caught**. Residual (reviewer-flagged) ordinal-fallback off-by-one blindness closed in 8cd3eaae with a marker assertion.

**2. HIGH — `_timestamped_files` + ±1 s tolerance re-derived `geecs_data_utils.native_files`** ("THE native per-shot naming contract… never widen [the ms canonicalisation] into a tolerance") — at 1 Hz a missing shot's *neighbour* would be served under its label.
→ Deleted wholesale; exact stat-probes over `native_file_name_from_key(timestamp_key_candidates(...))`; `test_missing_shot_never_serves_a_neighbour` pins refusal with a 0.4 s neighbour present. Portal added to `native_files`' consumer list.

**3. MED-HIGH — the `resolve_scan_folder` fall-through trusted the caller's `day`** (a bookmarked link with no `day` would resolve *today's* same-numbered scan — numbers restart daily).
→ `_run_day` derives the day from the run's own `start_time`; live-verified with no-day-param links; regression-pinned in 8cd3eaae (recorder test asserts the fallback receives the run's day even against a wrong `day` param).

**4. MEDIUM — `safe_name` re-derived wrongly (hyphenated diagnostic stems never matched → silent ordinal degrade); NaN rows served a neighbour.** Mutation-verified blind.
→ New `tiled_schema.device_acq_timestamp_column` (Data-Utils 0.18.0 — the mandated home per its CLAUDE.md), collapse-runs normalization with tests incl. `U_BCaveMagSpec-interpSpec`; column-present-but-NaN rows now 404 "device missed this shot". Reviewer re-ran mutation B — **now caught** (by the NaN-refusal test; noted honestly: the happy-path fixture remains ordinal-aligned, so the catch is via the refusal path, which is the realistic failure mode).

**5. LOW — placement** → resolved by 2/4: joins live in data-utils; portal keeps rendering + tiering only.

**6. LOW — dead ext entries (`.has` devices got "missing png" instead of the Tier C card), Pillow undeclared, float64 frames** → `_vendor_only` dir probe (+ tests), `pillow` declared + relocked (lock worktree-path-free), float32.

Verified non-findings: scan-folder invariant (mkdir mutation **caught** by the fallback pin), path traversal (membership guard sound incl. crafted names), console blast radius (Open button improved, contract text still true), corrupt-file containment, lazy-loading rule.

Suites at head: Portal 46, Data-Utils 222/3 skipped, Console 788, lint green; live-verified over VPN (real shots via the canonical join; bookmarked no-day links).

PR #714, opened 2026-08-30
portal review-fix wave B: one join/tier implementation in data-utils (0.20.0 / 0.4.1 / 1.17.2)

skbarber:
## What

**PR B of the #712 review-fix wave**: the shot↔file join and tier machinery gets ONE implementation each in GEECS-Data-Utils, and the two consumers (portal, ScanAnalysis) delegate — closing the review's "created a shared seam but left sibling copies alive" cluster, including one real parity break it exposed.

Per-concern breakdown (~437 ins / ~156 del over 18 files, ~90 are tests):

- **Stack join** (Data-Utils `io/scan_stack.py` +~110): `stack_frame_index_map` (**keep-first** duplicate-ms-key contract — the portal's private copy was keep-*last*, so a duplicate-key stack would have served a different frame in the portal than ScanAnalysis analyzed), `frame_index_for_timestamp` (exact-key ±1 ms canonicalisation probe), `read_shot_for_acq_timestamp` (single-shot join in **one** h5py open — the gallery's hot path was three opens per image request over SMB). Pinned by `TestSharedJoinHelpers` (keep-first, exact-key-never-window, single-open join).
- **Native probe** (Data-Utils `native_files.py` +~50): `probe_native_file` — the exact-key stat probe with the SMB stale-listing-cache rationale attached to the one copy (it lived only on ScanAnalysis's). Pinned by `TestProbeNativeFile`.
- **Extension taxonomy** (Data-Utils `scan_paths.py`/`io/images.py`): `VENDOR_ONLY_EXTS` + `DISPLAYABLE_IMAGE_EXTS` replace three uncoordinated consumer-local sets. **Judgment call (cheap veto): `_ACCEPTABLE_EXTS` gains `has`** — the missing sibling of the already-accepted `himg`; a `.has`-only HASO folder now infers its real extension instead of defaulting to `png`, which retires the portal's `_vendor_only` directory-scan workaround. Side effect: `geecs_bluesky` asset registration would also infer `.has` for such folders (currently it expects a `.png` that never exists) — an improvement in our reading, but it does change worker-side behavior for `.has`-only devices; veto → we re-add the portal-local workaround.
- **Matcher** (Data-Utils `tiled_schema.py`): `_normalize_token` → public `normalize_token` (docstring names it THE device↔column rule); ScanAnalysis's byte-identical private copy now delegates.
- **Tier ladder** (portal `resources.py` −~80): `device_kind` (now a `DeviceKind` NamedTuple carrying the ext) is the one ladder; `load_shot_image` dispatches on it, so the UI badge and the image endpoint can never disagree; `run_view` passes its device listing in (no second directory scan per page render).
- **Formatter**: `fmt_time_of_day` moved down beside `_fmt_datetime`; portal's `_fmt_hhmm` retired. (The console's B1/B4 swap onto the shared helpers is the next PR — its CHANGELOG note already points here.)
- ScanAnalysis `single_device_scan_analyzer.py`: three delegations, behavior preserved (the shared helpers encode its own semantics).

Versions: Data-Utils 0.19.0 → **0.20.0**, portal 0.4.0 → **0.4.1**, ScanAnalysis 1.17.1 → **1.17.2**.

**Deferred, recorded**: `geecs_bluesky/capture/diff.py`'s bulk set-reconciliation join keeps its local keep-first map build (correct + commented; adopting `stack_frame_index_map` means a GeecsBluesky bump/suite for a 4-line swap — next time that file is touched). Console B4/B1 rewire = next PR. Legacy MC `device_file_stem` handling = inventory-doc territory (the historic-ingester discussion changes its shape).

## Tests

`scripts/check.sh`: **Console 789 · Data-Utils 230 (+4 new pins) · Portal 60 · ScanAnalysis 192 passed**, lint clean.

## Hardware / live

No deployment change; behavior preserved by design (the keep-first fix changes portal behavior only for sub-millisecond duplicate stack timestamps — physically unreachable at fleet rep rates, but now contractually consistent).


skbarber:
**Adversarial review** (fresh-context subagent, three-lens brief per /land) — five findings, all **fixed in 1eb40bd0 and confirmed by the same reviewer** (5/5, arithmetic re-verified):

1. **Medium — unguarded second ScanPaths could 500 on an SMB blip**: the native path builder is now the one, guarded construction; `device_kind` needs no ScanPaths at all. **Fixed/confirmed.**
2. **Medium — vendor devices in non-canonical layouts regressed to "missing"**: new module-level `scan_paths.infer_device_dir_ext` (method delegates); tier classification is layout-independent, natives degrade to the layout card. Pinned by `test_non_canonical_layout_degrades_never_raises`. **Fixed/confirmed** (reviewer note: endpoint behavior = pre-PR parity; the badge in scratch folders is genuinely improved — CHANGELOG amended in the follow-up commit).
3. **Medium-low — the `.has` veto was framed on a false premise** (geecs_bluesky never calls `infer_device_ext`; its readback hardcodes png): taxonomy comment now names the real consumers (portal tier probe, `ScanData.add_expected_paths`) + the mixed-folder counting nuance; `stack_frame_index_map` docstring marks capture-diff as a deliberate non-adopter. **Fixed/confirmed.** *Owner note: the earlier veto framing in this PR's body overstated the blast radius — the `.has` addition touches only those two consumers.*
4. **Medium-low — boundary pins were vacuous** (the ±-values canonicalised to the same key): both pins now use genuinely divergent values (`…100.0005` renders `.001` but keys `.000`); deleting the candidate probe now fails the suite. **Fixed/confirmed.**
5. **Low — the listing-map candidate loop survived the consolidation**: `frame_index_for_timestamp` generalized to `Mapping[int, T]`; ScanAnalysis's `files_by_ms` fallback delegates, and the map build is keep-first per the shared contract (deliberate alignment, noted in the CHANGELOG). **Fixed/confirmed.**

Residual nit accepted as-is: `frame_index_for_timestamp` is a slight misnomer for the value-generic probe (docstring covers it). Refuted by the reviewer along the way: DeviceKind NamedTuple call-site breakage, the one-open join's exception surface, .is_file→.exists, fmt_time_of_day parity, all placement choices. No undispositioned findings remain.

PR #716, opened 2026-08-30
portal 0.5.0: within-scan prefetch caches (owner feature — eager within a scan, lazy across scans)

skbarber:
## What

**PR D — the owner's prefetch feature** (2026-08-29 request): per-shot round trips optimize for the wrong constraint when a diagnostic's per-scan data is ~100s of MB and RAM is free; navigation should serve from memory. Doctrine amendment recorded in CLAUDE.md: **eager within a scan, lazy across scans** (the scope doc's lazy rule still binds across scans — no day-thumbnailing).

- `geecs_portal/cache.py` (~250 LOC): **`CachingScanCatalog`** — delegating ScanCatalog caching `load_run` (completed runs LRU-8 forever, still-running runs 5 s TTL; KeyError misses never cached; `list_runs`/`probe` pass through so the day list stays live). Ends the repeated full-event-table Tiled reads (`__main__` wraps the real catalog). **`ShotDataCache`** — bytes-bounded LRU (~1.5 GB) of pixel data keyed `(uid, device)`, completed runs only: stack devices load the **whole frames array in one HDF5 read** on first touch; native devices cache decoded per-shot arrays, background-warmed by a daemon thread walking the run's event rows (timestamp-joined shots only).
- `resources.py`: cache-aware fast paths (a hit serves with zero filesystem access — no listing, no open); ordinal (listing-order) resolutions are **never cached** (same rule as the no-long-cache header from wave A).
- `app.py`: `run_image` passes the cache for completed runs; `run_view` schedules the native warm on first gallery touch (only when the device has an acq_timestamp column — ordinal-only devices never warm-loop).

Safety properties, each pinned: still-running runs never enter the pixel cache (live-run test deletes the file → 404, proving the share is still read); completed-run hits serve **after the underlying file is deleted** (the zero-filesystem proof); exact-key refusal (never-serve-a-neighbour) enforced identically from memory; budget eviction LRU.

Version: portal 0.4.2 → **0.5.0**.

## Tests

`scripts/check.sh`: **Portal 71 passed (+10 cache pins) · Console 790 · Data-Utils 231 · ScanAnalysis 192**, lint clean.

## Hardware / live

Deploys with the promotion (host pull + restart). Memory ceiling ~1.5 GB pixel budget + 8 run details — comfortable on the .14 box.

### DIFF
--- a/GEECS-DataPortal/CHANGELOG.md
+++ b/GEECS-DataPortal/CHANGELOG.md
@@ -3,6 +3,23 @@
 All notable changes to this package will be documented here.
 Format follows [Keep a Changelog](https://keepachangelog.com/en/1.1.0/).
 
+## [0.5.0] - 2026-08-29
+
+### Added
+
+- **Within-scan prefetch caches** (`geecs_portal/cache.py` — owner
+  feature request, amending the scope doc's blanket lazy rule to "eager
+  within a scan, lazy across scans"): `CachingScanCatalog` keeps
+  completed runs' details (LRU-8; still-running runs expire in seconds),
+  ending the repeated full-event-table Tiled reads; `ShotDataCache`
+  (bytes-bounded LRU, ~1.5 GB) keeps completed runs' pixel data — a
+  stack device's whole frames array in ONE HDF5 read on first touch, a
+  native device's decoded shots warmed by a background thread walking
+  the event rows — so stepping through a gallery serves from memory
+  with zero filesystem access (pinned by delete-the-file-then-serve
+  tests).  Still-running runs and ordinal (listing-order) resolutions
+  are never cached.  `__main__` wraps the real catalog.
+
 ## [0.4.2] - 2026-08-29
 
 ### Changed
--- a/GEECS-DataPortal/CHANGELOG.md
+++ b/GEECS-DataPortal/CHANGELOG.md
@@ -18,7 +18,14 @@ Format follows [Keep a Changelog](https://keepachangelog.com/en/1.1.0/).
   the event rows — so stepping through a gallery serves from memory
   with zero filesystem access (pinned by delete-the-file-then-serve
   tests).  Still-running runs and ordinal (listing-order) resolutions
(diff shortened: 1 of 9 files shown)


skbarber:
**Adversarial review** (fresh-context subagent, three-lens brief per /land) — three findings, all **fixed in c41b195d and confirmed by the same reviewer** (3/3):

1. **High — the byte budget did not bound the cache** (a single warming entry could grow to 5–16 GB on the box that also runs the DAQ stack; oversize stacks were admitted whole; doctrine-sized warm pairs could thrash-evict each other): per-entry cap = budget/3 enforced in `store_native_shot` (refuses at cap) + the warm loop (stops at cap) + `stack_frames` (size-gates from the dataset SHAPE before any bulk read; oversize → served per shot from disk, uncached). **Fixed/confirmed** — the bound is now real, thrash closed by cap + throttle.
2. **Medium — stop-doc-vs-finalization race could cache a truncated stack forever** (Tiled says complete seconds before the daemon stamps the stack): admission now requires the format contract's `finalized=True` (writer.py:139 — genuinely immutable once stamped); un-finalized stacks serve per shot from disk. **Fixed/confirmed**, pinned by a test reproducing the exact race window.
3. **Low-med — unthrottled warms + per-page hole re-probing**: max 2 warm threads (throttled keys reschedulable), once-per-key-per-process memo, and the threaded path now has real coverage. **Fixed/confirmed.**

Residuals accepted as non-blocking (reviewer concurrence): a near-cap warm pass reads-and-discards once; permanently inadmissible stacks pay one extra open per request for the admission re-check. Refuted along the way: protocol exactness of the wrapper, guard ordering (out-of-range 404 runs before the cache fast path), shared-DataFrame mutation (all consumers read-only), closure capture, lock coverage/TOCTOU, running→completed TTL transition. Placement concurrence: portal-local now; `CachingScanCatalog` is a cheap future promotion to data-utils if the console wants it. Suites at head: portal 76 passed, all others green.

PR #718, opened 2026-08-30
Planning: analysis-tabs backend decomposition (03) + inventory rulings + data-utils CLAUDE.md truth-up (0.20.2)

skbarber:
## What

The scouting/design step the owner green-lit: `Planning/data_portal/03_analysis_tabs_design.md` — every surface in the accepted mockup (artifact v4) mapped to its backend primitive with a four-status classification (**exists / extend / improve / new** — 'improve' added per owner: concept right, implementation predates current standards), grounded in a code-level recon of the actual implementations rather than the docs.

Per-concern breakdown:

- **03 doc (~230 LOC)**: eight recon findings that shape the plan (headline: `binned_scalars` has near-zero consumers — its only users are one docs notebook + one integration test, and the CLAUDE.md claims about ScanAnalysis renderers and a `sd.bin()` API were false — so the pure `bin_frame` rewrite is free; also: the s-file is read in 3 independent places, `from_date`'s defaults cost N_devices+1 SMB listings, the ImageAnalysis write-gate is a real-but-undocumented convention with one unconditional writer (HASO), `render_function` can't cross JSON, the event↔s-file bridge half-exists in `tiled_export`); the full decomposition tables per tab; the JSON endpoint sketch; the wave plan as 6 PRs with the wave-2-must-be-cheaper architecture checkpoint; open questions.
- **02 inventory (+~45 LOC)**: the post-mockup rulings transcribed as a superseding section (merged Plot tab, popup pattern, Images tab promotion, landing scaffold, in-rail steppers, raster traces, ephemeral processing + `computed` provenance, top-N eviction, statelessness).
- **GEECS-Data-Utils CLAUDE.md truth-up (0.20.1 → 0.20.2, docs-only)**: the Binning System section now shows the real stateful API, names the actual consumers, and points at the planned rewrite — the false `sd.bin(config)` example and the false ScanAnalysis-renderers claim are gone.

## Tests

Docs/planning only + a CLAUDE.md edit; `scripts/check.sh`: **GEECS-Data-Utils 231 passed, 3 skipped**, lint clean.

**Agent merges after review** (feature-branch target).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #719, opened 2026-08-30
W1a: read_sfile + the union-with-provenance scan_frame (data-utils 0.21.0)

skbarber:
## What

**Wave W1a of the analysis-tabs arc** (per 03_analysis_tabs_design.md): the substrate every tab consumes.

- `geecs_data_utils/data/sfile.py` (~90 LOC): **the** s-file reader + path convention in one home — `read_sfile(path)` and `sfile_path_for_scan(scan_folder)` (`{day}/analysis/s{N}.txt`, unpadded). `ScanData.load_scalars` now delegates (pinned by a spy test); the two ScanAnalysis duplicate reads converge in a trailing PR per the design doc.
- `geecs_data_utils/scan_frame.py` (~190 LOC): `scan_frame(detail, scan_folder) -> ProvenancedFrame` — the union-with-provenance frame. Event columns ∪ s-file columns, outer-joined on shot identity (**`Shotnumber == scan_event_index`, both 1-based** — the join contract the #718 review pinned). Per-column provenance dict (`run`/`sfile`; `computed` reserved), no name reconciliation (exact collisions suffix the s-file column), either provider optional, corrupt s-files degrade to run-only, strictly read-only.
- Canonical registry: `undulator_bluesky_1d` (2026-08-29 Scan 1 — the both-providers fixture) + integration tests.

## Tests

- Hermetic: 12 new (`test_scan_frame.py`) — join alignment by value pairing, outer-join one-sided shots, collision suffixing, run-only/sfile-only/empty, corrupt-file degradation, tree-untouched pin, the load_scalars delegation spy. Data-utils suite: **243 passed, 3 skipped**; root-tests 2 passed; lint clean.
- **Live (owner has lab access + mounted share)**: the existing canonical-scan integration baseline ran green before changes (13 passed, 93 s); the new integration tests pass against real data — `read_sfile` on the 2025-02-20 canonical s-file, and the full `TiledScanCatalog → resolve_scan_folder → scan_frame` union on the live catalog + share with the 1-based join asserted row-by-row. Post-change regression: 6 passed, 1 skipped (the skip = the union test under the root env, which lacks the tiled extra — `importorskip`; it passes from a tiled-capable env).

## Hardware / live

Covered above — integration verified on the mounted share + live Tiled, 2026-08-30.


skbarber:
**Adversarial review** (fresh-context subagent, three-lens brief per /land) — five findings, all **fixed (118cdb26 + f0efc862) and confirmed by the same reviewer** (5/5; the reviewer re-ran its original failure probes against head):

1. **Medium — a parses-but-corrupt s-file shot key raised out of `scan_frame`** (stray text cell → ValueError, non-integral float → TypeError), contradicting the degrade-to-run-only contract: key extraction now goes through `pd.to_numeric(errors="raise")` inside a guarded block that degrades with a warning; both repro cases pinned. **Fixed/confirmed.**
2. **Low — duplicate Shotnumber silently multiplied rows** in the outer merge: keep-FIRST with a counted warning (the shared join doctrine), pinned. Reviewer note accepted and made explicit in-code (2b6a532c): NA keys collapse too — blank-Shotnumber rows aren't shots. **Fixed/confirmed.**
3. **Low — the delegation silently removed a mkdir-on-read side effect** (the old path's `get_analysis_folder()` creates `analysis/ScanNNN`): CHANGELOG behavior note added with the owner's corrected framing — analysis-folder creation is permitted (only `scans/ScanNNN` is the hard invariant; the create-on-call getter serves ScanAnalysis's writers), the fix is that *reads* no longer write as a side effect — plus a tree-untouched pin on the delegated path. The reviewer's out-of-scope scans_database finding (has_analysis always True) is re-filed as its own task. **Fixed/confirmed.**
4. **Nit — the Shotnumber-collision join path was untested**: decoy-event-column test added with real mutation strength (keying on the decoy fails the value assertion). **Fixed/confirmed.**
5. **Nits — exports/hints/docstring**: `data/__init__` exports the sfile API, precise `dict[str, str]`/`list[str]` hints, canonical-scan docstring bullet. **Fixed/confirmed.**

Refuted by the reviewer along the way: placement (data/ + top-level scan_frame per the 03 spec), redundancy (the four s-file sites tell exactly the diff's story; DatasetBuilder is row-concat, not column-union), the 1-based join contract (re-verified against the plans and the exporter), the load_scalars path equivalence incl. Scan1000+, and the integration test's catalog assumptions. Suites at head: data-utils 246 passed/3 skipped, live integration 2 passed on the mounted share + lab Tiled. No undispositioned findings remain.

PR #720, opened 2026-08-30
W1b: OR-of-AND row filters (data-utils 0.22.0)

skbarber:
## What

**Wave W1b of the analysis-tabs arc** (per 03_analysis_tabs_design.md): GEECSplotter's "outer indexes OR, inner AND" filter model as shared vocabulary.

- `geecs_data_utils/data/row_filters.py` (~170 LOC): `RowFilters` → named `FilterGroup`s (enable flags; empty groups ignored) of `FilterCondition`s (`within`/`outside` inclusive bounds, validation-refused inverted bounds). `filter_mask(frame, filters)` is the composable primitive (`.sum()` = the rail's live pass count); `apply_filters` the frame view. Comparison kernel shared with `cleaning._OPERATORS` per the design doc (the proven kernel stays). **Explicit `nan_policy`** replaces the legacy op-dependent behavior for this path: `exclude` fails the condition under BOTH modes (the `outside` complement must not silently pass NaN rows), `keep` passes. No active groups = identity, never an empty result. Pydantic end to end — the same object is the endpoint query, the saved analysis config, and the "show the code" snippet.
- Exported from `data/__init__`.

## Tests

13 new hermetic tests: inclusive bounds, outside-complement-minus-NaN, both NaN policies, dtype-tolerant string coercion, unknown-column clear error, inverted-bounds validation, AND-within-group, OR-across-groups, disabled/empty-group ignoring, no-active-groups identity, mask/apply consistency, JSON round-trip, extra-fields-forbidden. Data-utils suite: **259 passed, 3 skipped**; root-tests 2 passed; lint clean.

## Hardware / live

None owed — pure frame algebra, hermetic by design (the live path is exercised end-to-end when W1d's endpoints compose this over W1a's union frame).


skbarber:
**Adversarial review** (fresh-context subagent, three-lens brief per /land) — five findings, all **fixed in 4710e366 and confirmed by the same reviewer** (5/5, original failure scenarios re-run against the fixed module):

1. **Medium — datetime/timedelta columns silently coerced to ~1e18 ns-ints** and produced wrong-but-plausible masks (`within` → nothing, `outside`+`exclude` → everything): refused loudly per the package's own `numeric_series` doctrine, pinned. **Fixed/confirmed** (loud refusal preferred over return-None here — the client named the column explicitly; object-dtype Timestamps checked safe: they coerce to NaN → all-fail, not the ns-int trap).
2. **Low-med — duplicated column labels escaped as an opaque pandas TypeError**: refused with our own column-naming ValueError, guard ordered before coercion. **Fixed/confirmed.**
3. **Low — NaN bounds validated and silently matched nothing** (pydantic parses JSON NaN; `low > high` never trips on NaN): `math.isnan` rejection in the validator, both constructor and JSON paths pinned; ±inf deliberately stays legal (half-open ranges verified both directions). **Fixed/confirmed.**
4. **Low — the "shared kernel" claim was inflated and the 03 doc's lowering sketch was silently abandoned**: `_OPERATORS` import removed (comparisons inline), the deviation owned in the module docstring (with the DatasetBuilder-composition future-work note), CHANGELOG rewritten, both 03-doc lowering claims amended with the W1b outcome. **Fixed/confirmed.**
5. **Info — GEECSplotter's per-condition enable vs this model's per-group enable**: porting-fidelity note recorded in the docstring (a saved LV filter set's disabled rows drop rather than carry state). **Confirmed.**

Refuted by the reviewer along the way (recorded so they aren't re-hunted): nullable-dtype Kleene logic comes out right under both NaN policies, index alignment survives chained filtering and duplicate index labels, `outside`+`exclude` correctly fails NaN, empty frames fine, and the test suite has real mutation strength (`|=`→`&=`, NaN-guard drop, inclusive-bounds mutants all killed). Suites at head: 16/16 row-filter tests, data-utils 262 passed/3 skipped (265 in tiled-capable envs). No undispositioned findings remain.

PR #721, opened 2026-08-30
W1c: pure bin_frame rewrite (data-utils 0.23.0)

skbarber:
## What

**Wave W1c of the analysis-tabs arc** — the flagship 'improve': the binned-plot core rewritten as a pure function, license granted by the recon finding that the old 145-line stateful property had **no production consumers** (one docs notebook + one integration test).

- `geecs_data_utils/data/binning.py` (~300 LOC): `bin_frame(frame, cfg) -> BinnedFrame` — same `BinningConfig` vocabulary (moved here; re-exported from `scan_data` so existing imports keep working), same `(col, {center, err_low, err_high})` schema, minus the warts: no `id(df)` cache key, label-aligned error assignment, **vectorized mad** (the per-group Python apply is gone), clean quantile extraction (the 50-line defensive unstack ladder is gone), one-shot output assembly (**the pandas PerformanceWarning is gone — pinned live with `-W error` against the canonical scan**), counts as a separate series (no `("count","center")` special-case in the core).
- **Deliberate semantic fixes** (each flagged): default `value_cols` excludes `Shotnumber` (as the config docstring always promised) while keeping the bin source column (the natural X axis); the `dropna` policy excludes the bin key (a grouping key is not a measurement — including it made `dropna="all"` vacuous) and the all-NaN-column guard now covers both policies.
- **`ScanData.bin(config)` exists** — the documented API, finally real; `binned_scalars` stays as the compat wrapper (re-attaches the legacy count column; cache keyed on df-version + config fingerprint, `id()` dropped).
- Docs notebook `basic_usage.ipynb` refreshed onto `sd.bin(...)` (edited cells' outputs cleared per notebook hygiene).

## Tests

**22 new hand-computed unit tests** (`test_binning.py`) — every err mode (median/mean centers, iqr/percentile asymmetric offsets + zero-clip, std/stderr, mad ± sigma scaling), min_count, both dropna policies incl. the all-NaN guard, numeric binning (width/edges/qcut, all four label modes, non-numeric identity fallback), empty input, missing bin col, and the ScanData compat + `sd.bin()` equivalence. These are the numbers the LabVIEW-source comparison will later verify (03 doc pin #1). Data-utils suite: **284 passed, 3 skipped**; root-tests 2 passed; lint clean.

## Hardware / live

Canonical-scan integration regression (`test_binning` on the real 2025-02-20 scan via the owner's mounted share) passes — **with PerformanceWarning promoted to error**, proving the fragmentation fix on real data.


skbarber:
## Adversarial review disposition (fix commit 0aa3028e)

Fresh-context reviewer ran an empirical old-vs-new harness (~45 scenarios: every err mode × agg, ddof variants, NaN placements under both dropna policies, numeric binning × all label modes, min_count, out-of-range values). **Numeric parity with the legacy implementation is bit-exact** everywhere except the documented deliberate deltas. Findings + disposition:

1. **HIGH — `expected_paths_by_bin` broken** (dangling call to the deleted `_compute_bin_key`; zero coverage hid it) → **fixed**: delegates to `compute_bin_key(df, self._bin_cfg)` + regression test. Retroactively justifies the `compute_bin_key` export.
2. **MEDIUM — stale bin cache on direct reassignment** (dropping `id(df)` from the key meant `sd.data_frame = df` served the previous frame's bins) → **fixed**: `data_frame` is now a property whose setter invalidates the cache (tolerant of the suite's `__new__` bare-instance pattern) + regression test.
3. **MEDIUM-LOW — empty results lost the column MultiIndex** → **fixed**: both empty paths build a proper 2-level schema; pinned at `bin_frame` and wrapper level.
4. **LOW — NaN-bin-row delta under-documented** → **fixed**: CHANGELOG symptom line, docstring Notes, pinning test.
5. **LOW — data-utils CLAUDE.md binning section stale** ("no sd.bin() yet") → **fixed**: rewritten around `bin_frame`/`BinnedFrame`/`sd.bin`; 03 spec row amended (bin col deliberately stays in default value_cols).

Nits (dead re-import, `BinnedFrame.counts` required, test cruft) all taken. Same reviewer re-verified every fix live: **all confirmed, no new defects**; suite 289 passed.

PR #723, opened 2026-08-30
W1e: Plot-tab polish from live feedback (portal 0.7.0, data-utils 0.24.0)

skbarber:
Owner feedback from first live use of the Plot tab (2026-08-30), item by item. Targets `feat/analysis-tabs`.

## What changed

| Feedback | Fix | Where |
|---|---|---|
| "can't set axes numerically, no log, no label/color editing" | **Display popup** (log X/Y, numeric axis ranges, marker size, per-trace colors — URL-carried) + Plotly `editable: true` (axis titles & legend names click-to-edit in place) | `run.html` |
| "columns all ts_ / telemetry_ prefixed" | Picker shows **pretty names first** (`telemetry_` stripped via the shared `display_name`, raw on hover); **`ts_` columns hidden behind an off-by-default toggle** — they're Tiled's per-key event-recording times, and they were 354 of the 668 columns on the live scan | `run.html`, `/api/columns` `timestamp` flag |
| "ts stamp cols should show sensible date/time in Pacific, LabVIEW protocol accounted" | Plotted timestamp columns arrive as **host-local ISO datetimes** on a Plotly date axis. New schema helpers pin the two-epoch convention: `ts_*` = Unix event times; any `acq_timestamp` spelling (event companions AND s-file headers) = LabVIEW wire epoch | `tiled_schema.is_key_timestamp_column` / `timestamp_epoch` (data-utils 0.24.0), `analysis.jsonable_datetimes`, `frame` `kinds` map |
| "day switch shouldn't go back to the main page; scan numbers should be a dropdown" | Rail gains a **scan dropdown** (the day's runs: number · mode · status) and the day steppers go through **`/run/jump/{day}`** — same scan number on the target day, else its newest run, whole analysis state carried; only an empty day falls back to the day page | `app.py`, `run.html` |
| "column math (×1000 etc.)" | **Deliberately deferred** to its own PR: that's the `computed` provenance the union frame reserves, and the expression vocabulary must stay notebook-identical — flagged for design, not bolted on here | — |

Plus one real bug found while testing: completed-run `/api` responses cache `immutable`, but their **shape changes across portal releases** — a browser upgrade would keep serving pre-upgrade payloads for a year. Fixed by keying every `/api` fetch with the installed portal version (immutable within a release, busts on upgrade).

## Live verification (lab Tiled)

ts-hide: 668 → 314 pool by default; a `ts_` X column renders `2026-08-29 12:37:17.246` local (matches the scan's 12:37 start); display popup log-Y + size 9 + custom color all applied and round-trip through the URL; `/run/jump/2026-08-28?prefer=1&y=…&view=bin` → 307 to the right run with state carried; scan dropdown lists the day.

Suites: portal 110 passed, data-utils 291 passed. Agent merges after review (feature-branch topology).


skbarber:
## Adversarial review disposition (fix commits d239b3b5, 0735cb4c)

Fresh-context reviewer, empirical probes on both packages. Findings + disposition:

1. **Medium — the new scan dropdown froze its URLs at render time** (picking columns/filters after load, then using the dropdown, silently dropped them — while the caption right under it promised persistence) **and `display` wasn't threaded through the server sticky query** → **fixed**: the dropdown navigates from the live `location.search` (state loss structurally impossible), and `display` rides `run_view`'s sticky state like the rest; pinned.
2. **Low-med — reproducibility**: the code snippet handed back raw seconds for a column the view showed as datetimes → **fixed**: the snippet emits the exact conversion (offset import only when a LabVIEW column is present); verified by executing it verbatim against the live catalog — matches to the millisecond. The binned view's deliberate raw-seconds behavior is now documented (CHANGELOG + CLAUDE.md).
3. **Low-med — CLAUDE.md stale** (missing `/run/jump`, `display`, the `timestamp`/`kinds`/`v`-key conventions) → **fixed**; reviewer re-audited every documented route against the decorators.
4. **Low — the epoch-name heuristic's false-positive mangle mode undocumented** → **fixed**: collision warning in the `timestamp_epoch` docstring (no such column exists in schema v1).
5. **Nit — `timestamp` flag vs `kinds` semantics** → clarifying comment landed.
6. **Nits** (extra-axis tick colors, log/range on datetime Y axes, color index shift on trace removal) → all fixed; the confirm pass's one residual (a hand-crafted URL with non-array `colors` could throw in `removeY`) → `Array.isArray` guard (0735cb4c).

Refuted by the reviewer: route shadowing, open redirect, param carry-through (repeated y, prefer semantics), `v`-key tolerance on all routes, jsonable_datetimes robustness, test honesty (`cam acq_timestamp` IS UI-reachable via plottable_columns), JS layout-mutation/stale-id concerns.

Final same-reviewer verdicts: **all findings CONFIRMED FIXED**. Suites: portal 112, data-utils 291.

PR #725, opened 2026-08-31
Multi-Y axis rendering fixed the library's way (portal 0.8.1)

skbarber:
Owner feedback: overlapping tick numbers on multi-column plots, grey always-there axis labels — "each little custom thing slightly off makes the UX look forced/cheap." Both symptoms traced to the same root: hand-rolled layout where Plotly has native mechanisms. Targets `feat/analysis-tabs`.

## Root causes → fixes

| Symptom | Root cause | Fix |
|---|---|---|
| Tick numbers of axes 3–4 rendered on top of axes 1–2 | Hand-set `position: 0.0/1.0` put free axes at the plot edges where the anchored axes already sit | Plotly's **`autoshift`** (built for exactly this) stacks them outward; `automargin` everywhere |
| Grey "Click to enter Y axis title" text always visible, overlapping when multi | Blanket `editable: true` renders placeholders for every unset title | **Granular `edits`** (legend names / annotations / shapes stay click-to-edit); axis titles are real, color-matched to their trace |
| — | Plotly does **not** shift a free-anchored axis's rotated title with the axis (measured live; `standoff` doesn't help) | Titles on the two anchored axes only; axes 3–4 identified by colored ticks + the legend (the standard dashboard pattern) |
| Redundant single-trace legend | — | Hidden when one column: the colored axis title names it |

## Verification (live, real scan, structural)

Bounding-box intersection over all tick blocks + titles with **4 columns picked: zero collisions, zero placeholders**; single column: colored title, no legend, no placeholders. Suite: 113 passed.

Note: this is deliberately a Plotly-native fix regardless of the pending renderer conversation (Altair/Vega-Lite vs plotly.py spec-authoring) — the lesson either way is lean on the library's mechanisms, not hand-positioning.

Agent merges after review.

### DIFF
--- a/GEECS-DataPortal/CHANGELOG.md
+++ b/GEECS-DataPortal/CHANGELOG.md
@@ -3,6 +3,25 @@
 All notable changes to this package will be documented here.
 Format follows [Keep a Changelog](https://keepachangelog.com/en/1.1.0/).
 
+## [0.8.1] - 2026-08-31
+
+### Fixed
+
+Multi-Y axis rendering (owner feedback: overlapping tick numbers,
+grey always-there labels):
+
+- Axes 3–4 stack outward via Plotly's native `autoshift` (the previous
+  hand-set `position` put them on top of axes 1–2's ticks).
+- Real color-matched axis titles on the two anchored axes; axes 3–4
+  rely on colored ticks + the legend (Plotly does not shift a
+  free-anchored axis's title with the axis — measured).
+- The grey "Click to enter …" placeholders are gone: in-place editing
+  is now granular (legend names, annotations, shapes) instead of
+  blanket `editable: true` — axis titles are auto-set, or settable via
+  the advanced layout box.
+- Single-trace plots hide the redundant legend (the colored axis title
+  names the trace); `automargin` on all axes.
+
 ## [0.8.0] - 2026-08-30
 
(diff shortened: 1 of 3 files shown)


skbarber:
## Adversarial review disposition (fix commit 1b8daee5)

Fresh-context reviewer verified the mechanics before hunting nits: **`autoshift` is genuinely implemented in the vendored 3.1.1 bundle** (attribute declaration + doAutoMargin integration + the repositioning pass located by grep — no silent no-op risk), call order safe in both draw paths, `showlegend`/titles consistent between shot and binned views, the display-layout passthrough still overrides titles as the intended escape hatch, and all n=1..4 / early-return edges hold. **No functional findings.**

Three doc-level nits, all taken (pure wording/symmetry — verifiable in the diff, no behavior change, suite 113):
1. The PLOT_CONFIG comment still advertised in-place *title* editing (this PR removed it) → corrected.
2. The popup hint's example set the plot title while the sentence implied axis titles → example is now `{"yaxis":{"title":{"text":"…"}}}`.
3. CHANGELOG said "automargin on all axes" but the binned view's x-axis lacked it → added for symmetry.

Noted-not-a-finding accepted as-is: dropping blanket `editable` also drops annotation-drag, which no UI surface ever advertised.

PR #726, opened 2026-08-31
GEECS-DataPortal 0.9.0: reverse-proxy mountability (OSPREY panel-tab feedback)

skbarber:
## What + why

OSPREY panel-tab integration feedback: the portal broke behind a reverse proxy because every navigation link, form action, image/script URL, JS `/api` fetch, and redirect was root-absolute. This makes the portal **prefix-agnostic** — it works at root and under any mount point (`proxy /portal → :8200`), the same way Grafana/JupyterHub solved it.

- **`_ForwardedPrefixMiddleware`** adopts the proxy's `X-Forwarded-Prefix` header as the ASGI `root_path`, per request. Values are validated (`_clean_prefix`: root-absolute, no `//`/whitespace/backslash/control chars; bare `/` = root) — malformed headers are ignored, never propagated into page links. The header, when present, wins over a static setting (the proxy is authoritative for where it mounted us; a client faking it only rewrites its own page's links).
- **One `root` context value** (a Jinja context processor) prefixes every template href/action/src; the page JS gets `const ROOT` and builds all `/api` fetches and the scan-jump dropdown from it. The `/`, `/go`, and `/run/jump` redirects prefix via `_root(request)`.
- **`--root-path` CLI fallback** → uvicorn `root_path`, for proxies that can't send the header.
- **`/health` already existed** (0.1.0) and is untouched — it's the panel health probe for `web.panels.dataview.health_endpoint`. It returns 200 always; the JSON `ok` field carries the catalog probe, so a down Tiled reads as degraded catalog, not dead portal.
- Docs: DEPLOYMENT.md gains a reverse-proxy section (nginx example, strip-prefix + header convention); CLAUDE.md pins the prefix-agnostic rule so future templates keep the `{{ root }}` discipline.

Design note (why prefix-injection rather than the feedback's literal `./relative` links + `<base href>`): relative links are depth-sensitive (`/run/{uid}` vs `/run/{uid}/image.png` resolve `./` differently) and `<base>` hijacks `href="#"` anchors the page uses for popups. One explicit per-request prefix covers every URL uniformly and degrades to exactly the current behavior at root.

## LOC breakdown (one concern; 244 insertions / 28 deletions)

| Piece | ~LOC |
|---|---|
| app.py: middleware + `_clean_prefix` + `_root` + context processor + 3 prefixed redirects | ~75 |
| templates: `{{ root }}` on 16 URL sites + `const ROOT` in the JS | ~30 changed |
| `__main__.py`: `--root-path` | ~15 |
| tests: `TestReverseProxy` (9 tests) | ~85 |
| CHANGELOG + CLAUDE.md rule + DEPLOYMENT.md section | ~65 |

## Tests

`./scripts/check.sh` green: **GEECS-DataPortal 122 passed** (113 existing + 9 new: prefixed links/forms/static/fetch-base/steppers, prefixed redirects incl. `/run/jump`, root-unchanged, trailing-slash normalization, malformed-header rejection, `/api` + `/health` under prefix), **GEECS-Data-Utils 294 passed**, root-tests 2 passed, lint/pre-commit clean.

## Hardware verification

No scan-execution or device surface. **OWED (integration, not hardware):** load the portal through an actual OSPREY panel / reverse-proxy mount and click through day → run → tabs → images; expected: all navigation and plots work identically under the prefix, and the health LED goes green once `web.panels.dataview.health_endpoint` points at `/health`.

### DIFF
--- a/GEECS-DataPortal/CHANGELOG.md
+++ b/GEECS-DataPortal/CHANGELOG.md
@@ -3,6 +3,24 @@
 All notable changes to this package will be documented here.
 Format follows [Keep a Changelog](https://keepachangelog.com/en/1.1.0/).
 
+## [0.9.0] - 2026-08-31
+
+### Added
+
+Reverse-proxy mountability (OSPREY panel-tab feedback): the portal now
+works at root **and** under any URL prefix — `proxy /portal → :8200`.
+
+- Every template href/form/img/script URL and the page JS's `/api`
+  fetch base (`const ROOT`) build through one per-request `root`
+  prefix; the `/`, `/go`, and `/run/jump` redirects carry it too.
+- The prefix auto-derives from the proxy's `X-Forwarded-Prefix` header
+  (validated: root-absolute, no `//`/whitespace/backslash — malformed
+  values are ignored, a bare `/` means root); a static fallback is
+  available as `geecs-data-portal --root-path /portal`. The header,
+  when present, wins.
+- `/health` (present since 0.1.0) is the panel health probe — wire
+  OSPREY's `web.panels.dataview.health_endpoint` at it.
+
 ## [0.8.1] - 2026-08-31
 
 ### Fixed
--- a/GEECS-DataPortal/CHANGELOG.md
+++ b/GEECS-DataPortal/CHANGELOG.md
@@ -14,10 +14,13 @@ works at root **and** under any URL prefix — `proxy /portal → :8200`.
   fetch base (`const ROOT`) build through one per-request `root`
   prefix; the `/`, `/go`, and `/run/jump` redirects carry it too.
 - The prefix auto-derives from the proxy's `X-Forwarded-Prefix` header
-  (validated: root-absolute, no `//`/whitespace/backslash — malformed
-  values are ignored, a bare `/` means root); a static fallback is
-  available as `geecs-data-portal --root-path /portal`. The header,
-  when present, wins.
+  (validated against a strict path-segment pattern — malformed values
+  are ignored, a bare `/` means root). The middleware also re-prefixes
+  the request path to the ASGI-canonical shape, so mounts named like a
+  route head (`/run`, `/api`, …) route correctly and trailing-slash
+  redirects keep the prefix. A static fallback is available as
+  `geecs-data-portal --root-path /portal` (without those two
+  guarantees — see DEPLOYMENT.md); the header, when present, wins.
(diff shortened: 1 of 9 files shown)


skbarber:

Reviewed through the correctness / redundancy / placement lenses against the whole repo, with the suite run and starlette 1.6.0 / fastapi 0.141.1 behavior verified empirically. Checked-and-refuted (not findings): header spoofing (self-only — immutable-cached /api JSON and PNGs embed no URLs), missed URL sites (every producer in the templates and app.py enumerated; page JS is `location`/`ROOT`-consistent), redundancy (no existing forwarded-prefix handling anywhere in the repo; uvicorn's ProxyHeadersMiddleware covers only For/Proto), placement (portal-local middleware is correct — no other package renders links).

### Findings + dispositions

1. **MEDIUM — mount prefix colliding with a route head double-strips → 404s that route family** (starlette strips `root_path` from the front of `path` wherever it happens to match; a `/run` mount broke every run page). **FIXED** in ea6a3fba: the middleware re-prefixes `scope["path"]` to the ASGI-canonical shape, making the strip exact. Pinned by `test_mount_name_colliding_with_a_route_head_works`; reviewer probed `/run`, `/api`, `/day`, `/health`, `/static`, `/go`, and a multi-segment mount — **confirmed fixed**.
2. **LOW-MEDIUM — starlette's trailing-slash redirect dropped the prefix** (Location built from the un-prefixed path). **FIXED** in ea6a3fba by the same re-prefix, pinned by `test_trailing_slash_redirect_keeps_the_prefix`; the `--root-path`-only fallback keeps both limitations, now documented as explicit caveats in DEPLOYMENT.md instead of an implicit full-coverage claim — **confirmed fixed** (residual documented, not hidden).
3. **LOW — `_clean_prefix` was looser than its stated contract** (`?`/`#`/quote characters passed). **FIXED** in ea6a3fba: strict whitelist regex + four new malformed-prefix test cases — **confirmed fixed**.

Reviewer's one constructed trade-off from the fix: a proxy that does NOT strip the prefix but still sends the header now 404s loudly instead of half-working — deliberate (any "already-prefixed" guard would reintroduce finding 1's ambiguity), matches the documented strip-style convention, and 349b0565 adds the reviewer's suggested runbook sentence stating it.

**Confirm/deny pass:** all three findings verified against the actual commit by the same reviewer — all CONFIRMED FIXED, no new findings. Suite after fixes: **124 passed**.

Issue #727, opened 2026-09-01
OSPREY integration: publish the ScanRequest schema (form-ready catalog) and a stable live document-stream contract


skbarber:
## Context

OSPREY now fronts the GEECS queueserver directly: its bluesky bridge runs in external-worker mode (als-apg/osprey#816, implemented in als-apg/osprey#817) as a *client* of the GeecsBluesky RE Manager — one RunEngine, one queue, GEECS-Console and OSPREY as peers. The plan catalog OSPREY's panel offers is now downloaded from the manager's `plans_allowed`, so it correctly shows `geecs_scan_request_plan` / `geecs_run_action_plan` with the docstring-derived parameter descriptions.

What's missing is *presentation richness* on the generic-client side: the panel can't render a real ScanRequest form, and live scan progress isn't streamed. The design position (discussed with @sbarber): **keep the ScanRequest funnel — it's the right facility front door — and publish its contract in the lingua franca** (JSON Schema + queueserver's own annotation machinery), rather than flattening it into many "standard" plans. Three items, in priority order:

## 1. Publish the ScanRequest schema + annotate the funnel plans

- Annotate `geecs_scan_request_plan` and `geecs_run_action_plan` with bluesky-queueserver's `parameter_annotation` decorator so `plans_allowed` carries typed parameters (this is upstream-supported metadata, no behavior change).
- Export `geecs_schemas.ScanRequest.model_json_schema()` as a generated artifact — a small script (~20 lines) emitting e.g. `scan_request.schema.json`, regenerated whenever `geecs_schemas` changes (same discipline as the channel-DB generators). Delivery options, either or both:
  - land it in the OSPREY deployment repo under `data/` (build-time artifact the bridge grafts onto the `request` parameter), and/or
  - serve it from GEECS-MCP (a `get_scan_request_schema` read tool), so any client can fetch it live.

Payoff: any JSON-Schema-aware client — OSPREY's panel, the agent's approval prompts, future web UIs — gets a full nested ScanRequest form, **optimization sub-model included**, for free.

## 2. (Cross-repo, tracked osprey-side) Graft the schema into OSPREY's external plan catalog

[...]

PR #730, opened 2026-09-01
Publish ScanRequest JSON Schema + annotate the funnel plans (#727 item 1)

skbarber:
Item 1 of #727: publish the ScanRequest contract in the lingua franca and annotate the two funnel plans so `plans_allowed` carries typed, described parameters — the GEECS side of the contract OSPREY's external plan catalog grafts onto (`als-apg/osprey#816` follow-ons consume it).

## Per-concern breakdown

**Concern 1 — GEECS-Schemas 0.13.0: the published JSON Schema artifact (~130 LOC module + tests, +570-line generated artifact)**

- `geecs_schemas/schema_export.py` — `ScanRequest.model_json_schema()` → committed artifact, mirroring `docgen`'s exact discipline (module constants, `render_*`/`write_*`, `python -m` CLI, regenerator script `tests/generate_scan_request_schema.py`, no-drift CI guard `tests/test_schema_export.py`).
- The artifact lands at `docs/geecs_schemas/scan_request.schema.json` — under `docs/` deliberately, so mkdocs publishes it as a static file and clients outside the repo can fetch it by URL (the "serve it live" delivery option; the OSPREY deployment repo can also vendor it build-time). Full nested vocabulary, **optimization sub-model included** (21 KB, 12 `$defs`) — pinned by `test_schema_carries_the_full_nested_vocabulary`.

**Concern 2 — GeecsBluesky 0.69.0: queueserver parameter annotations (~80 LOC + tests)**

- Annotation payloads (`SCAN_REQUEST_PLAN_ANNOTATION` / `RUN_ACTION_PLAN_ANNOTATION`) live in `plans/scan_request_plan.py` **as plain dicts, next to the plans** — that module stays queueserver-import-free per its own doctrine, and a signature change plus its annotation is one edit.
- `qserver/startup/startup.py` applies `parameter_annotation_decorator` and rebinds the module-level names — that rebinding is what the manager's plan discovery serves as `plans_allowed`. The `request` parameter's description points clients at the published schema artifact. Metadata only — the wrappers stay generator functions and delegate verbatim; no behavior change.
- The startup probe (`tests/_qserver_startup_probe.py`) now pins the decorated wrappers (`__wrapped__` is the real plan + annotation attribute present); new test `test_annotated_plans_carry_descriptions_and_still_validate` pins that the processed plan carries per-parameter descriptions AND that a real `queue add` item still validates against the decorated signature (the bare-namespace annotation-string constraint the existing signature test documents).

- Artifact under `docs/` rather than inside `GEECS-Schemas/` — chosen for by-URL delivery via the published site. Moving it is a one-constant change (`SCHEMA_ARTIFACT`).
- No version stamp inside the schema JSON — keeping the artifact pure means unrelated `poetry version` bumps don't force regeneration; provenance rides git history + the CHANGELOG.

## Not in this PR (explicitly)

- The GEECS-MCP `get_scan_request_schema` read tool (issue item 1's second delivery option) — natural follow-up PR in GEECS-MCP.
- Issue items 2 (osprey-side graft) and 3 (document-stream contract note) — item 3 is a docs PR against `GeecsBluesky/qserver/deploy/DEPLOYMENT.md` + `EVENT_SCHEMA.md`.

## Tests

`scripts/check.sh` green: lint, **GEECS-Schemas 269 passed** (3 new), **GeecsBluesky 654 passed, 1 skipped** (integration, expected off-network; startup probe + new annotation test included).

## Hardware verification

No device or scan-execution surface — the plan wrappers are metadata-only (delegation pinned hermetically by the probe and the manager-validation test). **OWED (worker deploy, not this PR's gate):** after the next worker restart on the deployed host, `qserver console` / a client `plans_allowed` fetch should show the two plans carrying the parameter descriptions — a 30-second check bundled with the next qserver deploy.


skbarber:

Reviewer verified the core design rather than assuming it: `parameter_annotation_decorator` uses `functools.wraps` (signature + `__wrapped__` survive, generator-ness preserved, `_custom_parameter_annotation_` set — exactly what the probe pins); the manager loads the startup dir directly (no committed plan-list file), so annotations propagate on worker restart; mkdocs copies the artifact verbatim for by-URL delivery; no overlapping implementation exists in the repo. Four findings survived refutation:

**1. (medium) Two startup tests hard-fail — instead of skip — without the `qserver` extra**, since `startup.py` now imports `bluesky_queueserver` at top level. → **FIXED** in 6f0d2968: `pytest.importorskip("bluesky_queueserver")` on both startup-executing tests (the file's existing discipline). Verifier confirmed empirically: no-extra env now reports 5 skipped, 0 errors; CI coverage unchanged (the Ubuntu job installs the extra).

**2. (medium-low) The artifact declared no `$schema` dialect but uses 2020-12-only vocabulary** (`prefixItems`, `$ref` siblings) — a draft-07-default validator would silently accept malformed tuples. → **FIXED** in 6f0d2968: `"$schema": "https://json-schema.org/draft/2020-12/schema"` injected first in `scan_request_json_schema()`, artifact regenerated. Confirmed.

**3. (low) Drift guard skipped when the artifact is missing, and the artifact path was pinned in two packages but tested in neither.** → **FIXED** in 6f0d2968: the guard now asserts existence (skips only outside a repo checkout, keyed on `mkdocs.yml`), and the GeecsBluesky annotation test pins the full path imported from `geecs_schemas.schema_export.SCHEMA_ARTIFACT` — a docs reorg now breaks tests instead of leaving a dead pointer. Confirmed; the verifier's Windows-separator nit taken in 89f07213 (`as_posix()`).

**4. (low) `schema_export.py` copies docgen's scaffolding (`_repo_root`, write/CLI shape) — second copy of the pattern.** → **WAIVED**: the mirror is deliberate (~40 LOC, keeps the two generated-artifact pipelines shape-identical); standing rule adopted per the review — a **third** generated artifact forces extraction of the shared render/write/CLI/no-drift pattern. Verifier had no objection.

Fix confirmation was done by a fresh-context verifier given only the findings and the fix commit (the original reviewer had completed; independence preserved per the /land substitution clause). All findings dispositioned; CI green on the final head.

PR #732, opened 2026-09-01
portal 0.11.0: binned-vs-X placement, plot-size control, copy-to-clipboard

skbarber:
## What + why

Owner live-feedback round on 0.10.0 (three asks, tested live on `feat/plot-tools`):

1. **Binned view plots against the X pick** (~90 LOC server+figure+snippet): bins still *group* by `bincfg.bin_col`, but each bin *plots at* the per-bin **mean** of the selected X column — real scan-parameter positions instead of bin indices ("bins group, X places"). Implementation is the same primitive twice: a second `bin_frame` with `replace(cfg, value_cols=(x,), agg="mean")`, so the "show the code" snippet mirrors it verbatim. Payload gains `x_centers` (only when `x` given); no `x` keeps bin labels exactly as before. Unplottable `x` 404s (same rule as `/frame`). X error bars deliberately deferred per the owner ("wouldn't make that the first move").
2. **Plot size control** (~30 LOC): `width`/`height` join the display vocabulary — popup inputs, same type-400/value-degrade boundary rules, applied in the shared `_apply_display`. A fixed size fixes the exported image size too.
3. **Copy plot to clipboard** (~60 LOC, client-only): a modebar button exporting the figure at 2× and writing the PNG to the clipboard ("a lot of times we copy paste graphics"). Caveat handled honestly: the async Clipboard API needs a secure context (https/localhost) — on plain-http lab hosts the button degrades to the 2× PNG download with a visible note. Built-in camera download bumped to 2× as well.

## Tests

157 passed (was 152): binned-x endpoint pins (hand-computed mean centers, no-x unchanged, unplottable-x 404, snippet mirror), figure-level x_values/x_label pin, width/height application + 400-ladder additions.

## Hardware verification

Live-verified over VPN against lab Tiled (U_S1H current scan `112e50d7…`, 2026-08-21): binned view plots at the real per-bin currents `[0.94995, 0.97492, 0.99989, 1.02509, 1.05006]` A (was bin indices 1–5), axis titled `U_S1H Current`; the binned snippet **ran verbatim** in the venv and produced a figure element-wise identical to the served one (x/y/error arrays compared). Copy button + size inputs verified present and wired in the browser. OWED: nothing — no scan execution or device I/O.

### DIFF
--- a/GEECS-DataPortal/CHANGELOG.md
+++ b/GEECS-DataPortal/CHANGELOG.md
@@ -3,6 +3,30 @@
 All notable changes to this package will be documented here.
 Format follows [Keep a Changelog](https://keepachangelog.com/en/1.1.0/).
 
+## [0.11.0] - 2026-09-01
+
+### Added
+
+Owner live-feedback round on 0.10.0:
+
+- **Binned view plots against the X pick** — bins still *group* by
+  `bincfg.bin_col`, but each bin now *plots at* the per-bin **mean of
+  the selected X column** (`/binned?x=…` → `x_centers` in the payload,
+  the figure's x positions, and the axis title). Same primitive, same
+  bins: a second `bin_frame` call with `replace(cfg, value_cols=(x,),
+  agg="mean")`, mirrored verbatim in the "show the code" snippet. No
+  X keeps the bin labels as the axis, exactly as before. (X error
+  bars deliberately deferred — mean placement is the first move.)
+- **Plot size control**: `width`/`height` join the display vocabulary
+  (popup inputs; same type-400/value-degrade rules). A fixed size also
+  fixes the exported image size.
+- **Copy plot to clipboard**: a modebar button exports the figure at
+  2× and puts the PNG on the clipboard — copy-paste is how plots
+  travel around the lab. Caveat: the async Clipboard API requires a
+  secure context (https or localhost); on plain-http lab hosts the
+  button degrades to the 2× PNG download with a note. The built-in
+  camera download is 2× now too.
+
 ## [0.10.0] - 2026-08-31
 
 ### Changed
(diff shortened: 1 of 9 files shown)


skbarber:

Reviewer probe-verified the diff in the package venv (suite green), refuted redundancy (first toast in templates; no console-side bin_frame consumers; the two-call binned-x design judged correct vs folding x into value_cols) and placement (three-layer contract respected), and confirmed hygiene/version chores. Findings, all dispositioned (fixes in 87179231, each re-probed and confirmed by the same reviewer):

1. **HIGH — `x_centers` could silently misalign with `bins`**: the x-placement `bin_frame` call's dropna/min_count run over x alone, so its surviving bins can differ from the y call's; positional zipping then plots points at the *wrong bin's x* (reviewer demonstrated 4 concrete scenarios, including a real union-frame case). **FIXED**: `x_result…reindex(result.frame.index)` on both the endpoint and the snippet — a bin missing an x center degrades to a null x (skipped point), never a shifted one. Reviewer re-ran all four probes: aligned or skipped, never shifted. Pinned by `test_binned_x_reindexes_onto_the_y_bins` (fails on pre-fix code — measured).
2. **MEDIUM — coercible-string columns 500'd the binned view** (dtype-tolerant telemetry passes the plottability guard, then `bin_frame` aggregates the raw object dtype). **FIXED**: both `bin_frame` call sites catch `TypeError` → 400 with an honest message; the same pre-existing hole on the y side is closed in the same commit. Pinned (x and y variants). **Owner follow-up flagged**: coerce-then-bin (so a pick that plots per-shot also bins) is a primitive-semantics decision, deliberately not made in a fix commit.
3. **LOW — timestamp X serves raw epoch seconds in binned view**. **WAIVED as deliberate, documented**: CLAUDE.md's binned route line + CHANGELOG now state datetime rendering is per-shot-only. Reviewer: WAIVER-OK.
4. **NIT — export scale hardcoded ×3 + misleading Safari fallback wording**. **FIXED**: one `EXPORT_SCALE` const; note reworded to "copy failed — downloaded PNG instead" (honest for both no-secure-context and gesture-expiry).

Suite 159 passed. Reviewer: "No new findings introduced by the fix commit… Nothing outstanding."

PR #735, opened 2026-09-01
Data-Utils 0.25.0 + ImageAnalysis 1.13.0: average_frames + the ephemeral run seam (W2a primitives)

skbarber:
First PR of the **W2a Images-tab arc** (`feat/images-tab` integration branch): the primitives, per the portal's adding-a-tab checklist ("if the tab needs new math, that PR comes first"). Two small pieces, one per package:

## GEECS-Data-Utils 0.25.0 (~35 LOC + 40 test LOC)

`geecs_data_utils.io.average_frames(frames)` — THE shared `nanmean` frame averager for per-bin averaged images. NaN pixels excluded per-position; empty input or inhomogeneous shapes degrade to `None` with a warning (the numpy-2.x `np.asanyarray` hard-error case), mirroring `ImageAnalyzerResult.average`'s guard.

**Placement judgment call (cheap to veto):** the design doc allowed "ImageAnalysis (or data-utils io)". Data-utils wins because the portal's plain per-bin averaged grid then needs **no ImageAnalysis dependency** — ImageAnalysis becomes an optional portal extra only when the processing selector lands (PR 3 of this arc). The three divergent averaging copies (ScanAnalysis `average_data`'s `np.mean`, `HIMG_with_average_saving`'s hand-rolled mean, `ImageAnalyzerResult.average`) converge on it in trailing PRs, not here.

## ImageAnalysis 1.13.0 (~120 LOC + 150 test LOC)

`image_analysis.ephemeral.run_diagnostic_ephemeral(name_or_path, frames, *, config_dir, overrides, auxiliary_data)` — the write-free run seam: load a diagnostic, instantiate its analyzer once, `analyze_image` per in-memory frame. The no-writes contract is **structural**, not conventional:

- In-memory frames only — the runner never receives or forwards a file path, so every path-gated writer analyzer stays dormant.
- `auxiliary_data["file_path"]` is **refused** (`ValueError`), never silently stripped — it is both the write gate five analyzers key on and a *read* trigger in `Standard1DAnalyzer`, so a caller passing it is confused about the contract.
- `EPHEMERAL_DENYLIST` refuses unconditional writers **by class-path string before import**. HASO is the one entry (five sidecars per shot from instance state); the pre-import check doubles as keeping the wavekit SDK off non-Windows hosts.

Contract documented in `ImageAnalysis/CLAUDE.md` ("Ephemeral runs" section) — including the forward obligation that new writer analyzers either gate on `file_path` or join the denylist.

**Judgment-call addition (cheap to veto):** `image_analysis.config.list_diagnostics(config_dir=...)` — a 3-line public wrapper over the existing `_discover_analyzers`, for the PR-3 picker UI.

## Testing

- ImageAnalysis: **235 passed, 11 skipped** (9 new in `tests/test_ephemeral.py`: per-frame results, nothing-written tree snapshot, overrides reach the pipeline, `file_path` refusal, pre-import denylist refusal, empty input, `list_diagnostics`).
- GEECS-Data-Utils: **298 passed, 3 skipped** (7 new in `tests/test_io_images.py::TestAverageFrames`: nanmean semantics, int→float, empty/mixed-shape degrade with warning, 1D traces).
- No scan-execution or device-facing surface — no hardware verification owed. Nothing here creates scan folders (the seam's whole point is the opposite).


skbarber:

The reviewer verified the write-gate claims against every analyzer source, probed edge cases, searched for duplicates, and ran both suites. **5 findings + 3 nits, no blocker; all dispositioned as FIXED in 6cf47616 and confirmed by the same reviewer.**

| # | Finding | Disposition |
|---|---|---|
| 1 | **Correctness (moderate):** `if not frames:` crashes on an already-stacked ndarray (`np.stack(bin_frames)`) — "truth value of an array is ambiguous". | **Fixed**: `len(frames) == 0`; pinned by `test_stacked_ndarray_input` + `test_empty_ndarray_returns_none`. |
| 2 | **Contract wording (moderate):** HASO writes its five sidecars from `load_image` only — `analyze_image` writes nothing; the CLAUDE.md/CHANGELOG claim misstated the trigger. Denylist entry stays valid for the real reasons (hard wavekit import; analyze-only output meaningless). | **Fixed**: wording corrected in CLAUDE.md, `ephemeral.py` denylist comments, CHANGELOG. |
| 3 | **Contract precision (low-mod):** `GrenouilleAnalyzer.analyze_image` unconditionally writes transient temp files + spawns a ~seconds 32-bit DLL subprocess per frame — un-gated, not denylisted, contradicting the "nothing written anywhere" promise. | **Fixed by denylisting Grenouille** (author call, cheap to veto): keeps the promise strong instead of narrowing it; contract rules now name transient files + subprocess spawns; runner error generalized; membership pinned in tests. Reviewer endorses: only the ephemeral seam is blocked — the scan pipeline runs Grenouille normally, and hosts without `frog_dll_path` failed at construction anyway. |
| 4 | **Redundancy (low):** `average_frames` duplicates `_safe_nanmean_arrays` line-for-line; the cheapest convergence was skipped. | **Fixed in this PR**: `_safe_nanmean_arrays` now delegates to `average_frames` (new keyword-only `label` for warning attribution, pinned by test). Reviewer checked no test anywhere asserts on the old message/logger. Remaining two copies (ScanAnalysis `average_data`, HIMG) stay for trailing PRs, documented in both CHANGELOGs. |
| 5 | **Placement (low):** `io/images.py`'s docstring charter ("path → array, no analysis logic") contradicted the new pure-math resident. | **Fixed**: module docstring amended — names both in-memory exceptions with the keep-beside-the-readers rationale. |
| nits | Aux-forwarding test proved nothing about delivery; `list_diagnostics` "loadable" overclaim; shallow-copy docstring nuance. | **All fixed**: test now records what `analyze_image` receives (value + per-frame identity distinctness); docstrings corrected. |

Verified non-findings worth recording: no denylist bypass via re-export/alias paths; the pre-import-denylist test genuinely discriminates (a post-import check would produce a different error); scan-folder invariant untouched; every other writer analyzer confirmed genuinely `file_path`-gated (line_stitcher, magspec_manual_calib, downramp, density_from_phase), so the denylist is sufficient.

Suites after fixes: **ImageAnalysis 235 passed / 11 skipped · GEECS-Data-Utils 301 passed / 3 skipped.** Reviewer's re-verification: "Nothing broken, nothing missed… PR #735 at 6cf47616 is clean from my side."

PR #722, opened 2026-08-30
W1d: the Plot tab — vendored Plotly + the /api layer (portal 0.6.0)

skbarber:
Fourth PR of the analysis-tabs arc (03 design doc wave plan) — the arc's first interactive surface, composing everything W1a–c built. Targets `feat/analysis-tabs`.

## What this is

**The Plot tab, live.** The scan page becomes rail + tabs (Overview / Plot / Images, per the accepted mockup): a type-to-filter column picker over the **union frame** (provenance-badged `run`/`sfile` columns), up to 4 Y columns on per-series axes, a **per-shot ⇄ binned** toggle (asymmetric error bands via `bin_frame`), the **OR-of-AND filters popup** (named chips in the rail, live would-pass preview), the bin-settings ⚙ popup, and **"show the code"** — every plot carries the notebook snippet that reproduces it exactly.

## Per-concern LOC breakdown

| Concern | Where | ~LOC |
|---|---|---|
| Vendored Plotly 3.1.1 (cartesian, MIT, 1.4 MB) | `geecs_portal/static/` + a pre-commit large-file exclusion scoped to that dir | asset |
| `/api` boundary chores (filters/bincfg parsing → 400, NaN→null, code snippets) | `geecs_portal/analysis.py` (new) | 210 |
| The four endpoints (`columns`/`frame`/`binned`/`filter-count`) + union helper + stepper neighbours | `app.py` | +190 |
| The scan page (rail, tabs, picker, popups, Plotly JS, URL state) | `templates/run.html` | 520 |
| Tests: hand-computed endpoint numbers, the 400/404 ladder, NaN→null, union provenance + tree-untouched pin, stepper state | `test_app.py` / `test_resources.py` | +230 |
| Doctrine: vendored-asset amendment, three-layer contract, **adding-a-tab checklist** | `CLAUDE.md` | docs |

## Doctrine notes

- **Vendored Plotly** is the owner-approved amendment (2026-08-30) now written into CLAUDE.md: one checked-in version-pinned asset, still no npm/CDN. The cartesian partial bundle covers the whole arc (scatter + heatmap for W2b traces).
- **Reproducibility**: every endpoint response's `code` field was run verbatim against the live catalog — it reproduces the endpoint's numbers exactly.
- **Statelessness**: all view state is URL-carried (`tab`/`y`/`x`/`view`/`filters`/`bincfg`); scan/day steppers keep the whole setup (JS mirrors state into the server-rendered links). A link IS the analysis — also the multi-user story.
[...]

PR #736, opened 2026-09-01
portal 0.12.0: Images tab per-bin averaged grid (the W2a checkpoint PR)

skbarber:
Second PR of the **W2a Images-tab arc** — the tab feature itself, and **the architecture checkpoint**: the doctrine says W2a must be dramatically cheaper than W1d or we stop and fix seams. Verdict below.

## What it adds (portal 0.9.1→0.11.0 was plot-tools; this is 0.11.0 → 0.12.0)

**The Images tab gains a per-shot ⇄ per-bin toggle** (the same URL-carried `view` state as the Plot tab — deliberate judgment call, cheap to veto: one binned⇄per-shot state for the whole analysis, "a binned link means binned everywhere"). Per-bin view renders a lazy grid of `nanmean`-averaged device images, one card per bin with `bin_col: label · n=count` captions.

Per the adding-a-tab checklist:

1. **Primitives** (already landed in #735): `average_frames`, plus this PR's `compute_bin_key` reuse — no new math anywhere in the portal.
2. **Endpoints** (~150 LOC in `app.py`):
   - `/api/run/{uid}/bin-images?device=&filters=&bincfg=` — membership JSON (bins, counts, member shots) over `compute_bin_key` + the same `groupby(dropna=False, observed=True)` semantics as `/binned`; carries a `code` snippet reproducing the numbers exactly (the pixel side is a sketch — the tier ladder is serving logic, not numerics).
   - `/run/{uid}/bin-image.png?device=&bin=<index>&filters=&bincfg=` — one bin's average, display-windowed once *after* averaging. `bin` is the index in `bin-images` order; both endpoints run the identical membership call, so the index is stable regardless of label serialization. Per-shot refusals carry over: events-bound guard, missed-shot (no timestamp) skip, vendor/unrenderable 404 — **never average a neighbour's image in**; any ordinal-resolved member downgrades the response to `no-cache`.
3. **Resource layer** (~70 LOC moved, not added): `load_shot_image` split into `load_shot_array` (`ShotArray` — the tier ladder resolving to raw pixels, riding `ShotDataCache` exactly as before) + a thin PNG wrapper. Averaging must precede windowing, so the array level is the natural seam; single-shot serving is unchanged (pinned by the untouched 34 pre-existing resource tests).
4. **Tab** (~90 LOC template/JS): segmented toggle + lazy `` grid, prefix-agnostic URLs via `ROOT`, boot-deduped fetch, filter/binset applies refresh the grid; all state URL-carried.
5. **Tests**: 7 new in `tests/test_resources.py::TestBinImages` — hand-computed averaged pixels (identity markers at half intensity, native + stack devices; the stack case pins that the leading pre-scan extra frame stays OUT of the average via the timestamp join), filters narrowing membership and pixels, the 400/404 ladder, immutable headers, tree-untouched invariant.

## The checkpoint verdict

W1d (#722) was ~2,600 added lines building the rail, three tabs, four endpoints, the vendored renderer, and the URL-state machinery. **W2a's tab feature is ~430 added lines** (including tests and docs), containing zero new numerics, zero new parsing vocabulary, and zero new rendering machinery — every piece composed from existing primitives (`compute_bin_key`, `average_frames`, `shot_axis_for_frame`, the tier ladder, `parse_filters`/`parse_bincfg`, `_png_headers`, the sticky URL state). The checklist held; the seams are earning their keep.

## Testing

Portal suite: **166 passed** (was 159; 7 new). Read-only view layer, no hardware verification owed. Live .14 verification happens at the arc's promotion per the settled staging strategy.

### DIFF
--- a/GEECS-DataPortal/CHANGELOG.md
+++ b/GEECS-DataPortal/CHANGELOG.md
@@ -18,8 +18,11 @@ Format follows [Keep a Changelog](https://keepachangelog.com/en/1.1.0/).
   once after averaging). Both run the same `compute_bin_key` +
   groupby membership call, so the `bin` index is stable between
   listing and render; per-shot refusals carry over (never average a
-  neighbour in: events bound, missed-shot skip, vendor 404), and a
-  bin whose only resolutions were ordinal serves `no-cache`.
+  neighbour in: events bound, missed-shot skip, vendor 404), and a bin
+  containing any native listing-order (ordinal-fallback) resolution
+  serves `no-cache` — the same per-shot rule. `min_count` applies to
+  the grid exactly as `bin_frame` applies it to `/binned` (per-bin row
+  counts), so the shared binset popup governs both tabs.
 - `resources.load_shot_array` (+ `ShotArray`): the tier ladder now
   resolves to raw pixels, with `load_shot_image` reduced to the
   render-one-shot wrapper — single-shot serving and per-bin averaging
--- a/GEECS-DataPortal/CHANGELOG.md
+++ b/GEECS-DataPortal/CHANGELOG.md
@@ -3,6 +3,28 @@
 All notable changes to this package will be documented here.
 Format follows [Keep a Changelog](https://keepachangelog.com/en/1.1.0/).
 
+## [0.12.0] - 2026-09-01
+
+### Added
+
+- **Images tab per-bin averaged grid** (W2a): a per-shot ⇄ per-bin
+  toggle mirroring the Plot tab's ONE URL-carried `view` state (a
+  binned link means binned everywhere — deliberate), rendering a lazy
+  grid of per-bin `nanmean` averages. Two new routes:
+  `/api/run/{uid}/bin-images?device=&filters=&bincfg=` (membership
+  JSON — bins/counts/member shots, notebook-reproducible via its
+  `code` snippet) and `/run/{uid}/bin-image.png?bin=<index>` (one
+  bin's average via the shared `average_frames`, display-windowed
+  once after averaging). Both run the same `compute_bin_key` +
+  groupby membership call, so the `bin` index is stable between
+  listing and render; per-shot refusals carry over (never average a
+  neighbour in: events bound, missed-shot skip, vendor 404), and a
+  bin whose only resolutions were ordinal serves `no-cache`.
+- `resources.load_shot_array` (+ `ShotArray`): the tier ladder now
+  resolves to raw pixels, with `load_shot_image` reduced to the
+  render-one-shot wrapper — single-shot serving and per-bin averaging
+  share one resolution path (and one `ShotDataCache` ride).
+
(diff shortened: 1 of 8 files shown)


skbarber:

The reviewer ran empirical probes against both endpoints (numeric/string/NaN bin labels, min_count configs, render shapes, XSS posture per the #729 precedent) and both suites. **2 MEDIUM + 2 LOW findings + nits; all dispositioned as FIXED in 5dda26a9 and re-verified by the same reviewer against its original probes.**

| # | Finding | Disposition |
|---|---|---|
| 1 | **MEDIUM:** `min_count` in the shared `bincfg` governed `/binned` but was silently ignored by the grid — the binset popup's "min shots/bin" did nothing on the Images tab, contradicting the one-view-state doctrine (empirically demonstrated). | **Fixed**: `_bin_groups` drops groups below `min_count` on the same row-count basis as `bin_frame`'s `g.size()`; pinned by a test that fails against the pre-fix code (JSON list + the PNG index following it). Residual nuance the reviewer noted as inherent, no action: `/binned`'s counts run after its y-column dropna, which the y-less grid cannot mirror. |
| 2 | **MEDIUM:** the `load_shot_array` refactor moved `to_display_png` outside the never-500 guards — a readable-but-unrenderable array (4-D `/image` h5, stacked `.npy` in a dev/scratch folder) turned the honest 404 into a 500. | **Fixed**: render guarded in the wrapper (degrades to the missing card, path/cacheable preserved) and in `run_bin_image` (404); pinned via the reviewer's exact vector. Bonus: the cached-stack memory path, unguarded even before the refactor, is now covered. |
| 3 | **LOW:** the grid's `IMG_KEY` memo could lock in a stale out-of-order response with no recovery path. | **Fixed**: superseded in-flight responses dropped in both `.then` and `.catch`; a failed current fetch still resets the memo. |
| 4 | **LOW:** the advertised cache behaviors (ordinal member → `no-cache`, running run → `no-cache`) were implemented but untested — deleting the logic left all tests green. | **Fixed**: two pins added (deleting the `cacheable` line now fails); CHANGELOG rephrased to name the native listing-order join specifically. Reviewer also verified ordinal members never enter `ShotDataCache`. |
| nits | Caption `n=` vs actually-averaged frames invisible client-side; `_image_folder` as a third day/folder-resolution variant. | **Fixed**: caption tooltip states the distinction; `run_image` adopts `_image_folder` (behavior-equivalent, one resolution path). |

Checked-and-clean worth recording: never-serve-a-neighbour holds under every averaging path (the stack test genuinely pins pre-scan-orphan exclusion); NaN bins serialize as `null` end-to-end and render; XSS posture matches #729 (all label/col/error sinks escaped, URLs via `URLSearchParams`, prefix-agnostic via `ROOT`); the membership snippet is a faithful transcription of the endpoint.

Portal suite: **170 passed** (11 new since 0.11.0). Reviewer's close: "No surviving findings."

PR #737, opened 2026-09-01
portal 0.13.0: Images tab ephemeral-processing selector (W2a PR 3)

skbarber:
Third and final feature PR of the **W2a Images-tab arc**: the ImageAnalysis link — the "processing" selector that runs a configured diagnostic **write-free** on the pixels the portal serves, via #735's `run_diagnostic_ephemeral` seam.

## What it adds (portal 0.12.0 → 0.13.0)

- **`processing=<diagnostic id>` URL state** on both image endpoints. Per-shot view renders the diagnostic's `processed_image` instead of the raw frame; per-bin view processes **each member shot first, then averages** — the correct order for nonlinear pipeline steps (thresholding, masking). One `run_diagnostic_ephemeral` call per request (one analyzer instantiation per batch, per the seam's design). Raw serving is byte-for-byte untouched when the param is absent (pinned).
- **A selector in the Images plotbar** (`raw` + every discovered diagnostic), rendered only when the feature is live; state rides the URL like everything else (`readState`/`writeState`/sticky links extended).
- **ImageAnalysis becomes an optional portal dependency** — the `analysis` extra (`pip install geecs-data-portal[analysis]`), mirroring GeecsBluesky's pattern. Without it the selector hides and the param 404s with an install hint; the base portal's footprint is unchanged.
- **Explicit-opt-in configs tree** (design judgment call, cheap to veto): `--processing-configs <tree>` / `create_app(processing_config_dir=…)` names the scan-analysis configs root, and the portal **never falls back to the global config resolution** — the 03 design doc's finding 7 names two competing resolution paths and says the portal should pass its own explicitly. No flag → feature off. (This also keeps the test suite hermetic — the fallback would have made pages depend on the host's `config.ini`.)
- **Error ladder**: unknown diagnostic → 404, denylisted (HASO/Grenouille per the #735 contract) → 400 with the contract message, analyzer failure → 400 honestly, render failure → 404 — never a 500. The write-free guarantee itself lives in ImageAnalysis (structural: frames only, `file_path` refused, pre-import denylist) — the portal's read-only doctrine is preserved by construction, not convention.

## Testing

Portal suite: **176 passed** (6 new in `TestProcessingSelector`): ROI-diagnostic crop verified in the decoded PNG for per-shot AND per-bin (markers at the hand-computed cropped coordinates), selector rendered only when configured, unknown/denylisted ladder, missing-extra degrade (selector hidden + honest 404), raw serving untouched. Hermetic — diag YAMLs are tmp_path trees.

Lock note: `poetry.lock` regenerated for the new optional dep inside the worktree and the worktree prefix stripped per the repo rule (0 `worktrees` occurrences).

Read-only view layer — no hardware verification owed; live .14 check at promotion (the systemd unit gains `--processing-configs` only when Sam opts the deployment in).


skbarber:

The reviewer audited every write site in ImageAnalysis runtime code to verify the write-free guarantee end-to-end (it holds: no `auxiliary_data` is passed, path-gated writers can't arm from constructor kwargs, denylist pre-import, `apply_background` read-only), verified the lock (`poetry check --lock` clean, 0 worktree paths, extras resolve), and confirmed hermeticity (suite passes with `HOME` pointed at an empty dir). **1 medium-high + 2 medium-low + 1 low finding + notes; all dispositioned in 0192b0fc and re-verified by the same reviewer.**

| # | Finding | Disposition |
|---|---|---|
| 1 | **Med-high:** processed PNGs cached `public, immutable` keyed by a URL that omits their second input — the mutable diagnostic YAML. Editing a config then reloading served the year-stale render, pinnable for all users by an intermediary cache under the reverse-proxy mount. | **Fixed**: processed responses always serve `no-cache` on both endpoints (chosen over ETag/config-fingerprint — bounded recompute, internal tool); raw paths keep the existing header ladder. Pinned three ways. |
| 2 | **Med-low:** the process-then-average test used a crop-only diagnostic — crop commutes with averaging, so the wrong order would pass. | **Fixed**: new threshold diagnostic (cutoff 600, between raw marker 1000 and averaged 500); the reviewer independently verified the wrong order renders an all-black image, so the test genuinely discriminates. |
| 3 | **Med-low:** a client-side selector pick was dropped by the shot steppers (not `a.act`, so never state-rewritten) and the Show form. | **Fixed**: steppers gained `class="carry"` (new class — `.act` is button-styled), `syncRailLinks` rewrites both, and `setProcessing` updates a new `processing` hidden input live. |
| 4 | **Low:** a typo'd `--processing-configs` silently no-opped into a hidden selector. | **Fixed**: startup WARNING naming both failure modes (also fires on a valid-but-empty tree — deliberate). |
| notes | Root dep graph missing the new optional edge; ImageAnalysis ships `pytest` as a main dependency (leaks into consumers' production installs); 404 details embed the configs-tree path. | Graph **fixed** (GeecsBluesky's convention); packaging debt **filed as #739** (pre-existing, surfaced not created here); error-detail internals **waived** as an internal-tool tradeoff — flagged for the owner. |

Redundancy/placement clean per the reviewer — with one arc-level note for later: `geecs_bluesky.analysis.image_analysis` runs analyzers over in-memory Tiled arrays *without* the ephemeral seam; a future convergence candidate, outside this diff.

Portal suite: **178 passed** (8 new since 0.12.0). Reviewer's close: "No surviving findings; the review is closed clean."

PR #740, opened 2026-09-01
portal 0.13.1: CI installs the analysis extra; processing tests importorskip (the #737 CI-red fix)

skbarber:
**Fixes the red `unit-tests` on `feat/images-tab`.** #737's `TestProcessingSelector` failed in CI because the workflow installs the portal with plain `poetry install --with dev` — no `analysis` extra, so `image_analysis` wasn't importable there and the six processing tests 404'd. My local runs had the extra installed, and the merge slipped past the red check because my CI watch was piped through `tail`, which discarded the failing exit status — a process error on my side, corrected (checks are now verified by exit code before any merge).

The fix, per the GeecsBluesky extras precedent (its workflow comment documents the same lesson from 2026-08-21):

- **Workflow**: the portal install step gains `--extras analysis`, with a comment naming the importorskip it guards — in CI these tests always **run**, never silently skip.
- **Tests**: `TestProcessingSelector`'s shared fixture gains `pytest.importorskip("image_analysis")`, so a minimal local env skips gracefully instead of failing.
- **/env-doctor**: the extras table gains the `GEECS-DataPortal` row.
- Portal 0.13.1 (test/CI-only patch, CHANGELOG entry).

Suite locally (with extra): 178 passed — the guard no-ops. The real verification is this PR's own CI run, which now exercises the previously failing path with the extra installed.

Review note: this is a 5-file test/CI patch reverting no behavior; per the /land ritual a fresh-context review still runs and its report lands as a comment before merge.

### DIFF
--- a/.claude/skills/env-doctor/SKILL.md
+++ b/.claude/skills/env-doctor/SKILL.md
@@ -50,6 +50,7 @@ diagnose the package the current task is about.
    | `GEECS-Core` | `poetry install` (self-contained) |
    | `GEECS-Console` | `poetry install` (the optimization stack is worker-side — no console extra) |
    | `GEECS-LogTriage` | `poetry install` |
+   | `GEECS-DataPortal` | `poetry install --extras analysis` (the processing-selector test class importorskips `image_analysis` without it) |
 
    Symptom of missing `ca`: the whole GeecsBluesky mock-device test layer
    collects as "1 skipped" (`pytest.importorskip("aioca")`).
--- a/.github/workflows/unit-tests.yml
+++ b/.github/workflows/unit-tests.yml
@@ -37,7 +37,10 @@ jobs:
 
       - name: Install GEECS-DataPortal dependencies
         working-directory: GEECS-DataPortal
-        run: poetry install --with dev
+        # --extras analysis: the Images tab's processing-selector tests
+        # (TestProcessingSelector) importorskip image_analysis and would
+        # silently skip without it (the #737 CI-red lesson, 2026-09-01).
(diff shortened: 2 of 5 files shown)


skbarber:

The reviewer performed failing-run forensics (all three red runs: one signature — 6 failures in `TestProcessingSelector`, each carrying the app's own "'analysis' extra is not installed" warning; no second cause hiding behind it), verified the workflow edit covers the failing path (same working-directory/env, no poetry cache on the ubuntu job, no other workflow touches the portal, the extra already locked by #737 so no relock owed), and watched this PR's own CI run the previously failing tests green.

| # | Finding | Disposition |
|---|---|---|
| 1 | **Low:** the class-wide `importorskip` needlessly retired two tests that never needed the extra — including the missing-extra degradation test, whose whole point is the extra-less env (both passed in #737's extra-less CI). | **Fixed** in 1cf7ed8b: guard moved to a per-test `analysis_extra` fixture on exactly the six importing tests. Reviewer verified empirically with a faithful absence simulation: **172 passed, 6 skipped, 0 failed** extra-less (mirroring #737's CI), **178 passed** with the extra. |
| 2 | **Informational:** "in CI these tests always run" is documented (workflow comment), not enforced — a future edit dropping the extra would skip green. | **Waived**: the established repo pattern (GeecsBluesky `qserver`/`qs-client`, GEECS-MCP — identically comment-guarded), per the reviewer's own disposition. |

Process note, on the record: #737 merged past its red `unit-tests` because the checks watch was piped through `tail`, discarding the failing exit status — corrected; merges are now gated on the unpiped `gh pr checks` exit code.

PR #741, opened 2026-09-01
portal 0.14.0: loadable-only selector + in-place image error surfacing (Sam's first live test)

skbarber:
Two defects from the owner's first live test of the processing selector (Amp4, 2026-09-01). The failure itself was a **legacy flat camera config** (`UNCLASSIFIED/UC_Amp4_IR_input.yaml` — no `image_analyzer`, no `pipeline`, pre-#412 `enabled:` flags) that discovery lists but `load_diagnostic` rightly refuses; the config has been migrated in GEECS-Plugins-Configs (owner to review/commit). What the portal got wrong:

1. **The selector offered an unloadable diagnostic.** `_processing_names` now validates each discovered stem with a real `load_diagnostic`, cached against the tree's YAML `(path, mtime)` fingerprint so re-validation happens only when the tree changes; an invalid config becomes an INFO log naming the file. A hand-edited URL still gets the honest 400 (pinned).
2. **The failure was invisible** — the endpoints returned an honest 400 with the exact validation error, but an `` can only render a broken icon. The per-shot image and every bin-grid card now carry an `onerror` hook that fetches the response's `detail` and shows it in place (cleared on the next processing change, so a retry isn't masked).

Tests: **178 passed** — the fixture gains a discoverable-but-unloadable legacy YAML; pinned that it's absent from the page while `UC_Crop` remains, and that forcing it by URL still 400s with "Invalid diagnostic". (The `onerror` path is browser behavior — the hook's presence ships in the template; verified live on :8214.)

### DIFF
--- a/GEECS-DataPortal/CHANGELOG.md
+++ b/GEECS-DataPortal/CHANGELOG.md
@@ -3,6 +3,25 @@
 All notable changes to this package will be documented here.
 Format follows [Keep a Changelog](https://keepachangelog.com/en/1.1.0/).
 
+## [0.14.0] - 2026-09-01
+
+### Fixed
+
+Sam's first live test of the selector (Amp4, 2026-09-01) — the
+diagnostic YAML was a legacy flat camera config, and the failure was
+invisible:
+
+- **The selector now offers only LOADABLE diagnostics**: each
+  discovered stem is validated with a real `load_diagnostic` (cached
+  against the tree's YAML mtimes), and an unloadable legacy config is
+  an INFO log line naming the file — never a pickable entry that can
+  only produce a broken image. A hand-edited URL still gets the
+  honest 400.
+- **Image failures surface their reason**: the per-shot image and
+  every bin card carry an `onerror` hook that fetches the endpoint's
+  4xx `detail` and shows it in place of the browser's broken-image
+  icon (cleared on the next processing change).
+
 ## [0.13.1] - 2026-09-01
 
 ### Fixed
--- a/GEECS-DataPortal/geecs_portal/app.py
+++ b/GEECS-DataPortal/geecs_portal/app.py
@@ -656,25 +656,62 @@ def create_app(
             raise HTTPException(status_code=404, detail=f"unknown device {device!r}")
         return folder, (run_day.isoformat() if run_day else None)
 
+    # (fingerprint → valid names): re-validated only when the tree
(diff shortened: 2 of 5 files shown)


skbarber:

**1 MODERATE + 3 LOW + 2 INFO; all six taken, all confirmed fixed by the same reviewer** (re-probed the race matrix, re-ran the suite: 178 passed).

| # | Finding | Disposition |
|---|---|---|
| 1 | **MODERATE:** `imgFailed`'s late-resolving error fetch could hide a successfully retried image behind a stale error (the fetch re-runs a heavy request, so the window is seconds). | **Fixed** (de3d65e7): `src` captured at entry; `show()` bails on `img.src !== src \|\| !img.isConnected`. Reviewer re-probed fail→switch→fail sequences and the detached-grid case: no hole. |
| 2 | **LOW:** fingerprint computed *after* discovery made the add-race sticky-stale. | **Fixed**: fingerprint-before-listing — the race now costs one harmless revalidation under a never-recurring key. |
| 3 | **LOW:** "cheap validation error" comment was false for analyzer-failure 400s (the onerror refetch re-runs loaded+processed work). | **Fixed**: comment states the re-run and the 2×-per-src bound; accepted cost for an error path. |
| 4 | **LOW:** cache read raced a concurrent `clear()` into a `KeyError` 500. | **Fixed**: `.get()` + `is not None` (deliberately not truthiness — a cached empty list for an all-invalid tree still hits). |
| 5 | **INFO:** mtime-only fingerprint misses same-quantum edits on coarse-mtime (SMB) trees. | **Taken**: tuples carry `st_size` too; nit follow-up also taken — one `stat()` per file (no torn tuples, half the syscalls). |
| 6 | **INFO:** a bookmarked URL naming a since-dropped diagnostic rendered the selector as "raw" with no one-click recovery. | **Taken**: a selected `"<name> (unavailable)"` fallback option renders (Jinja autoescape verified on both sinks); picking raw now fires a real change event. Pinned by a page-text test. |

Refuted probes worth recording: the missing-extra degrade path is unchanged (the widened import fails identically under the monkeypatch); error details reach the DOM via `textContent` only; the bin-card `nextElementSibling` adjacency is handled by the `imgerr` class guard; cache growth is capped at one entry and request-driven cache-busting is impossible (server filesystem state only — no DoS).

Suite: **178 passed** (+2 assertions). Reviewer's close: "From my side the PR is ready."

PR #742, opened 2026-09-01
portal 0.15.0: image colormaps + display windowing (owner's step 1)

skbarber:
The owner-requested "step 1" of the interactive-images plan (step 2 — the plotly Heatmap hover/zoom view — is planned post-promotion): **colormaps and window control on every served image**, with zero payload cost.

## What it adds (portal 0.14.0 → 0.15.0)

- `to_display_png` grows `cmap` (any matplotlib colormap; applied to the windowed 2D image, RGB inputs keep their own colors) and `plo`/`phi` percentile-window overrides (defaults 1/99.7). Value-degrade semantics per the display doctrine: unknown colormap → grayscale, insane window (inverted/out-of-range) → defaults — display state rides shared links and must never fail one. matplotlib is already a portal dependency (plot.png).
- Both image endpoints (`image.png`, `bin-image.png`) — raw, processed, and averaged paths alike — take the shared `display` state, whose curated vocabulary gains `cmap`/`plo`/`phi` (types 400 at `parse_display`, the existing boundary; the plot figures simply ignore the new keys — one display state, per doctrine).
- A small **"display…" popup in the Images plotbar** (colormap select + window inputs) merging into the same `display` JSON; the per-shot image, the bin grid, the steppers, and the grid's refresh key all carry it.

## Testing

Portal suite: **184 passed** (6 new): viridis pixel checked against matplotlib's own colormap value (no palette literals to go stale), unknown-cmap grayscale degrade, window override provably changing saturation, inverted-window byte-identical to defaults, the 400 ladder on the image endpoints (`notjson`, unknown field), and a colored bin-average render.

No cache-semantics change: `display` is in the URL, so distinct renders are distinct cache entries; the processed-path `no-cache` rule from #741's review is untouched.

Adversarial review to follow as a PR comment before merge. This is the last W2a feature PR — promotion follows its merge.

### DIFF
--- a/GEECS-DataPortal/CHANGELOG.md
+++ b/GEECS-DataPortal/CHANGELOG.md
@@ -3,6 +3,23 @@
 All notable changes to this package will be documented here.
 Format follows [Keep a Changelog](https://keepachangelog.com/en/1.1.0/).
 
+## [0.15.0] - 2026-09-01
+
+### Added
+
+- **Image colormaps + display windowing** (owner ask, "step 1" of the
+  interactive-images plan): the image endpoints (`image.png`,
+  `bin-image.png` — raw and processed alike) take the shared
+  `display` state's new curated fields `cmap` (matplotlib colormap
+  name) and `plo`/`phi` (percentile-window overrides, defaults
+  1/99.7). Types 400 at the parse boundary, values degrade (unknown
+  colormap → grayscale, insane window → defaults) — display state
+  rides shared links and must never fail one. A small "display…"
+  popup in the Images plotbar edits them; URLs carry them everywhere
+  (per-shot, grid, steppers). RGB inputs keep their own colors.
+  Step 2 (interactive per-shot Heatmap view with hover pixel values)
+  is planned post-promotion.
+
 ## [0.14.0] - 2026-09-01
 
 ### Fixed
--- a/GEECS-DataPortal/CLAUDE.md
+++ b/GEECS-DataPortal/CLAUDE.md
@@ -143,8 +143,11 @@ member shots that resolve to pixels average via the shared
 carry over, and an ordinal-resolved member downgrades the response to
 `no-cache`) · `/health` (catalog probe — the fleet-map health check).
 
-Both image endpoints also take `?processing=<diagnostic id>` (the
-`processing` URL state): the named ImageAnalysis diagnostic runs
+Both image endpoints also take `?display=` (the shared display state's
+image slice: `cmap` matplotlib-colormap name + `plo`/`phi` percentile
+window — types 400 at parse, values degrade to grayscale/defaults;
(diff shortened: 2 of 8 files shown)


skbarber:

**1 HIGH + 1 MEDIUM + 1 LOW-MED + nits; all taken, all re-verified by the same reviewer against ca50e1dc.** Empirical probes included refuting matplotlib thread-safety concerns (`mpl.colormaps[name]` returns a copy per call in the pinned 3.11.1 — no shared lazily-initialized state on the threadpool, no pyplot anywhere) and confirming the new display keys reach no figure/layout/attribute sink.

| # | Finding | Disposition |
|---|---|---|
| 1 | **HIGH:** one-sided window overrides silently no-opped — `float(None)` in the pairwise conversion discarded both values, and the popup's stores-defaults-as-absent convention made `{"phi": 50}` the *guaranteed common payload*; the reopened popup then displayed the "applied" value over an untouched image. Both original window tests passed both knobs, so the suite had zero discriminating power at the bug. | **Fixed**: per-field defaulting inside the try; pinned by `test_one_sided_window_override_applies` (fails against pre-fix code — the reviewer demonstrated the exact pre-fix byte-identity it now refutes). |
| 2 | **MEDIUM:** the Plot tab's display popup rebuilt the shared state from only its own inputs, silently deleting `cmap`/`plo`/`phi` — with the rendered image then diverging from the URL. | **Fixed**: `applyDisplay` seeds the image slice from the prior state before rebuilding its own keys (literal key names — no new pollution surface). Residual reviewer note, accepted: key reordering can cost one redundant grid refetch; cosmetic only. |
| 3 | **LOW-MED:** the shot "Show" form dropped `display` (the PR body's "carried everywhere" overclaimed). | **Fixed**: hidden `display` field added, and BOTH hidden fields are now centrally synced in `writeState` — every mutation path funnels through it, so the form can never go stale (the old per-callsite sync removed as redundant). Autoescape of JSON-in-attribute verified round-trip-safe. |
| 4 | Nits: H×W×4 float64 RGBA transient (~180 MB on a 5.5 MP frame); docstring's flat-image claim stale under a colormap. | **Fixed**: `colormap(norm, bytes=True)` (byte-identity verified over a 10k-value sweep); docstring states the floor-color behavior. |

Refuted probes worth recording: cache correctness holds (`display` rides every image URL — Jinja src, grid, `IMG_KEY` — so a cmap change is a distinct cache entry; `plot.png` correctly takes none of this); adding the keys to `_DISPLAY_FIELDS` is required, not incidental (an image-display link would otherwise 400 the Plot tab's shared-state fetches); junk inputs to the public `to_display_png` all degrade.

Suite: **185 passed** (7 new). Reviewer's close: "the branch is clean to merge from my side."

PR #745, opened 2026-09-01
Fleet map: checkout column, one-clone-per-service pattern, fresh-host bootstrap gotchas

skbarber:
Docs-only, from the owner's question after the W2a deploy ("are we accumulating clones at different states?" — answer: it's the deliberate pattern, but it deserved to be a documented fact rather than archaeology):

- **Checkout column** in the services table: which clone each service runs from, as service-account-relative paths (public-repo hygiene: no account names). Tiled/DB/NAS marked as non-repo services; the capture daemon's shared clone with the worker called out as the one deliberate exception (co-versioning is part of the co-location requirement).
- **"One clone per service" section**: the rule and its teeth — `Restart=on-failure` means a shared tree leaves every other service one crash away from restarting onto unvalidated code; per-service cadence; each clone as a pinned rollback point.
- **Fresh-host bootstrap gotchas** collected from the live deploys, ahead of the services-server consolidation: service-account installs, per-clone extras, config.ini, non-login-shell PATH (`bash -lc`), the CRLF trap on share-hosted config checkouts (`core.autocrlf true`, never rewrite in place — LabVIEW reads the same files), quoting space-bearing share paths in `ExecStart`, and the deploy-then-health-check-then-update-this-page loop.
- Snapshot note bumped to September 2026 (portal 0.15.x deployed).

`mkdocs build` clean (exit 0). No package code touched — no version bumps owed (fleet map lives in root docs/).

Landing to master per the owner's explicit instruction; adversarial review to follow as a PR comment before merge.

### DIFF
--- a/GEECS-DataPortal/CHANGELOG.md
+++ b/GEECS-DataPortal/CHANGELOG.md
@@ -3,6 +3,16 @@
 All notable changes to this package will be documented here.
 Format follows [Keep a Changelog](https://keepachangelog.com/en/1.1.0/).
 
+## [0.15.1] - 2026-09-01
+
+### Fixed
+
+- DEPLOYMENT.md now names the `analysis` extra and the
+  `--processing-configs` `ExecStart` flag (quoted — share paths carry
+  spaces) that the processing selector needs — the fleet map's
+  "each runbook names its extras" claim was false for this runbook
+  (docs-only; #745 review finding).
+
 ## [0.15.0] - 2026-09-01
 
 ### Added
--- a/GEECS-DataPortal/DEPLOYMENT.md
+++ b/GEECS-DataPortal/DEPLOYMENT.md
@@ -45,8 +45,21 @@ checks pass.
 ```bash
 cd ~/GEECS-Plugins-portal/GEECS-DataPortal
 poetry env use python3.11
-poetry install
+poetry install --extras analysis
 ```
+
+The `analysis` extra installs ImageAnalysis for the Images tab's
+**processing selector** (0.13.0+). The feature is explicit-opt-in
+twice over: the extra installs the code, and the unit's `ExecStart`
+must name the scan-analysis configs tree —
+
+```
+--processing-configs "/path/to/GEECS-Plugins-Configs/scan_analysis_configs"
+```
+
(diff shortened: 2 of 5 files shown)


skbarber:

The reviewer verified every factual claim against the runbooks and unit templates it names (including `Restart=on-failure` in all five templates — the section's central rationale), plus the anchor, hygiene, and a clean mkdocs build. **1 HIGH + 1 MED-HIGH + 1 MED + 1 LOW, all taken, all confirmed fixed in 482f6f22** (the reviewer re-verified the portal-runbook additions against the portal's *code*, not just the map).

| # | Finding | Disposition |
|---|---|---|
| 1 | **HIGH:** the PVA Checkout cell asserted a per-host clone that doesn't exist — the PVA runbook's model is the shared "Active Version" clone as the fleet pin with baked per-host venvs. Operationally misleading in exactly the way a fleet map must not be. | **Fixed**: cell states the runbook's truth verbatim-faithfully. |
| 2 | **MED-HIGH:** the MCP cell prescribed "own clone" against its runbook's deliberate design (worker's checkout for config-truth parity; isolation via a non-editable baked venv). | **Fixed**: cell matches the runbook; the exceptions bullet now carries all three isolation stories (worker+capture shared clone, MCP baked venv, PVA share-clone pin). |
| 3 | **MED:** "each runbook names its extras" was false for the map's own example — the portal runbook named none. | **Fixed at the root**: `GEECS-DataPortal/DEPLOYMENT.md` now names `--extras analysis` in install + upgrade rituals and the quoted `--processing-configs` flag (portal 0.15.1, docs-only patch bump per convention). Reviewer verified the new text against `__main__.py`/`app.py` behavior. |
| 4 | **LOW:** the qserver runbook's `/opt/geecs` layout diverged from the map's `~/<service>-checkout` convention without acknowledgment. | **Fixed**: "the path is a convention, not the invariant" clause names both layouts as the same pattern. |
| nit | The unit template's `ExecStart` carried no hint of the selector flag — extra installed, selector dark, no warning. | **Taken**: commented example added to the template. |

mkdocs build clean throughout. Reviewer's close: "the PR is clean to land."

PR #734, opened 2026-09-01
ScanRequest schema v2: CaptureSettings grouping + request/record split (Phase 1 of the schema refactor)

skbarber:
Phase 1 of `Planning/schema_refactor/00_overview.md` (the ScanRequest schema
refactor, amended this session with Sam: console-first priorities, `capture`
naming, required lifting validator, the `SubmissionRecord` vehicle, and the
shared-execution invariant). Phase 2 (three named plans) is NOT in this PR.

## What

**ScanRequest format v2** (GEECS-Schemas 0.14.0):

- The seven capture-concern fields (`shots_per_step`, `acquisition`,
  `save_sets`, `background_telemetry`, `native_image_save`,
  `trigger_profile`, `trigger_variant`) move into one new
  `CaptureSettings` sub-model at `ScanRequest.capture`. The bare-string
  `save_sets` coercion and the trigger-variant-needs-profile validator move
  with them.
- **Request/record split**: `submission` leaves the request document. The
  `SubmissionRecord` (type stays exported) now travels *beside* the request:
  `submit_scan(request, submission=...)` → the funnel plan's new optional
  `submission` kwarg → `md["submission"]` exactly as before.
- **v1 lifting validator (the migration story)**: a `mode="before"`
  validator accepts the flat v1 layout forever — lifts the seven flat fields
  into `capture`, drops an embedded v1 `submission` record, normalizes
  `schema_version` to 2. Saved presets, archived run-metadata documents, and
  stale clients keep validating with no lockstep upgrade. Mixing flat fields
  with an explicit `capture` block is rejected as ambiguous.

**Consumer adoption** (GeecsBluesky 0.70.0, GEECS-Console 0.26.0,
GEECS-MCP 0.8.0):

- GeecsBluesky: runner, queue plan, and qs_client preflight read
  `request.capture.*`; `run_scan_request`/`geecs_scan_request_plan` gain the
  optional `submission` parameter; `stamp_submission` →
  `build_submission_record` (returns the record instead of stamping the
  request); `metadata_submission` takes the record;
  `SCAN_REQUEST_PLAN_ANNOTATION` documents the new parameter.
- GEECS-Console: `request_builder` builds/reads the `CaptureSettings`
  sub-model — the doc's console-side acceptance test — and the submit
  pipeline sends the record beside the request.
- GEECS-MCP: `submit_scan` sends the record beside the request.
- GEECS-Data-Utils: **no changes** — `tiled_catalog` reads the engine's own
[...]

PR #746, opened 2026-09-01
Planning: schema refactor Phase 2 amendment (three steps after #734)

skbarber:
Planning-only. Amends `Planning/schema_refactor/00_overview.md` after #734 merged and was hardware-verified (Undulator Scan004/Scan005, 2026-09-01).

## What

- Status + Phase 1 marked **landed**, with the two things Phase 2 inherits: the versioning policy written down in #734 (integer `schema_version` bumps only with a migration) and the worker-first deploy order.
- New section **"Findings from reading the engine"**: the prologue exists twice (plan body + headless `run_scan_request`), the headless path's caller audit (only `GeecsSession.run`; alive as the hermetic-suite entry point, the hardware sweep test, and the documented scripting API — not a dead limb), `scan_request_runner.py` as a grab-bag module, scattered mode branches, and the two-tier cost of new plans vs new execution modes.
- **Phase 2 split into three ordered steps**: 2a one preamble (pure refactor, funnel signature unchanged), 2b the three named plans (unchanged from the original Phase 2, plus where each piece lives), 2c the OSPREY tail.
- Open items: #730 OWED restart closed by the #734 verification restart; worker-host checkout drift noted.
- Sequencing updated to three PRs with per-PR hardware verification.

## Why

Sam asked whether the schema change was smaller than imagined and where the per-mode split went; the answer is Phase 2, and reading the engine before building it showed the wrappers cannot be thin while two orchestrations exist. This records that so the implementer starts at 2a.

## Judgment calls flagged for veto (in the doc itself)

- 2a's runner module split by concern as a move-only commit in the same PR (large diff, zero logic change) — or deferred to the first new-mode PR.
- Per-mode spec-builder registry explicitly **deferred** until a fourth mode forces it.

## Tests / versions

No package code changed; no version bump; no test suites affected.

## Hardware verification

N/A (planning document).

### DIFF
--- a/Planning/schema_refactor/00_overview.md
+++ b/Planning/schema_refactor/00_overview.md
@@ -1,9 +1,11 @@
 # ScanRequest schema refactor + named scan plans
 
 Handoff design doc, 2026-09-01 (Sam + Claude session on the HTU/OSPREY integration).
-Status: **approved direction, implementation starting** (amended 2026-09-01 after
-review with Sam: priorities, naming, migration story, invariants). Written for
-whoever picks this up; assumes no context from the originating conversation.
+Status: **Phase 1 landed** (#734, merged 2026-09-01, hardware-verified on
+Undulator Scan004/Scan005 the same day); **Phase 2 amended 2026-09-01** after
+reading the engine with Sam (see "Findings from reading the engine" and the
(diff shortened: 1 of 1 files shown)


skbarber:
Folding into the Phase 2a build branch per Sam — the amendment lands with the code instead of as a separate planning PR.

PR #748, opened 2026-09-01
GEECS-MCP 0.8.1: warm the progress stream at HTTP-server startup (#685)

### LINKED ISSUE #685: GEECS-MCP: warm the progress stream at server startup (first scan after restart shows no counts)
Live finding from the v2 deployment checks (2026-08-24, deployed HTTP service): the `ProgressCache` stream threads start lazily on the first `scan_progress` call. The first check scan was submitted, then `scan_progress` was first called 12 s into the run — the dispatcher subscribed after the start document had passed, so the whole run showed `stream: {available: true}` with no scan number / shot counts. The second scan (warm cache) streamed perfectly (2→…→25/25, totals, exit status).

Fix shape: in `--transport http` mode (a long-lived service), call `progress_stream.get_progress_cache().ensure_started(...)` at server startup with the queue client's addresses, so the stream is warm before any run starts. Stdio sessions can keep the lazy start (short-lived, first call usually precedes the first submit). Small, self-contained; the latch/threading rules in `scans/progress_stream.py` already support this.

skbarber:
Closes #685.

## What + why

Live finding from the v2 deployment checks: the `ProgressCache` consumer threads (document stream + console-text stream) started **lazily on the first `scan_progress` call**. In `--transport http` mode — a long-lived systemd service — the first scan after a restart was typically submitted before any `scan_progress` poll, so the dispatcher subscribed *after* the start document had passed and that whole run showed `stream: {available: true}` with no scan number / shot counts. The second scan streamed fine (warm cache).

Fix, one concern:

- `geecs_mcp/__main__.py` — new `warm_progress_stream()`, called in the `http` branch of `main()` **before** `server.run(...)`. Resolves the queue client via `runtime.get_queue_client()` and starts the cache from its `doc_addr` / `info_addr`. Best-effort: a client that will not build is logged (`WARNING`) and the server still comes up; a missing address (stub client / unconfigured `[qserver]`) latches the cache's own honest `available=false` and logs why. Stdio sessions keep the lazy start (short-lived; first call usually precedes the first submit).
- `geecs_mcp/scans/progress_stream.py` — `start_for_client(client)`: the **one** address resolution (the two `getattr`s that lived inline in `control_tools._stream_snapshot`), now shared by the lazy path and the startup warm-up rather than duplicated. Returns the cache.
- `geecs_mcp/scans/control_tools.py` — `_stream_snapshot` calls `start_for_client` (idempotent no-op after the warm-up); docstring says so.
- `CLAUDE.md` — one sentence on the lazy-vs-warmed rule under the `scan_progress` stream paragraph.
- `CHANGELOG.md` 0.8.1 (patch — a bug fix, no API change).

LOC: runtime code +38 / −6 / +18 across the three modules; tests +133; docs +20.

**Judgment call (cheap to veto):** when `doc_addr` is `None` at startup the warm-up still calls `ensure_started(None, None)` and latches the cache unconfigured (the same picture `scan_progress` would latch a moment later, since the queue client is cached by `runtime` too — a config change already needs a restart per the cache's own docstring). The alternative — skipping the latch so a later lazy call could retry — buys nothing while the client stays cached, so I kept the single code path.

## Tests

`./scripts/check.sh` (lint + GEECS-MCP suite, the way CI runs it): **102 passed, 1 skipped** (was 97 + 1; 5 new):

- `tests/test_entry_point.py` (new, 4): http warms the cache with the client's addresses **before** `run()` is entered (order pinned via a shared event log); stdio does not touch the cache; a queue-client build failure logs and the server still runs; a client without addresses latches `(None, None)`, logs the honest warning, and still serves.
- `tests/test_progress_stream.py` (+1): `start_for_client` reads `doc_addr`/`info_addr` off the client, returns the cache, and resolves a bare object to `(None, None)`.

## Hardware verification

**OWED** (touches the deployed HTTP service, not scan execution): restart `geecs-mcp` on the qserver host with 0.8.1, confirm the startup log line `progress stream warming from document stream <host>:5568`, submit a scan and call `scan_progress` only *after* ~10 s into the run — expected: `stream.scan_number` set and `shots_done` advancing (e.g. `2 → … → 25/25` with `shots_total`) on that **first** run, not just the second. Also expected: the warm-up latches the addresses from `~/.config/geecs_python_api/config.ini` `[qserver]` — if the service's config points at the wrong host the stream reads available-but-empty exactly as before (documented honesty boundary, unchanged).

### DIFF
--- a/GEECS-MCP/CHANGELOG.md
+++ b/GEECS-MCP/CHANGELOG.md
@@ -4,6 +4,22 @@ All notable changes to `geecs-mcp` are documented here.
 
 Format follows [Keep a Changelog](https://keepachangelog.com/en/1.1.0/).
 
+## [0.8.1] - 2026-09-01
+
+### Fixed
+
+- **First scan after an HTTP-service restart streamed no counts**
+  (#685): the `ProgressCache` consumer threads started lazily on the
+  first `scan_progress` call, so a long-lived `--transport http` service
+  whose first poll came after the run's start document showed
+  `stream: {available: true}` with no scan number / shot counts. The
+  HTTP entry point now warms the cache at startup
+  (`__main__.warm_progress_stream`) from the queue client's own
+  `doc_addr` / `info_addr` — the same resolution `scan_progress` uses,
+  now shared as `progress_stream.start_for_client`. Best-effort: a
+  client that will not build or a missing address is logged and the
+  server still comes up. Stdio sessions keep the lazy start.
+
 ## [0.8.0] - 2026-09-01
 
 ### Changed
--- a/GEECS-MCP/CLAUDE.md
+++ b/GEECS-MCP/CLAUDE.md
@@ -192,6 +192,10 @@ geecs_mcp/
   console-text stream's failed-move line as the paused reason
   (surfaced only while actually paused — sticky otherwise).  The
   manager poll stays authoritative; `stream.available=false` names why.
+  The consumer threads start lazily on the first `scan_progress` call
+  (stdio) but the HTTP entry point warms them at startup (#685) — a
+  long-lived service must be consuming before its first start document
+  passes, or that run shows no counts.
 
 ## Testing
 
--- a/GEECS-MCP/CLAUDE.md
+++ b/GEECS-MCP/CLAUDE.md
@@ -192,10 +192,13 @@ geecs_mcp/
   console-text stream's failed-move line as the paused reason
   (surfaced only while actually paused — sticky otherwise).  The
   manager poll stays authoritative; `stream.available=false` names why.
-  The consumer threads start lazily on the first `scan_progress` call
-  (stdio) but the HTTP entry point warms them at startup (#685) — a
-  long-lived service must be consuming before its first start document
-  passes, or that run shows no counts.
+  The HTTP entry point warms the consumer threads at startup (#685) —
+  a long-lived service must be consuming before its first start
+  document passes, or that run shows no counts.  Stdio still starts
+  them lazily on the first `scan_progress` call (owner scope on #685),
+  so a stdio session that submits before its first poll has the same
+  first-run exposure; dropping the transport gate in
+  `__main__.main` is the one-line remedy if that bites.
 
 ## Testing
 
--- a/GEECS-MCP/geecs_mcp/__main__.py
+++ b/GEECS-MCP/geecs_mcp/__main__.py
@@ -24,8 +24,8 @@ def warm_progress_stream() -> None:
 
     HTTP mode only: a long-lived service must be consuming BEFORE its
     first run's start document passes — the lazy start behind
-    ``scan_progress`` (fine for a short-lived stdio session) left the
-    first scan after a service restart with no scan number or counts.
+    ``scan_progress`` (stdio keeps it — see ``_stream_snapshot``) left
+    the first scan after a service restart with no scan number or counts.
     Best-effort like the cache itself: nothing here can stop the server
     from coming up.  The cache logs what it did (consuming from which
     address, or latched unconfigured); the guard below only covers a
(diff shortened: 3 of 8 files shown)


skbarber:

Reviewer ran the suite in the worktree (`102 passed, 1 skipped`) and read the diff through the correctness / redundancy / placement lenses. Four findings, none invalidating the approach. Fix wave `28f58f8c` + docstring nit `ebe1997c`; every fix went back to the **same reviewer**, who confirmed all four dispositions against the actual diff and reported no new findings introduced by the fixes.

| # | Sev | Finding (reviewer's words, condensed) | Disposition |
|---|---|---|---|
| 1 | Medium | The transport gate leaves **stdio** with exactly the #685 failure on every session's first scan: submit-then-poll is the sequence any agent performs regardless of transport, the only runtime caller of `ensure_started` is `_stream_snapshot`, and a stdio process is spawned per osprey session. The new CLAUDE.md sentence codified "lazy is stdio's design". The console (the reference for these threading rules) starts its stream workers eagerly. Options: drop the gate, or warm in `_submit_scan_impl`. | **Waived** for the behavior change — #685's own text and the landing brief say stdio keeps the lazy start (owner scope). **Docs fixed** (`28f58f8c`): CLAUDE.md and the `_stream_snapshot` docstring now state the first-run exposure a stdio session keeps and name the one-line remedy (drop the transport gate in `__main__.main`). **Cheap to veto:** say the word and the gate goes. |
| 2 | Low | `warm_progress_stream` re-did `getattr(client, "doc_addr")` just to choose a log line — the duplication the PR claimed to eliminate; the cache is the only place that knows whether a thread was actually spawned. | **Fixed** (`28f58f8c`): the INFO ("consuming document stream …") and WARNING ("no document-stream address …") lines moved into `ProgressCache.ensure_started` (spawn path / latch path — the idempotent repeat call returns inside the lock before either, so no per-poll spam); `__main__` is one `start_for_client(...)` call inside the boot guard. The log assertion now pins the real cache in `test_progress_stream.py`. Reviewer confirmed. |
| 3 | Low (placement) | `fastmcp` 3.4.7 accepts `FastMCP(..., lifespan=...)` — a transport-agnostic startup hook on the module-level `mcp` in `server.py`; a future "warm the catalog/resolver at startup" feature will want it. | **Waived**: a transport-agnostic hook is finding 1's territory (owner scope). Noted here as the natural home if a general `warm_up()` ever arrives. |
| 4 | Low (test hygiene) | `test_warm_up_failure_never_stops_the_server` forces `runtime.get_queue_client` to raise, but by the seams' contracts it cannot (stub on missing/bad config; config errors swallowed). The `try/except` is defence against a path the layers promise not to take. Also `logging.basicConfig` per `main()` call adds a root handler per test — pre-existing. | **Fixed as a comment** (`28f58f8c`): the test and `warm_progress_stream`'s docstring say it pins the boot guarantee (defence in depth), not a reachable path. `basicConfig`-per-test **waived**: pre-existing, cosmetic. Reviewer confirmed. |

Residual nit from the confirmation pass — the `warm_progress_stream` docstring still called the stdio lazy start "fine" — fixed in `ebe1997c`.

**Checked and cleared by the reviewer:** no existing start-streams-from-a-client helper anywhere (console's `ScanMonitorController` takes addresses explicitly; `qs_client` exposes only the attributes; other `RemoteDispatcher` users are different consumers); the ordering pin in `test_entry_point.py` would fail if the warm-up moved after `run()` or into the stdio branch; the `_stream_snapshot` refactor is behavior-preserving; the WARNING now emitted under the cache lock cannot deadlock; no scan-folder creation, no lab hostnames, version + CHANGELOG travel with the change.

Suite after all fixes (`./scripts/check.sh`): **102 passed, 1 skipped** (baseline 97 + 1; 5 new).

PR #751, opened 2026-09-01
GeecsBluesky 0.70.1: drop asynchronous devices from capture_devices with a loud warning (#702)

### LINKED ISSUE #702: Async capture-eligible camera with empty variable_list escapes the active save-off

skbarber:
Fixes #702 (both shapes in the issue + its comment).

## What

`select_capture_devices` selected capture-owned cameras on devicetype + `save_nonscalar_data` alone, so an **asynchronous (snapshot-role)** camera of a capture devicetype landed in `capture_devices` / the toggle start-doc metadata and got `save_control_only=True` in the config copy — in two broken shapes:

1. **Empty `variable_list`**: `_build_request_detectors` skips it ("no scalars to record") before any device object exists, so `geecs_run_wrapper`'s eager `save="off"` loop never commands it off — the stale-path hazard the active off-write surface (#697) exists to close.
2. **Non-empty `variable_list`**: it got the `save` child (0.68.x) but no `acq_timestamp` column — `CaSnapshotReadable` has no acq machinery — so its capture stack could not be row-joined.

The remedy is the issue's named alternative: **async devices are dropped from `capture_devices` at the selection seam with a loud WARNING** that names the device, its devicetype and the reason (no save-control surface + no acq join column until the snapshot role grows them, #702). Native saving for them is left as-is — literally: `session._configure_saving` gates on `_save_nonscalar_data`, which the snapshot role never sets, so the engine never drove async native saving in the first place. Sync roles (the production case) are unchanged. No full async capture support is built.

Because both execution paths share `resolve_and_apply_capture_toggle` → `select_capture_devices`, one change covers every consumer: the run wrapper's off-write loop and capture-dir mkdir, the start-doc `capture_devices` key, `apply_native_image_save_off`'s `save_control_only` copy, `preflight_capture_liveness`'s roster coverage, and the daemon's explicit-list targeting.

## Why not an event-schema bump

`capture_devices` still means exactly "the devices the capture daemon owns for this run", and the daemon does the same thing with it. What narrowed is the engine's *selection policy*, not the key's shape or meaning — EVENT_SCHEMA.md's row and its `geecs_scalar_headers` note are truth-ups, not a contract change. Flagging in case the maintainer reads the skill's "semantics changes bump" rule more strictly.

## LOC (one concern; 264+/26−)

| piece | lines |
|---|---|
| `scan_request_runner.py` — the seam filter + warning + docstring | 35+/6− |
| runner-suite tests: async-cameras fixture + 3 end-to-end pins (replacing `test_native_image_save_off_wires_snapshot_branch`, which monkeypatched the seam into a now-unreachable state) | 171+/20− |
| `test_native_image_save.py` — seam unit pin (both shapes + missing-`synchronous`-key default) | 38+ |
| EVENT_SCHEMA.md / capture FORMAT.md / CLAUDE.md truth-ups | 20+ |
| CHANGELOG + `poetry version patch` (0.70.0 → 0.70.1) | 26+ |

- **`CaSnapshotReadable`'s `save_control_only` branch is left in place.** It is now unreachable from the runner (the seam never sets it for async roles) but is harmless, pinned in `test_ca_devices.py`, and removing it would widen this PR beyond one concern. Happy to delete it here or in a follow-up.
- The warning fires on every scan whose save set has such a camera (dual-write default included), not only toggle-off — the issue asked for loud, and a snapshot-role camera marked `images: true` is a save-set misconfiguration worth seeing each time.
- A missing `synchronous` key counts as asynchronous, the same default `_build_request_detectors` applies.

## Tests

- GeecsBluesky (changed-scope `scripts/check.sh`, lint all hooks Passed): **657 passed, 1 skipped, 4 deselected** (pre-existing integration/hardware layers).
- New: 4 pinning tests — `test_async_camera_without_scalars_is_dropped_from_capture` (shape 1), `test_async_camera_with_scalars_is_a_plain_snapshot_not_capture_owned` (shape 2), `test_toggle_off_with_only_async_eligible_camera_is_inert_and_unrefused` (sole eligible is async: no preflight, no `capture_devices` key, `native_image_save: false` stays visible), `test_select_capture_devices_drops_asynchronous_roles_loudly` (seam). All 4 verified to **fail on the unfixed code** (temporary revert) and pass with the fix. The three runner pins route the *real* seam through a fake devicetype provider, so the policy is exercised end to end hermetically.
- `check.sh --all` was not runnable in this fresh worktree (no envs for the other 11 packages); the diff is GeecsBluesky-only so CI's other suites are unaffected by construction — CI is the authoritative signal.

## Hardware verification

**OWED** — not on the lab network this session. To verify on the worker host (needs a save set containing a capture-devicetype camera in a snapshot role, e.g. `synchronous: false` / `role: snapshot`, plus one sync capture camera):

1. Toggle-off scan (`native_image_save: false`, daemon running): scan.log shows `<AsyncCam> is NOT capture-owned: its devicetype (<type>) is capture-eligible, but asynchronous ... (#702)`; run metadata `capture_devices` lists only the sync camera; `scans/ScanNNN/<AsyncCam>/` is NOT created by the wrapper and the daemon writes no stack for it; the sync camera's stack + `acq_timestamp` column are unchanged from 0.70.0; the async camera's native files behave exactly as before this PR (whatever its own save flag does — the engine does not touch it).
2. Same save set with the async camera as the ONLY capture-devicetype device + toggle-off: scan proceeds with no daemon-liveness refusal (nothing to capture), the existing `no capture-eligible devices resolved` warning appears alongside the #702 warning, and md carries `native_image_save: false` with no `capture_devices` key.
3. Dual-write default (toggle unset): the #702 warning still appears, the sync camera is captured as before.

## Coordination

Another in-flight PR (the #727 doc-stream note) also bumps GeecsBluesky; whichever merges second rebases and re-runs `poetry version patch` + CHANGELOG ordering.


skbarber:
## Adversarial review (fresh-context subagent, per `/land`) — report + dispositions

Reviewer read the diff against the runner, `session.py`, `devices/ca/snapshot.py`, `plans/run_wrapper.py`, `capture/daemon.py`, `qs_client/submit_preflight.py`, GEECS-Schemas `save_set.py`, the Planning arc spec and the touched tests; ran the two suites (111 passed); confirmed the seam is the single choke point for both execution paths and that the four new pins fail on the old selection. **Redundancy: none found** (checked `capture/discovery.py`, `preflight.py:122`, `submit_preflight.py:292`, `_build_request_detectors`' own `synchronous` check).

### 1. Medium — the loud warning is gated on the DB and on devicetype, but the misconfiguration it names is neither
> `role: snapshot` + `images: true` is a universal no-op (engine never natively saved a snapshot device; now the daemon never captures it either), yet the warning fires only after the devicetype lookup: a DB blip (fail-open `{}`) or a non-capture devicetype (HASO, scope) in a snapshot role is never warned. Better home: the DB-free role seam `save_set_to_devices_config` / `_build_request_detectors`, which `submit_preflight` also runs so console/MCP see it pre-submit. `SaveSetEntry.images` promises images with no role caveat.

**Disposition: WAIVED here, filed as #754.** The #702 warning is deliberately about *capture-ownership*, which is DB-gated by design (fail-open: on a blip nothing is capture-owned). The broader "images: true on ANY snapshot-role entry is a silent no-op, any devicetype" is pre-existing (the snapshot role never had `save_images`) and a schema/preflight concern — a second concern for this PR. Reviewer accepted the waiver on condition of the follow-up issue: **#754** names the DB-free home and the `SaveSetEntry.images` description gap. Residual for the record (reviewer): in dual-write mode on a DB blip neither this warning nor the toggle-off "no capture-eligible devices" warning fires, so the async-camera misconfiguration is silent there until #754 lands.

### 2. Medium — a formerly working shape becomes inert, and the PR/CHANGELOG described it as "left as-is"
> The issue-comment workaround (async camera listing `acq_timestamp` explicitly → joinable stack) and the 0.68.x save child + eager off-write on an async camera with scalars both go away; "native saving is left as-is" reads as "still saved" when it means "never commanded, never suppressed". The warning also names a "synchronous role" that is not in the schema vocabulary. Arguably `minor`, not `patch`.

**Disposition: FIXED via option (b) in `9ec6f3a9`** (reviewer CONFIRMED): warning, runner docstring, CHANGELOG, FORMAT.md and EVENT_SCHEMA.md now say plainly that `images: true` on a snapshot-role entry is ignored and the device's own save flag is neither commanded nor suppressed; the CHANGELOG states the trade explicitly; the action reads "Remove `role: snapshot` (legacy: `synchronous: false`)". Reviewer's follow-up wording nit ("so this scan writes no per-shot images" was a non sequitur — an out-of-band-left-on flag DOES write natively) **FIXED in `e3f5eb1a`**: "so the engine records no per-shot images for it". Option (a) — exempting async entries that list `acq_timestamp` — **WAIVED**: that is partial async capture support, which this PR's scope excludes (the other remedy #702 names). **Patch vs minor: left to the maintainer** — the brief and the issue treat this as a bug fix of a broken path; bumping to 0.71.0 is a one-liner if you read it as a behaviour change.

### 3. Low — the arc spec still listed #702 as an open gap
**Disposition: FIXED in `9ec6f3a9`** (reviewer CONFIRMED): `Planning/data_capture/01_central_pva_capture_scope.md` ~301 and ~444 now say the gap is closed by #702 / 0.70.1 and name the unbuilt remedy.

### Nit (no action)
`_select_through_real_seam` monkeypatches the module attribute because `resolve_and_apply_capture_toggle` has no `provider` parameter; threading one through would let the runner pins call the real seam without monkeypatching. **WAIVED** — test-only plumbing, would widen the diff.

No new findings from the fix commits. No undispositioned findings remain.

PR #757, opened 2026-09-02
GEECS-DataPortal 0.15.3: union-frame responses never immutable (s-file grows after the run)

skbarber:
## What

Responses computed over the **union frame** — `/api/run/{uid}/columns`, `frame`, `binned`, `filter-count`, `bin-images` and `bin-image.png` — are no longer served `Cache-Control: public, max-age=31536000, immutable` for completed runs. They are now `no-cache` — a plain re-fetch, no validator is emitted (an ETag over the s-file stat is the follow-up). The per-shot `image.png` and `plot.png`, which read the Tiled event table alone, keep their immutable headers unchanged.

## Why

The "completed run is immutable" premise holds for the event table but not for the s-file half of the union: ScanAnalysis **appends** its columns to `analysis/sN.txt` after the scan — hours later when it is re-run by hand. A browser that fetched the column list before analysis stayed pinned to the pre-analysis shape until the next portal release (the page keys `/api` fetches by package version only).

Live incident, 2026-09-01 (Undulator): scan 32 showed 7 columns while scan 33 showed 30. On disk both s-files had 30 columns by then — scan 32's FROG analysis ran at 19:19–19:54, three hours after the scan ended at 16:56 and after scan 33's at 19:32–19:43. The scanner-written `ScanDataScanNNN.txt` was identical (7 columns) for both.

## Changes (one concern, ~60 LOC net)

- `app.py`: module constant `_UNION_HEADERS` with the rationale; the four JSON endpoints and `bin-image.png` use it. `bin-image.png`'s per-member `cacheable` accumulation is removed — it only ever decided between `no-cache` and `immutable`, and the answer is now always `no-cache` (the per-shot `image.png` keeps its own `result.cacheable` gate). `_png_headers` and `_portal_version` docstrings updated to match.
- Tests: `test_completed_run_json_is_immutable` → `test_completed_run_json_never_immutable` (columns/frame/binned all `no-cache`); the bin-images JSON assertion and the bin-image ordinal-vs-joined test now pin `no-cache` on both devices.
- 0.15.2 → 0.15.3 + CHANGELOG.

**Judgment call, cheap to veto:** plain `no-cache` rather than an ETag keyed on the s-file's size/mtime. The server already re-reads the s-file per request (see `_union`), so the only thing an ETag would save is transfer bytes on unchanged data plus the bin-image `nanmean` recompute on revisits (the pixel cache stays warm). Happy to add the ETag as a follow-up if the Images tab feels slower on revisits.

## Tests

GEECS-DataPortal: 179 passed, 6 skipped (the `analysis`-extra tests), via `scripts/check.sh`.

## Hardware verification

None needed — no scan execution or device surface touched. Deploy check on the portal host after merge: `curl -sI http://<portal>:8200/api/run/<completed-uid>/columns | grep -i cache-control` should print `no-cache`; a hard reload of scan 32 (2026-09-01) should list 30 s-file-provenance columns.

https://claude.ai/code/session_01CxJgmVg7DB6xeZji9Hniw5

### DIFF
--- a/GEECS-DataPortal/CHANGELOG.md
+++ b/GEECS-DataPortal/CHANGELOG.md
@@ -15,9 +15,10 @@ Format follows [Keep a Changelog](https://keepachangelog.com/en/1.1.0/).
   when re-run by hand), so a browser that had fetched the column list
   before analysis stayed pinned to the pre-analysis shape for a year
   (scan 32 showing 7 columns while scan 33 showed 30, 2026-09-01).
-  These responses now always revalidate (`no-cache`); the per-shot
-  `image.png` and `plot.png`, which read the event table alone, keep
-  their immutable headers.
+  These responses (and `filter-count`, which had no header at all)
+  are now `no-cache` — a plain re-fetch, no validator is emitted; the
+  per-shot `image.png` and `plot.png`, which read the event table
+  alone, keep their immutable headers.
 
 ## [0.15.2] - 2026-09-01
 
--- a/GEECS-DataPortal/CHANGELOG.md
+++ b/GEECS-DataPortal/CHANGELOG.md
@@ -3,6 +3,22 @@
 All notable changes to this package will be documented here.
 Format follows [Keep a Changelog](https://keepachangelog.com/en/1.1.0/).
 
+## [0.15.3] - 2026-09-01
+
+### Fixed
+
+- Responses computed over the union frame — `/api/run/{uid}/columns`,
+  `frame`, `binned`, `bin-images` and `bin-image.png` — are no longer
+  served `immutable` for completed runs. The event table is frozen once
+  the run stops, but the s-file half of the union is not: ScanAnalysis
+  appends its columns to `analysis/sN.txt` after the scan (hours later
+  when re-run by hand), so a browser that had fetched the column list
+  before analysis stayed pinned to the pre-analysis shape for a year
+  (scan 32 showing 7 columns while scan 33 showed 30, 2026-09-01).
+  These responses now always revalidate (`no-cache`); the per-shot
+  `image.png` and `plot.png`, which read the event table alone, keep
+  their immutable headers.
+
 ## [0.15.2] - 2026-09-01
 
 ### Fixed
--- a/GEECS-DataPortal/CLAUDE.md
+++ b/GEECS-DataPortal/CLAUDE.md
@@ -61,10 +61,16 @@ this package; the architecture rules below are its distillation.
   executes server-side, and its prototype-pollution guard lives with
   it: cosmetic asks should land there, not as new per-knob code):
   a link IS the analysis, which is also the multi-user story
-  (statelessness; shared caches are a feature).  Every `/api` fetch
-  additionally carries `v=<portal version>` — completed-run responses
-  cache immutable, and the version key rolls browser caches over when
-  a release changes the payload shape.
+  (statelessness; shared caches are a feature).  Cache policy is
+  by INPUT, not by URL: the per-shot `image.png` and `plot.png` read
+  the event table alone and cache immutable once the run has a stop
+  doc; everything over the union frame (`columns`, `frame`, `binned`,
+  `filter-count`, `bin-images`, `bin-image.png`) is `no-cache`
+  (`_UNION_HEADERS`), because the s-file half of the union GROWS after
+  the run — ScanAnalysis appends its columns hours later (the
+  7-vs-30-columns incident, 2026-09-01).  Every `/api` fetch and
+  bin-image card still carries `v=<portal version>` so an upgrade
+  that changes a payload shape rolls over whatever a cache kept.
 - **Prefix-agnostic URLs.**  The portal works at root and under any
   reverse-proxy mount (OSPREY panel tabs).  Templates never write a
   root-absolute portal URL directly: every href/action/src carries the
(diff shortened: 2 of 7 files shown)


skbarber:
## Adversarial review (fresh-context subagent) — report + disposition

Reviewer's report, verbatim, at commit b4bad34e:

> Verified against the worktree at `b4bad34e`. The core change is right: `scan_frame` re-reads `analysis/sN.txt` on every request with no server-side memo (`scan_frame.py:144`, `data/sfile.py` has no cache), `CachingScanCatalog.load_run` caches only the Tiled `RunDetail` (`cache.py:91`), and the console does not consume `scan_frame` at all, so there is no second copy of this bug elsewhere in the repo. The new tests would fail against the old code.
>
> **1. Medium — the package contract file and the page JS still document the old policy.** `GEECS-DataPortal/CLAUDE.md` L64–67, L143–144, L183–184 ("Adding an analysis tab" step 2: "serve with the completed-run cache headers"), L194 (step 5: "immutable headers on completed runs"), and `run.html:519–520`. Concrete failure: the next contributor following the checklist builds a new union-frame endpoint with `_png_headers` and pins it immutable — reintroducing the 7-vs-30-columns incident for the new tab.
>
> **2. Low — `test_running_run_serves_no_cache` is now tautological.** Both endpoints it hits return `_UNION_HEADERS` unconditionally. The one endpoint still gated on completion **and** `resolved.cacheable` — `image.png` — has no running-run test at all. Retarget or delete.
>
> **3. Low — `_portal_version` docstring rests on a link that does not carry the key** (the per-shot `` at `run.html:241` has no `v`), and "revalidate" overstates what `no-cache` does here — no ETag/Last-Modified is emitted, so it is a full re-fetch and a full bin `nanmean` recompute per revisit, not a 304. `filter-count` returns a bare dict with no `Cache-Control` at all — the one union-frame endpoint the "four JSON endpoints" list omits. Follow-up note: the natural validator is the s-file `(mtime, size)`, which only `scan_frame` knows — `ProvenancedFrame` in GEECS-Data-Utils would need to expose it, so that piece lands below the portal.
>
> **Redundancy / placement:** no existing implementation elsewhere (checked `GEECS-Console/geecs_console`, `docs/`, `Planning/data_portal/`); the `cacheable` field on `ShotArray`/`ShotImage` still has a live consumer in `image.png`. `{"Cache-Control": "no-cache"}` is now spelled as a literal in three places plus `_UNION_HEADERS` — a `_NO_CACHE` alias would leave one place to change. Header policy belongs in the view layer, which is where it is.

### Disposition (04cbc663)

| # | Finding | Disposition |
|---|---|---|
| 1 | CLAUDE.md + run.html still describe the immutable policy | **Fixed** — all four CLAUDE.md passages and the `api()` comment rewritten to the by-input rule: `_UNION_HEADERS` for anything over the union frame, `_png_headers` only for event-table-only responses |
| 2 | Running-run bin test tautological; `image.png` gate untested | **Fixed** — retargeted as `test_shot_image_cache_gates_on_completion_and_join`: running → `no-cache`, completed timestamp-joined → `immutable`, completed ordinal (cam2) → `no-cache` |
| 3a | `_portal_version` docstring leans on a link without the key | **Fixed** — docstring no longer cites the per-shot link; adding `v=` to the `image.png` src is **waived** (pre-existing, immutable-by-input, not this PR's concern) |
| 3b | "revalidate" overstates `no-cache` | **Fixed** — `_UNION_HEADERS` comment and CHANGELOG now say re-fetch, no validator; s-file-stat ETag named as the follow-up, with the reviewer's placement note (below the portal, in `ProvenancedFrame`) |
| 3c | `filter-count` had no `Cache-Control` | **Fixed** — returns `JSONResponse` with `_UNION_HEADERS`; pinned in `test_completed_run_json_never_immutable` |
| R | `_NO_CACHE` alias for the three literals | **Waived** — cosmetic; folds naturally into the ETag follow-up, which is the change that would touch all of them |

Tests after fixes: GEECS-DataPortal 179 passed, 6 skipped (`analysis`-extra), `scripts/check.sh` green. Reviewer's confirm/deny on the fix commit follows in the next comment.

skbarber:
## Reviewer's confirmation of the fix commit (same fresh-context subagent, from `git show 04cbc663`)

> Verified from `git show 04cbc663` in the worktree, and by running the affected tests (`pytest tests/test_resources.py tests/test_app.py -k "cache or immutable or filter_count"` → 7 passed, 1 skipped).
>
> **Finding 1 (Medium — contract docs stale): Resolved.** All four `CLAUDE.md` sites are rewritten to the by-input policy; the bin-image.png contract line drops the "ordinal member downgrades" distinction; checklist step 2 says `_UNION_HEADERS` (never `_png_headers`) with the reason; step 5 says `no-cache` even on completed runs. `run.html` comment matches. The reintroduction path through the checklist is closed.
>
> **Finding 2 (Low — tautological running-run test): Resolved.** `test_shot_image_cache_gates_on_completion_and_join` exercises the only endpoint that still branches. I checked that the assertion is a real gate: the fixture's `cam` has the `acq_timestamp` dataset and timestamp-named files, so the `immutable` branch is reached via `_png_headers` × `result.cacheable`, and `cam2` has no acq column so it takes the ordinal branch. Each assertion would fail if its branch regressed.
>
> **Finding 3 (Low — docstring/wording, filter-count header): Resolved.** One residue outside the tree: the PR body still says "They now always revalidate" — worth a one-word edit to "re-fetch". Nothing in the code or docs is wrong.
>
> **Waived items (`_NO_CACHE` alias, `v=` on the per-shot `image.png` src): agreed.** No new findings from the fix commit.

PR body edited per the residue note. All findings dispositioned; nothing outstanding.

PR #758, opened 2026-09-02
Drop trigger-profile variants: GEECS-Schemas 0.15.0 (TriggerProfile v2, ScanRequest v3), GeecsBluesky 0.72.0, GEECS-Console 0.27.0

skbarber:
Follows #734/#755; the pre-2b cleanup Sam asked for. One concern across three packages: the trigger variant goes away.

## What

The schemas had two ways to say "this operating condition": a separate trigger-profile file, or a named *variant* overlaid inside one profile (`TriggerProfile.variants` + `CaptureSettings.trigger_variant`). Nobody adopted variants — every experiment kept one file per condition (`HTU-Normal`, `HTU-LaserOFF`, `HTU-NoGas`); the only corpus file with a `variants` key has it empty. So the variant cost a field on every scan form, a console combo, a validator, an adapter parameter, and a whole editor layer, for nothing. Deleted end to end:

- **GEECS-Schemas 0.15.0** — `TriggerVariant`, `TriggerProfile.variants`, the `variant=` parameter on `writes_for`/`defines_state`, `convert.merge_trigger_variant`, `CaptureSettings.trigger_variant` + its needs-a-profile validator. Per the README versioning policy a removed field is a migration: **TriggerProfile v2** (before-validator drops an empty v1 `variants` block, refuses a populated one with the remedy, normalizes `schema_version` ≤ 1 → 2) and **ScanRequest v3** (the existing lifting validator also drops an unset `trigger_variant` — flat v1 or inside `capture` — refuses a set one, normalizes ≤ 2 → 3). Artifact, reference page, and goldens regenerated.
- **GeecsBluesky 0.72.0** — `trigger_writes_from_profile(profile)` loses the parameter; the plan preamble and `validate_scan_request` stop reading it.
- **GEECS-Console 0.27.0** — the R1 variant combo (+ `.ui`), `ConsoleFormState.trigger_variant`, `ConsoleConfigs.trigger_variants`, the tooltip, and the shot-control editor's variant layer (layer combo, Add/Remove variant, per-layer description, `BASE_LAYER`, the layer handlers). The editor edits one profile: description + state write tree. This is the "console gets simpler" acceptance test the schema refactor set for itself.
- Planning doc: one amendment line (CaptureSettings is six fields; 2b builds on that).

**Corpus check:** both real files that carry the old keys lift cleanly — `HTU-test.yaml` (`variants: {}`) → TriggerProfile v2 with no `variants`; the July preset `basic test.yaml` (`trigger_variant: null`) → ScanRequest v3.

## Per-concern LOC

| Commit | Files | +/− |
|---|---|---|
| GEECS-Schemas (schema, converter, docgen, tests, goldens, artifact, reference, README, overview) | 19 | +273 / −410 |
| GeecsBluesky (adapter, plan call site, tests) | 6 | +28 / −84 |
| GEECS-Console (builder, main window + `.ui`, services, editor + `.ui`, tests) + planning note | 17 | +50 / −354 |

## Tests

- GEECS-Schemas: **273 passed** (6 integration-marked deselected), run the CI way from the repo root
- GeecsBluesky: trigger-touching suites (runner, validate, plan, listings) **143 passed, 1 skipped**
- GEECS-Console: **783 passed** (full suite, offscreen)
- GEECS-MCP / Data-Utils / DataPortal: no references, untouched (grep-verified)

## Compatibility

Old documents → new code: fine forever (both lifting validators). New documents → old code: a v3 request with no `trigger_variant` key validates on an old worker too (the field was optional), so **no deploy-order edge**; a v2 TriggerProfile likewise loads on an old reader. Console and worker can upgrade in either order.

## Hardware verification

Not owed: no scan-execution semantics change (the only production trigger path is "profile named → its writes", untouched). One console noscan after deploy is a nice-to-have, not a gate.

### DIFF
--- a/GEECS-Console/CHANGELOG.md
+++ b/GEECS-Console/CHANGELOG.md
@@ -4,6 +4,20 @@ All notable changes to GEECS-Console are documented here.  Format follows
 [Keep a Changelog](https://keepachangelog.com/en/1.1.0/); versioning is
 semantic.
 
+## [0.27.0] - 2026-09-02
+
+### Removed
+
+- **Trigger-profile variants** everywhere in the console (GEECS-Schemas
+  0.15.0 removed them): the R1 trigger-variant combo, `ConsoleFormState
+  .trigger_variant`, `ConsoleConfigs.trigger_variants`, and the shot-control
+  editor's variant layer (the layer combo, Add/Remove variant, per-layer
+  description). The editor now edits one profile: its description and its
+  state write tree. One profile file per operating condition, picked by
+  name in the trigger-profile combo, is the only shape; older presets with
+  `trigger_variant: null` load unchanged through the schema's lifting
+  validator.
+
 ## [0.26.0] - 2026-09-01
 
 ### Changed
--- a/GEECS-Console/CLAUDE.md
+++ b/GEECS-Console/CLAUDE.md
@@ -12,8 +12,8 @@ One main window, menu bar (Ops / Actions / Editors / Preferences / Help),
 status bar (gateway addr, configs path, version).  Object names in the `.ui`
 are prefixed by region (`r3_radio_1d`, `r5_start_button`, …).
 
-- **R1 session bar** — experiment combo, rep-rate field, trigger-profile +
-  variant combos, gateway/tiled/db health chips.
+- **R1 session bar** — experiment combo, rep-rate field, trigger-profile
+  combo, gateway/tiled/db health chips.
 - **R2 save sets** — available/selected lists, Add/Remove, union preview
   line ("union: N devices"), role-conflict/reference hint line.
 - **R3 scan form** — mode radios (No-scan / 1D / Grid / Optimization /
--- a/GEECS-Console/geecs_console/app/main_window.py
+++ b/GEECS-Console/geecs_console/app/main_window.py
@@ -378,9 +378,6 @@ class MainWindow(QMainWindow):
         self.trigger_profile_combo: QComboBox = self._child(
             QComboBox, "r1_trigger_profile_combo"
         )
-        self.trigger_variant_combo: QComboBox = self._child(
-            QComboBox, "r1_trigger_variant_combo"
-        )
         self.gateway_chip: QLabel = self._child(QLabel, "r1_gateway_chip")
         self.tiled_chip: QLabel = self._child(QLabel, "r1_tiled_chip")
         self.db_chip: QLabel = self._child(QLabel, "r1_db_chip")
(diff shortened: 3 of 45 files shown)


skbarber:

The reviewer exercised both lifting validators against every document shape (flat v1 preset with `trigger_variant: null`, v2 with `capture.trigger_variant: null`, quoted stamps, a future stamp with a nested null, `variants: {}` / `null` / populated, a future TriggerProfile v3) plus the two real corpus files that carry the legacy keys, ran the hermetic suites (Schemas incl. the no-drift artifact test, Bluesky runner/validate/plan, Console editor/tooltips/store/builder/actions-menu/main-window), grep-verified no live consumer of any removed name, and checked the `.ui` edits leave no orphaned widgets. Findings, then dispositions (fix commits **ecb8fcf9** + **cdf68709**, each fix confirmed by the same reviewer):

**P2 — The regenerated reference contradicted itself: the docgen examples still stamped the pre-bump `schema_version` (2 / 1) under table rows saying "Leave at 3 / 2", and the docgen test could not see it because the lifting validators normalize a stale stamp up.** → **FIXED**: examples stamp 3 and 2, reference regenerated; new pin `test_examples_stamp_the_current_schema_version` compares each example's raw stamp to the model default across the registry, so any future bump that forgets an example fails.

**P3 — Doc residue** (README "each kind versions on its own" bullet, Console CLAUDE.md R1 line, the `.ui` description tooltip). → **FIXED**, all three.

**P3 — Two refusal messages mis-described unlikely inputs.** (a) A flat `trigger_variant` beside a `capture` block hit the "mixed layout" branch with an empty field list, masking the v3 remedy. → **FIXED**: the removed field is judged on its own terms before the flat-vs-capture check (null → dropped, set → remedy); pinned. (b) A v1 variant with *no writes* — exactly what the retired editor's "Add variant" saved when never filled in — was refused with the split-file remedy. → **FIXED**: empty-states variants drop as no-ops, only populated ones are refused; pinned.

**P3 — The stale-version normalisation was copy-pasted into both validators.** → **FIXED**: one `stale_schema_version(data, current)` helper on `_base.py` (quoted-digit handling in one place), used by both.

**Verifier caveat on (b):** a non-mapping `variants` (e.g. a list) escaped as `AttributeError` instead of a validation error. → **FIXED** (cdf68709): refused as `ValueError` → `ValidationError`; pinned.

Refuted by the reviewer (not findings): the versioning policy is applied exactly as the README prescribes; mixed-deployment compatibility holds in both directions; resolver/store legacy detection keys on `schema_version`, not `variants`; the `.ui` edits are clean.

No undispositioned findings. Schemas hermetic suite: 277 passed.

PR #762, opened 2026-09-02
GEECS-Schemas 0.16.0: schema export registry (Phase 2b-i)

skbarber:
Phase 2b-i of `Planning/schema_refactor/00_overview.md`: the tooling groundwork the three named plans (2b-ii) need, split out so that PR is plans only.

## What

`geecs_schemas.schema_export` becomes a **registry**: `EXPORTED_SCHEMAS` (artifact name → model) renders one committed `docs/geecs_schemas/<name>.schema.json` per entry (`json_schema(model)`, `render_artifact(name)`, `write_artifacts`). The no-drift guard is parametrized over the registry, and a new **orphan check** fails CI if a `*.schema.json` is published whose model left the registry (it would otherwise stay fetched by URL while silently drifting). `SCHEMA_ARTIFACT` and `scan_request_json_schema()` stay as the ScanRequest aliases the worker's plan annotation (`GeecsBluesky/tests/test_qserver_startup.py`) and OSPREY's profile point at. The regenerator script is renamed `tests/generate_schema_artifacts.py`.

Today the registry has one entry, `scan_request`; the committed artifact is **byte-identical** (regenerated, no diff). 2b-ii adds the three plan parameter models as one line each.

## Scope

Tooling only — no schema, engine, or client change. GEECS-Schemas minor (0.16.0) for the new public API; README versioning-policy paragraph and the planning doc updated. No other package touched.

## Tests

- GEECS-Schemas: **280 passed** (6 integration-marked deselected), the CI way from the repo root — includes the parametrized guard, the orphan check, the alias pin, and a per-entry dialect/title check.
- GeecsBluesky `tests/test_qserver_startup.py` (imports `SCHEMA_ARTIFACT`): skips locally (no queueserver extra in this env); it runs in CI, which installs the extra — the alias it reads is pinned by `test_scan_request_stays_the_named_entry` regardless.

## Hardware verification

Not applicable (no runtime surface).


skbarber:

Reviewer read the diff against the whole repo, grep-verified every consumer of the old names (worker annotation test, plan docstring, skills, docs, workflows), reproduced the one runtime gap, argued both ways on the two-registries question, and ran the schema suite. Findings and what happened (fix commits **f6e36e8c** + **7cee5f29**, each confirmed by the same reviewer, who also mutation-checked the new test):

**P2 — `write_artifacts(root)` / `--root` failed on a fresh root (no parent creation), and the write path had no test.** → **FIXED**: `write_artifact` creates parents (docs tree, not a scan folder); `--root <empty tmpdir>` verified; `test_write_artifacts_creates_the_tree_under_a_fresh_root` fails without the `mkdir` line.

**P3 — `-o/--output` removed without a changelog line.** → **FIXED** (CHANGELOG "Changed"); `-o` not re-added — no consumer anywhere.

**P3 — orphan check hardcoded the directory and its docstring overclaimed.** → **FIXED**: globs `ARTIFACT_DIR`, docstring says "in the artifact directory".

**P3 — round-trip pin weakened to a title check (`__name__` is not a contract).** → **FIXED**: asserts `json.loads(text) == json_schema(model)`.

**P3 — `name="scan_request"` defaults invited writing the wrong artifact to a plan's path.** → **FIXED**: `name` required on `render_artifact` / `write_artifact`.

**P3 — two registries is right (config kinds vs published contracts), but the planning doc budgeted a third.** → **FIXED**: planning wording (docgen iterates `EXPORTED_SCHEMAS` for the plan-models page); new pin `EXPORTED_SCHEMAS["scan_request"] is SCHEMA_REGISTRY["scan_request"]`.

Refuted: every old-name consumer still resolves; artifact byte-identical; placement correct (pydantic-only, regenerator mirrors the docgen one).

No undispositioned findings. Schemas suite: 282 passed.

PR #764, opened 2026-09-02
GEECS-Console 0.28.0: R8 queue panel (running / waiting / finished, Clear queue) + BackgroundResult GUI hop

skbarber:
## What

**R8 queue panel** in the console's right column, below the NOW panel: what the RE Manager holds. bluesky-queueserver ships the queue as data (`queue_get` / `history_get`) and no operator view of it; the console's now panel shows the running scan but nothing behind it. The new panel shows:

- a summary line: `Running · 2 waiting` / `Queue empty` / `manager unreachable` / `unavailable (<error>)`
- a read-only table (State / Item / By / Detail): the running item, the waiting items front-first (any client's — the MCP's, a notebook's), then the last 10 finished items newest-first with the manager's `exit_status`, finish clock and first message line
- **Clear queue** (enabled only with waiting items; confirmation modal): the recovery verb for the failed-item-at-front trap, without going through a Start attempt

Item summaries (`summarize_item` / `summarize_request`) are pure functions tolerant of every ScanRequest schema version the history may hold (v1 flat `shots_per_step`, v2+ `capture`, step / grid / noscan / optimize / background, action plans, unknown plans).

**Refresh policy**: no new poll. The existing 1 s manager status poll (`_on_queue_status`) fans each snapshot into the panel, which refetches when the queue-shaped fields change (`items_in_queue` / `running_item_uid` / `re_state`) or every 5 s as a fallback — three bounded 0MQ reads on a `BackgroundResult` thread, one in flight. A failed fetch renders "unavailable", never an empty queue.

## BackgroundResult fix (bundled, same package — the panel exposed it)

Adding a fetch per status snapshot made the offscreen suite segfault at random positions (native stack: `QWidget::ensurePolished` walking a dangling child). Bisected to: a daemon-thread emission aimed at a consumer QObject races that consumer's destruction — keeping the controller alive forever removed the crash; slot body, payload type and the `dispose()` disconnect were irrelevant. `BackgroundResult` now hops the result onto the GUI thread through its own queued signal (receiver = the worker, held in a module-level in-flight set until the hop lands, released one event-loop turn later) and emits `result_ready` there. Consumer connections are unchanged; delivery takes two event-loop turns instead of one (one test that assumed the startup idle probe had already landed now waits for it). `HealthPoller` keeps its pre-existing emit-from-thread shape; noted in CLAUDE.md as the next candidate if it ever bites.

## LOC (≈1115 added / 18 removed)

| concern | files | LOC |
|---|---|---|
| queue panel controller + summarizers | `app/queue_panel.py` | ~430 |
| tests for it | `tests/test_queue_panel.py` | ~320 |
| window wiring, `.ui` R8 group, QSS table styling, tooltips | `main_window.py`, `main_window.ui`, `style.qss`, `tooltips.py` | ~130 |
| `BackgroundResult` GUI hop | `services/background.py` | ~50 |
| test fakes / one timing wait | `tests/test_main_window.py`, `tests/test_tooltips.py` | ~20 |
| CLAUDE.md + CHANGELOG + version | | ~90 |

- History depth 10 rows, fallback refresh 5 s — both module constants.
- Clear queue lives in the panel (not a menu item); single-item Remove is deferred because `qs_client` exposes no `item_remove` verb (would touch GeecsBluesky).
- The R6/R8 vertical stretch is 3:2; the log tail shrinks accordingly on small windows.

## Tests

GEECS-Console suite (main-checkout console env, `PYTHONPATH` on the worktree): **806 passed** ×3 runs (master baseline: 783). New: `tests/test_queue_panel.py` — 21 tests (summarizers, row ordering/limits, refresh gating on status change, unreachable / fetch-failure rendering, clear declined / confirmed / failed, dispose). Pre-commit clean on all changed files. `scripts/check.sh` could not run inside the worktree (no per-package venv there) — CI is the authoritative run.

## Hardware verification

**OWED** (Sam testing live now): with the worker up, submit a scan from the console and confirm the panel shows it as `running` with the console's user, then as `completed` in history with a finish time; queue a second scan while one runs (e.g. via the MCP) and confirm `1 waiting`; let a scan fail and confirm it appears at the queue front + Clear queue removes it and reports "queue cleared".

https://claude.ai/code/session_01UpBh2v5DycM2Twh3aUuoZT

### DIFF
--- a/GEECS-Console/CHANGELOG.md
+++ b/GEECS-Console/CHANGELOG.md
@@ -17,8 +17,12 @@ semantic.
   `app/queue_panel.py::QueuePanelController`; refreshed from the existing
   1 s manager status poll when the queue-shaped fields change (plus a 5 s
   fallback), three bounded 0MQ reads on a background thread, one in
-  flight. Item summaries are pure functions tolerant of every request
-  schema version the manager's history may hold.
+  flight; a change arriving mid-fetch is fetched as soon as that fetch
+  lands, and the fallback tick re-reads the queue only (the manager's
+  history is unbounded and can only change with the status key). Item
+  summaries are pure functions tolerant of every request schema version
+  the manager's history may hold, including the named plans' keyword
+  shape (GeecsBluesky ≥ 0.73.0).
 
 ### Fixed
 
--- a/GEECS-Console/CHANGELOG.md
+++ b/GEECS-Console/CHANGELOG.md
@@ -4,6 +4,33 @@ All notable changes to GEECS-Console are documented here.  Format follows
 [Keep a Changelog](https://keepachangelog.com/en/1.1.0/); versioning is
 semantic.
 
+## [0.28.0] - 2026-09-02
+
+### Added
+
+- **R8 queue panel** (right column, below the now panel): what the RE
+  Manager holds — a summary line, a read-only table of the running item,
+  the waiting items (front first, any client's) and the last 10 finished
+  items with their exit status / finish time / first message line, and a
+  **Clear queue** button (confirmation modal) as the recovery verb for a
+  failed item returned to the queue front. Owned by
+  `app/queue_panel.py::QueuePanelController`; refreshed from the existing
+  1 s manager status poll when the queue-shaped fields change (plus a 5 s
+  fallback), three bounded 0MQ reads on a background thread, one in
+  flight. Item summaries are pure functions tolerant of every request
+  schema version the manager's history may hold.
+
+### Fixed
+
+- `BackgroundResult` no longer emits toward its consumer from the daemon
+  thread: the result hops onto the GUI thread through the worker's own
+  queued signal (the worker is held alive until the hop lands) and
+  `result_ready` is emitted there. A daemon-thread emission aimed at a
+  consumer QObject races the consumer's destruction and segfaulted the
+  offscreen suite once the queue panel added a fetch per status snapshot.
+  Delivery now takes two event-loop turns; consumer connections are
+  unchanged.
+
 ## [0.27.0] - 2026-09-02
 
 ### Removed
--- a/GEECS-Console/CLAUDE.md
+++ b/GEECS-Console/CLAUDE.md
@@ -3,7 +3,7 @@
 The greenfield PySide6 operator console (decided 2026-07-10).  **The screen
 map is the spec**: `Planning/cutover_strategy/01_gui_feature_inventory.md`
 (the legacy-GUI capability inventory with dispositions) and the approved
-screen-map artifact (regions R1–R7).  This package is the "one working screen" half of
+screen-map artifact (regions R1–R7, plus R8 added in 0.28.0).  This package is the "one working screen" half of
 the commit/abort checkpoint's criterion (c).
 
 ## The screen map (regions → widgets)
@@ -83,7 +83,7 @@ are prefixed by region (`r3_radio_1d`, `r5_start_button`, …).
   re-arms on set_totals so a new scan's first line always renders); data
   signals never dedupe.
 - **R7 movable panel** (middle column, below R4/R5 since 0.19.1 — the
-  right column is R6's alone) — an editable combo (catalog scan-variable names
+  right column holds R6 and R8) — an editable combo (catalog scan-variable names
   first, then `device:variable` completions from `GeecsDbCompletions`),
   readback label, set field + button — owned by
   `app/movable_panel.py::MovablePanelController` since 0.19.0 (see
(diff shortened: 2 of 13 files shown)


skbarber:
## Adversarial review (fresh-context subagent, per /land) — findings on d39939d9

Reviewed `origin/master..HEAD` against the root/console `CLAUDE.md`, `qs_client/client.py`, `now_panel.py`, the MCP scan tools and the installed `bluesky_queueserver`; ran `tests/test_queue_panel.py` + `tests/test_tooltips.py` plus two ad-hoc scripts against the real controller.

1. **Medium — a queue change arriving while a fetch is in flight is dropped until the 5 s fallback (reproduced).** `on_status` commits `_last_key` before calling `refresh()`, and `refresh()` returns silently while `_fetch_inflight`; the change is consumed by the key compare but never fetched. Reproduced with a fake client whose last read blocks: "Queue empty" with Clear queue *enabled* for up to `FALLBACK_REFRESH_S`. Common case: the client's 0MQ recv timeout is 2 s and the poll is 1 s.
2. **Medium — `_INFLIGHT` is a set, so overlapping calls on one worker release the hold early; and the hop has no test of its own.** The now panel's per-experiment probes overlap on one worker across an experiment switch; the first landing's deferred discard removes the hold while the second is still out — the off-thread deletion the hold exists to prevent. Parented workers deleted mid-flight leave a dead wrapper in the set. Nothing pins the emission thread, the `_FAILED` release, or a consumer dropped mid-flight.
3. **Medium (scaling) — the whole manager history is re-read on every fallback tick, unbounded.** `history_items()` returns the entire Redis list and nothing in the repo caps or clears it; the history can only change when the status key changes, so the fallback tick should re-read `queue_get` only.
4. **Low — two identical `queue_get` round trips per refresh, non-atomic.** `running_item()` and `queue_items()` each call `queue_get`; the two reads can tear. Natural fix is a `queue_snapshot()` verb on `QueueClient` (GeecsBluesky) — a follow-up.
5. **Low / placement — plan vocabulary hard-coded in the GUI.** `summarize_item` recognises exactly two plan names; once the named plans (GeecsBluesky 0.73.0, `bluesky/named-plans`) land, every item submitted through them renders as `geecs_scan_plan (plan)`. The plan-name → argument-shape mapping belongs beside `qs_client`, where the MCP's `scan_history`/`scan_status` project the same dicts.
6. **Nit — third palette copy with a divergent value** (`_COLOR_GREY` differs from now_panel's) while the comment claims "the same values".
7. **Nit — window-level wiring untested.** Deleting `self._queue_panel.on_status(status)` from `main_window.py` would pass the suite.

Redundancy check: no findings (MCP read tools, GEECS-Schemas, request_builder, now_panel searched). All existing `result_ready` consumers connect `QueuedConnection`, so the hop is semantics-preserving. Scan-folder invariant and public-repo hygiene: untouched.

*Disposition follows in the next comment once the reviewer has confirmed the fix commit.*

skbarber:
## Review disposition (fix commit 8ada17b9, confirmed by the same reviewer)

| # | Finding | Disposition |
|---|---|---|
| 1 | Mid-fetch queue change dropped until the fallback | **Fixed** — `refresh()` queues a pending request while a fetch is in flight; `_apply_snapshot` runs it on both paths. Reviewer re-ran the reproduction: `fetches 2 / rows 1 / "1 waiting"`. Pinned by `test_change_arriving_mid_fetch_is_fetched_when_the_fetch_lands`. |
| 2 | `_INFLIGHT` set released overlapping holds early; hop untested | **Fixed** — counted hold (`_hold`/`_release`), dead-wrapper pop on the RuntimeError branch; new `tests/test_background.py` (emission thread, raise releases hold, overlap keeps hold, consumer dropped mid-flight). |
| 3 | Unbounded history re-read every fallback tick | **Fixed** — fallback tick re-reads the queue only; history cached, re-read on key change / after Clear. Pinned by `test_fallback_tick_rereads_the_queue_but_not_the_history`. |
| 4 | Two `queue_get` round trips per refresh, non-atomic | **Waived** — needs a `queue_snapshot()` verb on `QueueClient` (GeecsBluesky); follow-up, out of this console-only PR. |
| 5 | Named plans would render as `name (plan)` | **Fixed for keyword submissions** (`summarize_item` summarizes `geecs_*` kwargs carrying axes/capture/optimization; pinned). **Waived** for positional submissions + placement: the plan-name → argument map belongs in a shared `describe_item` beside `qs_client` (console + MCP), to be built in the named-plans PR; no client submits named plans yet. |
| 6 | Palette comment claimed identical values | **Fixed** — comment now states the deliberate text-vs-dot grey. |
| 7 | Window wiring untested | **Fixed** — `test_queue_panel_is_fed_by_the_status_poll_and_disposed_on_close` (own window, changed key, table/summary asserted, closeEvent disposes). |

Console suite at HEAD: **814 passed** (reviewer and author, multiple runs). CI green on both commits (pre-commit, unit-tests, console-windows).

**Reviewer observation, not against this PR**: `tests/test_main_window.py` run *alone* segfaults at random positions on a clean `origin/master` too (1 pass / 2 crashes of 3 on the reviewer's machine); the full suite never does. Same shape this PR diagnosed, and the remaining emitter of that shape is `HealthPoller` → window (`report_ready` from its daemon thread). Filed as a follow-up issue.

PR #765, opened 2026-09-02
GEECS-DataPortal 0.17.0: Analysis tab — run / re-run per diagnostic, state badges, error + log, artifacts inline

skbarber:
## What

PR 3 of the `feat/analysis-tab` arc (`Planning/data_portal/04_analysis_run_design.md`; backend = #763): the **Analysis tab** on the scan page, over the run endpoints #763 added. Portal 0.16.0 → 0.17.0.

- **`run.html`** (+~150 LOC, template + inline JS + CSS): a fourth tab. Each loadable diagnostic renders as a row — id, data device, state badge (`queued`/`running`/`done`/`failed`/`no_data`, or "not run"), run / re-run button (disabled while any run is active for the scan), error text and a collapsible captured log when failed, artifacts inline (raster images via `/run/{uid}/artifact`) or as download links, labels as plain text. Inapplicable analyzers (device not in this scan) collapse under "other analyzers". Polls the list endpoint every 1.5 s while a run is active and stops when it settles; a 409 on click just refreshes (the record is shown).
- **`app.py`** (+~20): `analysis_enabled` in the page context = feature configured + extra installed + folder resolvable; the tab button and pane render only then, and `tab=analysis` is accepted as URL state only then (otherwise a bookmarked link falls back to Plot, never a dead tab).
- CHANGELOG + CLAUDE.md routes paragraph.

## Judgment calls (cheap to veto)

- **Tab hidden rather than disabled** when runs are not possible — a read-only viewer deployment (no `--processing-configs`) shows no trace of the feature.
- **Artifacts shown = the job's returned list when the run is `done`, else the files on disk** under the analyzer's output dir — so a page after a portal restart still shows earlier outputs.
- **1.5 s poll** only while active; no websocket / SSE (no build chain, and runs are seconds-to-minutes).
- Rendered page JS was syntax-checked with `node --check` on the served HTML.

## Tests

- `tests/test_analysis_runs.py::TestAnalysisTab`: **4 new** — tab present + `tab=analysis` sticky when enabled; absent (and state falls back to `plot`) when the feature is off; absent when the folder is unresolvable; absent when the `analysis` extra is missing.
- Portal suite: **214 passed** (210 → 214).

## Hardware verification

**OWED at promotion** (unchanged from #763's list, plus): the tab renders the real listing on the worker host; a run's figures appear inline; a failed run shows the exception text and the log. No scan execution or devices touched.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

[...]

PR #770, opened 2026-09-02
GEECS-DataPortal 0.19.0: Analysis tab — summary figures automatically, per-bin visuals behind a stepper

skbarber:
## What

Owner feedback after the first live look at the Analysis tab (2026-09-02): "only automatically show the summary figs, not the individual bins … show one of the bins and the summary fig, but allow the user to cycle through the bins much like we do for individual images on the image panel."

- **`analysis_runs.classify_artifact`** (+ `describe_artifacts`): `kind` ∈ `summary` / `bin` / `other` and the `bin` number, from ScanAnalysis's own filename conventions (`<name>_<bin>_processed_visual.*` = bin; `_average_processed_visual`, `_averaged_image_grid`, `_summary_`, `_animation`/`.gif` = summary). Listing ordered summaries → others → bins by number (numeric, not lexical).
- **Listing endpoint**: `artifacts` now always describes the files on disk under the analyzer's output dir plus the finished job's non-file labels (before: the job's own list when done, which was the summaries only — the bins only appeared after a portal restart, inconsistently).
- **Tab**: summaries and unclassified files render automatically; bins render ONE at a time behind "‹ prev bin / bin N (i of M) / next bin ›", the Images tab's stepper idiom. Stepper state is per analyzer, in page memory (a poll re-render keeps it; a stepper click re-renders from the cached payload without a fetch). Handlers take the id via `data-` attributes (the #765 rule).
- Portal 0.18.0 → 0.19.0, CHANGELOG, CLAUDE.md.

- The bin stepper is page state, not URL state — a link does not carry which bin you were looking at. Cheap to add to the URL vocabulary if wanted.
- Labels changed from "← prev / next →" to "‹ prev bin / next bin ›" so the Images tab's "next →" absence test stays meaningful.

## Tests

- `test_artifact_kinds_follow_scan_analysis_naming` (classification + ordering) and the described-artifact shape assertions updated; **222 passed** (219 → 222 incl. the parametrised additions). Stepper markup evaluated in node from the served page: one bin image shown, prev/next disabled at the ends, ids via `data-id`.

## Hardware verification

Visual check on the worker host (already serving `feat/analysis-tab`) once merged: scan 32's FROG analyzer row shows the averaged figure automatically and a single bin with a working stepper.

https://claude.ai/code/session_018K1DmUzf1ZPjM4NxX2QRX3

### DIFF
--- a/GEECS-DataPortal/CHANGELOG.md
+++ b/GEECS-DataPortal/CHANGELOG.md
@@ -3,6 +3,24 @@
 All notable changes to this package will be documented here.
 Format follows [Keep a Changelog](https://keepachangelog.com/en/1.1.0/).
 
+## [0.19.0] - 2026-09-02
+
+### Changed
+
+- Analysis tab artifacts (owner ruling after the first live look,
+  2026-09-02): the scan-level figures — averaged image, image grid,
+  1D summary, animation — show automatically; the per-bin visuals
+  (`<name>_<bin>_processed_visual.*`) show **one at a time** behind a
+  prev / next stepper ("bin N (i of M)"), like the Images tab's shot
+  stepper, instead of every bin at once. The classification
+  (`kind` ∈ summary / bin / other, `bin` number) is decided server-side
+  by ScanAnalysis's own filename conventions
+  (`analysis_runs.classify_artifact`) and the listing is ordered
+  summaries → others → bins by number. The listing now always
+  describes what is on disk under the analyzer's output dir plus the
+  finished job's non-file labels (before: the job's own list when done,
+  which was the summaries only).
+
 ## [0.18.0] - 2026-09-01
 
 ### Added
--- a/GEECS-DataPortal/CHANGELOG.md
+++ b/GEECS-DataPortal/CHANGELOG.md
@@ -13,13 +13,17 @@ Format follows [Keep a Changelog](https://keepachangelog.com/en/1.1.0/).
   (`<name>_<bin>_processed_visual.*`) show **one at a time** behind a
   prev / next stepper ("bin N (i of M)"), like the Images tab's shot
   stepper, instead of every bin at once. The classification
-  (`kind` ∈ summary / bin / other, `bin` number) is decided server-side
-  by ScanAnalysis's own filename conventions
-  (`analysis_runs.classify_artifact`) and the listing is ordered
-  summaries → others → bins by number. The listing now always
-  describes what is on disk under the analyzer's output dir plus the
-  finished job's non-file labels (before: the job's own list when done,
-  which was the summaries only).
+  (`kind` ∈ summary / bin / other, `bin` number) is ScanAnalysis's
+  own output-name contract, parsed by its
+  `renderers.config.parse_output_filename` (ScanAnalysis 1.18.0, next to
+  the code that builds the names; imported lazily) — a bin's data file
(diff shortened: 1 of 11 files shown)


skbarber:
## Adversarial review (fresh-context subagent) — report + dispositions

Reviewer report (222 tests green at review time; regex vs digit-bearing device names, XSS, the `ANALYSIS_LAST`/`force` path, done-job dedupe, invariants all checked and cleared):

1. **Medium-high — per-bin `.h5` files classified `other` and therefore auto-shown**: the renderers write `_<bin>_processed.h5` next to each `_<bin>_processed_visual.png`, so a 30-bin scan would auto-render the grid plus 30 download links — contradicting the ruling this PR implements; the PR's own test pinned the wrong answer.
2. **Medium — the naming contract lived only in the portal**, pinned by partly fictitious names (`_animation.gif`, `_summary_per_shot`); ScanAnalysis records no consumer of its names and already parses them ad hoc twice (`parts[-2].isdigit()` in both wrappers). Better home: next to `get_filename`.
3. **Low-medium — a stepper click rebuilt the whole tab**, re-requesting every inline (no-cache) artifact.
4. **Low — `list_artifacts(limit=200)` capped lexically**, so summaries (`_average…` sorts after every `_<digits>_…`) were the first to drop at 100+ bins.
5. **Trivial — the clamped stepper index was not written back.**

### Dispositions (287420aa)

- **#1 fixed** — `_(\d+)_processed(?:_visual)?\.<ext>$` is bin-kind; the stepper groups by bin number and shows every file of the current bin (visual inline, data file as a link); tests pin `_16_processed.h5 → ("bin", 16)`.
- **#2 fixed (placement)** — `parse_output_filename` now lives in `ScanAnalysis/scan_analysis/analyzers/renderers/config.py` next to `get_filename` (ScanAnalysis **1.17.3 → 1.18.0**, CHANGELOG, `tests/test_renderer_output_names.py` incl. a round-trip of the `get_filename` shape and only real names — the `_animation.` marker is gone). The portal classifies through it (lazy import, `other` without the extra). **Waived for scope**: the two ad hoc `parts[-2].isdigit()` parsers in the wrappers are not migrated here; follow-up.
- **#3 fixed** — `stepBin` replaces only that analyzer's row from the cached payload.
- **#4 fixed** — listing uncapped; `describe_artifacts(max_bins=200)` caps bins only, after classification; test pins summaries surviving a `max_bins=1` cap.
- **#5 fixed** — clamp written back; verified in node with an out-of-range index.

Tests: **235 passed** (portal 223 + the new ScanAnalysis contract test 12). Reviewer re-verifying 287420aa; CI watching.

skbarber:
Reviewer confirmed the five dispositions on 287420aa and found three residuals: (A) `max_bins` counted bin *files* and could split the last bin; (B) my test-file edit had dropped the `class TestRun:` header (the five tests ran as `TestListing` members); (C) the listing cost was unbounded after removing the lexical cap. Fixed in e519202e: cap by distinct bin numbers (a bin is never split — the `max_bins=1` test now expects both bin-2 files), header restored, listing-derived paths skip the per-file stat (`known_files`) and the listing keeps a sanity bound of 2000; also the inline visual now sorts before its data file within a bin, and ScanAnalysis/CLAUDE.md names `parse_output_filename` as a consumed contract. 235 passed; reviewer re-verifying; CI watching.

PR #775, opened 2026-09-03
Planning: site profile + fleet uniformity execution plan

skbarber:
Planning doc only (`Planning/site_profile/00_overview.md`), no code — the execution plan for making the services fleet uniform and portable before the new services box arrives (~late Sep 2026) and before other GEECS facilities adopt it.

**What it fixes:** `/fleet-status` (pending on `feat/fleet-status-skill`) showed the current box drifted from the fleet map: RE Manager / doc proxy / MCP hand-started with no unit, capture daemon as a `--user` unit, 774 staged files in the gateway clone, MCP venv not at its runbook path.

**The design move:** one home per facility value — `config.ini` client-side (exists), `site.env` host-side (new) consumed by systemd `EnvironmentFile=`; unit templates carry install-time placeholders for paths (systemd does not expand env vars in `WorkingDirectory`/executable paths) and `${GEECS_*}` for runtime values; a `deploy/bootstrap_host.sh` makes the fleet map's bootstrap list executable.

**Phases:** 0 land fleet-status → 1 contract docs + `site.env.example` + `/land` literal check → 2 templates + bootstrap (scratch-root live check, non-invasive) → 3 promote the current box (maintenance window, owner's sudo steps, fleet map rewritten to observed truth) → 4 remaining literals (DB-driven PVA roster, TZ defaults, default experiment) → 5 the new box = run the bootstrap. Containers and passwordless deploy sudo explicitly deferred; four open decisions for the owner listed at the end.

No version bumps (Planning/ only). Skipped the adversarial-review step of `/land` for a planning document; the plan itself is the thing to review.

https://claude.ai/code/session_01N8myqX2WHYm38s14FBwn8E

### DIFF
--- a/Planning/site_profile/00_overview.md
+++ b/Planning/site_profile/00_overview.md
@@ -1,0 +1,251 @@
+# Site profile + fleet uniformity
+
+Execution plan, 2026-09-03 (Sam + Claude session that built `/fleet-status`).
+Status: **planned, nothing built.** Written for whoever picks it up; assumes
+no context from the originating conversation.
+
+## Why now
+
+`/fleet-status` (PR pending on `feat/fleet-status-skill`) showed the
+current services box as it is, not as the fleet map describes it:
+
+| Service | Observed | Fleet map says |
+|---|---|---|
+| RE Manager :60615, doc proxy :5568 | hand-started processes, no unit (parent = a dead shell / PID 1) | systemd `geecs-qserver` |
+| GEECS-MCP :8100 | hand-started, baked venv at `~/geecs-mcp-venv` | systemd unit, venv at `/opt/geecs-mcp-venv` |
+| Capture daemon | `systemctl --user` unit (linger on) | system unit `geecs-capture` |
+| CA gateway clone | 774 files staged against HEAD | clean pinned clone |
+| PVA fleet | 4 of 13 roster hosts unreachable | 9 hosts |
+
+Three forcing functions make fixing this now worth more than waiting:
+
+1. **A higher-end services box arrives in ~3 weeks** (late Sep 2026) and
+   the fleet migrates to it. The recipe is what migrates, not the
+   processes. Making the current box uniform *is* rehearsing that recipe
+   on a host we can afford to break.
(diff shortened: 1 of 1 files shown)


skbarber:
## Response to the Codex review

The plan was written before the build; #777 (Phases 1+2) already avoided P1 and the trailing-comment issue, and the document is now corrected to the built behaviour (12d08f11). P2 was a real gap in #777 and is fixed there.

| # | Finding | Disposition |
|---|---|---|
| P1 | `Environment=` will not expand `${...}` from `EnvironmentFile=` | **Fixed** in the doc (12d08f11): site.env carries the downstream names (`TZ`, `EPICS_*`, `QS_EXPERIMENT`) and no template has an `Environment=` line for a site value — which is how #777 was built (`deploy/site.env.example`, the five units). |
| P1 | Inline comments in the site.env example | **Fixed** in the doc: own-line comments only, with the syntax rule stated; #777's `deploy/site.env.example` was already written that way and its `site_env_lib.sh` mirrors systemd's parser. |
| P2 | "only if absent" contradicts site.env as root truth | **Fixed** in #777 (f4c93f58): when a `config.ini` exists the bootstrap prints the diff against the rendered form (comments/blanks/`api_key` excluded, `(none)` when they agree) so a stale file is reconciled, never silently kept; still never overwritten because the file also holds the hand-entered Tiled key. Doc updated to match. |
| P3 | Scratch-root live check cannot have only env diffs | **Fixed** in the doc: the check as actually performed on 2026-09-03 rendered with the host's real account/home/root/poetry path, so path lines compare like for like; expected residual diffs are enumerated (EnvironmentFile line, older Description text, the gateway's future `gateway-checkout` path). That diff is what surfaced the portal's hand-typed `--processing-configs` path. |

skbarber:
Fixed the last stale reference: Phase 4's timezone bullet now reads "the portal unit inherits `TZ` from `site.env` via `EnvironmentFile=` (no template hole)" — no `GEECS_TZ` remains outside the two sentences that say it does not exist.

PR #774, opened 2026-09-03
GEECS-DataPortal 0.20.0: browsing JSON API (day list, run overview, device tier, day jump)

skbarber:
## What

Everything the portal's pages show, as JSON — so a script or an AI agent (the OSPREY HTU assistant's new `geecs-data-portal` skill) can read what a browser shows and press what a browser presses without scraping HTML. The Plot/Images tabs already had a JSON `/api` layer; the day table, the rail, the Overview tab and the device tier were HTML-only.

New routes (all `GET`, `no-cache`):

| Route | Page equivalent |
|---|---|
| `/api/day/{iso}?experiment=&filter=` | the day page's table (newest first, same filter haystack) |
| `/api/run/{uid}?day=` | the rail + Overview table verbatim (`metadata_rows`) + start/stop docs + `devices` + stepper `neighbours` / `day_runs` + `processing_options` + `analysis_enabled` |
| `/api/run/{uid}/device?device=` | one device's tier / path / ext (the click on a device name — same `device_kind` probe) |
| `/api/run/jump/{iso}?prefer=` | the day steppers' target, as data instead of a redirect |

Plus `version` on `/health`, and a test pinning that `/openapi.json` lists the whole agent surface.

## Why it is shaped this way

- The day listing, neighbour and folder-resolution logic moved into shared closures (`_list_day`, `_neighbours`, `_resolved_folder`) that the templates now call too — one implementation, so the HTML and JSON surfaces cannot drift.
- Documents pass through `analysis.jsonable_document` (NaN / numpy / set → JSON-safe): a document key must never 500 a run's detail.
- No new numerics, no new primitives: every payload is the template's own data.

## Tests

`cd GEECS-DataPortal && poetry run pytest tests -q` → **240 passed** (18 new: `TestBrowsingApi` in test_app.py, two device-tier/read-only pins in test_resources.py, one proxy-prefix pin from the review). Pre-commit clean.

## Hardware verification

None needed — no scan execution or device path. **Live check done 2026-09-02**: the portal host runs this branch (0.20.0) under systemd; `/health` reports `version: 0.20.0`, `/api/day/2026-09-02`, `/api/run/<uid>`, `/api/run/<uid>/device?device=UC_ModeImager` and `/api/run/jump/2026-09-02` all answered against the live catalog, and the maintainer confirmed the existing pages are unchanged in the browser. The host stays on this branch until merge, then flips back to master.

## Review

[...]

PR #776, opened 2026-09-03
/fleet-status: observed deployed-fleet picture (host → service → checkout/branch/commit/version)

skbarber:
## What

A repo skill + script that reports what the fleet is **actually running**, as opposed to the fleet map's intended picture: for every service on every host, which checkout it runs from, which branch/commit and package version, whether the running process matches the code on disk, and how each deployed commit relates to `origin/master` and the local worktrees. Gated on `/lab-status` tier 1 first (per owner request); read-only and bounded throughout — no PV writes, no restarts, no pulls.

- `scripts/fleet_status.sh` — four stages: (0) `lab_status.sh` reachability gate; (1) each service's self-report over its own protocol (Tiled HTTP, portal `/health`, queueserver/MCP ports, CA gateway identity PVs via `lab_status.sh --hardware`, PVA fleet `version`/`heartbeat` PVs from the checked-in Phoebus roster); (2) per-host ssh discovery — system **and user** systemd units plus **whoever owns each fleet port** (a hand-started queueserver has no unit and must still show), mapped to its clone via cwd (or the baked venv's `direct_url` for the MCP pattern) → branch/sha/staged/unstaged, pyproject-vs-installed version, "checkout moved after the process started", and **"files on disk = commit X, not HEAD"** (tree-hash search — HEAD can lie when a ref advances without a checkout); (3) local cross-reference (fetch, worktrees, branch tip/containment, distance from master). `--summary` renders one box table + attention list; `--watch SECS` makes a terminal pane the dashboard; `--full` keeps the log.
- `scripts/fleet_table.py` — stdlib-only renderer for the summary table (reads the tab-separated `role=` records the script emits).
- `.claude/skills/fleet-status/SKILL.md` — when to use it, what each stage proves, the warnings-to-act-on table, reporting + boundaries.
- Cross-references: fleet map intro (intended vs observed), skills overview row, root `CLAUDE.md` skill inventory, `CONTRIBUTING.md`, `/lab-status` pointer.

Hosts/aliases come from `config.ini` (`[tiled] uri`, `[qserver] host`) and `~/.ssh/config` HostName matching — nothing lab-specific is committed beyond the fleet map's port numbers and the existing `.bob` roster.

## Why

During rapid development hosts run feature-branch checkouts for live checks, checkouts get pulled under running services, and version bumps land without reinstall; the fleet map cannot show any of that. First live run found: RE Manager / doc proxy / MCP hand-started with no unit, the capture daemon as a `--user` unit, and the gateway clone's working tree three weeks older than its HEAD pointer (fixed on the host today). Those findings drive the follow-on plan in #775.

## LOC by concern

| Concern | Files | LOC |
|---|---|---|
| Probe script | `scripts/fleet_status.sh` | +589 |
| Table renderer | `scripts/fleet_table.py` | +192 |
| Skill | `.claude/skills/fleet-status/SKILL.md` | +130 |
| Cross-references | fleet_map, skills overview, CLAUDE.md, CONTRIBUTING.md, lab-status skill | +17 / −7 |

No package code changed → no version bumps.

## Tests

`./scripts/check.sh --base origin/master`: lint clean (ruff, ruff-format, pydocstyle, pre-commit hooks); no package suites implied by the diff (repo-level `scripts/` + skills + docs). Offline paths exercised (network-down → local picture only, `--local-only`), plus a stubbed end-to-end run (fake lab_status/ssh/gateway) for stages 1–3 formatting.

## Hardware verification

**Live on the lab network 2026-09-03**, full run ~18 s: stage 1 reported Tiled 0.2.14, portal 0.20.0, CA gateway 0.20.0 with 98→99 devices, 9 PVA gateways at 0.4.4 (4 roster hosts unreachable); stage 2 found all 7 services on the host (3 UNMANAGED, 1 user unit) with correct clones/versions; the disk≠HEAD detection named the gateway clone's real tree (9c9681de vs HEAD 28749780) — confirmed by hand, reset, gateway restarted, row now clean. Dashboard pane running `--watch 300` since this afternoon.

Judgment-call additions (cheap to veto): `--watch`/`--summary`/`fleet_table.py` were added on request during the session rather than in the original brief; the disk-vs-HEAD tree search (`rev-list -400 --all` per dirty clone) is the one probe that costs more than a few ms.

https://claude.ai/code/session_01N8myqX2WHYm38s14FBwn8E

### DIFF
--- a/.claude/skills/fleet-status/SKILL.md
+++ b/.claude/skills/fleet-status/SKILL.md
@@ -62,18 +62,24 @@ versions` line means a rollout is incomplete or a box's
 pull-on-restart no-oped — the GeecsPvaGateway runbook's known failure.
 
 **Stage 2 — host checkouts (ssh).** On each host derived from
-config.ini, every `geecs-*` (+ `tiled`) systemd unit is mapped to the
-clone it runs from (`WorkingDirectory`, or the baked venv's recorded
-source path for the MCP pattern), then: branch, short sha, commit date,
-dirty count, pyproject version vs the version installed in the unit's
-venv, and whether HEAD moved after the unit started. The three warnings
-to act on:
+config.ini, services are discovered two ways and deduplicated by pid:
+every `geecs-*` (+ `tiled`) systemd unit in **both** system and user
+scope, and **whoever owns each fleet port** — a queueserver started by
+hand in tmux has no unit and must still show up. Each process is mapped
+to the clone it runs from (its cwd, or the baked venv's recorded source
+path for the MCP pattern), then: branch, short sha, commit date,
+staged/unstaged counts, pyproject version vs the version installed in
+the running interpreter's venv, and whether HEAD moved after the process
+started (for a baked venv: after the install). The warnings to act on:
 
 | Warning | Meaning | Remedy (on the host, by the user or with their go) |
 |---|---|---|
-| `pyproject says X but the venv has Y installed` | a bump landed in the checkout without reinstall | `poetry install` in that package dir, then restart |
-| `checkout moved <time> after the service started` | the running process predates the code on disk | `systemctl restart <unit>` — after confirming that checkout is the intended one |
-| `DIRTY: N modified tracked file(s)` | someone edited the deployed tree in place | look before touching: it may be a live hotfix nobody committed |
+| `no systemd unit owns this process` (UNMANAGED) | started by hand; dies with its tmux/ssh session or a crash, no restart on reboot | install the runbook's unit when the dev phase ends; until then, know it |
+| `(user unit)` in the name | runs under `systemctl --user`, not the system scope the runbook describes | fine for development; note that `systemctl status` without `--user` will not see it |
+| `pyproject says X but the venv has Y installed` | a bump landed in the checkout without reinstall | `poetry install` in that package dir (pip reinstall for a baked venv), then restart |
+| `checkout moved <time> after the process started` | the running process predates the code on disk | restart it — after confirming that checkout is the intended one |
+| `checkout moved <time> after the baked install` | the MCP-style venv was baked from an older checkout state | pip reinstall from the checkout, then restart |
+| `STAGED: N` / `UNSTAGED: N` | the deployed tree differs from its HEAD | look before touching: it may be a live hotfix nobody committed, or a stray `git add` |
 
 Any unit whose clone is on a branch other than `master` is a fact to
 surface, not a fault — feature-branch deploys for live checks are
@@ -114,3 +120,7 @@ this page in the same PR".
   device roster is healthy — that is `/lab-status --hardware` and the
   gateway DB audit.
 - Tiled has no checkout (pip install); its version is the whole story.
+- Port-owner discovery needs `ss -p` to see the pid, which it does only
+  for processes owned by the ssh user (or root). A fleet port that shows
+  as listening in stage 1 but has no stage-2 record was started by a
+  different account.
--- a/.claude/skills/fleet-status/SKILL.md
+++ b/.claude/skills/fleet-status/SKILL.md
@@ -31,14 +31,18 @@ and bounded: no PV writes, no restarts, no pulls.
 
 | Situation | Command |
 |---|---|
-| The normal case | `scripts/fleet_status.sh` |
+| The normal case (full log, for reasoning) | `scripts/fleet_status.sh` |
+| One box table + attention list (for the user) | `scripts/fleet_status.sh --summary` |
+| Persistent dashboard pane (cmux/tmux) | `scripts/fleet_status.sh --watch 300` (summary every 300 s; add `--full` for the log) |
 | Off the lab network / just want the local worktree picture | `scripts/fleet_status.sh --local-only` |
 | No key-based ssh to the hosts from this machine | `scripts/fleet_status.sh --no-ssh` |
 | ssh alias not discoverable from `~/.ssh/config` | `scripts/fleet_status.sh --ssh <ip>=<alias>` |
 | config.ini has no `[Experiment]` | add `--experiment <Name>` (needed for the CA gateway PVs) |
 | Avoid the GitHub round-trip | `--no-fetch` (distances then use the last-fetched `origin/master`) |
 
-`$ARGUMENTS`, if given, passes straight through. Budget: ~10 s offline,
+The summary table is rendered by `scripts/fleet_table.py` (stdlib only)
+from the same probe records; the full log is what you reason over, the
+table is what you show. `$ARGUMENTS`, if given, passes straight through. Budget: ~10 s offline,
 under a minute on the lab network (each ssh host 25 s max, each PVA
 host 2 s per get).
 
--- a/.claude/skills/fleet-status/SKILL.md
+++ b/.claude/skills/fleet-status/SKILL.md
@@ -1,0 +1,116 @@
+---
+name: fleet-status
+description: >
+  Observed picture of the deployed fleet: which host runs which service,
+  from which checkout, on which branch/commit and package version, and
+  whether the running process matches the code on disk. Use when the user
+  asks "what is deployed where", "which version is the portal / gateway /
+  worker running", "is the host still on the feature branch", "did the
+  deploy take", "does the fleet map match reality", before deploying or
+  restarting a service, or after a PR merges that a host is expected to
+  pick up. Wraps scripts/fleet_status.sh, which gates on /lab-status
+  first. For plain reachability ("am I on the lab network") use
+  /lab-status; for the intended picture read docs/platform/fleet_map.md.
+---
+
+# /fleet-status — what code is the fleet actually running?
+
+`docs/platform/fleet_map.md` is the **intended** fleet: hosts, ports,
+units, runbooks. `scripts/fleet_status.sh` is the **observed** one. The
+two drift during rapid development — a host runs a feature-branch
+checkout for a live check, a checkout gets pulled but the unit never
+restarted, a version bump lands without a `poetry install` — and this
+skill exists to see that drift instead of remembering it.
+
+The script owns the probes, timeouts, and how endpoints are derived
+(config.ini `[tiled]`/`[qserver]`, ssh aliases from `~/.ssh/config`).
+Do not restate or hardcode hosts here. Everything it does is read-only
+and bounded: no PV writes, no restarts, no pulls.
+
+## Invocation
+
+| Situation | Command |
+|---|---|
+| The normal case | `scripts/fleet_status.sh` |
+| Off the lab network / just want the local worktree picture | `scripts/fleet_status.sh --local-only` |
+| No key-based ssh to the hosts from this machine | `scripts/fleet_status.sh --no-ssh` |
+| ssh alias not discoverable from `~/.ssh/config` | `scripts/fleet_status.sh --ssh <ip>=<alias>` |
+| config.ini has no `[Experiment]` | add `--experiment <Name>` (needed for the CA gateway PVs) |
+| Avoid the GitHub round-trip | `--no-fetch` (distances then use the last-fetched `origin/master`) |
+
+`$ARGUMENTS`, if given, passes straight through. Budget: ~10 s offline,
+under a minute on the lab network (each ssh host 25 s max, each PVA
+host 2 s per get).
+
+## Stages and what each one proves
+
+**Stage 0 — reachability.** Runs `scripts/lab_status.sh` tier 1 (the
+same gate `/lab-status` uses). If the network is DOWN the script stops
+after the local picture and says `remote: UNKNOWN`. Report exactly that:
+an unreachable service is *unknown*, not down. Never bypass this by
+calling the services directly.
+
+**Stage 1 — self-reported versions.** Each service over its own
+protocol: Tiled's library version (HTTP), the Data Portal's `/health`
+(`ok`, catalog probe, package version), MCP port liveness (no version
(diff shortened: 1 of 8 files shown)


skbarber:

Static review only: `bash -n`, ruff / ruff-format / pydocstyle on `scripts/fleet_table.py` (all clean), the renderer run against synthetic stdin, and the diff read against `scripts/lab_status.sh`, the `deploy/*.service` templates, `GeecsPvaGateway/server.py` + `deploy/gen_fleet_status.py`, `GEECS-MCP/deploy/DEPLOYMENT.md`, and `docs/platform/fleet_map.md`. Nothing on the lab network was touched.

### 1. Summary table merges records from *different processes* into one row (Medium, Correctness)

**Scenario / evidence.** `fleet_table.parse()` merges by `role` with first-non-empty-wins. Stage 2 can legitimately emit two records for one role: a loaded-but-dead unit (`state=inactive/dead`, `pid=0`, clone taken from the unit's `WorkingDirectory`) and the UNMANAGED process that actually owns the port (the tmux-started worker the PR itself found live). Reproduced with synthetic stdin (`scripts/fleet_table.py`): the row shows **`systemd` / `~/qserver-checkout master aaaaaaaa`**, drops `UNMANAGED`, the feature branch and the 3 unstaged files of the process that is really serving :60615, and fabricates a `venv 0.41.0 ≠ pyproject 0.40.0` note by combining `installed` from record 3 with `pyproject` from record 2. This is exactly the state the #775 follow-on produces (units installed, then someone stops one and hand-starts from a worktree). The full log is correct; only `--summary`/`--watch` — the view the user is meant to look at — lies.

**Suggested fix.** In `parse()`, key stage-2 records by `(role, pid|svc)` and, when a role has several, pick the one whose `state` starts with `active/`/`running` for the row (or emit one row per record); never let `installed`/`pyproject`/`staged` cross records.

### 2. Package identification is a cwd / first-match heuristic and can name the wrong package (Medium, Correctness)

**Scenario / evidence.** Two paths in `REMOTE_SNIPPET::emit`:
- Baked venv (MCP): `for d in m.distributions(): ... if "GEECS" in direct_url.json: break`. `pip install <checkout>/GEECS-MCP` (per `GEECS-MCP/deploy/DEPLOYMENT.md`) pulls `geecs-bluesky`, `geecs-data-utils`, `geecs-core`, `geecs-schemas` from `file://` path deps (`GEECS-MCP/pyproject.toml` lines 34-40), each with its own `direct_url.json`; under `/opt/geecs/GEECS-Plugins/...` or `~/qserver-checkout/GEECS-Core` several match `"GEECS"`. `distributions()` iterates in filesystem order, so the MCP row's `pkg`/`pyproject`/`installed` become e.g. geecs-core's — and `fleet_table.version()` prints the bare number for non-Bluesky roles, so nothing signals the swap.
- cwd path: `pkgdir_up "$cwd" "$clone"` stops at the first `pyproject.toml`. A hand-started process launched from the clone root (`cd ~/x-checkout && poetry -C GeecsBluesky run ...`) resolves to the **root** `pyproject.toml` → `pkg=geecs-docs pyproject=0.1.0`, `installed=?` → spurious `geecs-docs: pyproject says 0.1.0 but the venv has ? installed`.

The live run's "correct clones/versions" is consistent with cwd luck (services started from their package dirs) and doesn't exercise the baked branch (MCP was found via cwd, not `direct_url`).

**Suggested fix.** Derive the package from the process, not the directory: take `-m <module>` / the console-script name from `/proc/pid/cmdline`, map it to a distribution via `importlib.metadata.packages_distributions()`, then read *that* dist's `direct_url.json` and its `pyproject.toml`. Fall back to `pkgdir_up` only when that fails, and never accept the clone root's pyproject.

### 3. Stage-2 partial-output guard is dead code; stderr noise becomes `[DOWN]` rows (Medium-low, Correctness)

**Scenario / evidence.** `scripts/fleet_status.sh:501` — `grep -q '^svc='` — but every record starts with `role=` (`emit` builds `rec="role=$role\tsvc=..."`; the header comment "one `svc=...` record per service" is stale). So when `bounded` kills a slow ssh at 25 s (rc 137) after emitting records, the intended "keep what we got" branch never fires and the host is reported as `ssh failed`. Separately, `2>&1` on the ssh call means any non-record line — `Warning: Permanently added '<host>' (ED25519) to the list of known hosts.` on the first `accept-new` connect, or a remote `git`/`date` warning — flows into `fmt_host_records`, which has no `role=` filter: `state=""` → `tag="[DOWN]"` → a blank `  [DOWN]                                    since ?` line in the log **and** in the `--summary` Attention list.

**Suggested fix.** `grep -q '^role='` at line 501; in `fmt_host_records` `case "$line" in role=*) ;; *) continue ;; esac` (or log it as `info`).

### 4. Printed remedy contradicts the skill's own rule and is destructive (Low-medium, Correctness / policy)

**Scenario / evidence.** The disk≠HEAD warning ends with `Remedy: git reset --hard HEAD, then restart`, while `SKILL.md`'s table says STAGED/UNSTAGED means "look before touching: it may be a live hotfix nobody committed". "Index tree equals an older commit's tree" is precisely what `git revert -n HEAD` (a staged rollback someone didn't commit) or `git reset --soft` leaves behind; `reset --hard` would discard that and any unstaged edits with it. Also the search runs `git write-tree`, which writes a tree object into the host's `.git` — a (harmless) write from a script whose contract is "never writes", and the `rev-list -400 --all` + per-commit `rev-parse` loop spawns up to 400 processes per dirty clone inside the 25 s ssh budget.

**Suggested fix.** Drop the command from the warning (state the fact + "confirm which tree is intended before touching"), keep the remedy in the skill table where the caveat is. Replace the loop with one `git log -400 --all --format='%h %H %cs %T' | awk -v t="$idx_tree" '$4==t'`; if the write-tree is kept, say so in the read-only claim.

### 5. `(disk)` sha gets its own table row instead of annotating the service (Low, Correctness)

**Scenario / evidence.** `note_sha "${role} (disk)"` then `rec "role=$label	master_rel=..."` — reproduced: a separate row `CA gateway (disk) | ? | — | ? | 2 behind master` appears, and the real CA gateway row carries no master distance. This is the exact case the live run hit (gateway clone at 9c9681de vs HEAD).

**Suggested fix.** Emit `rec "role=$role	disk_master_rel=$rel"` and render it in `notes()` as `disk: 2 behind master`.

### 6. Verbatim duplication of `lab_status.sh` plumbing, plus a text-scrape contract between the two scripts (Low, Redundancy)

**Evidence.** `ini_get` (identical awk), `ok/bad/skip`, the `[tiled] uri` → `LAB_HOST`/`TILED_PORT` sed pair, and the `[Experiment] expt|exp_name` lookup are copied from `scripts/lab_status.sh`; `port_open` is a second implementation of the same watchdog. Tiled's version is fetched by stage 0 (lab_status prints `(vX)`) and again by stage 1; `lab_status.sh --hardware` re-runs all of tier 1 (three more 2 s probes + ping) just to reach tier 2, whose result then crosses scripts via `grep 'gateway alive'` + `sed 's/.*version=([^, ]+)'` on human-formatted output that no test pins.

**Suggested fix.** `scripts/lib/lab_env.sh` (ini_get, endpoint derivation, printers, `bounded`/`port_open`) sourced by both; give `lab_status.sh` a `--tier2-only` (or a `key=value` record line) so fleet_status consumes a stated contract instead of scraping prose.

### 7. PVA fleet probe re-derives GeecsPvaGateway's naming contract in a bash heredoc (Low, Placement)

**Evidence.** The heredoc decodes host IPs with `re.sub("_", ".", pv.split(":")[2])` — the inverse of `server.py::_instance_token` (`normalize_component(cameras[0].host)`, with a `gethostname()` fallback that this inverse cannot decode) — and takes its roster from the *generated* `fleet_status.bob` although `gen_fleet_status.py` declares `HOSTS` the roster of record and already sketches the owed DB derivation. That knowledge belongs in GeecsPvaGateway (e.g. `python -m geecs_pva_gateway.fleet --experiment X`, emitting the same `key=value` records), which the script would call the way it calls `lab_status.sh --hardware`; the future DB-driven roster then lands in one place and a token-format change cannot silently break the probe.

### 8. Fleet map paragraph added by this PR sits above a snapshot the PR's own live run shows is stale (Low, docs consistency)

**Evidence.** `docs/platform/fleet_map.md` now says "reconcile drift back into this table", while the Snapshot admonition two lines below still says Data Portal `0.15.x`; the PR's hardware section reports portal 0.20.0 (matches merged #774). Update the snapshot in the same PR — it is the rule the new sentence asks readers to follow.

### Checked and refuted

- Syntax/lint: `bash -n` clean; ruff, ruff-format, pydocstyle clean on `fleet_table.py`; pre-commit excludes do not skip `scripts/`.
- Public-repo hygiene: only `192.168.6.14` (allowed) and the pre-existing `/home/geecs` in unit templates; ssh aliases come from `~/.ssh/config`, none committed.
- Versioning: no package code touched → no bump/CHANGELOG owed. Scan-folder invariant: n/a, nothing writes to the share.
- PVA PV parsing: after stripping `pva://`, `split(":")[2]` is the IP token and `base + ":heartbeat"` matches `server.py:301-302`.
- `${counts%%<tab>*}` uses a literal tab (confirmed with `sed -n l`); `rev-list --left-right --count` output is tab-separated.
- `[qserver] host` and `[tiled] uri` match the keys documented in `docs/tutorials/getting_started.md`.
- All `deploy/*.service` names match `role_for_unit` globs; `launch_re_manager.sh` ends in `exec start-re-manager`, so the qserver MainPID is the venv python and the version check applies.
- `bounded` stdin handling works for the pipe (ssh), heredoc (PVA) and `--watch` (tty) cases; the watchdog is detached from stdout so command substitutions do not block.
- `--watch N` argument stripping is correct in any position; `--full` outside `--watch` is a harmless no-op that coincidentally makes the footer hint right.
- `${venv/#$HOME/\~}` prints `\~` under macOS bash 3.2 but the snippet only runs on Linux bash 5 (quote removal applies); cosmetic at worst, unverified here.
- `.bob` roster vs `gen_fleet_status.HOSTS`: generated and committed together, so reading the artifact is consistent today (see 7 for the placement point).
- Skill rosters (`CLAUDE.md`, `CONTRIBUTING.md`, `docs/skills/overview.md`, `/lab-status` pointer) are updated consistently; `get-started` / `writing_a_skill.md` mention `/lab-status` in prose, not as rosters.

---

skbarber:

| # | Finding | Disposition |
|---|---|---|
| 1 | Summary table merged records from different processes into one row | **Fixed** 285da075 (per-process records, live one shown, others noted); confirmed by the reviewer |
| 2 | Package identification was a cwd / first-match heuristic | **Fixed** 285da075 (dist resolved from the process's own cmdline via its interpreter; root pyproject refused); confirmed |
| 3 | Partial-output guard dead code; stderr noise became `[DOWN]` rows | **Fixed** 285da075 (`^role=` guard, non-record lines logged as info); confirmed |
| 4 | Printed remedy destructive + 400-process tree search + unstated write | **Fixed** 285da075 (fact + skill pointer, one `git log`+awk, header names `write-tree`); confirmed |
| 5 | `(disk)` sha got its own table row | **Fixed** 285da075 (annotates the service row via `disk_master_rel`); confirmed |
| 6 | Verbatim duplication of `lab_status.sh` plumbing + text-scrape contract | **Waived** — `lab_status.sh` is untouched by this PR; the scrape is on our own script's stable line; a shared `scripts/lib/lab_env.sh` + a tier-2 record line is a follow-up when either script next changes (noted in #775) |
| 7 | PVA probe re-derives GeecsPvaGateway's naming in a bash heredoc | **Waived** — the DB-driven roster per experiment is Phase 4 of #775; that PR moves the probe's knowledge into GeecsPvaGateway (`fleet` subcommand) as suggested |
| 8 | Fleet map snapshot said Data Portal 0.15.x | **Fixed** 285da075; confirmed |
| A | Stage-3 `master_rel` still merged across processes | **Fixed** c4e46953 (`for_sha=` records, attached by sha); confirmed |
| B | Process state overrode the stage-1 verdict (crash-looping unit rendered ✓) | **Fixed** c4e46953 (`proc_state` beside the verdict; ✗/! rules); confirmed |
| C | cwd clone beat the distribution's source clone | **Fixed** c4e46953 (source clone wins); confirmed |
| nit | `--short=8` can exceed 8 chars → distance fails to attach | **Fixed** ed1c7985 (prefix match) |

Live re-verified after the fixes on the lab host: all seven services identified through their own processes, one row per role, disk≠HEAD detection exercised on the gateway clone earlier today (now clean).

skbarber:
**Response to the Codex review (2 findings) — both fixed.**

| # | Finding | Disposition |
|---|---|---|
| P2 | `--local-only --summary` printed "network DOWN" and hid the local section | **Fixed**: in summary mode, `--local-only` prints the stage-3 section (worktrees + origin/master) instead — verified: `scripts/fleet_status.sh --local-only --summary` now shows the local checkouts |
| P3 | `--watch --full` consumed `--full` as the interval → `sleep --full` error loop | **Fixed**: `--watch` takes the next token only if it is all digits; a following flag leaves the 300 s default; a non-numeric value (`--watch abc`) exits 2 with a message — all three verified |

skbarber:
**Codex re-review finding — fixed.** The watch loop now drops the token after `--watch` only when it is all digits (the interval the parser actually consumed); a flag there is passed to the child. Verified by running watch mode briefly for each case and checking the child's output: `--watch --local-only --no-fetch` and `--watch --experiment Undulator --local-only --no-fetch` rerun with their flags intact (no usage error), `--watch 3 …` drops the `3` and refreshes on that interval, `--watch --full …` keeps `--full`.

PR #759, opened 2026-09-02
ImageAnalysis 1.13.2: untrack third_party_sdks/wavekit_43 (gitignore rule never applied)

skbarber:
## What

Removes `ImageAnalysis/image_analysis/third_party_sdks/wavekit_43/` from the git index — 538 files, ~720 MB on disk (Imagine Optic WaveKit 4.3 Python bindings + Intel MKL DLLs). Files stay on disk in existing checkouts (`git rm --cached`); fresh clones stop pulling them.

## Why

The SDK was committed in a single commit (a3e77ff7, 2025-03-13). The root `.gitignore` rule for `wavekit_43/**` was added afterwards, and git only ignores *untracked* paths, so the rule has been a no-op ever since. Seven of the eight largest blobs in the repo's history are these MKL DLLs. First item of the weekly cleanup-day effort.

## Changes (per concern)

| Concern | LOC |
|---|---|
| Untrack `wavekit_43/` | −538 files (binary + vendor py) |
| `third_party_sdks/README.md` — what lives there, import path, install layout, history note | +25 |
| `ImageAnalysis/CLAUDE.md` — one tree line pointing at that README | +1 |
| `ImageAnalysis/CHANGELOG.md` + `pyproject.toml` patch bump 1.13.1 → 1.13.2 | +14 / 1 |
| Root `.gitignore` | see below |

**Judgment-call addition (cheap to veto):** the gitignore rule is widened from the two `…/wavekit_43/**` lines to `**/third_party_sdks/**` with `!**/third_party_sdks/README.md`, so any future vendor SDK dropped into a `third_party_sdks/` directory is ignored by construction rather than needing its own rule. Revert to the narrow rule if you'd rather keep ignores explicit.

## ⚠️ Pulling this on an existing checkout deletes the SDK from disk

`git rm --cached` keeps the files in the checkout that made the commit, but every *other* clone that pulls master afterwards will have git remove `wavekit_43/` from the working tree (tracked → absent between commits; ignore rules do not protect that). **On the HASO analysis box and on your main checkout: copy `ImageAnalysis/image_analysis/third_party_sdks/wavekit_43/` aside before pulling, move it back after.** Documented in the new README (second commit).

## Not in this PR

- **History purge.** The blobs remain in every commit since March 2025; `.git` stays ~590 MB. Rewriting history invalidates every clone and open branch on a public repo — separate decision, separate day.
[...]

PR #781, opened 2026-09-04
scripts/commit.sh: skip staged deletions in the re-stage loop (#761)

### LINKED ISSUE #761: scripts/commit.sh: staged deletion of a now-gitignored file that still exists on disk aborts the commit
Found while landing #759 (untracking `third_party_sdks/wavekit_43/`).

**Repro:** `git rm -r --cached <dir>` where `<dir>` stays on disk and is matched by `.gitignore`; `git add` the other files; run `scripts/commit.sh -m ...`.

**What happens:** the re-stage loop keeps every staged path that still exists on disk and passes it to `git add --`. The staged deletions exist on disk, are now ignored, so `git add` refuses them ("Use -f if you really want to add them") and `set -e` aborts before `git commit`. HEAD does not advance; the tail of the output looks like a normal hook run.

**Fix:** skip paths whose index status is `D` (or run `git check-ignore -q` and skip matches) in the re-stage loop. One-liner; a pinning test would stage a deletion of an ignored-on-disk file.

**Workaround used:** move the directory aside for the commit, move it back after.

skbarber:
## What

`scripts/commit.sh`'s re-stage loop now filters on the **index status** of each staged path (`git diff --cached --name-status --no-renames -z`) and skips status `D` outright, instead of relying only on whether the path still exists on disk.

Pinning test: `tests/test_commit_sh.py` (root `tests/`, unmarked, so CI's root step runs it) drives the script end to end in a throwaway git repo — a copy of the script inside the repo (it `cd`s to its own `../`), a stub `pre-commit` on PATH that rewrites staged files like a real fixer would, and a stub `poetry` that fails so runner selection lands on the stub. Four cases: the #761 shape (`git rm --cached` of a now-ignored file that stays on disk), the same untrack of a *non*-ignored file (which the old loop silently re-added — see below), the plain `git rm` shape, and a staged `git mv` + a following staged edit (pins `--no-renames`: without it the three-field rename record desynchronises the status/path pairing and the fixer's rewrite of the destination and every later path is dropped).

## Why

The loop kept every staged path that exists on disk and handed it to `git add --`. A deletion staged with `git rm --cached` of a file that stays on disk and is now matched by `.gitignore` therefore reached `git add`, which refuses ignored paths ("Use -f if you really want to add them"), and `set -e` aborted before `git commit` — HEAD did not advance while the output tail looked like a normal hook run. Hit while landing #759.

The review surfaced a quieter variant of the same bug: when the untracked file was **not** ignored, `git add` accepted it and silently re-added it — the commit went through with the user's staged untracking reverted. Keying on the index status fixes both.

`--no-renames` keeps every record a `<status>\0<path>\0` pair (a detected rename would emit three fields); a rename then shows as `D src` + `A dst`, which is exactly the shape the loop wants. The on-disk existence test stays as a second guard (a staged add/modify whose file was since removed would otherwise be turned into a deletion by `git add`) — flagging that as a judgment call, cheap to drop.

## Scope

Tooling only (`scripts/` + root `tests/`) — no package code changed, so no version bump / CHANGELOG. No hardware surface.

- `scripts/commit.sh`: +20 / -8
- `tests/test_commit_sh.py`: +206 (new)

## Tests

Tooling only — no hardware verification applies.

- `./scripts/check.sh` (lint + root-tests) at c7820da5: **6 passed, 62 deselected** (4 of them `test_commit_sh.py`); lint clean.
- Mutation-checked locally: against the unfixed `master` script the #761 test fails (`git add` refusal → non-zero exit); removing the `D` skip fails both untrack tests (2 failed); removing `--no-renames` fails the rename test (1 failed).
- `./scripts/check.sh --all` at 558f7006 (root-env layer, the one this diff can affect): root-tests **4 passed, 62 deselected** (6 at c7820da5, above); ImageAnalysis **250 passed, 1 skipped, 10 deselected**; ScanAnalysis **213 passed, 1 skipped, 11 deselected**; GEECS-Data-Utils **325 passed, 3 skipped**; GEECS-Schemas **286 passed, 6 deselected**; `pre-commit run --all-files` clean.
- The eight own-env suites (GeecsBluesky, GeecsCAGateway, GeecsPvaGateway, GEECS-Console, GEECS-Core, GEECS-DataPortal, GEECS-LogTriage, GEECS-MCP) errored at collection with `ModuleNotFoundError` for their own deps — the fresh worktree has no per-package venvs (env-shaped per /env-doctor, not code). This diff touches none of those packages; CI runs each from its own installed env and is the authoritative result for them.

Closes #761

### DIFF
--- a/scripts/commit.sh
+++ b/scripts/commit.sh
@@ -32,7 +32,7 @@ set -euo pipefail
 # Run from the repo root, wherever the script was invoked from. This pins
 # `poetry` below to the ROOT env (a package cwd would resolve that package's
 # env instead), and makes the root-relative paths that `git diff --cached
-# --name-only` emits line up with the `git add` in the re-staging loop —
+# --name-status` emits line up with the `git add` in the re-staging loop —
 # which silently re-staged nothing when invoked from a subdirectory.
 # Consequence: any pathspec args you pass through to `git commit` are
 # interpreted relative to the repo root, not your shell's cwd.
@@ -101,6 +101,8 @@ pc run || true
 # matched by .gitignore): the path exists, so an existence test would pass it
 # to `git add`, which refuses ignored paths ("Use -f if you really want to add
 # them") and, under set -e, aborted the script before `git commit` (#761).
+# Worse, when the untracked file was NOT ignored, `git add` silently re-added
+# it: the commit went through with the user's staged untracking reverted.
 # The existence test stays as a second guard: a staged add/modify whose file
 # was since removed from the worktree would otherwise be turned into a
 # deletion by `git add` — something the user never staged.
--- a/scripts/commit.sh
+++ b/scripts/commit.sh
@@ -40,11 +40,14 @@ REPO_ROOT="$(cd "$(dirname "$0")/.." && pwd)"
 cd "$REPO_ROOT"
 
 # The files staged for this commit, recorded NUL-delimited in a temp file
-# (space-safe; bash variables cannot hold NUL, so a file — not a var).
+# as <status>\0<path>\0 pairs (space-safe; bash variables cannot hold NUL, so
+# a file — not a var). --no-renames keeps every record a pair: a detected
+# rename would otherwise emit three fields (R<score>, src, dst), and
+# splitting it into D src + A dst is exactly the shape the loop below wants.
 staged="$(mktemp)"
 existing="$(mktemp)"
 trap 'rm -f "$staged" "$existing"' EXIT
-git diff --cached --name-only -z >"$staged"
+git diff --cached --name-status --no-renames -z >"$staged"
 if [ ! -s "$staged" ]; then
     echo "commit.sh: nothing staged — 'git add' your changes first." >&2
     exit 1
@@ -91,11 +94,18 @@ fi
 pc run || true
 
 # Re-stage exactly the originally-staged files so the fixes are included —
-# but only those that still exist on disk. A staged *deletion* is in neither
-# the index nor the worktree, so `git add` (even with -A) rejects its pathspec
-# as fatal; and there is nothing to re-stage anyway, since no hook can have
-# rewritten a file that does not exist. The deletion stays staged as-is.
-while IFS= read -r -d '' path; do
+# except staged *deletions* (index status D), which have nothing to re-stage:
+# no hook can have rewritten a file the commit removes, and the deletion stays
+# staged as-is. Filtering on the index status rather than on-disk existence
+# matters for `git rm --cached` (untracking a file that stays on disk, now
+# matched by .gitignore): the path exists, so an existence test would pass it
+# to `git add`, which refuses ignored paths ("Use -f if you really want to add
+# them") and, under set -e, aborted the script before `git commit` (#761).
+# The existence test stays as a second guard: a staged add/modify whose file
+# was since removed from the worktree would otherwise be turned into a
+# deletion by `git add` — something the user never staged.
+while IFS= read -r -d '' status && IFS= read -r -d '' path; do
+    case "$status" in D*) continue ;; esac
     if [ -e "$path" ] || [ -L "$path" ]; then
         printf '%s\0' "$path"
     fi
--- a/tests/test_commit_sh.py
+++ b/tests/test_commit_sh.py
@@ -15,16 +15,26 @@ import os
 import shutil
 import stat
 import subprocess
+import sys
 from pathlib import Path
 
 import pytest
 
+# The script is bash and the harness stubs are bash shebang scripts; on a
+# Windows dev host this is a skip, not a failure (CI runs root tests/ on Linux).
+pytestmark = pytest.mark.skipif(
+    sys.platform == "win32" or shutil.which("bash") is None,
+    reason="commit.sh and its test harness need bash",
+)
+
 REPO_ROOT = Path(__file__).resolve().parents[1]
 COMMIT_SH = REPO_ROOT / "scripts" / "commit.sh"
 
-# What the stub pre-commit appends to keep.txt, standing in for a real
-# auto-fixing hook (ruff-format, end-of-file-fixer) rewriting a staged file.
+# What the stub pre-commit appends to each of HOOK_TARGETS that exists,
+# standing in for a real auto-fixing hook (ruff-format, end-of-file-fixer)
+# rewriting a staged file.
 HOOK_SUFFIX = "rewritten-by-hook\n"
+HOOK_TARGETS = ("keep.txt", "new.txt")
 
 
 def _write_executable(path: Path, body: str) -> None:
@@ -39,9 +49,9 @@ def repo(tmp_path: Path) -> tuple[Path, dict[str, str]]:
     Layout after the fixture: ``keep.txt`` and ``vendor/blob.txt`` are
     tracked and committed; ``scripts/commit.sh`` is a copy of the real one
     (untracked — irrelevant to the scenarios); ``stubbin/`` holds a
-    ``pre-commit`` that appends :data:`HOOK_SUFFIX` to ``keep.txt`` and a
-    ``poetry`` that fails, which steers the script's runner selection onto
-    the stub pre-commit.
+    ``pre-commit`` that appends :data:`HOOK_SUFFIX` to each existing
+    :data:`HOOK_TARGETS` file and a ``poetry`` that fails, which steers the
+    script's runner selection onto the stub pre-commit.
     """
     work = tmp_path / "work"
     (work / "scripts").mkdir(parents=True)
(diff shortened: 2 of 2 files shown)


skbarber:

Reviewed from `.claude/worktrees/commit-sh-ignored-deletions` against `origin/master`, with the root `CLAUDE.md`, `CONTRIBUTING.md`, `scripts/check.sh`, `.github/workflows/unit-tests.yml`, `.pre-commit-config.yaml`, the `/land` `/check` `/env-doctor` skills, and `tests/test_skill_frontmatter.py` as context. `poetry run pytest tests/test_commit_sh.py -q` → 2 passed. I also drove `master`'s script, the PR's script, and two mutants (`D*` skip removed; `--no-renames` removed) through scenarios the tests do not cover, in a throwaway repo with a stub fixer. The fix itself is correct in every scenario I constructed; the surviving findings are about what the new test pins and two small consistency nits.

### Findings (ranked)

**1. Medium — `--no-renames` is load-bearing but unpinned; both new tests pass with it removed.**
Git's `diff.renames` defaults to `true` for the porcelain `git diff` since 2.9, so without `--no-renames` a staged `git mv old.txt new.txt` emits `R100\0old.txt\0new.txt\0` and the two-field `read status && read path` loop desynchronizes. Reproduced with the mutant (`--name-status -z`, no `--no-renames`):

- pure rename, stub hook rewrites `new.txt` → `rc=0`, commit created, but `HEAD:new.txt` lacks the hook's fix and the worktree ends dirty (` M new.txt`);
- rename + a staged edit of `zed.txt` (the record *after* the rename) → **both** files' fixes are dropped (` M new.txt`, ` M zed.txt`) because every subsequent pair is offset by one field.

In real use the commit-time pre-commit run then fails with "files were modified by this hook" — exactly the failure the script exists to prevent — or, for hooks that are not idempotent-checked, commits stale content. Raw shapes for the record: renames-on → `R100~old.txt~new.txt~M~zed.txt~`; `--no-renames` → `A~new.txt~D~old.txt~M~zed.txt~`. The PR's script handles both rename scenarios correctly (`HEAD:new.txt = old|HOOK|`, `HEAD:zed.txt = z|edit|HOOK|`, clean worktree), so this is purely a test gap: one more case in `tests/test_commit_sh.py` that does `git mv` + asserts the destination carries `HOOK_SUFFIX` would make the flag's removal fail CI. Note the stub fixer needs to rewrite the *destination* path for this to bite (a `--diff-filter=AMT` stub, as I first wrote, filters out `R` and passes vacuously).

**2. Low — stale comment left inside the same file.** `scripts/commit.sh` lines 34–35 still read "the root-relative paths that `git diff --cached --name-only` emits line up with the `git add` in the re-staging loop", but the diff changed that command to `--name-status --no-renames -z`. Internal contradiction introduced by this diff; one-word fix.

**3. Low — the fix cures a nastier, silent variant that neither the PR body nor a test records.** On `master`, `git rm --cached blob.txt` (file **not** ignored, still on disk) plus a staged edit to `keep.txt` → `rc=0`, HEAD advances, and `blob.txt` is still tracked: the re-add loop silently reverted the user's staged untracking and the commit "succeeded". With the PR's script: `rc=0`, `blob.txt` untracked. This is arguably the stronger motivation for keying on index status (it is data-changing, not merely aborting), and it is a two-line third test case (`git rm --cached` without touching `.gitignore`, assert path absent from `ls-files` and HEAD advanced). Worth a sentence in the "Why" and/or the comment block, since the comment currently frames the change only around the ignored-path refusal.

**4. Low — test portability on Windows dev machines.** The test hard-requires `bash` on PATH, `#!/usr/bin/env bash` shebang stubs made executable via `chmod`, `git init -b` (git ≥ 2.28) and `GIT_CONFIG_GLOBAL` (git ≥ 2.32). CI is fine — the root `tests/` step runs only on the Linux job; the `windows-latest` job runs Console + `qs_client` only — but a developer on a Windows host running `poetry run pytest tests` will get a hard failure, not a skip. A `pytest.mark.skipif(shutil.which("bash") is None or sys.platform == "win32", ...)` (or a `pytest.skip` in the fixture) keeps the test honest without changing what CI checks. Judgment call; flagging because the fleet does include Windows hosts.

**5. Nit — duplicate git runner in the test file.** The `repo` fixture defines an inner `git(*args)` closure and the module defines `_git(work, env, *args)` doing the same `subprocess.run([...], check=True, capture_output=True, text=True)`. Having the fixture use `_git(work, env, ...)` removes the duplicate.

### Checked and not raised

- **Correctness of the loop itself**: `--no-renames` also disables copy detection (`detect_rename = 0` covers `-C`), so statuses are single letters and `D*` ≡ `D`. `-z` puts NUL between status and path, so the pair read is right. Staged `A` whose file was since deleted (the "second guard") commits the staged blob rather than converting it to a deletion — verified (`gone.txt@HEAD present: yes`) on both master and the fix; behaviour unchanged. `U` (unmerged) paths reach `git add` and would be marked resolved — pre-existing, not touched by this diff, and `git commit` refuses unmerged trees anyway.
- **Mutant `D*` removed**: `test_staged_untrack_of_ignored_file_still_on_disk_commits` fails (`git add` refusal, `rc=1`), confirming the PR's claim that the primary test is falsifiable.
- **Redundancy**: `scripts/check.sh:117` is the only other staged/changed-file enumerator (`git diff --name-only -z "$MB"`, a different purpose — changed-package detection against merge-base); no other shell-script test harness exists in the repo (`tests/test_commit_sh.py` is the only `subprocess ["bash", ...]` test outside `.claude/`). No duplication.
- **Placement**: root `tests/` is the right home — precedent `tests/test_skill_frontmatter.py` (unmarked, repo-tooling test), and both `scripts/check.sh` `root-tests` and the CI root step collect it with `-m "not integration and not gui"`.
- **Policy docs**: `CLAUDE.md:118`, `CONTRIBUTING.md:108`, `docs/tutorials/getting_started.md:166`, `/land` step 5, `/env-doctor:81`, `/check:47` describe commit.sh only at the "apply fixes, re-stage, commit" level — nothing contradicts the new behaviour. No version bump / CHANGELOG owed (tooling only), consistent with the release policy.
- **Public-repo hygiene**: `test@example.invalid`, no hostnames/paths.
- Pre-existing, not this PR's: the "Root tests/ is currently all integration-marked" comments in `.github/workflows/unit-tests.yml:29` and `scripts/check.sh:229` were already stale once `test_skill_frontmatter.py` landed; this PR makes them a little more stale. Optional touch, not a finding against the diff.

### Suggested disposition
Merge-ready on correctness. Finding 1 is the one I'd fix before merge (a rename case is ~10 lines and pins the only non-obvious flag in the change); 2 is a one-word comment edit; 3–5 are optional.

skbarber:
## Review disposition (per /land)

All five findings **fixed** in c7820da5 and each confirmed by the **same reviewer subagent** via independent mutation testing on scratch copies (its confirm report is summarised inline):

| # | Finding | Disposition | Reviewer confirm |
|---|---|---|---|
| 1 | `--no-renames` load-bearing but unpinned | **Fixed** — `test_staged_rename_re_stages_the_destination` (`git mv keep.txt new.txt` + a later staged `zed.txt`); stub fixer rewrites every existing `HOOK_TARGETS` file incl. the rename destination | Mutant without `--no-renames`: 1 failed / 3 passed at `HEAD:new.txt` lacking the hook suffix — the exact desync |
| 2 | Stale `--name-only` comment | **Fixed** — now `--name-status` | Only `--name-status` remains (lines 35, 50) |
| 3 | Silent re-add of a non-ignored `git rm --cached` unrecorded | **Fixed** — `test_staged_untrack_of_unignored_file_is_not_silently_reverted` + a sentence in the loop's comment block; PR "Why" updated | Mutant without the `D*` skip: 2 failed / 2 passed (`vendor/blob.txt` re-added; #761 case rc=1) |
| 4 | Hard failure on Windows / bash-less hosts | **Fixed** — module-level `pytestmark = skipif(win32 or no bash)` | Simulated bash-less PATH: 4 skipped before the fixture runs |
| 5 | Duplicate git runner in the test | **Fixed** — fixture uses module-level `_git` | No `def git` closure remains |

Reviewer's non-finding note, **waived**: the on-disk existence "second guard" is unpinned (a mutant replacing it with `if true` still passes all 4 tests). It is the judgment-call addition already flagged in the PR body as cheap to drop, and pinning it would mean testing a state (`A`/`M` staged, file since deleted) the script never promised to handle — owner's call whether to keep or drop the guard.

Pre-existing, out of scope: the "Root tests/ is currently all integration-marked" comments in `.github/workflows/unit-tests.yml` and `scripts/check.sh` were already stale since `test_skill_frontmatter.py` landed — left for a separate tidy.

Root-tests at c7820da5: 6 passed, 62 deselected; lint clean. Ready for maintainer merge once CI reports on c7820da5.

PR #782, opened 2026-09-04
bootstrap_host.sh: detect a linked git worktree posing as a service clone

skbarber:
## What

`deploy/bootstrap_host.sh` distinguishes a linked `git worktree` from a real clone (`git rev-parse --git-dir` ≠ `--git-common-dir`), warns, and leaves it untouched instead of trying to `git clone` into the existing directory (the previous `-d .git` test is false for a worktree, whose `.git` is a file). `docs/platform/site_profile.md` records the rule: a clone means a clone, not a worktree. No package code; `deploy/` + docs only.

## Why

Phase 3 preflight on the interim host (today): `qs-checkout` is a worktree of the gateway's clone `~/GEECS-Plugins`, both with `master` checked out. A `git pull` in one advances the shared branch pointer without touching the other's files — which is the root cause of the 2026-09-03 gateway disk≠HEAD incident (`/fleet-status` found the gateway running an August tree under a September HEAD). Phase 3 replaces the worktree with a real clone during the maintenance window; the bootstrap must not mistake it for one in the meantime.

## Tests

`bash -n`; lint clean. Local reproduction: a scratch root where `qs-checkout` is a linked worktree of `gateway-checkout` — `bootstrap_host.sh --dry-run --no-install` fetches the clone, plans the missing clone, and prints the WARNING for the worktree (verified output in the commit message's scenario).

## Hardware verification

Non-invasive live use is next: this is the version of the bootstrap that runs on the host for Phase 3 staging (a fresh `gateway-checkout` clone + env; `qs-checkout` is replaced in the cutover window).

### DIFF
--- a/deploy/bootstrap_host.sh
+++ b/deploy/bootstrap_host.sh
@@ -82,7 +82,18 @@ echo "  git $(git --version 2>/dev/null | awk '{print $3}'), $(python3.11 --vers
 say "clones (one per service family, at $REF)"
 for c in $(for s in $SERVICES; do wanted "$s" && clone_of "$s"; done | sort -u); do
     dir="$GEECS_CHECKOUT_ROOT/$c"
-    if [ -d "$dir/.git" ]; then
+    if [ -e "$dir/.git" ]; then
+        gitdir="$(git -C "$dir" rev-parse --git-dir 2>/dev/null || true)"
+        common="$(git -C "$dir" rev-parse --git-common-dir 2>/dev/null || true)"
+        if [ -n "$gitdir" ] && [ "$gitdir" != "$common" ]; then
+            # A linked worktree shares its object store AND its branch refs with
+            # another clone: `git pull` here advances that clone's branch pointer
+            # without touching its files (the 2026-09-03 gateway disk≠HEAD
+            # incident). Not a clone of its own — replace it in a maintenance
+            # window: stop its services, mv it aside, clone, poetry install.
+            echo "  $c: WARNING — linked worktree of $(cd "$dir" && cd "$common/.." && pwd), not a clone of its own; left untouched (see the site profile page)"
+            continue
+        fi
         echo "  $c: exists ($(git -C "$dir" rev-parse --abbrev-ref HEAD) $(git -C "$dir" rev-parse --short=8 HEAD)) — fetching, NOT moving HEAD (a pull is a deploy; do it per service, deliberately)"
         run git -C "$dir" fetch --quiet origin
     else
--- a/deploy/bootstrap_host.sh
+++ b/deploy/bootstrap_host.sh
@@ -80,21 +80,33 @@ echo "  git $(git --version 2>/dev/null | awk '{print $3}'), $(python3.11 --vers
 [ -d "$GEECS_CONFIGS_ROOT" ] && echo "  configs repo at $GEECS_CONFIGS_ROOT" || echo "  WARNING: configs repo not at $GEECS_CONFIGS_ROOT"
 
 say "clones (one per service family, at $REF)"
+# Clone dirs that are NOT usable clones are recorded here; every later stage
+# (environments, units, root steps) skips their services, so the run cannot
+# end with a clean tail after a warning scrolled off the top.
+SKIP_CLONES=" "
+skipped_clone() { case "$SKIP_CLONES" in *" $1 "*) return 0;; *) return 1;; esac; }
 for c in $(for s in $SERVICES; do wanted "$s" && clone_of "$s"; done | sort -u); do
     dir="$GEECS_CHECKOUT_ROOT/$c"
     if [ -e "$dir/.git" ]; then
         gitdir="$(git -C "$dir" rev-parse --git-dir 2>/dev/null || true)"
         common="$(git -C "$dir" rev-parse --git-common-dir 2>/dev/null || true)"
-        if [ -n "$gitdir" ] && [ "$gitdir" != "$common" ]; then
-            # A linked worktree shares its object store AND its branch refs with
-            # another clone: `git pull` here advances that clone's branch pointer
-            # without touching its files (the 2026-09-03 gateway disk≠HEAD
-            # incident). Not a clone of its own — replace it in a maintenance
-            # window: stop its services, mv it aside, clone, poetry install.
-            echo "  $c: WARNING — linked worktree of $(cd "$dir" && cd "$common/.." && pwd), not a clone of its own; left untouched (see the site profile page)"
-            continue
+        if [ -z "$gitdir" ]; then
+            echo "  $c: WARNING — .git present but git cannot read it (a worktree whose main clone was removed, or corrupt); skipped — fix by hand"
+            SKIP_CLONES="$SKIP_CLONES$c "; continue
         fi
-        echo "  $c: exists ($(git -C "$dir" rev-parse --abbrev-ref HEAD) $(git -C "$dir" rev-parse --short=8 HEAD)) — fetching, NOT moving HEAD (a pull is a deploy; do it per service, deliberately)"
+        case "$gitdir" in
+            "$common"/worktrees/*)
+                # A linked worktree shares its object store AND its branch refs
+                # with another clone. If the same branch is checked out in both
+                # (git allows that only via --ignore-other-worktrees/--force —
+                # how the 2026-09-03 state arose), a pull in one moves the
+                # other's HEAD without a checkout, and that service runs an older
+                # tree than its HEAD claims. Not a clone of its own — see the
+                # site profile page for the replacement recipe.
+                echo "  $c: WARNING — linked worktree of $(cd "$dir" && cd "$common/.." && pwd), not a clone of its own; skipped (see the site profile page)"
+                SKIP_CLONES="$SKIP_CLONES$c "; continue ;;
+        esac
+        echo "  $c: exists ($(git -C "$dir" rev-parse --abbrev-ref HEAD 2>/dev/null) $(git -C "$dir" rev-parse --short=8 HEAD 2>/dev/null)) — fetching, NOT moving HEAD (a pull is a deploy; do it per service, deliberately)"
         run git -C "$dir" fetch --quiet origin
     else
         run git clone --quiet --branch "$REF" "$GEECS_REPO_URL" "$dir"
@@ -105,6 +117,7 @@ say "environments (inside each package dir, with that service's extras)"
 for s in $SERVICES; do
     wanted "$s" || continue
     dir="$GEECS_CHECKOUT_ROOT/$(clone_of "$s")/$(pkgdir_of "$s")"
+    skipped_clone "$(clone_of "$s")" && { echo "  $s: skipped — $(clone_of "$s") is not a usable clone (see the clones stage)"; continue; }
     [ "$INSTALL" -eq 1 ] || { echo "  $s: skipped (--no-install)"; continue; }
     if [ ! -d "$dir" ]; then
         echo "  $s: clone not present yet ($dir) — rerun after the clone step has run for real"; continue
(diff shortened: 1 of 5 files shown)


skbarber:

Scope: `deploy/bootstrap_host.sh` (+11/-1) and `docs/platform/site_profile.md` (+9). Exercised with `bash -n` and `--dry-run` against scratch roots holding a real clone, a linked worktree (with and without the package dirs/unit templates a real checkout has), a dangling worktree whose main clone was deleted, a `--separate-git-dir` clone, a submodule, a main clone with a worktree hanging off it, a stray empty `.git`, and a PATH without `git`. Git 2.39.

### 1. "Left untouched / leaves it alone" is only true of the clones stage — the worktree is still installed into, baked from, and given a unit and an `enable --now` line

**Severity:** Medium  **Lens:** Correctness (script ↔ doc contract)

**Scenario / evidence:** With `qs-checkout` a linked worktree of `gateway-checkout` and package dirs present (RUN 8, `--dry-run --only qserver,capture,mcp`), the output after the one-line WARNING was:

```
== environments ...
  [dry] poetry install --quiet --extras ca tiled qserver        (inside <root>/qs-checkout/GeecsBluesky)
  [dry] poetry install --quiet --extras ca tiled qserver capture
  [dry] python3.11 -m venv <root>/geecs-mcp-venv
  [dry] .../pip install --quiet --upgrade <root>/qs-checkout/GEECS-MCP[analysis-run]
== units ...
  [dry] render_units.sh ... <root>/qs-checkout/GeecsBluesky/qserver/deploy/geecs-qserver.service ...
== root steps ...
  sudo systemctl enable --now geecs-qserver / geecs-capture / geecs-mcp
rc=0
```

Lines 107–125 and 189–190 key off `-d "$dir"` / `-f "$t"` only, so the `continue` at line 95 skips just the fetch. The site-profile text says the bootstrap "warns, and leaves it alone"; in fact it fully provisions the worktree-backed services (poetry env, baked MCP venv, staged units, enable lines) and exits 0, with the warning scrolled off several sections earlier. An operator running the printed root steps would enable three units on the checkout the docs just told them is not a clone.

**Suggested fix:** Pick one contract and make both files say it. Either (a) record flagged clone dirs (e.g. `SKIP_CLONES="$SKIP_CLONES $c"`) and have the environments/units/root-steps loops skip services whose `clone_of` is flagged, printing a one-liner per service; or (b) keep provisioning (defensible — the tree on disk *is* what the service runs) but change "left untouched" to "git state left untouched", re-emit the warning in the root-steps block (and to stderr), and reword the doc sentence. Either way the run should end with a visible reminder, not a clean rc=0 tail.

### 2. An unreadable `.git` (dangling worktree, corrupt dir) now routes through the "exists" branch, prints `exists ( )` with leaked `fatal:` lines, and aborts the rest of the bootstrap

**Severity:** Low–Medium  **Lens:** Correctness

**Scenario / evidence:** `qs-checkout` is a linked worktree whose main clone was removed (RUN 3/4). `-e "$dir/.git"` is true (it is a file); both `rev-parse` calls fail (`|| true` → empty), `"" != ""` is false, so we fall into the exists branch:

```
fatal: not a git repository: <S>/tmpmain/.git/worktrees/qs-checkout   (×2, from the un-silenced rev-parse on line 97)
  qs-checkout: exists ( ) — fetching, NOT moving HEAD ...
  $ git -C <root>/qs-checkout fetch --quiet origin
rc=128        (set -e; environments/config.ini/units for every service never run)
```

Pre-PR this shape also aborted (`git clone` into a non-empty dir), so it is not a regression — but the PR widened the gate from `-d` to `-e` precisely to catch `.git`-file shapes, and this is the other common `.git`-file shape after a "move it aside" repair goes wrong (finding 5). The two anomalies ("is a worktree" vs "git can't read it") are handled inconsistently: one warns-and-continues, the other kills the run half-way with an opaque message.

**Suggested fix:** Add an explicit branch before the comparison: `if [ -z "$gitdir" ]; then echo "  $c: WARNING — .git present but git cannot read it (dangling worktree whose main clone was removed, or corrupt); left untouched — fix by hand"; continue; fi`. Optionally `2>/dev/null` the two `rev-parse` calls on line 97.

### 3. The `--git-dir != --git-common-dir` test can misfire when the checkout root sits inside an enclosing repository — exactly the relative-vs-absolute quirk you asked about

**Severity:** Low  **Lens:** Correctness

**Scenario / evidence:** From the top level of a clone both outputs are `.git`; from a linked worktree both are absolute — so far so good. But from any *subdirectory* of a repo, git 2.39 prints `--git-dir` absolute and `--git-common-dir` relative (`../.git`), so they differ. The `-e "$dir/.git"` gate normally guarantees `$dir` is a repo top level — except when `$dir/.git` exists but is not a valid gitdir (git then walks upward). RUN 7: `GEECS_CHECKOUT_ROOT` is itself a repo (a git-managed `$HOME` is a real pattern, and `/home/geecs` is the example root) and `qs-checkout/.git` is a stray empty dir →

```
  qs-checkout: WARNING — linked worktree of <root>, not a clone of its own; left untouched
```

Wrong diagnosis (and the service is then silently skipped for fetch while still provisioned per finding 1). With no enclosing repo (RUN 6) the same input falls into finding 2 instead.

**Suggested fix:** Test the actual linked-worktree shape rather than inequality: `case "$gitdir" in "$common"/worktrees/*) …linked…;; esac` (both are absolute in the linked case, so this is exact and immune to the relative form), or additionally require `[ "$(git -C "$dir" rev-parse --show-toplevel)" = "$(cd "$dir" && pwd -P)" ]` before trusting either branch.

### 4. The doc's mechanism sentence describes a state git refuses to create by default; as written it will not survive a reader's own test

**Severity:** Low  **Lens:** Correctness (docs)

**Scenario / evidence:** "`git pull` in one advances the other's branch pointer without touching its files" requires *both* worktrees to have the same branch checked out. Verified: `git worktree add … master` while master is checked out → `fatal: 'master' is already checked out`; `git checkout master` in the linked tree → same refusal; `git branch -f master origin/master` and `git fetch origin master:master` from the other tree → refused. Only after `checkout --ignore-other-worktrees` (or `--force`) does the described failure occur — and then it does exactly as stated (`gw HEAD` moved, `gw/f` stale, `git status` shows `M  f`). With the default shape (one detached, one on master) a pull moves nothing in the other tree. A maintainer who tests the sentence, sees git refuse, and concludes the rule is folklore is the failure mode here.

**Suggested fix:** One clause: "…shares its object store and branch refs; if the same branch ends up checked out in both (git only allows this via `--ignore-other-worktrees`/`--force`, which is how the 2026-09-03 state arose), a pull in one moves the other's HEAD with no checkout — the other service then runs an older tree than its HEAD claims." Same for the script comment at lines 89–93 if you want them to agree.

### 5. The replacement recipe leaves worktree debris behind and duplicates what the bootstrap already automates

**Severity:** Low  **Lens:** Correctness (docs, executable as written?)

**Scenario / evidence:** "stop its services, move it aside, clone, `poetry install`". After `mv`, the main clone still carries `.git/worktrees/qs-checkout/` (its `gitdir` file now points at the old path → `git worktree list` in `gateway-checkout` shows a prunable phantom), and the moved directory's `.git` file still points into the shared store, so any later `git` command inside it operates on the gateway clone's refs — the very hazard being removed. And "clone, poetry install" is the bootstrap's own job (the page says two sections later that rerunning it after a clone appears is the intended flow).

**Suggested fix:** "stop its services, `git -C <main clone> worktree remove --force <dir>` (or `mv` it aside and `git -C <main clone> worktree prune`), then rerun `deploy/bootstrap_host.sh` — it clones and installs — and restart."

### 6. The rule lives only where the symptom was found; the canonical rule section and the routine observer don't carry it

**Severity:** Low  **Lens:** Placement

**Scenario / evidence:** `docs/platform/site_profile.md` explicitly defers the rule and its reasons to `docs/platform/fleet_map.md` § "One clone per service" (line 91), and that section ends with the operative sentence "Never `git pull` a clone that isn't the one your service runs from" — yet the worktree corollary (a worktree *looks* like its own clone and isn't) is added only to the site profile. Likewise, `scripts/fleet_status.sh`'s remote `emit` (lines ~411–450) is what the runbook says to trust after a deploy ("every row systemd / clean / matching"); it reports branch/sha/dirty and even has disk≠HEAD detection via `write-tree` (which would have shown the *symptom* in the incident), but has no notion of "this checkout is a worktree of X", and it runs on every check whereas the bootstrap runs on fresh hosts/reruns. Nothing else in the repo detects worktree-ness of host clones (checked: `scripts/*.sh`, `deploy/*`, `GeecsBluesky/{qserver,capture}/deploy/DEPLOYMENT.md`, `GEECS-DataPortal/DEPLOYMENT.md`, `GEECS-MCP/deploy/DEPLOYMENT.md`; `fleet_status.sh` lines 595–636 list *local* worktrees only; root `CLAUDE.md` line 102 is policy prose about the sibling clone) — so this is a gap, not a duplication.

**Suggested fix:** One bullet in fleet_map's "One clone per service" ("a linked `git worktree` is not a clone — see the site profile"), and the same two `rev-parse` lines in `fleet_status.sh`'s `emit` producing a `worktree_of=<main>` field so the per-row check flags it on every run. The latter can be a follow-up, but it is where a future "why does this row say clean while the code is stale" investigation will want it.

### Checked and refuted

- **`--separate-git-dir` clone** (`.git` is a file): `--git-dir` and `--git-common-dir` both print the same absolute path → correctly treated as a clone. **Submodule**: same, equal absolute paths. **Bare repo**: no `$dir/.git`, falls to the pre-existing clone-into-non-empty-dir failure; not a realistic service checkout shape.
- **Main clone with linked worktrees hanging off it** (`portal-checkout` in RUN 3): `.git` == `.git` → treated as a clone and fetched. Correct — it *is* the real clone.
- **Relative vs absolute from different cwds**: `git -C "$dir"` pins cwd to the clone root, so the clone case is always `.git`/`.git` and the linked case always absolute/absolute; the only route to a mixed form is finding 3.
- **Display expression `$(cd "$dir" && cd "$common/.." && pwd)`**: correct for absolute `common` (→ `<root>/gateway-checkout`) and for the relative form (RUN 7 printed the enclosing repo, consistently with the misdiagnosis). For a worktree of a `--separate-git-dir` main it prints the gitdir's parent rather than the main tree — but `git worktree list --porcelain` prints the gitdir itself in that case, so there is no better one-liner; informational only.
- **`set -e` interaction**: the substitution failing inside an `echo` argument does not abort the script (exit status of the substitution is discarded); confirmed by RUN 3.
- **`git` missing**: real mode exits at the prerequisites stage (`prereq_fail`); dry-run degrades to `exists ( )` + `[dry] git fetch` for every dir, identical in spirit to pre-PR behaviour (RUN 5b).
- **Public-repo hygiene**: the new text names no hostnames, accounts, or home paths. **Versioning**: `deploy/` is not a Poetry package; no bump or CHANGELOG owed. **Contract files traveling with behaviour**: the script's "see the site profile page" pointer and the doc paragraph land in the same commit.
- **Redundancy**: no existing implementation of host-clone worktree detection anywhere in the repo (files listed under finding 6).

---

skbarber:

| # | Finding | Disposition |
|---|---|---|
| 1 | "Left untouched" held only for the fetch; later stages still provisioned the worktree's services | **Fixed** d21e10eb: `SKIP_CLONES` — environments, unit rendering and the enable lines skip a flagged clone's services, the run ends with the warning repeated on stderr; confirmed |
| 2 | Unreadable `.git` aborted the run half-way with leaked `fatal:` lines | **Fixed** d21e10eb: explicit warn-and-skip branch, rev-parse silenced; confirmed |
| 3 | `--git-dir != --git-common-dir` could misfire (relative/absolute forms) | **Fixed** d21e10eb: exact shape `gitdir == common/worktrees/*`; b5efc53a adds the `--show-toplevel == pwd -P` guard for the stray-`.git`-inside-another-repo case; confirmed |
| 4 | Doc described a state git refuses by default | **Fixed** d21e10eb: same-branch precondition + `--ignore-other-worktrees`/`--force` route stated in the page and the script comment; confirmed |
| 5 | Replacement recipe left worktree debris and duplicated the bootstrap | **Fixed** d21e10eb: `worktree remove --force` / prune, then rerun the bootstrap; confirmed |
| 6 | Rule lived only where the symptom was found; the routine observer could not see it | **Fixed** d21e10eb: fleet-map bullet; `scripts/fleet_status.sh` emits `worktree_of=` and the table notes `WORKTREE of <clone>`; confirmed |
| A | `$SKIP_CLONES—` → unbound variable on bash 3.2 under UTF-8 | **Fixed** b5efc53a (`${SKIP_CLONES}—` + comment); confirmed rc=0 in C and UTF-8 |
| B | Install lines printed when nothing was rendered | **Fixed** b5efc53a (gated on rendered units); confirmed |

Reviewer's closing statement: no surviving findings across normal clone, separate-git-dir clone, submodule, main clone with worktrees, linked worktree, dangling worktree, stray `.git` with/without an enclosing repo, symlinked root, git absent, bash 3.2 in C and UTF-8.

Issue #779, opened 2026-09-04
Scan axes fire shots before the stage arrives: set-ack is not motion-complete, readback confirm is opt-in and nothing opts in (data integrity)


skbarber:
## Summary

Every Undulator scan axis is a **plain `CaSettable`** today: the move "completes" when the gateway `:SP` put returns, and the engine fires the shots. Live data from 2026-09-03 shows the MCC stepper (`U_StretchterXMCC Position.Ch3`, scan variable `StretcherMIStaged`) **acknowledges a set before the motion finishes**, so shots were recorded against the commanded position while the stage was somewhere else. The readback-confirm machinery (`kind: motor` → `CaMotor`, `confirm:` → `CaConfirmSettable`) exists and is tested, but it is **opt-in per variable and nothing in the deployed config opts in** — the legacy `scan_devices.yaml` / `composite_variables.yaml` pair carries no `kind:` at all, and the converter default is `setpoint`.

This is a **data-integrity** bug, not just an operator annoyance: the scan grid says one position, the device was at another.

## Evidence (Undulator, 2026-09-03, s-files)

`ScanDataScan005.txt` — 1-D step scan 40→100 by 5, 3 shots/step, first item after Scan004 ended at 100:

| bin | target | readback on all 3 shots |
|---|---|---|
| 1 | 40.0 | **99.999974** (stage had not moved) |
| 2 | 45.0 | 45.000044 (after operator resume) |
| 3+ | 50 … | correct |

`ScanDataScan004.txt` — same axis, 50→100 by 10, mid-scan:

| bin | target | readback on all 3 shots |
|---|---|---|
| 2 | 60.0 | 60.000059 |
| **3** | **70.0** | **60.000059** (stage still at 60) |
| 4 | 80.0 | 79.999954 |

`scan.log` timings for the same moves: set→"done" 0.2 s for 100→40 and 60→70 (impossible physically; ~2.5 units/s otherwise), 2–5 s for the moves the device really did block on. The set to 45, issued while the stage was still travelling toward 40, blocked >10 s (the `GatewaySetpointPut` default budget) → `FAILED MOVE - pausing for operator` → the console showed the second scan "starting paused"; Resume retried and landed instantly because the stage had arrived meanwhile.

So the premise in `CaSettable`'s docstring — *"the :SP put rides GEECS's native blocking set, which completes only when the device reports convergence"* — does **not** hold for this device class. The set acknowledgement is not motion-complete.

## What exists today

[...]

PR #777, opened 2026-09-03
Site profile: site.env + templated units + host bootstrap

skbarber:
## What

The contract that lets this fleet be stood up on a new box, a different network, or another GEECS facility without forking a file — Phase 1+2 of the plan in #775 (bundled: the docs page would otherwise describe scripts that don't exist yet).

- **One home per facility value.** Client side: `config.ini` (exists). Host side: **`/etc/geecs/site.env`** (new), consumed by systemd `EnvironmentFile=`. `deploy/site.env.example` is the HTU/Undulator worked example, every key commented; syntax is systemd's (comments on their own lines only — a trailing `# comment` becomes part of the value).
- **Unit templates.** The five service units (`GeecsCAGateway/deploy`, `GEECS-DataPortal/deploy`, `GeecsBluesky/qserver/deploy`, `GeecsBluesky/capture/deploy`, `GEECS-MCP/deploy`) carry `@PLACEHOLDER@` holes for the account / home / checkout root / poetry path — systemd does **not** expand variables in `WorkingDirectory=`, `User=`, `Environment=`, or the exec path — and `${VARIABLE}` arguments for runtime values (`--experiment ${GEECS_EXPERIMENT}`, `--doc-addr ${GEECS_QS_DOC_ADDR}`). EPICS addressing, `TZ`, `QS_EXPERIMENT` need no hole: `EnvironmentFile=` delivers them under the names the services already read.
- **`deploy/render_units.sh SITE_ENV OUT_DIR`** — one `sed`, refuses to leave a placeholder unfilled, prints the root install lines.
- **`deploy/bootstrap_host.sh SITE_ENV`** — idempotent, unprivileged: per-service clones (`<root>/gateway-checkout`, `portal-checkout`, `qs-checkout` — the fleet map's one-clone-per-service rule with its two exceptions), `poetry install` inside each package dir with that service's extras (table mirrors the runbooks), the MCP venv baked non-editably from `qs-checkout` at `<root>/geecs-mcp-venv` (no `/opt`, no sudo), `config.ini` rendered from `site.env` **only if absent**, units rendered to staging, root steps printed. `--ref`, `--only`, `--no-install`, `--dry-run`.
[...]

PR #785, opened 2026-09-04
GeecsBluesky 0.73.3 + GEECS-Schemas 0.17.1 + GEECS-MCP 0.8.6: warn on images: true for snapshot-role save-set entries (#754)

### LINKED ISSUE #754: images: true on a snapshot-role save-set entry is a silent no-op (any devicetype) — warn at the DB-free role seam
Split out of the #751 adversarial review (finding 1, waived there as a second concern) — pre-existing, not introduced by #751.

## The gap

`images: true` on a **snapshot-role** save-set entry (`role: snapshot`, legacy `synchronous: false`) is a silent no-op for ANY devicetype, DB or no DB:

- the engine never drove native saving for the snapshot role — `session.snapshot` has no `save_images`, `CaSnapshotReadable` sets no `_save_nonscalar_data`, so `session._configure_saving` skips it;
- since #751 (0.70.1) the capture daemon does not own it either.

#751's warning covers only the capture-ownership slice and is deliberately DB-gated (it fires after the devicetype lookup, fail-open), so the misconfiguration stays silent when:

- the device is a non-capture devicetype (HASO, scope) in a snapshot role with `images: true` — never warned, images never written;
- the DB blips in dual-write mode — `GeecsDbDeviceTypes` fail-opens to `{}`, and the existing toggle-off "no capture-eligible devices" warning is gated on `native_image_save=off`, so nothing fires.

## Where the loud signal belongs (DB-free)

- `save_set_to_devices_config` in `GeecsBluesky/geecs_bluesky/scan_request_runner.py` (~line 235), at the `SaveRole.SNAPSHOT` → `synchronous=False` derivation — the one role→mechanics seam. It is also what the client-side `qs_client/submit_preflight._resolve_devices_config` runs, so a console/MCP operator sees the finding **pre-submit** instead of in the worker's scan.log. A `PreflightOutcome` warning check is the existing vocabulary.
- `SaveSetEntry.images` (`GEECS-Schemas/geecs_schemas/save_set.py` ~line 167) and the generated `docs/geecs_schemas/schema_reference.md` promise images for any entry with no role caveat — the description should say the snapshot role ignores it (or the schema should reject the combination outright; owner's call).

## Not in scope

Building real image support for the snapshot role (save child + `acq_timestamp` join column) is the other remedy #702 named — a separate decision.

skbarber:
Closes #754 (split out of the #751 adversarial review, finding 1).

## What

`images: true` on a **snapshot-role** save-set entry (`role: snapshot`, legacy `synchronous: false`) is a silent no-op for every devicetype, DB or no DB: the engine never drove native saving for the snapshot role (`session.snapshot` has no `save_images`, so `_configure_saving` skips it) and since #751 the capture daemon never owns it either. The #751 warning covers only the capture-ownership slice and is DB-gated by design, so a HASO/scope in a snapshot role, or any snapshot camera during a DB blip, stayed silent.

This PR makes the misconfiguration loud at the **DB-free role→mechanics seams** (the save-set derivation, and — after review — the optimizer-requirements merge that can re-create the combination worker-side):

- **GeecsBluesky 0.73.2** — `save_set_to_devices_config` (`scan_request_runner.py`) logs one warning per save set naming the affected devices, right at the `SaveRole.SNAPSHOT → synchronous=False` derivation. `merge_optimizer_device_requirements` (the second seam: it preserves `synchronous=False` and flips save on, or provisions a new async + save device) warns for what the merge *introduced* — the #520 NaN class, which the client preflight cannot see because requirements are provisioned worker-side (review finding 1). The condition and the wording live in two small helpers beside it, `snapshot_images_ignored(devices_config)` / `snapshot_images_ignored_message(...)`, evaluated on the derived config (no DB, no schema import). Because the worker runs this seam on every scan, the warning lands in `scan.log`; because `qs_client/submit_preflight._resolve_devices_config` runs the same seam, the client-side preflight reuses the helper as a new **`snapshot_images`** check — a `PreflightQuestion` (warning the operator can continue past; `passed` outcome when clean), **never a refusal**. The combination is still accepted and the entry's scalars are still recorded.
- **GEECS-MCP 0.8.5** — `_ACKNOWLEDGEABLE_CHECKS` gains `snapshot_images`. This is the third package by necessity, not scope creep: the tuple is the MCP's fail-closed drift guard ("update BOTH when adding a check" in its own comment). Without it, `submit_scan` would refuse such a request with `policy_refusal` *and* refuse the acknowledgement as an unknown name — an unsubmittable request through the MCP.
- **GEECS-Schemas 0.17.1** — `SaveSetEntry.images` / `.role` descriptions and the `SaveRole.SNAPSHOT` docstring now say the snapshot role ignores `images`; `docs/geecs_schemas/schema_reference.md` regenerated with `GEECS-Schemas/tests/generate_schema_reference.py`. Description-only: no validation change.

### Deliberately NOT done (owner's calls, per the issue)
- The schema does **not** reject `role: snapshot` + `images: true` — left open.
- No snapshot-role image support (save child + `acq_timestamp` join) — that is the other remedy #702 names.
- `PreflightOutcome.check`'s description in `scan_request.py` lists example check names ("e.g."); not extended, to keep the JSON-schema artifacts untouched.

## Per-package LOC (328+ / 7−, 14 files; after review fixes 844ac76c)

| Package | Code | Tests | Version/CHANGELOG/docs |
|---|---|---|---|
| GeecsBluesky | `scan_request_runner.py` +71, `submit_preflight.py` +40 | `test_scan_request_runner.py` +99, `qs_client/test_submit_preflight.py` +23 | pyproject 1/1, CHANGELOG +27 |
| GEECS-MCP | `control_tools.py` +1 | `test_control_tools.py` +29 | pyproject 1/1, CHANGELOG +12 |
| GEECS-Schemas | `save_set.py` +7/−2 (descriptions) | — (existing no-drift test covers the regenerated page) | pyproject 1/1, CHANGELOG +11, `schema_reference.md` 2/2 |

## Tests (worktree envs, `scripts/check.sh`)

New pins: merge warning fires for images asked of a save-set snapshot device and for a new async + save device, not for a synchronous one, and does not repeat a save-set offender; runner warning fires for snapshot + `images: true` (names the device and save set, not the synchronous or image-less snapshot entries) and the entry is still derived; no warning for synchronous + images; preflight raises a `snapshot_images` question (not a refusal, no `passed` outcome when raised) and records `passed` on the all-green fixture; MCP refuses-then-accepts once `snapshot_images` is acknowledged and stamps `continued`.

Full run in the worktree (lint `--lint`: OK; suites via `check.sh` explicit list = every suite with a local env):

| Suite | Result |
|---|---|
| GeecsBluesky | 685 passed, 4 deselected (re-run after 844ac76c; MCP/root-env suites unchanged by that commit) |
| GEECS-MCP | 125 passed |
| GEECS-Schemas | 286 passed, 6 deselected |
| GEECS-Data-Utils | 325 passed, 3 skipped |
| ImageAnalysis | 250 passed, 1 skipped, 10 deselected |
| ScanAnalysis | 213 passed, 1 skipped, 11 deselected |
| root tests | 2 passed, 62 deselected |

Not run locally (no worktree env; none of them import the changed code — Console tests build their own `PreflightReport`s): GEECS-Console, GeecsCAGateway, GeecsPvaGateway, GEECS-Core, GEECS-DataPortal, GEECS-LogTriage — CI covers them.

## Version notes
- Rebased onto `master` @ 85fa45f7 (#775–#777 merged mid-work); bumps are relative to the post-#777 versions (GeecsBluesky 0.73.1 → 0.73.2, GEECS-MCP 0.8.4 → 0.8.5, GEECS-Schemas 0.17.0 → 0.17.1).
- A GeecsBluesky fix for #779 is landing in parallel — if it also bumps to 0.73.2, whichever merges second rebases + re-bumps.

## Hardware verification
**OWED:** submit a scan whose save set has a snapshot-role camera with `images: true`; expect the `snapshot_images` warning in the console modal / MCP `needs_acknowledgement` pre-submit and the `IGNORED for snapshot-role entries` line in the worker `scan.log`; the scan is otherwise unchanged (scalars recorded, no images for that device — as before).


skbarber:
## Adversarial review (fresh-context subagent, per `/land`) — reviewer's report, verbatim

Reviewed at 9cfddc04 (the initial commit of this PR).

---

**Read:** full diff of 9cfddc04 vs `origin/master` (14 files) plus the surrounding code it plugs into — `save_set_to_devices_config` / `merge_optimizer_device_requirements` / `_build_request_detectors` / `select_capture_devices` in `scan_request_runner.py`, both plan call sites in `plans/scan_request_plan.py`, the full `submit_preflight.py` flow and fixtures, `scan_log.py`'s pre-claim buffer, the MCP ack loop, the Console's question renderer (`main_window.py:1710`), the legacy save-element converter, and the three CHANGELOGs. **Ran:** `tests/test_scan_request_runner.py` + `tests/qs_client/test_submit_preflight.py` (92 passed), GEECS-MCP (125 passed), GEECS-Schemas (291 passed, 1 failed — `test_corpus_integration.py::TestFullCorpus::test_every_scan_variable_catalog_converts`, `@pytest.mark.integration`, fails identically on `master` @ 85fa45f7 because the local configs corpus no longer has legacy `scan_devices.yaml`; not this PR's).

The core claims check out: the `synchronous` key is always set by the seam, the pre-claim root-logger buffer (`begin_pre_scan_capture`) does carry the warning into `scan.log`, the Console renders questions generically so no vocabulary edit is needed there, and the new tests do fail if the condition is inverted or the device filter is dropped.

### 1. (Medium, correctness) The optimize path recreates snapshot + images *after* the seam, and neither surface sees it

`scan_request_plan.py:873` derives `devices_config` (seam runs, warning evaluated), then `:878` calls `merge_optimizer_device_requirements`, which:

- on an **existing** snapshot-role device, flips `save_nonscalar_data` to `True` while deliberately keeping `synchronous=False` (`scan_request_runner.py:399-401`, pinned by `test_existing_device_keeps_save_set_synchronous_semantics`);
- on a **new** device, copies the requirement's `synchronous`/`save_nonscalar_data` verbatim (`:370-373`), so a requirement declaring `synchronous: False, save_nonscalar_data: True` lands as-is.

Concrete scenario: the operator's save set lists the objective camera `UC_TopView` with `role: snapshot` (scalars only); the evaluator's auto-generated requirements want `UC_TopView` images. Merged config = async + save → `_build_request_detectors` builds `session.snapshot(...)` with no `save_images` → no images → every objective evaluates to NaN. That is exactly the #520 incident class the merge was added to prevent, and it is silent again: the worker warning already ran on the pre-merge config, and the client preflight's `_resolve_devices_config` never sees requirements at all ("provisioned worker-side; nothing to check here"). The PR body's "the one DB-free role→mechanics seam, and nowhere else" does not hold for this branch.

Fix is small: run `snapshot_images_ignored` on the *merged* config in the optimize branch (one call after `:878`), or move the worker-side `logger.warning` out of `save_set_to_devices_config` into the two plan call sites (after the merge in the optimize branch). The second option also removes the incidental double-logging in Finding 3. Keep the pure helper shared either way.

### 2. (Low, redundancy) Two near-identical warnings for a capture-eligible camera in snapshot role

`select_capture_devices` (`scan_request_runner.py:1348-1362`) already logs, per device: "...Its `images: true` is ignored: the engine neither commands nor suppresses the device's own save flag, so the engine records no per-shot images for it. Remove `role: snapshot` (legacy: `synchronous: false`)...". The new `snapshot_images_ignored_message` repeats that sentence nearly verbatim. For a snapshot camera with `images: true` and a live DB, the worker now emits both (seam at `:237-239`, then `select_capture_devices` at `:1641`). The PR body frames #702's warning as the "capture-ownership slice", which is true — but now that the general warning exists, the #702 branch could shrink to its capture-specific sentence (or an INFO `continue`) so `scan.log` carries one remedy per device, not two paraphrases.

### 3. (Low, noise) The client process also logs the worker's warning, twice through the MCP

`_resolve_devices_config` (`submit_preflight.py:110`) calls the same seam, so every client run emits the `logger.warning` into the Console/MCP-server log *in addition to* raising the question. The MCP re-runs preflight on the acknowledged resubmit, so one MCP submission logs it twice. Harmless, but "one warning per save set" is not what those logs show. Folding the `logger.warning` into the plan (Finding 1's second option) makes the helper side-effect-free for clients.

### 4. (Nit, placement) Schema description promises engine behavior

`SaveSetEntry.images` now says the scanner "warns about the combination at submit time and in the scan log". GEECS-Schemas is the pydantic-only vocabulary; the submit-time warning exists only for clients that run `run_submit_preflight` (Console, MCP) — a script submitting through `qs_client` directly gets none. Suggest keeping the schema sentence to the contract ("Ignored for role 'snapshot' … records scalars only") and leaving *where* it warns to GeecsBluesky's docs/CHANGELOG, so the regenerated `schema_reference.md` does not go stale when the warning surface moves (e.g. after Finding 1).

No hygiene, scan-folder-invariant, or test-strength findings survived.

skbarber:
## Adversarial review — dispositions

Every finding in the report above is dispositioned; fixes went back to the **same reviewer** (continued subagent), which re-read the fix commit and re-ran the suites.

### 1. Medium — optimize path re-creates snapshot + images after the seam
**FIXED in 844ac76c — reviewer CONFIRMED.** `merge_optimizer_device_requirements` (the second role→mechanics seam) snapshots `snapshot_images_ignored(devices_config)` before the loop and, after it, warns via the shared message for the devices the merge *introduced* — a save-set snapshot device whose save flipped on, or a newly provisioned async + save device — with a suffix naming the optimizer's `device_requirements` as the source and the objective's diagnostics as the loss (the #520 NaN class). Save-set offenders already warned at `save_set_to_devices_config` are not repeated. Two pins: `test_merge_requirements_warns_when_images_land_on_a_snapshot_device`, `test_merge_requirements_does_not_rewarn_save_set_offenders`. Reviewer also confirmed the casefold-match branch is covered (the set is keyed by the configured spelling). The reviewer's alternative — moving the worker-side warning out of the seam into the plan call sites — was **not taken**: the seam is the DB-free home #754 names and is what covers `GeecsSession.run` / direct callers; the PR body and CHANGELOG no longer claim a single seam. Reviewer's follow-up wording nit (for an auto-provisioned device with no save-set entry the remedy "remove `role: snapshot`" points at an edit that doesn't exist) **FIXED in d3e5077c** using the reviewer's own suggested text (94 runner + preflight tests pass; text-only, so not re-routed).

### 2. Low — two near-identical warnings for a capture-eligible camera in snapshot role (live DB)
**WAIVED.** The #702/#751 `select_capture_devices` warning is DB-gated and carries capture-specific information the general warning does not (the devicetype, daemon ownership, the #702 remedy) with its own pinned wording; shrinking it is a text edit on #751's surface, a follow-up if the owner wants one line per device in `scan.log`. Reviewer: "safe to waive."

### 3. Low — the client process also logs the worker's warning (twice through the MCP's resubmit)
**WAIVED.** A log line beside the modal / `needs_acknowledgement` is harmless duplication, and keeping the warning in the seam is exactly what covers non-preflight callers. Reviewer: "safe to waive."

### 4. Nit — schema description promised engine behaviour
**FIXED in 844ac76c — reviewer CONFIRMED.** `SaveSetEntry.images` now stops at the contract ("...the scanner neither commands nor suppresses the device's own save flag."); `schema_reference.md` regenerated (GEECS-Schemas 286 passed, 6 deselected).

No new findings from the fix commits; no undispositioned findings remain. The pre-existing `integration`-marked corpus failure the reviewer hit locally also fails on `master` @ 85fa45f7 (local configs checkout lacks legacy `scan_devices.yaml`) and is unrelated.

skbarber:
Rebased onto master after #783 merged (Codex P1 / version collision): GeecsBluesky 0.73.2 → **0.73.3**, GEECS-MCP 0.8.5 → **0.8.6**; both CHANGELOGs now carry #783's entry under the old number and this PR's under the new one, cross-references updated. Merge + re-bump only, no code change. Local: GeecsBluesky 685 passed / 4 deselected, GEECS-MCP 125 passed.

Issue #787, opened 2026-09-04
GEECS-Console: a dropped MainWindow lives until the cyclic GC, which can delete it under Qt's event dispatch (residual of #767)


skbarber:
## What was found while fixing #767

The `HealthPoller` GUI-thread hop (branch `console/healthpoller-gui-hop`, GEECS-Console 0.28.1) was **not** the main cause of the isolated `QT_QPA_PLATFORM=offscreen pytest tests/test_main_window.py` segfaults: on this machine (macOS/arm64, PySide 6.11.1, pytest-qt 4.5.0) the file crashed 5/5 before the hop and still 6/8 after it.

### Evidence

- **Native stacks** (three macOS crash reports, all the same shape): `QMainWindow::event` → `QWidget::ensurePolished` recursing into children → `QStyleSheetStyle::polish` → crash inside a child — `QComboBox::changeEvent` building its popup container (`setPalette_helper`), `QLineEdit::changeEvent` → `QTextEngine::attributes`, `QAbstractItemView::event`. Qt dispatching into a widget tree that is being freed under it. The Python side of every crash is `pytestqt/plugin.py::_process_events` inside `pytest_runtest_call` (after the test body), twice with `Garbage-collecting` as the innermost frame.
- **Ablations** (4–6 isolated runs each): GC disabled → 0 in-test crashes; scan monitor (1 Hz status poll) off → 0/4; health poller off → 3/4; window ignores status snapshots → 0/5; queue panel ignores status → 0/5; hold release without `QTimer.singleShot` → 3/5 (not the singleShot); `gc.collect()` at teardown start → 2/4 (too late); **`gc.collect()` after the test body, before pytest-qt's `processEvents` → 0/6 in-test**.

### Mechanism

[...]

Issue #778, opened 2026-09-03
GEECS-Console: startup thread import race kills the document stream on every launch (bluesky _DeadlockError, partially-initialized geecs_data_utils)


skbarber:
## Symptom

On **every** GEECS-Console launch since 2026-08-24 (per `~/.config/geecs_console/logs/console.log`, 9/9 launches), two errors land within ~1 s of the window appearing:

```
WARNING geecs_console.app.scan_monitor: document stream at 192.168.6.14:5568 ended
Traceback (most recent call last):
  File "geecs_console/app/scan_monitor.py", line 89, in _run
    from bluesky.callbacks.zmq import RemoteDispatcher
  ...
  File "bluesky/utils/__init__.py", line 40, in <module>
    from bluesky._vendor.super_state_machine.errors import TransitionError
_frozen_importlib._DeadlockError: deadlock detected by _ModuleLock('bluesky._vendor.super_state_machine.errors')
```

```
INFO geecs_console.services.ops_paths: geecs_data_utils unavailable: cannot import name
'read_tiled_config' from partially initialized module 'geecs_data_utils.tiled_catalog'
(most likely due to a circular import)
```

Consequences for the whole session:

- The bluesky **document stream is dead** — `stream_failed` fires ("live per-shot progress is off"). The state pill and progress then ride only the 1 s manager status poll; start/stop documents never arrive.
- The R6 idle display ("Scan NNN (previous)") is blank because `ops_paths.todays_scan_folder` reports data-utils as unavailable and returns `None`.

The engine, the Start gate and submission are unaffected (verified 2026-09-03 headless against the live manager), so the console *looks* fine — which is why this went unnoticed for ten days.

## Cause

A **concurrent-import race** at startup, not a real circular-import bug. Several daemon threads spawn at construction and each lazily imports a package that has module-level import cycles inside it:

| thread | lazy import |
|---|---|
| `qs-doc-stream` (`scan_monitor.py:89`) | `bluesky.callbacks.zmq` → `bluesky/__init__` → … → `bluesky._vendor.super_state_machine.errors` |
| status poll / console stream (`scan_monitor.py:151`, `ZmqQueueClient._manager`) | `bluesky_queueserver_api` → `bluesky_queueserver` → `bluesky` |
| health poller (`services/health.py:182`) | `geecs_bluesky.tiled_integration.read_tiled_config` → `geecs_data_utils.tiled_catalog` → package `__init__` |
[...]

PR #784, opened 2026-09-04
GEECS-Console 0.28.1: warm cycle-bearing imports on the GUI thread before monitor threads start (#778)

skbarber:
## What

`MainWindow.__init__` now calls a new `services/warm_imports.py::warm_imports()` as its very first statement, importing one entry per cycle-bearing package the console's daemon threads lazily enter — `bluesky.callbacks.zmq` (all of `bluesky`), `geecs_data_utils.scan_paths` and `geecs_bluesky.tiled_integration` (both halves of the `geecs_data_utils` cycle) — once, synchronously, on the GUI thread, before any controller spawns a thread.

The lazy import sites themselves are untouched: this is a load-**order** fix, not a dependency change (each module's offline import-safety is unchanged; a missing dependency still surfaces at the call site that needs it). The warm-up is best-effort — a failed import is logged at WARNING and skipped, never raised.

Closes #778

## Why

On every launch since 2026-08-24 (9/9 in `console.log`) the bluesky document stream died with `_frozen_importlib._DeadlockError: deadlock detected by _ModuleLock('bluesky._vendor.super_state_machine.errors')` and the R6 idle "Scan NNN (previous)" display went blank with `cannot import name 'read_tiled_config' from partially initialized module 'geecs_data_utils.tiled_catalog'`.

Root cause, confirmed by reading the current code and reproduced cold by the adversarial reviewer (see the review comment): Python's per-module import lock serializes two threads importing the *same* module, but two threads entering a package through *different* submodules that its `__init__` imports deadlock — importlib raises `_DeadlockError` in one and the other sees a "partially initialized" module. Two such pairs spawn during window construction:

| package | racer A | racer B | observed error |
|---|---|---|---|
| `bluesky` | `qs-doc-stream`: `DocumentStreamWorker._run` → `bluesky.callbacks.zmq` | `console-health-poll`: `GatewayTiledDbHealth._check_gateway` → `geecs_bluesky.devices.ca._pv` → ophyd_async → `bluesky.protocols` | `_DeadlockError` on `bluesky._vendor.super_state_machine.errors` (reproduced 4/4 cold) |
[...]

PR #788, opened 2026-09-04
GEECS-Console 0.28.2: HealthPoller emits report_ready on the GUI thread (#767)

### LINKED ISSUE #767: GEECS-Console: HealthPoller emits from its daemon thread straight at the window — isolated test_main_window.py runs segfault on master
## Symptom

`QT_QPA_PLATFORM=offscreen pytest tests/test_main_window.py` (the file **alone**) in GEECS-Console segfaults at random positions on a clean `master` — observed 2 of 3 runs on macOS/arm64 during the #764 review. The full console suite does not crash, and CI (which runs the full suite) is green, so this is latent in isolated-file runs today.

## Diagnosis (from #764)

#764 bisected an identical crash (native stack: `QWidget::ensurePolished` walking a dangling child) to **a daemon-thread signal emission aimed at a consumer QObject racing that consumer's destruction** — Qt's C++ connection bookkeeping survives the race, PySide's Python-slot delivery does not. `BackgroundResult` was fixed in #764 by hopping the result onto the GUI thread through the worker's own queued signal (worker held alive until the hop lands) and emitting `result_ready` there.

`HealthPoller` (`services/background.py`) still has the original shape: its daemon thread emits `report_ready` straight at the window's `_apply_health_report` / `_on_queue_status` (via `ScanMonitorController`) — the busiest emitter in the app (1 Hz status poll + 5 s health poll), and the last one of that shape. Test windows are torn down while a poll is out, so isolated runs hit the race.

## Proposed fix

Give `HealthPoller` the same hop as `BackgroundResult` (queued self-signal + `_INFLIGHT` hold, emit `report_ready` on the GUI thread) — the `_busy` in-flight skip stays. Consumers already connect `QueuedConnection`; delivery becomes two event-loop turns like `result_ready`. Pin it in `tests/test_background.py` beside the `BackgroundResult` hop tests. Verify by running `tests/test_main_window.py` alone ≥5 times.

Reference: GEECS-Console `CLAUDE.md`, the `BackgroundResult` paragraph (0.28.0 note); PR #764 review thread.

skbarber:
Closes #767. Residual findings filed as #787.

## What

**Concern 1 — the #767 fix (`services/background.py` ±138 LOC, `tests/test_background.py` +118).** `HealthPoller` no longer emits `report_ready` from its daemon thread toward its consumers (the window's `_apply_health_report`, the scan monitor's `status_ready` → `_on_queue_status`). It takes the same GUI-thread hop `BackgroundResult` got in #764: the daemon thread emits the worker's own queued `_landed` signal, the worker is held in the counted `_INFLIGHT` registry until the hop lands, and `report_ready` is emitted on the GUI thread. The `_busy` in-flight skip stays and now lasts until the report has landed, so the flag is only ever written on the GUI thread and at most one hop per poller is pending. The hop lives once, in the private `_GuiHopWorker` base both workers subclass (`_hop` on the daemon thread → `_forward` → the subclass's `_deliver` on the GUI thread) — a judgment call over duplicating the ~15 lines into the poller; cheap to veto. Consumers connect `QueuedConnection` as before; delivery takes two event-loop turns. Six new tests pin the poller hop beside the `BackgroundResult` ones: emission thread, in-flight skip lasting until landing, raising probe / `None` report emit nothing but release hold + busy, consumer dropped mid-flight, poller dropped mid-flight survives until the hop lands.

**Concern 2 — the larger part of the crash, found while verifying (`tests/conftest.py` +26).** The isolated `tests/test_main_window.py` run still segfaulted after concern 1. Native stacks (three macOS crash reports) all show `QMainWindow::event → ensurePolished → QStyleSheetStyle::polish` crashing inside a child widget, with pytest-qt's post-call `processEvents` as the only Python frame; ablations: GC disabled → 0 crashes, scan-monitor poll off → 0, window ignoring status → 0, queue panel ignoring status → 0, `gc.collect()` after the test body → 0. Mechanism: a `MainWindow` built in a test body is Python-owned and cyclic (window ↔ controllers via injected bound-method callbacks), so dropping it at `return` leaves it to the cyclic GC, which can run inside `processEvents` while Qt is dispatching to that window — shiboken deletes the C++ window under the running dispatch. The 1 Hz status poll's fetch/render only supplies the allocations that make the GC land there. Fix: a `pytest_runtest_call` wrapper in `tests/conftest.py` collects garbage after the body, before pytest-qt's own (`tryfirst`) hook processes events, so a dropped window dies at a safe point. The product-side debt (the cycles themselves) and an interpreter-exit segfault seen twice are in #787.

**Docs (`CHANGELOG.md` +27, `CLAUDE.md` ±42):** 0.28.1 entry; the `BackgroundResult` paragraph now says `HealthPoller` rides the hop (it previously named it the remaining exception) and that new workers subclass `_GuiHopWorker`; the Testing section documents the safe-point collection. Version: `poetry version patch` 0.28.0 → 0.28.1 — **collides with the #778 PR's console bump; whichever merges second rebases + re-bumps.**

## Why

Isolated `QT_QPA_PLATFORM=offscreen pytest tests/test_main_window.py` runs segfaulted at random positions on a clean master (#764 review observation → #767). Both causes are the same hazard class the console `CLAUDE.md` already names: a Python-owned Qt object destroyed while Qt or PySide is still using it.

## Crash counts (this machine, macOS/arm64, PySide 6.11.1, `tests/test_main_window.py` alone)

| State | Crashes / runs |
|---|---|
| master (before) | **5 / 5** |
| concern 1 only (poller hop) | **6 / 8** |
| concern 1 + concern 2 | **0 / 8** (133 passed each) |

## Tests

- `./scripts/check.sh` (scoped: lint + GEECS-Console): **820 passed, 2 warnings** (the warnings are the pre-existing `report_ready.disconnect()` on a never-connected poller in `test_scan_monitor.py`, noted in #787).
- `./scripts/check.sh --all` was **not run locally**: the fresh worktree has a Poetry env for GEECS-Console only, the diff touches nothing else, and no package imports GEECS-Console; `pre-commit run --all-files` passes except the pre-existing notebook-output hook on untouched notebooks. CI's `unit-tests` runs every suite.

## Hardware verification

**OWED:** console launch against the live manager — health chips still update at the 5 s poll and the R6 pill / R8 queue panel still follow the 1 Hz status poll (two event-loop turns of added latency, not observable).


skbarber:

Reviewed commit `57348722` (`console/healthpoller-gui-hop` vs `origin/master`), read-only, with the root and `GEECS-Console/CLAUDE.md`, `scan_monitor.py`, `movable_panel.py`, `editors/base.py`, `main_window.py`, the pytest-qt 4.5.0 plugin source in the venv, and issue #787 as context. `tests/test_background.py` + `tests/test_scan_monitor.py` pass (17 passed, the two known `report_ready.disconnect()` warnings).

### 1. Medium — the "last daemon-thread emitter aimed at a consumer" claim is false, and the CLAUDE.md now contradicts itself

CHANGELOG (0.28.1 entry), the `_GuiHopWorker` docstring, and `GEECS-Console/CLAUDE.md:381-388` ("it was the last daemon-thread emitter aimed at a consumer … never emit from a thread toward a consumer again") describe a state the code is not in. Still emitting from a non-GUI thread toward a consumer QObject, in this same package:

- `geecs_console/app/scan_monitor.py:95,189,193` — `DocumentStreamWorker.document.emit` / `ConsoleStreamWorker.line.emit` / `pause_reason.emit` run on the zmq daemon threads, connected `QueuedConnection` to `ScanMonitorController` (a Python-owned, unparented QObject). That is exactly the pre-PR `HealthPoller` shape, and `CLAUDE.md:199-201` still documents it approvingly ("Stream workers emit on worker-owned signals connected `QueuedConnection`") ten paragraphs above the new "never again" rule.
- `geecs_console/app/movable_panel.py:218-223` — `self.value_ready.emit` is invoked from the aioca loop thread (`subscribe_many(..., self.value_ready.emit)` / the lambda). That is a *controller-owned* signal emitted from a foreign thread — the "window-owned signal from a daemon thread" shape the same CLAUDE.md paragraph names as the original segfault.
- `geecs_console/editors/base.py:196` — `completions_ready.emit` on a daemon thread toward the dialog (already self-documented as known debt to consolidate onto `BackgroundResult`; `_GuiHopWorker` is now the thing it should consolidate onto).

None of these is exercised by the test suite (window tests dispose the monitor and pass `doc_addr=None`), so the PR's crash counts are honest; but a live console closing while a document or readback arrives is the same race class, and the text says the class is closed. Suggested fix: scope the claim to "the last emitter of that shape *in `services/background.py`*", keep the "new workers subclass `_GuiHopWorker`" rule, and list the three residual emitters in #787 so the hazard is inventoried rather than declared gone.

### 2. Medium — the "in-flight skip lasts until landed" test does not pin what it is named for (confirmed by mutation)

I ran the six new `TestHealthPollerGuiHop` tests against a mutant that restores the pre-PR shape — `_run` resets `_busy = False` in a `finally` on the daemon thread and `_deliver` only emits — in the package venv (scratch file, no repo edits): **12/12 pass**, `test_poll_in_flight_is_skipped_until_the_report_has_landed` included. The test's second `poll_async()` happens while the probe gate is still closed, which is the plain in-flight skip the old code also had; between `gate.set()` and the drain nobody calls `poll_async()`, so nothing observes whether `_busy` was cleared on the daemon thread or after landing. The CHANGELOG/CLAUDE.md statement "`_busy` is only ever written on the GUI thread" therefore has no test behind it. Concrete strengthening: after `gate.set()`, wait for the `console-health-poll` thread to exit **without** pumping events (e.g. `threading.enumerate()` name filter, or a `done` `Event` set at the end of `GatedProbe.poll`), then call `poll_async()` and assert `probe.calls == 1` and `background._INFLIGHT[poller] == 1`; only then `_drain` and assert exactly one delivery. That fails under the mutant and passes under the PR.

### 3. Low — declared pytest floor is now stale

`GEECS-Console/pyproject.toml:39` says `pytest = ">=7.0"`, but `@pytest.hookimpl(wrapper=True)` in the new `tests/conftest.py` is the pluggy ≥ 1.1 / pytest ≥ 8.0 new-style wrapper (pytest 7.0 ships with pluggy 1.0, where the kwarg raises at registration). The lock pins pytest 9.1.1 / pluggy 1.6.0 and CI installs from the lock, so nothing breaks today; the root `pyproject.toml` already says `>=8.1`. One-line bump or ignore.

### Verified and not raising

- **Hook ordering**: pytest-qt 4.5.0's `pytest_runtest_call` is `@pytest.hookimpl(wrapper=True, tryfirst=True)` (plugin.py:185), so its post-yield `_process_events()` does run *after* the conftest wrapper's `finally: gc.collect()`, as the docstring claims.
- **The hop itself**: hold taken on the GUI thread in `poll_async`, released one turn after `_forward`; the only Python reference drop that could kill the poller mid-hop is covered by `_INFLIGHT`; `HealthPoller` is unparented so the `_hop` `RuntimeError` branch (which would strand `_busy=True`) is unreachable in practice. `test_set_finishing_after_window_deletion_is_swallowed` (the one mid-body window drop) uses explicit `deleteLater`, so the new "collect yourself" rule does not bite it.
- **Redundancy**: no other GUI-hop implementation in the repo — the browser already uses `BackgroundResult`, `editors/base.py`'s thread is the pre-existing documented debt (finding 1), nothing outside GEECS-Console is Qt. `HealthPoller` as a subclass rather than composition over `BackgroundResult` is a taste call, not duplication.
- **Placement**: `_GuiHopWorker` belongs in `services/background.py` (the blessed worker home per CLAUDE.md); the safe-point collection belongs in `tests/conftest.py` (pytest-qt exposes no equivalent option). No scan-folder or public-repo-hygiene surface in this diff.

*Disposition follows in the next comment once the reviewer has confirmed the fix commit.*

skbarber:
## Review disposition (fix commit 2a6868bb)

The original reviewer subagent became unreachable after the fix (API rate limit), so per the `/land` skill a **fresh-context verifier** was given only the finding texts and the fix commit — substitution noted here.

| # | Finding | Disposition |
|---|---|---|
| 1 | "Last daemon-thread emitter aimed at a consumer" claim false; CLAUDE.md self-contradictory | **Fixed** (2a6868bb) — CHANGELOG, the `_GuiHopWorker` docstring and `CLAUDE.md` scope the claim to `services/background.py` and name the residual emitters (`scan_monitor.py` stream workers, `movable_panel.py` readback callback, `editors/base.py` completions thread); the absolute "never again" rule is replaced by "new workers subclass `_GuiHopWorker`; move a residual emitter onto it when it bites". The three residuals are inventoried in a comment on #787. Verifier: confirmed the texts no longer contradict the scan-monitor paragraph and that each cited emit site exists. |
| 2 | In-flight-skip test did not pin "until landed" (mutant passed) | **Fixed** (2a6868bb) — the test now waits for the `console-health-poll` thread to exit *without* pumping events, then asserts a `poll_async()` is still skipped (`probe.calls == 1`, `_INFLIGHT[poller] == 1`) before draining. Verifier re-ran the mutant: `1 failed` at the new assertion (`assert 2 == 1`), only that test sensitive; restored: `10 passed`; soundness judged deterministic (queued event cannot dispatch during a non-pumping wait), 25/25 isolated runs. |
| 3 | `pytest = ">=7.0"` floor stale for `hookimpl(wrapper=True)` (needs pytest ≥ 8) | **Waived** — the lock pins pytest 9.1.1 / pluggy 1.6.0 and CI installs from the lock, so nothing can hit it today; bumping the constraint requires a relock, which this repo forbids from a session worktree (the `poetry-lock-no-worktree-paths` hook). Cheap follow-up: bump to `>=8.0` in the next console PR that relocks anyway. |

Console suite at 2a6868bb: **820 passed, 2 warnings** (pre-existing `report_ready.disconnect()` warning, #787). CI green on both commits (pre-commit, unit-tests, console-windows).

skbarber:
Rebased onto master after #784 merged (Codex P1 / version collision): GEECS-Console 0.28.1 → **0.28.2**; the CHANGELOG carries #784's entry under 0.28.1 and this PR's under 0.28.2, the four in-tree '0.28.1' references for the HealthPoller hop (CLAUDE.md, background.py docstring, test_background.py) updated. Merge + re-bump only, no code change. Local: console suite 827 passed, 2 pre-existing warnings.

PR #791, opened 2026-09-04
render_units.sh: refuse a unit file that is not a site-profile template

skbarber:
## What

`deploy/render_units.sh` refuses a template that lacks the `@SERVICE_USER@` / `@SITE_ENV@` placeholders, naming the file and telling the operator to pull that clone forward. `docs/platform/site_profile.md` states the rule. `deploy/` + docs only, no package code.

## Why

Phase 3 cutover, 2026-09-04: the bootstrap renders each unit from its own service's clone (by design, #777 review finding 3). The portal's clone was pinned at a1a23ae1, before #777, so its unit file was the pre-profile hand-edit copy with no placeholders. The unfilled-placeholder check had nothing to flag, the file was installed verbatim (`User=geecs`, `/home/geecs/GEECS-Plugins-portal`), and the portal crash-looped with `status=217/USER` until the clone was pulled to master and re-rendered. Every other unit rendered correctly. This closes the gap between "render from the service's clone" and "that clone may predate the templates".

## Tests

`bash -n`; lint clean. Negative: rendering a1a23ae1's portal unit through the script now exits 1 with `not a site-profile template … pull it to a version with them`. Positive: the five current templates still render from the example.

## Hardware verification

Live: the incident above is the reproduction; the corrected render (portal clone at master) produced the unit now installed.

### DIFF
--- a/GEECS-DataPortal/DEPLOYMENT.md
+++ b/GEECS-DataPortal/DEPLOYMENT.md
@@ -167,7 +167,7 @@ without the processing selector.)
 | `/health` reports a catalog error | Tiled down, or `[tiled]` uri/api_key wrong — `curl http://<tiled-host>:8000/api/v1/` |
 | Day pages load, images 404 | share not mounted (or moved) at `geecs_data_local_base_path`; a 404 on one shot with others fine is the exact-match rule working (that device missed the shot) |
 | Slow day listings | measure `list_runs` against the catalog first — the fix is a portal-side cache, not a schema change (scope doc, open questions) |
-| Unit crash-loops at start | wrong absolute Poetry path in `ExecStart` (`status` shows 203/EXEC); env installed by a different account than `User=` (empty venv — reinstall as the service account); or port 8200 already taken. A down Tiled does **not** exit the service — that shows up as the `/health` row above |
+| Unit crash-loops at start | `status` shows **217/USER** — the installed unit is a pre-profile file (or a copy from an old staging run) with the generic `User=`; the clone it came from predates the templated units — pull it forward, re-render with `deploy/render_units.sh`, reinstall (site profile page). Wrong absolute Poetry path in `ExecStart` (`status` shows 203/EXEC); env installed by a different account than `User=` (empty venv — reinstall as the service account); or port 8200 already taken. A down Tiled does **not** exit the service — that shows up as the `/health` row above |
 | Evening scans 404 (or resolve oddly) while daytime scans work | host timezone differs from the scanner hosts' — daily folders are named by the scanner's local date. `site.env` sets `TZ`; keep it matching the lab's zone |
 
 The fleet-map page (`docs/platform/fleet_map.md`) carries the
--- a/GeecsCAGateway/deploy/README.md
+++ b/GeecsCAGateway/deploy/README.md
@@ -51,6 +51,11 @@ systemctl status geecs-ca-gateway
 journalctl -u geecs-ca-gateway -f     # expect the startup lines, then quiet
 ```
 
+`status=217/USER` in a crash loop means the installed unit is not a rendered
+one (generic `User=` — a pre-profile file, or a copy from an old staging
+run): the clone it was rendered from predates the templated units. Pull that
+clone forward, re-render, reinstall.
+
 Then the smoke tests from `DEPLOYMENT.md` §4 — from a *different* machine, so
 subnet scoping and firewalls are exercised:
 
--- a/deploy/bootstrap_host.sh
+++ b/deploy/bootstrap_host.sh
@@ -211,12 +211,16 @@ for s in $SERVICES; do
     t="$GEECS_CHECKOUT_ROOT/$(clone_of "$s")/$(template_of "$s")"
     if [ -f "$t" ]; then
         if ! is_unit_template "$t"; then
-            echo "  $s: WARNING — $(clone_of "$s") predates the templated units ($t has no placeholders); no unit rendered or enabled — pull that clone forward (a deploy of $s), then rerun"
+            echo "  $s: WARNING — $(clone_of "$s") predates the templated units ($t: directives lack User=@SERVICE_USER@ / EnvironmentFile=@SITE_ENV@); no unit rendered or enabled — pull that clone forward (a deploy of $s), then rerun"
             SKIP_SERVICES="$SKIP_SERVICES$s "; continue
         fi
         TEMPLATE_PATHS+=("$t")
     else echo "  $s: clone absent — template from this clone ($REPO_ROOT)"; TEMPLATE_PATHS+=("$REPO_ROOT/$(template_of "$s")"); fi
 done
+# The staging dir holds only THIS run's units: a stale .service from an
+# earlier run (a service since skipped) must never ride along on the
+# install line. cutover/other files in the dir are left alone.
+[ -d "$STAGE" ] && rm -f "$STAGE"/*.service
 if [ "${#TEMPLATE_PATHS[@]}" -eq 0 ]; then echo "  nothing to render (every wanted service was skipped)"
 elif [ "$DRY" -eq 1 ]; then echo "  [dry] render_units.sh $SITE_ENV $STAGE ${TEMPLATE_PATHS[*]}"
 else RENDER_QUIET=1 "$REPO_ROOT/deploy/render_units.sh" "$SITE_ENV" "$STAGE" "${TEMPLATE_PATHS[@]}" | sed 's/^/  /'; fi
--- a/deploy/bootstrap_host.sh
+++ b/deploy/bootstrap_host.sh
@@ -200,11 +200,22 @@ template_of() { case "$1" in
     qserver) echo "GeecsBluesky/qserver/deploy/geecs-qserver.service";; capture) echo "GeecsBluesky/capture/deploy/geecs-capture.service";;
     mcp) echo "GEECS-MCP/deploy/geecs-mcp.service";; esac; }
 TEMPLATE_PATHS=()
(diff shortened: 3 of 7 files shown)


skbarber:

Reproduced the incident first: `HEAD~1`'s `render_units.sh` renders `a1a23ae1`'s portal unit "clean" (rc 0, `User=geecs`, `WorkingDirectory=/home/geecs/GEECS-Plugins-portal/...`); the PR's script refuses it with the intended message (rc 1); all five current templates still render from `site.env.example` (rc 0). `bash -n` passes on both scripts. The findings below are about the shape of the check, not whether it fires.

### 1. One pre-profile clone aborts the whole bootstrap, contrary to the script's own per-service-skip doctrine

**Severity:** Medium
**Lens:** Correctness (behaviour under `bootstrap_host.sh`)
**Scenario / evidence:** `deploy/bootstrap_host.sh:211` runs `RENDER_QUIET=1 render_units.sh ... "${TEMPLATE_PATHS[@]}" | sed 's/^/  /'` under `set -euo pipefail`. Simulated: the refusal propagates through the pipe (pipefail), `set -e` kills the bootstrap at that line — nothing after it runs. Concretely, with templates `[gateway(good), portal(pre-profile), qserver]` the staging dir ends up holding only `geecs-ca-gateway.service`; qserver/capture/mcp are never rendered, and the "root steps" tail (site.env install line, `daemon-reload`, per-service `enable --now`) is never printed. Compare the design stated at `bootstrap_host.sh:83-85` and `docs/platform/site_profile.md` ("skips every stage for the services that clone hosts ... ending with the warning again"): a linked-worktree clone degrades to a per-service skip with a loud tail; a pre-profile clone now hard-stops everything. The realistic trigger is not the portal but `gateway-checkout` — "control-room-critical, moves rarely", first in the template list, and the clone most likely to sit at a pre-#777 pin — in which case a bootstrap rerun on that host cannot stage *any* unit until the operator either pulls the gateway clone forward (a deploy of the control-room service, forced by an unrelated bootstrap — the coupling the one-clone-per-service rule exists to prevent) or knows to pass `--only`, which neither message mentions.
**Suggested fix:** Keep the refusal in `render_units.sh` (right home — see refuted list), but have the bootstrap's units stage pre-check the template at the `[ -f "$t" ]` spot (`bootstrap_host.sh:207`), record the service in a skip list, and print in the root steps `# geecs-data-portal: NOT enabled — portal-checkout predates the templated units (pull it forward = a deploy, then rerun)`, exactly like `skipped_clone`. Factor the predicate into `deploy/site_env_lib.sh` (e.g. `is_unit_template FILE`) so render and bootstrap share one definition. Optionally mention `--only` in render's hint.

### 2. The pre-check scans comment lines; the post-render check deliberately does not

**Severity:** Low–Medium
**Lens:** Correctness (predicate)
**Scenario / evidence:** `render_units.sh:84` strips `^\s*#` lines before looking for `@[A-Z_]*@` ("Comments may mention @PLACEHOLDER@ by name; only directive lines count"). The new check at line 71 greps the whole file. Constructed: the a1a23ae1 portal unit with one added header line `# Template header: @SERVICE_USER@ and @SITE_ENV@ are filled at render time.` passes the new check and renders rc 0 with `User=geecs` — the incident file again. Every current template header already lists placeholder names in prose (`@PLACEHOLDER@`, MCP's mentions `@CHECKOUT_ROOT@`), so a directive-line hand-revert that keeps the header is a plausible future shape. Secondary: the two names are hard-coded here, in the `sed` list (77–81), and in the docs table (`site_profile.md:71`); a renamed hole needs three coordinated edits, and a future unit that legitimately carries no `EnvironmentFile=` gets the misleading diagnosis "predates the templated units".
**Suggested fix:** Apply the same comment filter as line 84 and test the directive the incident actually hinged on: `grep -v '^[[:space:]]*#' "$src" | grep -q '^User=@SERVICE_USER@'` (or the generic `grep -q '@[A-Z_]*@'` on directive lines if you want to avoid naming holes at all; the line-84 check still catches partial templates either way). One shared `directive_lines()` helper for both checks removes the asymmetry.

### 3. The regression has no test, although the repo has the convention for exactly this

**Severity:** Low
**Lens:** Correctness (would a test fail if the code were wrong)
**Scenario / evidence:** The PR body's negative case ("rendering a1a23ae1's portal unit ... exits 1") lives only in the PR body. `tests/test_commit_sh.py` already pins a bash script's silent-failure mode from pytest with a throwaway fixture; `scripts/check.sh` runs root `tests/`. A ten-line test — write a placeholder-free unit (or `git show a1a23ae1:GEECS-DataPortal/deploy/geecs-data-portal.service`), run `render_units.sh deploy/site.env.example <tmp> <file>`, assert rc 1 and the `not a site-profile template` string; plus assert the five in-tree templates render rc 0 — would have failed before this PR and would catch finding 2's comment-line bypass if written against a commented variant.
**Suggested fix:** `tests/test_render_units_sh.py` alongside `test_commit_sh.py`.

### 4. The docs sentence sits where a *bootstrap reader* looks, not where the *symptom* is looked up

**Severity:** Low
**Lens:** Correctness (docs placement)
**Scenario / evidence:** The render-time error is self-explanatory and `site_profile.md:130-134` quotes the exact string, so anyone grepping the message finds the rule — that part is fine. The incident's operator-visible symptom, though, was `status=217/USER` on an already-installed unit, and the place an operator looks for that is `GEECS-DataPortal/DEPLOYMENT.md:170` ("Unit crash-loops at start" → lists `203/EXEC`, empty venv, port taken) — `217/USER` is absent there and in every other runbook (grepped `GeecsCAGateway/DEPLOYMENT.md`, `GEECS-MCP/deploy/DEPLOYMENT.md`, `GeecsBluesky/{qserver,capture}/deploy/DEPLOYMENT.md`, `docs/platform/fleet_map.md`). A unit installed by hand from an old clone, or a copy from an earlier staging run, still produces 217/USER with no pointer to the cause.
**Suggested fix:** Add to the portal's crash-loop row: "`217/USER` — unit rendered from (or copied out of) a pre-profile clone; `User=` is the generic account. Pull that clone forward and re-render (site profile page)." The gateway runbook is the other candidate given finding 1's gateway-pin scenario.

### Checked and refuted

- **Renamed-hole scenario** (template has `@SERVICE_USER@`, another hole renamed): the existing directive-line check at `render_units.sh:84` still fails loudly on the unfilled `@NEW_NAME@`; the new predicate does not weaken that.
- **MCP unit shape:** `GEECS-MCP/deploy/geecs-mcp.service` carries both `@SERVICE_USER@` and `@SITE_ENV@` (as do the other four); all five render rc 0 from the example env.
- **Redundancy:** no existing template-validity check anywhere — checked `deploy/bootstrap_host.sh`, `deploy/site_env_lib.sh`, `scripts/fleet_status.sh` (reports versions/branches, never template-ness), and the five `DEPLOYMENT.md` runbooks. Line 84 is the complement of this check (unfilled vs. absent placeholders), not a duplicate.
- **Placement:** `render_units.sh` is the correct home — it is also invoked standalone from the runbooks (`GeecsCAGateway/DEPLOYMENT.md:458`, `GeecsBluesky/qserver/deploy/DEPLOYMENT.md:150`, `GEECS-MCP/deploy/DEPLOYMENT.md:105`), which the bootstrap-only alternative would not cover. Finding 1 asks for the bootstrap to *also* handle it, not to move it.
- **Public-repo hygiene:** the diff carries no hostnames, account names, or home paths; the incident reference is by PR number and date.
- **Versioning:** `deploy/` + `docs/` only, no package code — no version bump or CHANGELOG owed, consistent with the earlier deploy-only commits (`b5efc53a`, `d21e10eb`).
- **Docs text accuracy:** "has no template to render" / "refuses it by name" match the implemented behaviour (the message names the source path).
- **Cosmetic, not filed:** under the bootstrap the refusal lines go to stderr and so bypass the `| sed 's/^/  /'` indent — they print flush-left, which if anything makes them more visible.

---

skbarber:

| # | Finding | Disposition |
|---|---|---|
| 1 | One pre-profile clone aborted the whole bootstrap (render refusal under `set -e`) | **Fixed** 1f9336ac: per-service pre-check → `SKIP_SERVICES`, `# <unit>: NOT enabled` root line, stderr summary; render's hint mentions `--only`; confirmed (real run: four units staged, portal skipped, rc 0) |
| 2 | Pre-check scanned comment lines; predicate names hard-coded in three places | **Fixed** 1f9336ac: shared `is_unit_template` on directive lines (`unit_directives`) in `site_env_lib.sh`, used by both render checks; the commented-header bypass is refused; confirmed |
| 3 | No test for the regression | **Fixed** 1f9336ac: `tests/test_render_units_sh.py` (3 tests); the reviewer's mutation check shows the comment-bypass test fails against the old predicate; confirmed |
| 4 | Symptom (`217/USER`) not in the runbooks | **Fixed** 1f9336ac: portal crash-loop row + gateway deploy/README Verify note; confirmed |
| nit | Bootstrap WARNING said "no placeholders" while the predicate is stricter | **Fixed** (this commit): wording matches the render's message |
| obs | `deploy-staging` never cleaned — a stale unit could ride the install line | **Fixed** (this commit): the units stage removes `*.service` from staging before rendering, so it holds only this run's units |

The reviewer's other observation (the environments stage still installs inside a pre-profile clone) is left as is: idempotent and consistent with "a pull is a deploy".

skbarber:
**Addendum:** re-review confirmed 95b75b05 (wording + staging holds only this run's units) and d0c95201 (cleanup skipped under `--dry-run`); its one semantic note — a `--only` rerun re-stages only that subset — is now stated in the script's `--only` usage text (this commit). No open findings.

Issue #793, opened 2026-09-04
geecs-qserver: a running service does not mean ready — nothing opens the RE worker environment, and the empty plan list reads as "plan not in the list of allowed plans"


skbarber:
## Symptom (live, 2026-09-04)

A noscan submitted from GEECS-Console 0.28.2 was refused by the RE Manager:

```
Submission failed: Request failed: Failed to add an item: Plan validation failed:
Plan 'geecs_scan_request_plan' is not in the list of allowed plans.
```

The message points at the plan name, so the first hypothesis was that the
named-plans work (#769 — `geecs_noscan_plan` / `geecs_scan_plan` /
`geecs_optimize_plan`, GeecsBluesky 0.73.0) had renamed or de-permissioned the
funnel. It had not. The repo side was correct throughout:
`qserver/user_group_permissions.yaml` allows `:geecs_.*` (operator) and `:.*`
(admin/primary), and `qserver/startup/startup.py` still exports
`geecs_scan_request_plan` in `__all__` alongside the three named plans.

## Root cause

The manager's plan list was **empty**, so *every* plan name failed validation
identically. Queried against `tcp://192.168.6.14:60615` at the time of failure:

```
worker_environment_exists: False,  worker_environment_state: 'closed',  re_state: None
plans_existing: {}
plans_allowed: {}
```

`qserver environment open` fixed it immediately — environment and RE both
`idle`, and `plans_allowed` came back with all five GEECS plans
(`geecs_noscan_plan`, `geecs_optimize_plan`, `geecs_run_action_plan`,
`geecs_scan_plan`, `geecs_scan_request_plan`). The scan then ran.

Why it started now: the allowed-plans list is not in git. bluesky-queueserver
keeps it in `existing_plans_and_devices.yaml`, defaulting to the *startup
directory* (`bluesky_queueserver/manager/config.py:520` — `aux_dir`), and only
writes it when the worker environment is opened
(`--update-existing-plans-devices` default `ENVIRONMENT_OPEN`,
`manager/start_manager.py:416`). The site-profile arc's Phase 3 re-rendered
`geecs-qserver` onto `<root>/qs-checkout` — a fresh clone with no such file —
so the manager came back up knowing zero plans. Redis preserved the queue and
the 72-item history, which is why `qserver status` otherwise looked healthy and
`/fleet-status` reported the service as up.

Nothing opens the environment on service start: not the rendered unit, not
[...]

PR #794, opened 2026-09-04
Tooling hygiene wave: lock-path scrub hook (#753), handshake-completing MySQL probe (#790), queueserver readiness in /fleet-status (#793 part 3)

### LINKED ISSUE #753: poetry.lock files serialize the maintainer's absolute home path in geecs-bluesky extras lines (public-repo hygiene)
Surfaced by the #749 review. The root \`poetry.lock\`, \`GEECS-MCP/poetry.lock\` and \`GEECS-Console/poetry.lock\` carry the geecs-bluesky \`analysis\`/\`optimize\` extras as absolute URLs of the form

    imageanalysis @ file:///Users/<user>/.../GEECS-Plugins/ImageAnalysis

i.e. Poetry serializes a path-dep extra as an absolute \`file://\` URL resolved from wherever \`poetry lock\` ran — which is a user home path. These lines already exist on \`master\`; #749 re-touched them (picking up the \`capture\` extra) without changing the prefix.

This conflicts with the public-repo hygiene rule in root \`CLAUDE.md\` / \`CONTRIBUTING.md\` (no user home paths in committed files). The existing \`poetry-lock-no-worktree-paths\` pre-commit hook only guards the \`.claude/worktrees/\` suffix, not the home-path prefix.

Open questions for the fix: whether Poetry can be made to emit a relative/portable form for extras that reference path deps (it does for direct \`url = "../ImageAnalysis"\` entries but not for the extras strings), or whether the extras lines should be post-processed / the hook widened to reject \`file:///Users/\` and \`file:///home/\`. Note the path is also load-bearing for \`poetry install\` on that machine only if the extras are selected, so a generic placeholder is not obviously safe — needs a test on a clean checkout.

### LINKED ISSUE #790: lab_status / fleet_status: bare TCP probe of MySQL 3306 trips the server's max_connect_errors block (error 1129) for the whole VPN pool
## Symptom (live, 2026-09-04)

GEECS-Console launched from a VPN'd Mac showed the DB health chip DOWN and logged, on every completions fetch and health poll:

```
device completions unavailable for 'Undulator': 1129: Host '192.168.8.0' is blocked because of many connection errors; unblock with 'mysqladmin flush-hosts'
```

MySQL error 1129 is the server's `max_connect_errors` protection (default 100): it counts connections from a host that open TCP to 3306 and drop **without completing the MySQL handshake**, and once the count passes the limit it refuses every connection from that host until an admin runs `FLUSH HOSTS`. Successful logins reset the counter; failed authentication does not count. Only connect-and-drop does.

Fixed for now server-side (`FLUSH HOSTS`, 2026-09-04 morning); the chip is green again. This issue is the client-side fix so it does not recur.

## Cause

Our reachability probes are exactly the pattern MySQL counts:

- `scripts/lab_status.sh` `port_open()` — `exec 3<>/dev/tcp/$HOST/3306`, then the subshell exits. No handshake, no `COM_QUIT`. Tier 1 runs it on every `/lab-status` and every `/fleet-status` (stage 0 gates on it).
- `scripts/fleet_status.sh:134` has its own copy of the same `/dev/tcp` probe (`bounded ... exec 3<>/dev/tcp/$1/$2`).
- `fleet_status.sh --watch N` repeats the gate every N seconds — a watch pane alone reaches 100 aborted connects in well under an hour at short intervals.

Aggravating factor: the server sees VPN clients as `192.168.8.0` (the VPN pool's NAT address, not the individual Mac), so the counter is **shared across every VPN user** — one person's watch loop blocks the DB for everyone on VPN, and the block outlives the loop.

## Fix

1. **Replace the raw TCP probe for the DB with a handshake-completing one.** Bounded, read-nothing, but a real MySQL login + clean close using the `config.ini` credentials through the same connector `geecs_core.db.geecs_db` uses (`_connect_mysql`, `connect_timeout` ≈ 2 s), e.g. a `python -c` snippet from the `lab_status.sh` Tier-1 block, or `mysqladmin ping`-equivalent. A failed login is still a valid "port reachable" signal and does not count against the host. Keep the `/dev/tcp` probe for every non-MySQL port (Tiled, portal, MCP, CA, 0MQ) — it is only MySQL that penalizes it.
2. **Deduplicate the probe**: `fleet_status.sh` should call `lab_status.sh`'s helper (or source it) rather than carry its own `/dev/tcp` copy, so the DB special case lives in one place.
3. **Document the trap** in `scripts/lab_status.sh`'s header and the `/lab-status` skill: never probe 3306 with a bare TCP connect; and in `docs/platform/` next to the DB facts: what 1129 means, the shared-VPN-NAT consequence, and the `FLUSH HOSTS` remedy (needs DB admin on the server host).
4. Optional, server-side, owner's call: raise `max_connect_errors` in the MySQL server config so a single misbehaving client cannot lock out the VPN pool.

## Pins

- A test of the lab_status DB probe against a stub TCP listener asserts the client sends a MySQL handshake response (or `COM_QUIT`) before closing — i.e. the connection is not a bare connect-and-drop.
- `fleet_status.sh` contains no second `/dev/tcp` probe for `$DB_PORT`.

## References

- `scripts/lab_status.sh` `port_open()`; `scripts/fleet_status.sh:134`
- `GEECS-Core/geecs_core/db/geecs_db.py` `_connect_mysql` (the blessed connector path)
- MySQL: `max_connect_errors`, `performance_schema.host_cache` (`COUNT_HANDSHAKE_ERRORS`), `FLUSH HOSTS`

skbarber:
Three tooling-only fixes from the 2026-09-04 issue wave, one commit each. No package Python source changes, so no version bumps: everything lives in `scripts/`, `tests/`, `.pre-commit-config.yaml`, the two skills and the docs.

Closes #753
Closes #790
Part of #793 (part 3 only — the readiness unit (part 1) and the console `worker_ready` preflight (part 2) are GeecsBluesky/Console work for the next PR)

## Per-commit breakdown

| Commit | Concern | LOC |
|---|---|---|
| `13748adf` | **#753** `scripts/scrub_lock_paths.py` auto-fixing pre-commit hook: rewrites path-dep `file://` URLs in `poetry.lock` to `file:///GEECS-Plugins/<Pkg>`, exits 1 on a rewrite (the ruff-format contract `scripts/commit.sh` re-stages); `poetry-lock-no-worktree-paths` widened to reject `file:///Users/` and `file:///home/`; the three affected locks (root, GEECS-MCP, GEECS-Console) byte-rewritten — 9 substitutions on 6 lines, `poetry check --lock` passes in all three; root `CLAUDE.md` relock paragraph + `CONTRIBUTING.md` sentence; 7 pinning tests | 8 files, +353/−25 |
| `ee603783` | **#790** `scripts/mysql_probe.py` — a bounded real MySQL handshake (same pure connector as `geecs_core.db.geecs_db`, GEECS credentials when they resolve, else a throwaway probe user whose `Access denied` still proves the port and is not counted); `scripts/lib/net_probes.sh` — `bounded` / `port_open` / `mysql_probe` shared by both scripts, `port_open` refuses 3306; `fleet_status.sh` drops its own `bounded` + `/dev/tcp` copies; `lab_status.sh` MySQL line: OK / WARN blocked (1129) / `--` not probed / DOWN; docs: script header, `/lab-status` skill (trigger symptom, rule 4, capability row, failure vocabulary), `docs/platform/fleet_map.md` admonition + DB health-check cell; 5 tests incl. a stub MySQL server that records the HandshakeResponse | 7 files, +484/−47 |
| `17b39ae4` | **#793 part 3** `scripts/qserver_probe.py` — 0MQ `status` + `plans_allowed` → `ready` / `notready` / `down`; `fleet_status.sh` queueserver row prints `ready — manager idle · environment idle · RE idle · 5 plans allowed (primary) · queue 0`, or `[WARN] NOT READY — worker environment CLOSED …` (Attention list, `!` glyph), falling back to port liveness labelled "readiness UNKNOWN" without a local bluesky-queueserver env; `/fleet-status` skill stage-1 text + 2 warning rows; fleet-map health-check cell; 6 tests on the pure verdict + wiring | 6 files, +274/−8 |

- **#753 rewrite rule**: any `file://` URL whose *last path component* is a top-level package dir of this repo is rewritten, regardless of prefix — so macOS, Windows (`file:///C:/…`) and worktree prefixes all normalize, and a checkout named anything (`qs-checkout`, `portal-checkout`) too. URLs pointing anywhere else are left for the reject hook. The design report's `git rev-parse --show-toplevel` approach would miss a lock generated in a *different* checkout.
- **#753 CLAUDE.md**: the bold rule changed from "never `poetry lock` in a worktree" to "lock files carry no checkout paths" — the hook makes a worktree relock safe, so the manual strip-the-prefix remedy is gone. The `poetry check --lock` verification was run in all three dirs; `poetry install --dry-run` was not (the design report already did that for the same byte change).
- **#753 side finding, deliberately not touched**: GEECS-Console's lock is stale vs the path deps on disk (geecs-bluesky 0.62.0 locked vs 0.74.x). Its own relock PR.
- **#790 credentials**: the probe prefers GEECS `Configurations.INI` credentials (a real login, resets the server's counter) and falls back to user `geecs-lab-status-probe` with an empty password. Host always comes from the caller (`[tiled] uri`'s host, the script's existing "one box" assumption), not from `Configurations.INI`.
- **#790 interpreter**: `probe_python MODULE` looks in GEECS-Core → GeecsBluesky → root env → PATH `python3` (in-project `.venv` first, then `poetry env info`); `GEECS_PROBE_PYTHON` pins one (tests, worktrees). A machine with none reports `[ -- ] not probed` — never a bare TCP fallback. A blocked host (1129) still counts as `network: UP` (the route is fine; the DB is the one refusing).
- **#793 user group**: `plans_allowed` is asked for `primary` (the CLI/console default, admin-equivalent per `user_group_permissions.yaml`). `--user-group` overrides.
- **#793 table semantics**: a *ready* queueserver puts nothing in the Notes column (`✓`); the readiness summary is in the full log. Only NOT READY / UNKNOWN become notes.
- **Facility literals**: none added. `3306` and `60615` are protocol/fleet-map ports already in these scripts.

## Tests

- Root suite (`poetry run pytest tests -m "not integration and not gui"`): **27 passed, 62 deselected** (18 new: 7 scrub, 5 mysql probe, 6 qserver probe).
- `scripts/check.sh --lint --base origin/master`: OK. `mkdocs build`: the fleet-map admonition renders, no new errors.
- GEECS-Console and GEECS-MCP suites are not run locally: their only change is the lock's extras URL strings (verified by `poetry check --lock`); CI runs them.

## Hardware / live verification

Run from a VPN'd Mac against the deployed fleet, 2026-09-04:

- **#790**: `lab_status.sh` tier 1 → `[ OK ] MySQL 192.168.6.14:3306 (login, geecs credentials)` in 0.7 s; `fleet_status.sh --no-ssh` stage 0/1 otherwise unchanged. The 1129 and down paths are pinned by the stub-server tests (no live way to trigger the block without causing it).
- **#793 part 3**: `qserver_probe.py` → `ready manager idle · environment idle · RE idle · 5 plans allowed (primary) · queue 0`; the fleet table shows `✓`.
- **OWED**: the NOT READY path against the live worker — `qserver environment close`, run `scripts/fleet_status.sh --summary`, expect `[WARN] Queueserver … NOT READY — worker environment CLOSED …` in Attention and `!` in the table; then `qserver environment open`. Not done here because it takes the worker out of service.

## Review

Adversarial review per `/land` done — dispositions (C1–C5, R1–R3, P1, P2: eight fixed in `288de4b8` / `69b94a78` / `f1fe06b9`, three waived with reasons) in https://github.com/GEECS-BELLA/GEECS-Plugins/pull/794#issuecomment-5549317201. After the fixes: 39 targeted / 48 root-suite tests pass, lint OK, `mkdocs build` clean.


skbarber:
## Adversarial review dispositions

Findings from the fresh-context adversarial review (per `/land`), each **fixed** (commit) or **waived** (reason). Three fix commits on top of the original three; still no package source and no version bumps.

| # | Finding | Disposition |
|---|---|---|
| **C1** | `scripts/mysql_probe.py`'s "a server answered" branch classified any connector error with errno not None / not ≥ 2000 / not in `_DOWN_ERRNOS` as a server-side refusal — wrong for connector-raised errors such as `NotSupportedError` (errno −1, an auth plugin the pure connector lacks): the probe exited 0 "ok" on a connect-and-drop, the exact pattern #790 exists to eliminate. | **fixed** `288de4b8` — the server-answered branch requires `exc.sqlstate` (every server ERR packet carries one) or `0 < errno < 2000`; the connector's own 2xxx range is `down`; anything else is the new exit 5 `unverified` (never ok). Stub-server test: a greeting advertising `authentication_ldap_sasl_client` yields `unverified`, and the server receives no HandshakeResponse (confirmed the pre-fix probe printed `ok auth-refused (MySQL -1)` on that greeting). |
| **C2** | `scripts/qserver_probe.py::summarize` returned exit 0 `ready` when `plans_allowed` was `None` (second 0MQ request timed out) or `success: False`, contradicting the docstring and the fleet-map row. Also: the manager address was hand-built as `tcp://<host>:60615`, ignoring a `[qserver] control_addr` override. | **fixed** `69b94a78` — exit 6 `unknown … plans_allowed unanswered` (a closed environment still wins as `notready`); `fleet_status.sh` renders it `[WARN] readiness UNKNOWN` with a table note, so `!`, never `✓`. `--addr` is optional: the probe resolves `[qserver]` from `~/.config/geecs_python_api/config.ini` with `read_qserver_config`'s precedence (`control_addr` verbatim, else `tcp://<host>:60615`) — through `geecs_bluesky.qs_client` when it imports in the probe python, parsed directly otherwise; `fleet_status.sh` honors the same override. Tests for both verdict paths and the precedence table. Live: `qserver_probe.py` with no `--addr` → `ready … 5 plans allowed (primary)`. |
| **C3** | `mysql_probe._credentials` read `Configurations.INI` off the data share before connecting; a stalled SMB mount blocks past the 12 s `bounded` wall → rc 137 → `lab_status.sh` renders DOWN and sets `NET_UP=0`, stopping `fleet_status` at stage 0. | **fixed** `288de4b8` — the lookup runs on a daemon thread bounded by `--creds-timeout` (`TCP_TIMEOUT`, 2 s) and falls back to the probe user on timeout or any failure (pinned by a stalled-lookup test); `lab_status.sh` renders rc 137 as `[ -- ] not probed … killed at its wall … not a DB verdict` without flipping `NET_UP` (verified: a hanging probe python returns 137 at the wall). |
| **C4** | Root `CLAUDE.md`, `CONTRIBUTING.md` and the `scrub_lock_paths.py` docstring claimed "a bare name in an extras string resolves to PyPI on the next consumer install"; the #753 design report's Q2 table shows a hand-substituted bare name installing fine — the PyPI fallthrough appeared only in the PEP 621 restructuring experiment. | **fixed** `f1fe06b9` — reworded to that precise statement in all three places; the never-hand-edit instruction stays. |
| **C5** | `fleet_status.sh --summary` redirected only stdout to the log file, so a manager-timeout `logger.warning` (+ traceback) from `bluesky_queueserver` and probe stderr printed above the box table. | **fixed** `69b94a78` — the summary `exec` takes stderr along (`2>&1`, restored with stdout for the table), the way `--watch` already captures; the probe also quiets the `bluesky_queueserver` logger for its own expected timeouts so `--full` stays clean (a bogus `--addr` now prints only `down RE Manager did not answer`). |
| **R1** | The error-1129 narrative was written out five times (`lab_status.sh` header, `net_probes.sh` header, `mysql_probe.py` docstring, `/lab-status` rule 4, the fleet-map admonition). | **fixed** `288de4b8` (scripts) + `f1fe06b9` (skill, admonition) — the fleet-map admonition is the single owner and says so; the other four are one-line pointers; the skill no longer restates the port number anywhere (asserted in the edit). |
| **R2** | `tests/test_scrub_lock_paths.py::test_real_locks_are_clean` hardcoded 3 locks. | **fixed** `f1fe06b9` — globs `**/poetry.lock` excluding `extras/` and dot-dirs: 14 locks pinned, all clean. |
| **R3** | `probe_python` sits beside the older `poetry -C … env info` discovery in `lab_status.sh` tier 2 and the `fleet_status.sh` PVA block. | **waived** — migrating the older sites is outside this PR; follow-up. |
| **P1** | The readiness verdict belongs in `geecs_bluesky.qs_client`; the script should be a thin CLI over it. | **waived** for this PR — the concurrent GeecsBluesky readiness PR #795 (#793 parts 1+2) adds a pure readiness verdict to `qs_client`; rewiring `scripts/qserver_probe.py` onto it is a follow-up once #795 lands (needs `bluesky_queueserver_api` in the probe python's env list). Noted on #793. |
| **P2** (host part) | The DB probe overrode the INI's `[Database] ipaddress` with the Tiled-URI host. | **fixed** `288de4b8` — the probe targets `[Database] ipaddress`/`port` when the INI resolves (the server the real clients use), the Tiled-derived host only as fallback; the status line names the host:port actually probed and `lab_status.sh` prints that. Tests run the probe with an empty `HOME` so the stub, never a real INI target, is probed. Live tier 1: `[ OK ] MySQL <ini host>:3306 (login, geecs credentials)`. |
| **P2** (API part) | A public `geecs_core.db.probe_connection()` replacing the private `_find_credentials` import + reimplemented connect. | **waived** — a GEECS-Core API addition and version bump is outside a no-bump tooling PR; follow-up. |

Verification after the fixes: `poetry run pytest tests/test_scrub_lock_paths.py tests/test_mysql_probe.py tests/test_qserver_probe.py -q` → 39 passed; root suite (`-m "not integration and not gui"`) → 48 passed, 62 deselected; `scripts/check.sh --lint --base origin/master` → OK; `mkdocs build` → clean. Live from a VPN'd Mac: `lab_status.sh` tier 1 all OK with the INI-sourced DB target; `fleet_status.sh --summary --no-ssh` with a `bluesky-queueserver` env pinned → Queueserver `✓`, nothing on stderr.

skbarber:
**Review confirm round (same reviewer):** C1, C2, C3, C4, C5, R1, P2-host all CONFIRMED fixed. One new problem introduced by the R2 fix — the recursive `**/poetry.lock` glob at module import walked every `.venv`/`.git`/worktree in a developer checkout (~23 s of collection per root pytest run). **Fixed** in df62c88f: one-level glob (committed locks sit at depth ≤ 1), computed inside the test function; same 14 locks, ~0.05 s.

skbarber:
## Live verification 2026-09-04 (VPN, worker env idle, queue empty)

- `lab_status.sh` tier 1 from this branch: MySQL **OK (login, geecs credentials)** 0.7 s, Tiled OK, CA gateway OK, data mount OK — the handshake probe against the real server, no 1129.
- `qserver_probe.py` (Bluesky env python): **ready** — 5 plans allowed, queue 0.
- **NOT READY path (the owed check):** `environment_close` on the idle manager → probe **rc 5 notready** "worker environment CLOSED"; `fleet_status.sh --summary` queueserver row **`!` NOT READY: worker environment CLOSED** + the Attention WARN line. After the environment reopened (via #795's `geecs-qserver-ensure-ready`): probe rc 0, fleet row `✓`, no Attention line.

**Findings for the next review (not fixed here):**
1. With the environment closed, the manager still answers `plans_allowed` with the 5 plans (it preloads the on-disk `existing_plans_and_devices.yaml`, present on the worker since the first open). The notready line therefore contradicts itself: "… CLOSED — **0 plans**, every submission is refused … · **5 plans allowed (primary)**". The verdict (closed ⇒ not ready) is right; the "0 plans" wording describes only the fresh-clone case.
2. From a worktree with no envs, `probe_python` falls back to `python3` and the probe prints **`down` bluesky-queueserver is not importable** (rc 1) — an environment problem rendered as the service being down; `fleet_status` renders the same case as "readiness unknown (port only)", which is right. Same class as the C3 finding (env/mount trouble ≠ DOWN). Tier 2's "[DOWN] CA gateway … GeecsBluesky poetry env not installed" is the same misattribution.
3. Pre-existing fleet drift still visible: the queueserver row reports **2 processes for this role** (an unmanaged manager on :60615 beside the systemd one).

Issue #772, opened 2026-09-03
Scan-variable editor accepts targets that name no real device variable (free-text + completer, no save-time check); manual moves skip the unserved-variables preflight


skbarber:
## Symptom (live, 2026-09-02)

A composite scan variable (`grating_angle_w_DMsurf_TRA`) was saved from the console's scan-variable editor with the target `U_Grating2Rotation:Position.Axis1`. The ESP301 device type names that variable `Position.Axis 1` (with a space; the Aerotech compressor stage is the one that uses `Position.Axis1`). The editor accepted it, Save accepted it, and the failure only surfaced when moving the composite from the movable panel — as a 20 s ophyd-async connect timeout on the worker:

```
_setpoint_0: NotConnectedError: ca://undulator:u_grating2rotation:position_axis1:SP
_target_readback_0: NotConnectedError: ca://undulator:u_grating2rotation:position_axis1
```

Nothing in that message says "no such variable", so it was first chased as a gateway/device connectivity problem (the device was *also* missing from the gateway roster — added to the DB after the gateway's last start — which masked the real cause until a restart).

## Root cause

`GEECS-Console/geecs_console/editors/scan_variable_editor.py`: the device and variable fields (simple target, confirm target, and every composite component row) are plain `QLineEdit`s with a `QCompleter` attached. The completer *suggests* names from the DB device-variable listing but never constrains input — any text is accepted. The completer is contains-matching, so a near-miss like `Position.Axis1` matches nothing and the popup goes silently quiet.

`_save_draft` validates only the pydantic shape (`ScanVariables.model_validate`) — that a target is a `device:variable` string — not that the device or variable exists.

`save_set_editor.py` shares the same base (`editors/base.py` completions plumbing) and the same free-text + completer pattern, so it very likely has the same gap.

## Proposed fix

1. **Save-time target validation in the editors** (scan-variable editor; check the save-set editor for the same shape). When the completions listing has loaded (`self._device_vars` non-empty), an unknown device or unknown variable-for-that-device blocks Save with an inline error naming the offending variable and row, with a near-miss hint where one exists ("did you mean `Position.Axis 1`"). Device match should be case-insensitive (the completer already tolerates case — see `_refresh_variable_model`). When the listing is unavailable (offline / DB down / fetch failed), downgrade to a warning and allow the save so the editor stays usable off-network.
[...]

Issue #490, opened 2026-07-11
CA put-completions to a device :SP intermittently vanish in loaded session processes (action set-steps hang 30s)


skbarber:
Found during the M4 step-(i) hardware smoke (2026-07-10). An action set-step to `Undulator:U_ESP_JetXYZ:Position_Axis_1:SP` hangs for its full CA timeout at scan closeout, deterministically — while the device, gateway, circuit, and event loop are all demonstrably healthy at the same instant.

## Fingerprint
- `aioca CANothing: ECA_TIMEOUT` after the full client timeout (put-with-callback never completes)
- The write **never executes** in the scan context (zero-writer discriminator: readback pinned at 5.1000 through the whole 30 s window with the closeout targeting a distinct 5.02)
- Deterministic at scan closeout; reproducible without a scan via `compile_action_plan` in a session with devices connected (**18/18 failures** across three transports); flaky-to-absent in minimal processes

## Every layer exonerated by direct observation (all during an active hang)
| Probe | Result |
|---|---|
| Fresh-process `caput` (own CA context) | ✅ completes in 2 s, device executes |
| Same-process, same-loop `caget` | ✅ |
| Same-process, same-loop, same-channel second `caput` (threadsafe task) | ✅ completes ~1 s |
| Device TCP push stream | ✅ streaming throughout |
| Gateway per-device UDP lock | ✅ free (parallel put would otherwise queue — `_exchange` holds the lock across ACK+exe) |

The 30 s (not 1.5 s) timing rules out a lost/rejected fast-ACK path client-visibly; the free gateway lock + no execution says the write request never reaches the gateway at all.

## Transport-independence (the part that kills every simple theory)
In the no-scan harness (session + CaMotor connected, puts via `RE(compile_action_plan(...))`, benign 5.00↔5.02 toggles):
- (a) raw `aioca.caput` + dtype-inferred probe signal: **6/6 fail**
- (b) raw `caput`, no probe: **6/6 fail**
- (c) fresh `epics_signal_rw(float)` ophyd put: **6/6 fail**
- Yet `CaMotor.set()` (the *same* ophyd put-with-completion via the session's own signal) works 12/12 mid-scan every run, and `CaPutSetter` raw caputs (DG645, cameras) complete at finalize seconds before the failing put.

[...]

PR #796, opened 2026-09-05
GEECS-Console 0.30.0: editor save-time target check (#772), Ops → Restart gateway (#773), every cross-thread signal on the background hop (#787)

### LINKED ISSUE #773: GEECS-Console: expose a gateway restart (cagateway:restart) from the console — Ops action with confirmation, scan-active gate, heartbeat narration
## What

Expose the CA gateway's restart verb in the GEECS console: an **Ops → Restart gateway…** action (and/or a small button beside the R1 gateway health chip) that writes `Restart` to `{experiment}:cagateway:restart` after a confirmation modal, then narrates the bounce in the status bar / R6 log tail until the heartbeat is back.

## Why

The gateway's one client-writable PV (`PV_CONTRACT.md` § `cagateway:restart`) is the operator fix for the frozen-readback symptom — a stale GEECS TCP subscription after a device-app restart (half-open socket; the supervisor only reconnects on socket *close*). Live incident 2026-08-17 on u_s1h: Phoebus readbacks frozen while `:SP` writes still moved hardware; fixed by writing `Restart` from a terminal. It is also the DB-resync mechanism after a `expt_device_variable` edit. Today the only ways to trigger it are `caput` or Phoebus; the console — the operator's DAQ surface, already holding the gateway health chip — has no button for it. Issue #611 (staleness self-detection) is the long-term fix; this is the operator escape hatch until then.

## Design sketch

- **Where**: Ops menu item "Restart gateway…" (enabled only while an experiment is selected and the gateway chip is not UNKNOWN); optionally a compact ⟳ beside the gateway chip.
- **Guard**: confirmation modal via `_ask_binary` stating the cost — a few seconds of fleet-wide CA disconnect (every client's monitors reconnect on their own), and a refusal while a scan is active on the manager (`_scanning()` — the same gate Start uses), since the worker's ophyd-async signals would see the disconnect mid-scan.
- **Write path**: `GatewaySetpointPut`-style put through the blessed CA helpers (`ca_pv(experiment, "CAGateway", "RESTART")` — never a hand-built PV name, issue #490 rule), on a `BackgroundResult` worker (the 0.28.0 GUI-hop shape), result to the status bar.
- **Narration**: after the put, the R1 health poll's heartbeat check does the rest — chip goes DOWN then OK; log tail lines "gateway restart requested" / "gateway back (uptime N s)" would make the bounce legible. `cagateway:uptime` dropping to ~0 is the confirmation the restart actually happened (a `Restart` write on a gateway not run under a restart-on-86 supervisor leaves it *down* — the contract's exit-code-86 deployment note; worth surfacing in the confirmation text).

## Acceptance

- Ops action present and gated as above; pinned by tests in `tests/test_main_window.py` (gating on experiment / chip state / scanning, confirmation abort leaves nothing written, confirm writes exactly one put to the restart PV, put failure reported).
- Hardware: trigger from the console on the deployed gateway; chip DOWN → OK within ~10 s, `devices_connected` recovers to its prior count, a previously frozen readback resumes.

## References

- `GeecsCAGateway/PV_CONTRACT.md` § `cagateway:restart` (exit code 86, `Idle`/`Restart` enum, DB-resync semantics)
- Incident + fix pattern: 2026-08-17 frozen readbacks (see #611 for the self-detection follow-up)
- Console rules: `GEECS-Console/CLAUDE.md` — Ops menu, `GatewaySetpointPut` as the one put primitive, `BackgroundResult` hop

skbarber:
## Summary

GEECS-Console 0.29.0 → **0.30.0**, three commits, one bump, one CHANGELOG entry. The operator-surface PR of the 2026-09-04 fix wave (PR A = tooling, PR C = GeecsBluesky readiness).

| Commit | Issue | Product | Tests | Docs/meta |
|---|---|---|---|---|
| `649452e6` editors: save-time device/variable check with near-miss hints | #772 part 1 | +268 / −6 | +251 | — |
| `91d97250` Ops → Restart gateway… | #773 | +270 / −2 | +204 / −7 | — |
| `e3eb39b6` every cross-thread signal rides the background hop | #787 | +290 / −146 | +191 / −3 | +102 / −25 |

Closes #773.
Part of #772 (part 1 — the editor gate; part 2, the manual-move unserved-variables preflight, is in the GeecsBluesky readiness PR).
Part of #787 (emitter consolidation; the window ↔ controller reference cycles and the once-seen interpreter-exit segfault remain open).

## What each commit does

**#772 part 1 — editor save gate.** The scan-variable editor (simple target, confirm target, every composite component row) and the save-set editor (entry device, listed scalars) check every name against the fetched completions listing at Save. An unknown device (case-insensitive match, like the completer) or an unknown variable for that device blocks the save with an inline error naming the variable and row plus a difflib "did you mean" hint. The live case, `Position.Axis1` for `Position.Axis 1`, scores 0.97 and hints correctly. Free text stays: with the DB unreachable a real provider that came back empty saves with a "names were not checked" note; the `EmptyCompletions` default (no experiment) saves quietly. Shared helpers (`target_problem`, `resolve_device`, `near_miss`) live once in `editors/base.py`.

**#773 — Ops → Restart gateway….** Writes `Restart` to `{experiment}:cagateway:restart` (the gateway's one client-writable PV) through `GatewaySetpointPut` with `wire_value` coercion on a `BackgroundResult` worker; PV name via `ca_pv`/`bare_pv` only (#490). Enabled only with an experiment selected and a *known* gateway chip (OK/WARN/DOWN; UNKNOWN = no poll yet); refused while a scan is active on the manager (`_scanning()`, the Start gate); one `_ask_binary` confirmation stating the cost (fleet-wide CA disconnect for a few seconds; a gateway outside its restart-on-exit-86 supervisor stays down). The health poll then narrates the bounce in the log tail: "gateway restarting — heartbeat down" on the first DOWN, "gateway back — heartbeat OK" on the first OK, then disarms.

**#787 — hop consolidation.** The three residual daemon-thread → consumer emitters moved onto `services/background.py`: the scan monitor's stream workers subclass `_GuiHopWorker` (zmq thread posts tagged payloads; `document`/`line`/`pause_reason`/`stream_failed` fire on the GUI thread; `stop()` gates *both* sides of the hop so an in-flight payload lands and is dropped; threads stay abandoned), the movable panel's aioca readback callback posts through a new public `GuiRelay` (controller `value_ready` removed), and the editors' completions fetch is a plain `BackgroundResult` (`completions_ready` removed). Long-lived producers take the in-flight hold on their own thread, so `_INFLIGHT` is now lock-guarded. `ScanMonitorController.dispose()` severs only the public signals (a whole-object `QObject.disconnect()` cut the internal hop and stranded an in-flight hold) and the never-connected-poller `libpyside: Failed to disconnect (None)` RuntimeWarning is gone (`disconnect_quietly`).

## Tests

- `./scripts/check.sh --base origin/master`: lint clean; **GEECS-Console 877 passed** (was 865 on master; +12 net, 35 new tests across `test_editor_target_check.py`, `test_gateway_restart.py`, `test_scan_variable_editor.py`, `test_save_set_editor.py`, `test_main_window.py`, `test_scan_monitor.py`, `test_background.py`).
- Isolated `QT_QPA_PLATFORM=offscreen pytest tests/test_main_window.py` (the #767/#787 segfault shape), macOS arm64, PySide 6.11.1: baseline on master **3/3 clean** (142 passed), after this branch **4/4 clean** (142 passed). The conftest safe-point collection from #788 already holds this file stable here; this PR removes the last three emitters of the racing shape rather than fixing a currently reproducing crash.
- Suite run with `-W error::RuntimeWarning` on the #787-touched files: clean.

## Judgment calls

- **Save-set editor had the same gap** (free text + completer, no save-time check) — fixed in the same commit, mapped onto its document shape (entry device, then that entry's scalars; an unknown device skips its scalars since they cannot be checked).
- **Variable match is case-sensitive, device match is not.** The issue asked for device case-insensitivity (the completer's behavior). A case-only variable mismatch gets a hint ("did you mean 'Velocity'") rather than silent acceptance, since the target string is stored verbatim.
- **"Unchecked" note only when a listing was expected**: a pre-existing pin (`test_error_clears_after_a_good_save`) showed that nagging on every save of an `EmptyCompletions` editor is wrong; the note now fires only when a real provider returned empty (offline/DB down).
- **Restart PV value** is the label `"Restart"` (contract: label or index 1); the put is `GatewaySetpointPut(setpoint_pv=<bare restart pv>)` even though the PV has no `:SP` suffix — it is the control PV itself, and the primitive normalizes the name.
- **The restart put runs on `asyncio.run`**, not the shared one-shot loop (`geecs_bluesky.devices.ca.oneshot` serves reads and keeps its loop private). aioca caches channels per loop, so each click strands one channel; accepted for an incident-rate verb, documented in the module.
- **Restart action enabled on DOWN**: a DOWN gateway under its supervisor is exactly when a restart may help; only UNKNOWN (no experiment / no poll yet) disables.
- **`apply_operator_tooltips` now runs after `_build_menus`** — the catalog names the new action and raises on a missing name.
- **Ops-menu pin reads the menu through `window._menus`**: a QMenu reached via `menuBar().actions()[i].menu()` is torn down with its actions when that temporary wrapper is collected, and since #773 the window re-gates the action on every health poll — the old enumeration surfaced as a RuntimeError after the test.
- **`GuiRelay` is public, `_GuiHopWorker` stays private**: the stream workers need typed signals so they subclass the base directly (the sanctioned pattern per CLAUDE.md); the movable panel needs only a payload relay.
- **`sever()` instead of `QObject.disconnect()`** in the monitor's dispose — the blanket disconnect was silently cutting the hop's own connection.

## Hardware verification (owed)

Everything here is hermetic; the #773 action needs the deployed gateway. Steps:

1. On a console with the lab network (VPN is fine), select the experiment, wait for the gateway chip to read `ok`, confirm **Ops → Restart gateway…** is enabled and that it is disabled while a scan is running.
2. Note `caget <experiment>:cagateway:devices_connected` and `…:uptime`.
3. Click **Restart gateway…**, confirm. Expect in the log tail: "restarting the <experiment> gateway…", then "gateway restart requested (<pv>) — expect a few seconds of CA disconnect".
4. Within ~10 s: chip `down` then `ok`; log tail "gateway restarting — heartbeat down" (may be skipped if the bounce falls between 5 s polls, in which case the "back" line says so) and "gateway back — heartbeat OK".
5. `…:uptime` reads ~0 and `…:devices_connected` recovers to its prior count; a previously frozen readback (if one exists) resumes.
6. Negative: with a scan running, click the action — status bar "Gateway restart refused — a scan is active on the manager", nothing written.

The #772 editor check can be tried live too: in Editors → Scan Variables, set a target to `U_Grating2Rotation:Position.Axis1` and Save — expect the inline error with the `Position.Axis 1` hint (DB completions must have loaded; offline the save goes through with the note).

## Review

Adversarial review done; findings 1–7 dispositioned (6 fixed on this branch in `33bd2e0c`, `b21fc131`, `faa78e1b`, `b81f1ad3`; the placement half of finding 1 waived as a cross-package follow-up) in https://github.com/GEECS-BELLA/GEECS-Plugins/pull/796#issuecomment-5549421897 — the same comment records the pre-existing isolated-run segfault (reproduces on the pre-fix head `e3eb39b6`, ~1 in 12).

### DIFF
--- a/GEECS-Console/CHANGELOG.md
+++ b/GEECS-Console/CHANGELOG.md
@@ -4,6 +4,61 @@ All notable changes to GEECS-Console are documented here.  Format follows
 [Keep a Changelog](https://keepachangelog.com/en/1.1.0/); versioning is
 semantic.
 
+## [0.30.0] - 2026-09-04
+
+### Added
+
+- **Save-time device/variable check in the editors** (#772 part 1).  The
+  scan-variable editor (simple target, confirm target, every composite
+  component row) and the save-set editor (entry device, listed scalars)
+  check every name against the fetched completions listing when Save is
+  clicked: an unknown device (case-insensitive, like the completer) or an
+  unknown variable for that device blocks the save with an inline error
+  naming the variable and row plus a difflib "did you mean" hint — the
+  live case was `Position.Axis1` for `Position.Axis 1`, which used to
+  save cleanly and fail 20 s into the first move as a bare connect
+  timeout.  Free text stays: with the DB unreachable a real provider
+  that came back empty saves with a "names were not checked" note, and
+  the `EmptyCompletions` default (no experiment) stays quiet.  Shared
+  helpers in `editors/base.py` (`target_problem`, `resolve_device`,
+  `near_miss`).
+- **Ops → Restart gateway…** (#773): writes `Restart` to the
+  experiment's `cagateway:restart` PV (the gateway's one client-writable
+  PV — the operator fix for frozen readbacks and the DB-resync after a
+  device edit) through `GatewaySetpointPut` on a `BackgroundResult`
+  worker (`services/gateway_restart.py`, import-safe offline, PV name
+  via `ca_pv`/`bare_pv`).  Enabled only with an experiment and a known
+  gateway chip; refused while a scan is active on the manager; one
+  confirmation modal states the cost (fleet-wide CA disconnect for a few
+  seconds; a gateway outside its restart-on-exit-86 supervisor stays
+  down); the health poll then narrates the bounce in the log tail
+  ("gateway restarting — heartbeat down" / "gateway back — heartbeat
+  OK").  Operator tooltips are now applied after the menus are built,
+  since the catalog names this action.
+
+### Changed
+
+- **Every cross-thread signal now rides the `services/background.py`
+  hop** (#787).  The three residual daemon-thread emitters inventoried
+  after 0.28.2 moved: the scan monitor's stream workers are
+  `_GuiHopWorker`s (the zmq thread posts tagged payloads; `document` /
+  `line` / `pause_reason` / `stream_failed` fire on the GUI thread, and
+  `stop()` also gates a payload already in flight), the movable panel's
+  aioca readback callback posts through the new public `GuiRelay` (the
+  controller's `value_ready(int, object)` signal is gone — `delivered`
+  carries `(index, value)`), and the editors' completions fetch is a
+  plain `BackgroundResult` (the dialogs' `completions_ready` signal is
+  gone).  Long-lived producers take the in-flight hold on their own
+  thread, so the counter is now lock-guarded.  `ScanMonitorController.
+  dispose()` severs the workers' public signals instead of
+  `QObject.disconnect()` on the whole worker (which cut the internal hop
+  and stranded an in-flight hold), and disconnects quietly — the
+  `libpyside: Failed to disconnect (None)` RuntimeWarning on a
+  never-connected poller is gone (`disconnect_quietly`).  Delivery takes
+  one more event-loop turn; tests that assumed a same-turn landing wait
+  for it.  Still open on #787: the window ↔ controller reference cycles
+  and the once-seen interpreter-exit segfault.
+
 ## [0.29.0] - 2026-09-04
 
 ### Removed
--- a/GEECS-Console/CHANGELOG.md
+++ b/GEECS-Console/CHANGELOG.md
@@ -60,6 +60,14 @@ semantic.
   and the once-seen interpreter-exit segfault.
 
 ### Fixed
+- Editors (#772): a device listing still *loading* no longer reads as
+  "DB unavailable" — Save refuses with "device listing still loading" until
+  the fetch lands, instead of persisting unchecked names one poll early
+  (`completions_pending`; Codex review of PR #796).
+- Ops → Restart gateway: the scan-active gate also reads the
+  document-driven state pill, so the window between a start document and
+  the matching status poll can no longer wave a restart through (Codex
+  review of PR #796).
 
 - **Restart gateway… (adversarial review of PR #796)**: the restart put
   rides the device-panel backend's persistent CA loop (new
--- a/GEECS-Console/CHANGELOG.md
+++ b/GEECS-Console/CHANGELOG.md
@@ -59,6 +59,25 @@ semantic.
   for it.  Still open on #787: the window ↔ controller reference cycles
   and the once-seen interpreter-exit segfault.
 
+### Fixed
+
+- **Restart gateway… (adversarial review of PR #796)**: the restart put
+  rides the device-panel backend's persistent CA loop (new
+  `GatewayDevicePanel.put_pv`, which the R7 `set` now delegates to)
+  instead of a per-call `asyncio.run` — aioca caches channels per loop,
+  so the gateway's own CONN_DOWN used to print `RuntimeError: Event loop
+  is closed` from the CA thread on every click; the gate is re-checked
+  after the confirmation modal (status polls keep landing during its
+  nested `exec()`) and also refuses while the movable panel's manual
+  set/move is in flight (a worker-side move never changes `re_state`);
+  and the bounce narration ignores reports from health polls that began
+  before the put completed (`HealthReport.sequence`, stamped by
+  `HealthPoller`) — a pre-put "OK" landing late no longer reads as
+  "gateway back".  The Ops menu now renders its actions' tooltips.
+- The save-set editor's scalar completer resolves the entry's device
+  case-insensitively, matching its save-time check; the scan-variable
+  editor's completer uses the shared `resolve_device` helper.
+
 ## [0.29.0] - 2026-09-04
 
 ### Removed
--- a/GEECS-Console/CLAUDE.md
+++ b/GEECS-Console/CLAUDE.md
@@ -215,17 +215,24 @@ are prefixed by region (`r3_radio_1d`, `r5_start_button`, …).
   over `Submitter.status()`), the document stream (`RemoteDispatcher`
   on the worker's proxy out-port — per-shot progress, totals, scan
   number, terminal states), and the manager console-output stream (log
-  tail + failed-move pause reasons).  Stream workers emit on
-  worker-owned signals connected `QueuedConnection`;
-  `stop()`/`dispose()` only gate emission and **abandon** the daemon
-  threads — a zmq socket must never be closed from another thread (a
-  libzmq assertion aborts the process; #653 review).  Stream setup
-  failure emits `stream_failed(str)`, which the window surfaces via
-  `_report` — degraded mode (no progress/log stream) is never silent.
+  tail + failed-move pause reasons).  **Since 0.30.0 (#787) the stream
+  workers are `_GuiHopWorker`s**: the zmq thread posts a tagged payload
+  through the worker's own queued hop and `document` / `line` /
+  `pause_reason` / `stream_failed` are emitted on the GUI thread (the
+  window still connects them `QueuedConnection`); `stop()`/`dispose()`
+  gate emission on **both** sides of the hop (a payload already in
+  flight lands and is dropped) and **abandon** the daemon threads — a
+  zmq socket must never be closed from another thread (a libzmq
+  assertion aborts the process; #653 review).  `dispose()` severs the
+  public signals only (`sever()`), never `QObject.disconnect()` on the
+  whole worker — that would cut the internal hop and strand an
+  in-flight hold.  Stream setup failure emits `stream_failed(str)`,
+  which the window surfaces via `_report` — degraded mode (no
+  progress/log stream) is never silent.
 - **Movable panel (R7)** — owned by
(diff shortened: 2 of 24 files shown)


skbarber:
## Adversarial review dispositions

Findings from the adversarial review of this PR (correctness + redundancy + placement), each fixed on this branch or waived with a reason. GEECS-Console stays at 0.30.0; the CHANGELOG entry gained a **Fixed** section for findings 1–3.

| # | Finding | Disposition |
|---|---|---|
| 1 | `gateway_restart.py` ran the put under a per-call `asyncio.run`; aioca caches channels per loop and `Channel.on_ca_connect` posts to that loop unguarded, so the gateway's own exit (this click) fires CONN_DOWN on the stranded channel and the CA thread prints `RuntimeError: Event loop is closed` on every restart. | **fixed** `33bd2e0c` — `GatewayDevicePanel.put_pv` is the console's one CA put, on the panel's persistent daemon-thread loop; `set()` delegates to it and `request_gateway_restart(experiment, backend=...)` writes through it (the offline stub refuses with a clear message). Pinned by a no-`asyncio.run` source check, a same-loop test, and a window test over the backend. **Placement half waived** — moving the put primitive into `geecs_bluesky.devices.ca` (a `caput_once` beside `gateway_put`/`liveness`, for the MCP and #611) is a cross-package follow-up; the console reuses its existing persistent loop instead. |
| 2 | `_on_restart_gateway` checked `_scanning()` only before the confirmation modal, whose nested `exec()` keeps status polls landing. | **fixed** `b21fc131` — `_restart_refusal()` runs before AND after the modal; test flips the scan state inside the confirm callback and asserts no put. |
| 3 | `_restart_pending` armed in `_on_restart_result`, but a health poll already out before the click could land after it with a pre-put OK and narrate "gateway back" while the gateway was going down. | **fixed** `faa78e1b` — `HealthPoller` numbers its polls (`polls_started`, GUI thread) and stamps each delivered `HealthReport` with its poll's `sequence` (exact: one poll in flight, stamped on the GUI thread before the next can start); the window records `polls_started` when the put completes and narrates only later sequences. Chips still follow every report. Tests: a pre-arm report after arming, an end-to-end gated-probe run, poller stamping. |
| 4 | `_scanning()` reads `re_state` only; a worker-side `move_variable` (`function_execute`) never changes it, so a restart during the movable panel's in-flight set was permitted. | **fixed** `b21fc131` — `MovablePanelController.set_in_flight` accessor, read by `_restart_refusal()`; test drives a real gated set and asserts the refusal, then the restart once it lands. |
| 5 | `tooltips.py` catalogs `restart_gateway_action`, but QMenu renders action tooltips only under `setToolTipsVisible(True)`. | **fixed** `b81f1ad3` — set on the Ops menu; pinned in `TestOpsMenu`. |
| 6 | `GEECS-Console/CLAUDE.md` still said the Ops menu has four items. | **fixed** `b81f1ad3` — fifth item documented with its gating (both sides of the modal, manual move), the persistent-loop put, the poll-sequence narration and the tooltip flag. |
| 7 | Redundancy: (a) `scan_variable_editor._refresh_variable_model` duplicated `resolve_device`; (b) `save_set_editor._refresh_variable_completer` was exact-match while its save check is case-insensitive; (c) `disconnect_quietly` was introduced but the diff's own dispose sites kept the try/except form. | **fixed** `b81f1ad3` — (a) uses the helper; (b) case-insensitive via the helper, pinned by a new test; (c) `movable_panel.dispose`, the window's `closeEvent` (poller + the three workers incl. `_restart_worker`) and `editors/base.py::closeEvent` use `disconnect_quietly`. |

### Tests after the fixes (macOS arm64, PySide 6.11.1, offscreen)

- Full GEECS-Console suite: **892 passed** (was 877; +15 new tests across `test_gateway_restart.py`, `test_device_panel.py`, `test_main_window.py`, `test_background.py`, `test_save_set_editor.py`).
- `scripts/check.sh --lint --base origin/master`: clean.
- Isolated `tests/test_main_window.py` (148 tests): first three runs 2/3 clean — one segfault. Chased it: **12 runs on this branch → 1 segfault; 12 runs with the six new window tests deselected → 1 segfault; 12 runs on the pre-fix PR head `e3eb39b6` (via `git archive`, baseline package on `PYTHONPATH`, verified from the trace paths) → 1 segfault.** Same shape every time: `Fatal Python error: Segmentation fault`, main thread `Garbage-collecting` inside pytest-qt's post-call `_process_events`, a `bluesky_queueserver_api.console_monitor._thread_receive_msgs` zmq `poll` thread alive — the #767/#787 GC-during-`processEvents` window (the reference cycles the PR body lists as still open on #787), not introduced by this PR and not correlated with the new tests. The PR body's "4/4 clean" was a small sample; the rate here is ~1 in 12. Reported as pre-existing; not fixed in this PR.

skbarber:
**Review confirm round (same reviewer):** findings 1–7 all CONFIRMED fixed; no new problems. One layering note recorded, not a defect: `services/background.py` now imports `HealthReport` from `services/health.py` to stamp the poll sequence in `HealthPoller._deliver` (no cycle; a future kit extraction would move the stamp into a subclass or callback). Residual isolated `test_main_window.py` segfault measured as **pre-existing**: fixed head 0/18, pre-fix PR head 1/12, master 421528d0 itself 2/12 — tracked on #787.

skbarber:
## Codex review dispositions

1. **`targets_unchecked` conflated "fetch not landed" with "DB unavailable"** — **fixed** f3ab30b2: new `completions_pending` (real provider, fetch not yet applied) is distinct from `targets_unchecked` (fetch landed empty); both the scan-variable and save-set editors refuse Save with "device listing still loading — Save again in a moment" while pending, then run the ordinary check once it lands. Pinned in both editor suites with a gated provider (save refused while pending, near miss blocks after).
2. **`_restart_refusal()` read only the polled status** — **fixed** f3ab30b2: it also refuses while the document-driven pill is not quiet (`_QUIET_PILL_STATES` = idle/done/aborted/unknown), covering the start-document-before-poll window. Pinned: idle poll + RUNNING pill refuses; a terminal pill lets the put through.
3. **Open question — action-library target validation:** deliberately deferred. #772 named the scan-variable and save-set editors (the two whose targets name device variables); action steps also carry non-DB verbs (waits) and need their own shape. The `editors/base.py` module doc now says so explicitly.

Full console suite 895 passed; lint OK.

PR #797, opened 2026-09-05
GeecsPvaGateway 0.5.0: DB-driven PVA fleet roster with a not-deployed state (site-profile Phase 4, item 1)

skbarber:
## What

Site-profile arc **Phase 4, item 1** (`Planning/site_profile/00_overview.md`): the PVA image-gateway fleet roster comes from the GEECS DB per experiment, and a roster host can be **not deployed** instead of alarming as DOWN every run.

- **`geecs_pva_gateway.fleet`** (new): `fleet_roster(experiment)` = every endpoint IP hosting an enabled image device for the experiment (the same camera test `PvaGatewayConfig` applies per host, now via a shared `config.image_variables`), each marked `deployed` by membership in the client `config.ini`:
  ```ini
  [pva]
  addr_list = 192.168.6.100 192.168.7.161 ...   # the camera servers running an instance
  ```
  That list is the clients' `EPICS_PVA_ADDR_LIST` for cross-subnet search (mirrors `[epics] ca_addr_list`), so "deployed" and "where clients search" are one value with one home. Absent key = every roster host deployed (the pre-key behaviour).
- **`deploy/gen_fleet_status.py --experiment X`** writes `fleet_status_<x>.bob` from that roster; the hand-curated `HOSTS`/`EXPERIMENT` literals are gone. Not-deployed rows render as a label with no live PVs and no restart button. `fleet_status.bob` (13 hosts) → `fleet_status_undulator.bob` (HTU, the reference deployment: 11 hosts in the DB, 9 deployed).
- **`scripts/fleet_status.sh`** reads the roster live from the DB through the PVA env instead of grepping the checked-in screen; not-deployed hosts print as `[ -- ]` rows.
- Docs kept true in the same PR: `GeecsPvaGateway/DEPLOYMENT.md` (rollout + client access), `CLAUDE.md` (package + root residue list), `docs/geecs_gateway/image_pvs.md`, `docs/platform/fleet_map.md`, `docs/platform/site_profile.md`, `docs/tutorials/getting_started.md`, the `/fleet-status` skill.

## Why

The old `HOSTS` list was a facility literal in code (root CLAUDE.md § "Facility values have one home") and had drifted: two of its 13 hosts (`.7.203`, `.8.207`) no longer have image devices in the DB at all, and two (`.6.66`, `.6.73`) host one camera each but never had the service installed — `/fleet-status` reported all four as DOWN on every run. Decisions (Sam, 2026-09-04): mark home = `config.ini [pva] addr_list`; script reads the DB live; HTU screen stays committed.

## LOC by concern (+645 / −304 over 16 files)

| Concern | Files | Δ |
|---|---|---|
| roster module + shared camera test | `fleet.py` (new), `config.py` | ~+150 / −8 |
| generator rewrite + regenerated screen | `gen_fleet_status.py`, `fleet_status_undulator.bob` (renamed, 58% similar) | ~+190 / −230 |
| observed-fleet probe | `scripts/fleet_status.sh` | +38 / −26 |
| tests | `tests/test_fleet.py` (new) | +110 |
| docs + skill + version/changelog | 9 files | ~+120 / −40 |

## Tests

`./scripts/check.sh`: lint clean; **GeecsPvaGateway 24 passed** (was 14 — 10 new in `test_fleet.py`: roster derivation + IP ordering, absent-key = all deployed, addr_list marks not-deployed + keeps/warns a stale listed host, instance-PV naming contract, config.ini experiment fallback, generated rows deployed vs not, generator exit codes).

## Hardware verification

**Done, 2026-09-04, from this worktree on the lab network:**
- `gen_fleet_status.py` against the live DB: 11 rows, 9 deployed, 2 not deployed (matches the hand count).
- `scripts/fleet_status.sh --no-ssh --experiment Undulator`: 9 `[ OK ]` on 0.4.4 with heartbeats, `[ -- ] 192.168.6.66 not deployed (UC_Stretcher_MI)`, `[ -- ] 192.168.6.73 not deployed (U_FROG_Grenouille)`, **no DOWN rows** (previously 4).
- No fleet restart involved — the served code is unchanged; only client-side tooling moves.

OWED (Sam): open `fleet_status_undulator.bob` in Phoebus once to eyeball the not-deployed rows and the subtitle line.

- `addr_list` also accepts commas between entries.
- A listed address with **no cameras in the DB** is kept as a deployed row with an empty camera list and logged as a warning (a stale entry you can see), rather than silently dropped.
- Sam's own `~/.config/geecs_python_api/config.ini` got the `[pva]` section with the 9 deployed hosts during the live check (backup at `config.ini.bak-2026-09-04`) — needed for the generated screen to be true.
- **Not touched**: `deploy/bootstrap_host.sh` renders the *service host's* config.ini from `site.env` and does not know `[pva]`; the capture daemon's PVA discovery is unchanged. If the host should carry the list too, that is a one-line `site.env` key + render line — a follow-up, flagged here rather than bundled.

## Facility-literal check

New lab literals appear only in: the regenerated `fleet_status_undulator.bob` (a generated reference artifact, as agreed), the `getting_started.md` config example block, and DEPLOYMENT.md's labelled HTU reference numbers. The code paths read the DB and config.ini.

## Adversarial review

Posted as a PR comment below (either way).

https://claude.ai/code/session_014uheL9fSUG1CFmSrzsi7Ei


skbarber:

Ran from the worktree: `poetry -C GeecsPvaGateway run pytest` (24 passed), plus scripted probes of `fleet_roster` against the fake DB with hand-written `config.ini` variants.

### 1. HIGH (correctness + docs) — the PR's own `config.ini` example crashes the reader

`docs/tutorials/getting_started.md:113`, `GeecsPvaGateway/DEPLOYMENT.md:194`, and the `fleet.py` module docstring (`GeecsPvaGateway/geecs_pva_gateway/fleet.py:12`) all show the key with a trailing inline comment:

```ini
addr_list = 192.168.6.100 192.168.7.161 192.168.8.197   # the camera servers running an image gateway
```

`configparser.ConfigParser()` in `fleet.read_config` (`fleet.py:84`) is built with the default `inline_comment_prefixes=None`, so the comment is part of the value. `deployed_addr_list` (`fleet.py:110`) then yields `["192.168.6.100", …, "#", "the", "camera", "servers", …]`; each word becomes a "stale listed host" (`fleet.py:134-139`), and the final `sorted(hosts, key=_ip_key)` (`fleet.py:140`) compares `(192, 168, 6, 100)` against `("#",)` and raises `TypeError: '<' not supported between instances of 'int' and 'str'`. Reproduced: a user who copies the tutorial block verbatim gets a traceback from `deploy/gen_fleet_status.py` and `[DOWN] PVA fleet roster unreadable from the DB (TypeError: …)` from `scripts/fleet_status.sh` — attributed to the DB, which is not the cause. It is the only inline comment in the tutorial's example block (the other sections have none), so the doc change introduced it. Fix either side: drop the inline comments from the three examples, or construct the parser with `inline_comment_prefixes=("#", ";")` in `read_config` — and add a test with a commented value, since `test_fleet.py` never exercises one.

### 2. MEDIUM (correctness) — `_ip_key` raises on any non-dotted-quad entry the key is documented to accept

`fleet.py:59-61` (`_ip_key`) mixes `int` and `str` tuple members. The docstring and `docs/tutorials/getting_started.md:137` say the list is the clients' `EPICS_PVA_ADDR_LIST`, whose syntax allows hostnames and `host:port`. Reproduced: `addr_list = 192.168.6.100 camserver7` and `addr_list = 192.168.6.100:5076 192.168.7.161` both raise the same `TypeError` from the final sort (the `host:port` case also fails the membership test at `fleet.py:130`, so the real host is marked *not deployed* and a phantom `192.168.6.100:5076` row is appended). `test_addr_list_marks_not_deployed_and_flags_stale` only uses a well-formed `192.168.9.9`, so it cannot catch this. Fix: a homogeneous sort key (e.g. `(0, int(p))` for digit parts, `(1, p)` otherwise), and either strip a `:port` suffix before membership or state "dotted-quad IPs only" in the three docs and reject anything else with a clear error.

### 3. LOW (docs accuracy) — "who reads `[pva] addr_list`" names readers that do not exist

`docs/tutorials/getting_started.md:137` lists "cross-subnet PVA clients" as readers and says the key "doubles as `EPICS_PVA_ADDR_LIST`"; `docs/geecs_gateway/image_pvs.md:121` says the fleet list "has one home". Nothing in the repo applies `[pva] addr_list` to the environment (grep: the only reader is `fleet.py:108`), unlike `[epics] ca_addr_list`, which `geecs_bluesky/epics_env.py:69-73` applies at import. p4p clients and Phoebus still get the expansion by hand (`DEPLOYMENT.md:185-188` still says so). Either add the applier (an `epics_env`-style function in `geecs_pva_gateway` for PVA clients) or reword the two doc lines to "copy it into `EPICS_PVA_ADDR_LIST` / the Phoebus settings; only the fleet tooling reads it automatically".

### 4. LOW (redundancy / placement) — a fourth client-`config.ini` reader, and the instance-PV name composed in two places

- `fleet.py:37, 82-89, 92-100` (`USER_CONFIG_PATH`, `read_config`, `default_experiment`) duplicate `GeecsBluesky/geecs_bluesky/epics_env.py` (`_USER_CONFIG_PATH` + same parse), `GEECS-MCP/geecs_mcp/runtime.py:69-77` (`_read_experiment`), `GEECS-Data-Utils/geecs_data_utils/config_roots.py:36-56`, and the path literal at `GEECS-Core/geecs_core/db/geecs_db.py:44`. GeecsPvaGateway cannot import the bluesky/MCP ones (dependency graph), so the duplication is structural: the natural home for `USER_CONFIG_PATH` + "experiment from config" is GEECS-Core (which already knows the path and is imported by every one of those packages). Not a blocker for this PR, but the site-profile arc is the right time — also note `fleet.py` accepts `exp_name` as an alternate key while the Python readers (`GeecsPathsConfig`, MCP) accept only `expt`; the shell scripts accept both.
- `FleetHost.instance_pv` (`fleet.py:53-55`) re-composes `pv_name(experiment, "pvagateway", normalize_component(ip)) + ":name"`, which `server.py:270-274, 299-303` composes independently. `test_instance_pv_follows_naming_contract` pins only the fleet side; `test_server.py:312` pins only the server side. One helper (e.g. `instance_pv_prefix(experiment, host)` in `config.py`, used by both) would make drift impossible; the server's `socket.gethostname()` fallback (`server.py:274`) already means a listed host with no DB cameras — the case `fleet_roster` deliberately keeps — would never answer at the IP-token name the probe uses.

No other surviving findings. Checked and refuted: `image_variables` extraction (behaviour-preserving, `sorted` moved into the helper); scan-folder invariant (no path creation); public-repo hygiene (only `192.168.x.y` literals, in a generated artifact and labelled examples); stale references to `fleet_status.bob` (only historical CHANGELOG entries); `fleet_status.sh` `EXPERIMENT` resolution (`ini_get` strips whitespace, and the new skip message covers the empty case); `HEADER.format` placeholder safety (tests pass); the committed `.bob` matches the generator's current templates (11 rows, subtitle "11 … 9 deployed").

skbarber:
## Review disposition

| # | Finding | Disposition |
|---|---|---|
| 1 | HIGH — the PR's own config.ini example (inline `#` comment) crashed the reader | **Fixed** in e966a38f: `read_config` parses with `inline_comment_prefixes=("#", ";")`; the three examples drop the comment; pinned by `test_addr_list_tolerates_inline_comment_port_suffix_and_hostname`. Reviewer confirmed. |
| 2 | MEDIUM — `_ip_key` raised on a hostname / `host:port` entry | **Fixed** in e966a38f: homogeneous sort key; a `host:port` entry counts by its host. Same test covers both shapes. Reviewer confirmed (note: a hostname that resolves to a roster IP is still not matched — docs show IP entries only). |
| 3 | LOW — docs named readers of `[pva] addr_list` that do not exist | **Fixed** in e966a38f: getting_started.md, image_pvs.md, DEPLOYMENT.md and the module docstring say the value is copied into `EPICS_PVA_ADDR_LIST` / Phoebus by hand; only the fleet tooling reads the key. Reviewer confirmed. |
| 4a | LOW — a fourth client-config.ini reader | **Waived**: consolidating `USER_CONFIG_PATH` + experiment-from-config into GEECS-Core spans four packages and belongs to the site-profile arc as its own PR (recorded as a follow-up). Reviewer agrees with the disposition. |
| 4b | LOW — instance PV name composed in server.py and fleet.py independently | **Fixed** in e966a38f: `config.instance_pv_prefix` is the one composition, used by both; both existing tests now exercise it. Reviewer confirmed. |

Re-verification by the same reviewer: 25 tests pass, ruff clean, no new findings.

No open findings. Ready for the maintainer's merge once CI on e966a38f is green.

skbarber:
## Codex review disposition

| # | Finding | Disposition |
|---|---|---|
| P2 | `render()` interpolated DB/config strings unescaped into XML text, tooltips and comments (`Cam&A` → invalid .bob; `--` breaks the comment) | **Fixed** in the commit below: every interpolated string goes through `xml.sax.saxutils.escape`; the per-row comment carries only the host and a camera count, no names. |
| P3 | Widget suffixes assumed IPv4 (`split(".")[2:]`), so hostname entries rendered `host_` and collided | **Fixed**: suffixes are `normalize_component(host)`, the same component the instance PVs use. |

Pinned by `test_screen_is_well_formed_xml_with_hostile_names` (camera names with `&`, `<>`, `--`, two hostname entries, experiment `R&D`): the rendered screen parses with `xml.etree`, widget names are unique. `fleet_status_undulator.bob` regenerated and parses. `./scripts/check.sh`: lint clean, GeecsPvaGateway 26 passed.

PR #795, opened 2026-09-04
GeecsBluesky 0.75.0: execution-surface readiness — delete geecs_bluesky.analysis (#786), unserved-target move refusal (#772), worker_ready preflight + geecs-qserver-ready unit (#793)

skbarber:
One GeecsBluesky bump (0.74.0 → 0.75.0), four concern commits + two follow-up commits (review fixes, shared readiness verdict) — the "execution-surface readiness" half of the 2026-09-04 fix wave (PR A carries the tooling half: #753, #790, #793 part 3).

Closes #786 (item 1; item 2 — a declared-vs-imported dependency pin — is deferred, see the issue comment).
Part of #772 (part 2; part 1, editor save-time validation, is the console PR).
Part of #793 (parts 1 and 2; part 3, `/fleet-status` observing environment state, is PR A).

## Per-commit breakdown

| Commit | Concern | LOC |
|---|---|---|
| 9e9ef690 | **#786** — delete `geecs_bluesky.analysis` + the `analysis` extra + every doc reference to the derived-run-to-Tiled path (owner decision 2026-09-04: abandoned, not paused). `imageanalysis` stays optional under `optimize` only. Relocked (no `file://`/worktree paths). | +54 / −2831 (21 files) |
| aab53971 | **#772 part 2** — `GeecsSession.move_variable` runs `check_movable_served` (scan_request_runner) before any device is built: every `(device, variable)` written — each pseudo component — vetted against the gateway's served set (the scan path's provider). Unserved → `GeecsUnservedVariablesError` "… — move not started"; served set unknown → warn + proceed. | +228 / −4 |
| 1441b4d8 | **#793 part 2** — `worker_ready` preflight check beside `gateway_liveness`: manager answers, environment exists, `geecs_scan_request_plan` in `plans_allowed`. Closed env / missing plan → refusal naming the recovery gesture; unreachable / stub → skipped (fail-open). New import-light `geecs_bluesky.plan_names`; startup `__all__` built from it; `QueueClient.allowed_plan_names()` + `close()`. | +477 / −21 |
| 191cdd01 | **#793 part 1** — `geecs-qserver-ensure-ready` entry point (`qserver_ready.py`) + `geecs-qserver-ready.service` oneshot template (`Requires=`/`PartOf=geecs-qserver`), `render_units.sh` / `bootstrap_host.sh` render+enable both queueserver units, runbooks + fleet map. | +716 / −36 |
[...]

Issue #786, opened 2026-09-04
GeecsBluesky: the `analysis` extra has no in-package consumer that imports without `optimize` (base_evaluator hard-imports scan_analysis); consider a declared-vs-imported dependency pin


skbarber:
Surfaced by the #783 review (the #752 matplotlib fix) — two related follow-ups, both deliberately kept out of that metadata-only PR.

## 1. `analysis` extra vs. `optimization/`

`GeecsBluesky/pyproject.toml` declares `analysis = ["imageanalysis"]` separately from `optimize = ["xopt", "gest-api", "scananalysis", "imageanalysis"]`, but the only modules under `geecs_bluesky/` that import `image_analysis` are `optimization/base_evaluator.py` (which also hard-imports `scan_analysis.config.create_scan_analyzer` at module top level) and `optimization/config_models.py` (which hard-imports `xopt`). So in a fresh `poetry install --extras "ca tiled analysis"` env, `image_analysis.config.load_diagnostic` now imports fine (#752/#783), but `geecs_bluesky.optimization.base_evaluator` still fails — with `ModuleNotFoundError: No module named 'scan_analysis'`. Verified 2026-09-04 in a fresh env on the #783 branch.

Options, one of which should be chosen deliberately:
- make `analysis` imply `scananalysis` too (or fold it into `optimize` and drop the separate extra if nothing installs it alone — the worker runbooks install `ca tiled qserver` / `ca qserver tiled capture`, never `analysis`);
- or import `scan_analysis` / `xopt` lazily inside the evaluators so the `analysis` extra has a working in-package consumer.

## 2. Declared-vs-imported pin (optional)

#739 (pytest in main) and #752 (matplotlib in dev; nptdms/boto3 unused; h5py test-only) were both found by manual audit. A test that AST-walks `image_analysis/**/*.py`, collects third-party import roots, maps them via `importlib.metadata.packages_distributions()` and compares against `[tool.poetry.dependencies]` would have caught all of them in one shot. `GeecsBluesky/tests/test_dependency_direction.py` already has the AST-walk shape to adapt. Worth generalising to every package if it is built at all.

Waived on #783 as out of scope (Python change in a dependency-metadata PR).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

skbarber:
[...]

PR #799, opened 2026-09-05
GeecsBluesky 0.76.0: timezone defaults are the host zone, not America/Los_Angeles (site-profile Phase 4; stacked on #795)

skbarber:
> **Stacked on #795 — merge #795 first.** This PR's base is `bluesky/readiness-wave-786-772-793`. GitHub auto-closes a PR whose base branch is deleted, so if #795 is merged with `--delete-branch`, retarget this PR to `master` **before** deleting that branch (or merge without deleting; the PR retargets to master automatically on a plain merge). Diff vs #795: 8 files.

## What

Site-profile arc **Phase 4** (`Planning/site_profile/00_overview.md`, the timezone bullet + the default-experiment bullet): the last `America/Los_Angeles` literals in package defaults go.

- **GeecsBluesky 0.76.0** (0.75.0 is #795's) — the three `timezone="America/Los_Angeles"` defaults in `assets/tiled_readback.py` become `timezone: str | None = None`. `None` = the process's own zone rules: `TZ` from `site.env` for the systemd units (there since #777), the OS setting on a client. A new `_zone()` helper resolves the name; `_day_bounds` computes **both** bounds as naive local midnights passed through `astimezone()`, which applies the host's rules for that date — DST-correct, unlike `datetime.now().astimezone().tzinfo` (a fixed offset that would shift a January day by an hour when queried in July), and unlike `start + 1 day` (which kept the fall-back day's offset and lost that day's last local hour — a bug the review caught in the first push). An explicit IANA name still wins. The two `analysis/` copies of the default are deleted by #795 (#786), so they are not in this diff.
- **GEECS-Data-Utils 0.26.1** (docstring only) — `GeecsPathsConfig.__init__` said `default_experiment` defaults to `"Undulator"`. The code has read `[Experiment] expt` from config.ini (or left it None) since the experiment became optional; only the docstring lied. Patch bump per the docs-only convention.
- The plan bullet in `Planning/site_profile/00_overview.md` no longer prescribes `now().astimezone().tzinfo` (it described the approach this PR rejects).

## Why

Root CLAUDE.md § "Facility values have one home" lists these as the known residue scheduled for Phase 4. The service host already carries `TZ` in `site.env`, so the defaults were a second, code-baked home for the same value. **The residue paragraph in root CLAUDE.md is deliberately not edited here**: #797 edits the same paragraph, and the Phase 4 narrative purge drops the list entirely once items 1–3 land.

## LOC by concern (diff vs #795)

| Concern | Files | Δ |
|---|---|---|
| host-zone default + `_zone` helper + both-bounds fix | `tiled_readback.py` | ~+30 / −10 |
| tests | `test_tiled_asset_readback.py` (2 new, 4 lookups made zone-explicit) | ~+75 |
| docstring fix | `geecs_paths_config.py` | +2 / −1 |
| plan bullet | `00_overview.md` | +5 / −3 |
| versions + changelogs | 4 files | ~+25 |

## Tests

`./scripts/check.sh` on top of #795 (root env + GeecsBluesky env): lint clean; **GEECS-Data-Utils 325 passed, 3 skipped; GeecsBluesky 417 passed, 39 skipped** (2 new: `test_timezone_default_is_the_host_zone` runs one instant under `TZ=America/Los_Angeles` and `TZ=UTC` and expects different calendar days; `test_host_zone_day_bounds_follow_dst` pins −8 h in January vs −7 h in July, a 23:30 run on the fall-back day 2026-11-01 inside the bounds, and both DST-day stops equal to the true next local midnight). The readback suite also run under `TZ=Pacific/Honolulu`, `UTC`, `Asia/Tokyo`: 14 passed each. The four existing lookups that build Pacific timestamps now pass `timezone="America/Los_Angeles"` explicitly — under the old default they passed by coincidence of the runner's zone (Honolulu exposed the fourth).

## Hardware verification

Not applicable — no device or scan-execution path. Worker-side: the units run with `TZ=America/Los_Angeles` from `site.env`, so Tiled readback on the host resolves the same days as before; a client without `TZ` set uses its OS zone (previously always Pacific, regardless of where it ran). No service on the host calls these functions at runtime (the reviewer checked); the notebooks now follow the laptop's zone, which is the intended semantics.

## Facility-literal check

No new lab literals outside tests (which use `America/Los_Angeles`, `UTC`, `Pacific/Honolulu`, `Asia/Tokyo` as fixtures) and the CHANGELOG.

- `_day_bounds` for the host zone goes through `astimezone()` on naive datetimes (DST-correct via the OS) rather than a fixed offset. If a future caller needs a `ZoneInfo` for the host, that is a separate helper.
- The data-utils docstring fix rides along as the second Phase 4 bullet rather than its own PR (two lines).
- `_zone("")` means the host zone (it used to raise in `ZoneInfo("")`).

## Adversarial review

Posted as PR comments below: the report on the first push, then the disposition after re-verification.

https://claude.ai/code/session_014uheL9fSUG1CFmSrzsi7Ei


skbarber:
## Adversarial review (fresh-context subagent) — on the first push (d61e0c29, based on master)

### 1. HIGH (process) — Collides with open PR #795 on version, changelog, and two files #795 deletes

- `GeecsBluesky/pyproject.toml:4` — both #799 and #795 bump `0.74.0 → 0.75.0`.
- `GeecsBluesky/CHANGELOG.md:7` — both add a `## [0.75.0] - 2026-09-04` section at the same location.
- `GeecsBluesky/geecs_bluesky/analysis/camera.py:47` and `analysis/assets.py:47` — #799 edits two of the `analysis/` files that #795 deletes outright (−306 / −294 lines; the #786 decision "abandoned, not paused").
- `GeecsBluesky/tests/test_tiled_asset_readback.py:519-553` — #795 renames `test_find_geecs_run_ignores_derived_analysis_runs`; #799 adds a `timezone=` line inside the same test body.

Whichever merges second gets a modify/delete conflict on the two `analysis/` files, a same-hunk conflict on the changelog, and — worse — a *silent* version collision: two different 0.75.0s. The PR body cites the root CLAUDE.md residue paragraph and #797 as the coordination hazard but not #795, which is the actual one. The second-lander must rebump to 0.76.0 and drop the `analysis/` hunks (they are dead the moment #795 lands; the timezone change in `tiled_readback.py` alone is the entire runtime effect). Consider rebasing #799 onto #795's branch or landing #795 first.

### 2. MEDIUM (correctness) — Host-zone `_day_bounds` stop bound is one hour early on the fall-back day; a scan in the last local hour is unfindable

`GeecsBluesky/geecs_bluesky/assets/tiled_readback.py:648-653`: `start = local_midnight.astimezone()` yields a datetime carrying a **fixed-offset** `timezone` tzinfo (verified: `type(start.tzinfo).__name__ == 'timezone'`), so `start + timedelta(days=1)` keeps that day's offset instead of resolving the next local midnight. Probed under `TZ=America/Los_Angeles`:

| day | host-zone stop | true next midnight | Δ |
|---|---|---|---|
| 2026-03-08 (spring fwd) | `03-09T00:00-08:00` | `03-09T00:00-07:00` | +1 h (harmless: `_date_matches` rejects the extra hour) |
| 2026-11-01 (fall back) | `11-02T00:00-07:00` | `11-02T00:00-08:00` | **−1 h** |

Concrete failure: a run with `start.time` = 2026-11-01 23:30 Pacific, `find_geecs_run(..., day=1, timezone=None)` on the worker (`TZ=America/Los_Angeles` from `site.env`). `_search_catalog` (line 546-552) applies `Key("start.time") < stop.timestamp()`, which excludes the run before `_run_matches` ever sees it → `LookupError: No Tiled run found`. Probe: `hs <= inst < he` → `False` while `_date_matches(...)` → `True`. The explicit-name path is correct (`ZoneInfo` arithmetic is wall-clock), so this is a regression the PR introduces only for the new default. The new `test_host_zone_day_bounds_follow_dst` (test file line 600-611) asserts `start` offsets only and would not catch it. Fix: compute `stop` the same way as `start` — `datetime.combine(date + timedelta(days=1), time.min).astimezone()` — and pin the fall-back day in the test.

### 3. LOW (tests) — One date-filtered lookup is still zone-dependent; fails for any host west of UTC−7

`GeecsBluesky/tests/test_tiled_asset_readback.py:439-451` (`test_load_asset_from_tiled_finds_run_and_loads_text_array`) builds a start doc at Pacific midnight and calls `load_asset_from_tiled` without `timezone=`. The PR made three sibling lookups zone-explicit ("under the old default they passed by coincidence of the runner's zone") but missed this one. Verified: `TZ=Pacific/Honolulu poetry run pytest tests/test_tiled_asset_readback.py` → `1 failed, 13 passed` (`LookupError: No Tiled run found for experiment='Undulator', 2026-06-23 Scan042`); passes under unset/UTC/Asia/Tokyo because Pacific midnight is 07:00 UTC, which stays on the same calendar day for every zone from UTC−7 eastward. Not a CI risk (ubuntu = UTC) but exactly the class the PR claims to close. Add `timezone="America/Los_Angeles"` at line 444.

### 4. LOW (docs) — The plan bullet still specifies the approach the PR rejects

`Planning/site_profile/00_overview.md:252-255` says "the four `America/Los_Angeles` defaults … default to the host zone (`datetime.now().astimezone().tzinfo`)". The PR (a) fixes five, and (b) deliberately does *not* use `now().astimezone().tzinfo` (the fixed-offset trap called out in `tiled_readback.py:649-650`). The PR body explains the root CLAUDE.md paragraph is left for #797/the narrative purge, but says nothing about the plan bullet, which now describes a design this code contradicts. One-line edit, or an explicit note that the plan is frozen history.

---

**Checked and not surviving:** call sites — no package outside GeecsBluesky calls `find_geecs_run` / `load_asset_from_tiled` / `load_camera_image_from_tiled` / `run_tiled_*` (grepped Console, Portal, MCP, ScanAnalysis, docs, notebooks; only the two GeecsBluesky notebooks, which pass no `timezone` and so now follow the laptop's OS zone — the intended semantics). `TZ` delivery — all five units (`capture`, `qserver`, `data-portal`, `mcp`, `ca-gateway`) carry `EnvironmentFile=@SITE_ENV@` and `deploy/site_env_lib.sh:31` lists `TZ` in `SITE_RUNTIME_KEYS`; none of those services actually invoke these functions at runtime, so the worker-side claim is true but moot. Windows CI — `console-windows` runs only `GEECS-Console` + `GeecsBluesky/tests/qs_client`; the `tzset` fixture skips there anyway. Redundancy — no existing host-zone helper anywhere (`grep astimezone|ZoneInfo|tzlocal`); the rest of the repo already resolves local dates via naive `fromtimestamp` (`GEECS-DataPortal/geecs_portal/app.py:203`, `geecs_data_utils/tiled_catalog.py:524`), so the new default is consistent with existing practice. Placement — `_zone` belongs where its only consumer is. Data-utils docstring — matches `geecs_paths_config.py:95-97` (`config["Experiment"].get("expt")`, may stay `None`); `GEECS-Data-Utils/tests` 325 passed, 3 skipped. Public-repo hygiene — no new lab literals outside tests/CHANGELOG. `_date_matches` host path (`fromtimestamp(ts, None)`) is DST-correct per date; `_zone("")` now means host zone where `ZoneInfo("")` used to raise — noted, not a finding.

skbarber:
## Review disposition (re-verified by the same reviewer on 1ba33b44)

| # | Finding | Disposition |
|---|---|---|
| 1 | HIGH — collided with open #795 (same 0.75.0, same changelog hunk, two `analysis/` files #795 deletes) | **Fixed**: rebased onto #795's branch and retargeted (stacked, "merge #795 first"); the `analysis/` hunks are gone, GeecsBluesky is 0.76.0 above #795's 0.75.0. Reviewer confirmed; notes GitHub auto-retargets a stacked PR when the base branch is deleted through a merge, so the body's warning is conservative. |
| 2 | MEDIUM — host-zone `_day_bounds` stop was `start + 1 day`, an hour early on the fall-back day | **Fixed** in 1ba33b44: both bounds are naive local midnights through `astimezone()`. Reviewer probed both 2026 DST days (Δ = 0 h) and mutation-checked that the old implementation fails the strengthened test. |
| 3 | LOW — a fourth date-filtered lookup test was zone-dependent (failed under `TZ=Pacific/Honolulu`) | **Fixed** in 1ba33b44: passes `timezone="America/Los_Angeles"`. Suite green under unset/UTC/Honolulu/Tokyo. Reviewer confirmed. |
| 4 | LOW — plan bullet prescribed `now().astimezone().tzinfo` | **Fixed** in 1ba33b44: the bullet describes the naive-local-midnight approach and the fixed-offset caveat. Reviewer confirmed. |

No open findings. Ready for the maintainer's merge **after #795**, once CI on 1ba33b44 is green.

PR #802, opened 2026-09-06
Redis becomes a fleet fact: fleet-status reports its supervision, bootstrap installs it (site-profile Phase 4, last gap)

skbarber:
## What + why

Redis was the one service on the box that no part of our tooling knew about,
and that is how it went unnoticed for two weeks. The queueserver was running
against a **source-built Redis 8.10.1 in no unit**, started once by hand, in
nobody's dependency graph:

- `deploy/bootstrap_host.sh` never installed or enabled it, so a fresh box
  gets none.
- `scripts/fleet_status.sh` had **zero** references to it, so `/fleet-status`
  could not report it — even though `docs/platform/fleet_map.md` already
  states that "an inactive `redis-server` unit on a service host is a
  finding". The doc asserted a finding nothing checked.

What makes the gap self-concealing: `launch_re_manager.sh` starts its own
`${QS_REDIS_SERVER:-redis-server} --daemonize yes` whenever nothing answers
on 6379, and `geecs-qserver.service` merely orders `After=` the Redis unit
without requiring it. So a host that never got the package does not fail
loudly — it silently acquires an **unsupervised** Redis that dies with
whatever started it, is absent after a reboot, and comes back empty.

The host was migrated to the packaged Redis under its unit on 2026-09-06.
This PR closes the two code gaps that let it drift, plus the runbook trap
that migration hit. It is the last site-profile Phase 4 item.

## Per-concern LOC breakdown

| Concern | Files | +/- |
|---|---|---|
| fleet-status reports Redis supervision | `scripts/fleet_status.sh`, `scripts/fleet_table.py` | +43 |
| bootstrap installs/enables the package | `deploy/bootstrap_host.sh` | +30 |
| runbook: the RDB trap + launcher fallback | `GeecsBluesky/qserver/deploy/redis.conf-notes.md`, `CHANGELOG.md`, `pyproject.toml` | +71/-1 |
| fleet map: the finding is now detected | `docs/platform/fleet_map.md` | +5/-1 |
| tests | `tests/test_fleet_status_redis_sh.py`, `tests/test_bootstrap_host_sh.py`, `tests/test_fleet_table.py` | +310 |

## Supervision is judged by the unit, never by a pid

Worth flagging because the first attempt got it wrong and a live run caught
it. Comparing `redis-server.service`'s `MainPID` against the pid owning 6379
reports a **false absence**: `ss -p` reveals pids only for the ssh user's own
processes (or root), and the packaged Redis runs as the `redis` account. A
perfectly healthy supervised Redis showed as `[DOWN] nothing on 6379`. The
listener is now read without `-p` and supervision comes from the unit state,
both visible to any account. `tests/test_fleet_status_redis_sh.py::
test_packaged_redis_owned_by_another_account_is_not_a_false_absence` pins it.

The Redis row is emitted only where it is relevant — a queueserver unit on
the host, or something already on 6379 — so other fleet hosts stay quiet.

## The runbook trap

Carrying the spike's `dump.rdb` onto the package failed: Redis 8.10.1 writes
**RDB format version 15**, the jammy package is 6.0.16 and reads to version
9. It exits with `Can't handle RDB format version 15`, logged to
`/var/log/redis/redis-server.log` and **not** the journal, which shows only
`status=1/FAILURE` and a restart-throttle. The notes now send you to that
logfile first and say to start empty instead, spelling out what that
discards: `qs_default_plan_queue` (the **pending queue**) and
`qs_default_plan_history` hold data nothing recreates, while permissions come
back from `user_group_permissions.yaml` via the launcher's
`--user-group-permissions` on every start. (An earlier revision of this PR
claimed only history was at risk; adversarial review caught it, and the key
list was verified against the `bluesky_queueserver` installed on the host.)

## Tests

Run from this worktree:

- `tests/test_fleet_status_redis_sh.py` — **10 passed**. Extracts the remote
  ssh snippet and drives it against stubbed `systemctl`/`ss`/`redis-cli`:
  supervised-and-enabled, supervised-but-disabled, listening-without-the-unit
  (the two-week state), nothing-listening, no-Redis-row-on-an-unrelated-host,
  plus `bash -n` on the snippet and an assertion that it contains no bare
  quote (which would break `ssh bash -s`).
- `tests/test_bootstrap_host_sh.py` — **5 passed**. Stubs `systemctl` so both
  branches are deterministic: the root step appears when the unit is not
  enabled and precedes the `geecs-qserver` enable line, is absent when it is
  enabled, and is skipped under `--only portal`.
- `tests/test_fleet_table.py` — **7 passed** (1 new): `Redis` sorts below its
  consumers in `ROLE_ORDER`. The row's glyph and notes are asserted in the
  snippet tests against records the script really emits, so no fixture can
  fossilize apart from it.
- `tests/test_fleet_status_sh.py` — **3 passed** (new): script-wide snippet
  invariants (valid bash, no bare single quote, valid script).
- `./scripts/check.sh --lint` — **OK**, all hooks pass.
- `./scripts/check.sh` root suite — **64 passed, 1 skipped, 63 deselected**.
- `./scripts/check.sh GeecsBluesky` — **not runnable in this worktree**
  (`ModuleNotFoundError: No module named 'bluesky'`; the worktree has no
  GeecsBluesky env). The GeecsBluesky change is docs + version bump with no
  Python touched, so CI covers it.

## Live verification

`scripts/fleet_status.sh --summary` against the services box, after the
migration:

```
│ ✓ │ Redis        │ systemd    │ —                        │ 6.0.16         │ loopback 6379, enabled │
```

All other rows unchanged and green. The pre-fix run on the same host is the
false-absence bug described above, which is what prompted the pid rewrite.

No hardware verification section: this touches no scan execution or device
code.

## Adversarial review

Reviewed at commit eef3922c; seven findings, all fixed in ef8057c8 and
re-verified by the same reviewer. The report and disposition are in the PR
comments. Two were serious: `systemctl` prints a state *and* exits non-zero,
so `$(cmd || echo default)` captured both and split the one-line record in
two; and `fmt_host_records` never rendered `note=`/`info=`, so the default
full log showed a bare `[ OK ]` for the very state this PR exists to catch.
Each fix is mutation-tested — reintroducing the bug fails a named test.

### DIFF
--- a/.claude/skills/fleet-status/SKILL.md
+++ b/.claude/skills/fleet-status/SKILL.md
@@ -80,10 +80,17 @@ versions` line means a rollout is incomplete or a box's
 pull-on-restart no-oped — the GeecsPvaGateway runbook's known failure.
 
 **Stage 2 — host checkouts (ssh).** On each host derived from
-config.ini, services are discovered two ways and deduplicated by pid:
-every `geecs-*` (+ `tiled`) systemd unit in **both** system and user
+config.ini, services are discovered three ways. Two are deduplicated by
+pid: every `geecs-*` (+ `tiled`) systemd unit in **both** system and user
 scope, and **whoever owns each fleet port** — a queueserver started by
-hand in tmux has no unit and must still show up. Each process is mapped
+hand in tmux has no unit and must still show up. The third is **Redis**,
+deliberately *not* judged by pid: `ss -p` reveals a pid only for the ssh
+user's own processes (or root) and the packaged Redis runs as the `redis`
+account, so a pid comparison reports a healthy service as absent. Its
+listener is read without `-p`, its supervision comes from the unit state,
+and `redis-cli info server` supplies the version and the server's own pid.
+The `Redis` row appears only where it is relevant: a queueserver unit on
+that host, or something already on 6379. Each process is mapped
 to the clone it runs from (its cwd, or the baked venv's recorded source
 path for the MCP pattern), then: branch, short sha, commit date,
 staged/unstaged counts, pyproject version vs the version installed in
@@ -101,6 +108,10 @@ started (for a baked venv: after the install). The warnings to act on:
 | `Queueserver … NOT READY — worker environment CLOSED` | the unit is up but nothing opened the RE worker environment (every `geecs-qserver` restart recreates this until #793's readiness step lands); the console gets "Plan … is not in the list of allowed plans" | `qserver environment open` from any client env (GeecsBluesky's), then re-run; the plan name in the console's error is a red herring |
 | `Queueserver … readiness UNKNOWN (port only)` | no local env with `bluesky-queueserver` to ask the manager | `/env-doctor` for GeecsBluesky; the port is listening but that proves nothing about plans |
 | `Queueserver … readiness UNKNOWN — plans_allowed unanswered` | the manager answered `status` but not the plan list (a wedged or overloaded manager, or a refused user group) — readiness is an assertion about plans, so none is made | re-run; if it repeats, read the manager's journal on the host before touching the environment |
+| `Redis … answering on 6379 but redis-server.service is <state>` | the queueserver's state store is the **launcher's fallback**, not a supervised service: `launch_re_manager.sh` starts its own daemonized `redis-server` whenever nothing answers, and the unit only orders `After=` it. It dies with whatever started it, is gone after a reboot, and returns empty — the queue and history with it | install the distro package and enable the unit ([qserver runbook](https://github.com/GEECS-BELLA/GEECS-Plugins/blob/master/GeecsBluesky/qserver/deploy/redis.conf-notes.md)); note the pending queue and history do not survive, and **never** carry a `dump.rdb` from a newer Redis onto an older one |
+| `Redis … is disabled — no Redis after a reboot` | the unit runs now but is not enabled, so the next boot has none and the launcher's fallback takes over silently | `sudo systemctl enable redis-server.service` |
+| `Redis … active as pid N but the server on 6379 is pid M` | the unit is up yet something else holds the port; the queueserver talks to the one the unit does not supervise | read both before acting — stop the stray, then confirm the unit owns 6379 |
+| `Redis … nothing answering on 6379` | no state store at all; the next queueserver start manufactures an unsupervised one | install/enable the package **before** starting `geecs-qserver` |
 
 Any unit whose clone is on a branch other than `master` is a fact to
 surface, not a fault — feature-branch deploys for live checks are
--- a/GeecsBluesky/CHANGELOG.md
+++ b/GeecsBluesky/CHANGELOG.md
@@ -15,10 +15,11 @@ Format follows [Keep a Changelog](https://keepachangelog.com/en/1.1.0/).
   package is 6.0.16 (reads to version 9): given the newer dump it exits with
   `Can't handle RDB format version 15`, logged to
   `/var/log/redis/redis-server.log` and *not* the journal, which shows only
-  `status=1/FAILURE`. The notes now say to start the packaged Redis empty
-  (only `qs_default_plan_history` is unrecoverable; permissions come from
-  `user_group_permissions.yaml` via the launcher's
-  `--user-group-permissions` every start) and record why a missing package
+  `status=1/FAILURE`. The notes now say to start the packaged Redis empty and
+  spell out what that discards — `qs_default_plan_queue` (the pending queue)
+  and `qs_default_plan_history` hold data nothing recreates, while
+  permissions come back from `user_group_permissions.yaml` via the launcher's
+  `--user-group-permissions` every start — and record why a missing package
   does not fail loudly: `launch_re_manager.sh` starts an unsupervised
   `redis-server --daemonize yes` whenever nothing answers on 6379, and
   `geecs-qserver.service` only orders after `redis-server.service` without
--- a/GeecsBluesky/CHANGELOG.md
+++ b/GeecsBluesky/CHANGELOG.md
@@ -4,6 +4,27 @@ All notable changes to `geecs-bluesky` are documented here.
 
 Format follows [Keep a Changelog](https://keepachangelog.com/en/1.1.0/).
 
+## [0.76.1] - 2026-09-06
+
+### Documentation
+
+- **`qserver/deploy/redis.conf-notes.md`: never carry a `dump.rdb` across
+  Redis major versions, and the launcher starts its own Redis when none
+  answers.** Migrating the interim host off its source-built Redis failed
+  because the spike binary was 8.10.1 (RDB format version 15) and the jammy
+  package is 6.0.16 (reads to version 9): given the newer dump it exits with
+  `Can't handle RDB format version 15`, logged to
+  `/var/log/redis/redis-server.log` and *not* the journal, which shows only
+  `status=1/FAILURE`. The notes now say to start the packaged Redis empty
+  (only `qs_default_plan_history` is unrecoverable; permissions come from
+  `user_group_permissions.yaml` via the launcher's
+  `--user-group-permissions` every start) and record why a missing package
+  does not fail loudly: `launch_re_manager.sh` starts an unsupervised
+  `redis-server --daemonize yes` whenever nothing answers on 6379, and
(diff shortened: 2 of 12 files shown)


skbarber:
## Adversarial review (fresh-context reviewer, per `/land` step 7)

Reviewed at commit `eef3922c`. **Seven findings.** Redundancy: nothing
duplicated. Placement: no objection. Verbatim report:

---

### F1 — `|| echo` on `systemctl is-active` / `is-enabled` appends instead of substituting, so both finding branches emit a split, truncated record (reproduced)

`systemctl is-active` exits **3** while printing `inactive`, and `is-enabled` exits **1** while printing `disabled`. So `"$(systemctl … || echo inactive)"` captures *both* strings with a newline between them. (`scripts/fleet_status.sh:497-498`, `deploy/bootstrap_host.sh:95`.) Reproduced by running the extracted `REMOTE_SNIPPET` against stubs that use systemd's real exit codes — two lines, not one. The record protocol is one line per probe: `fleet_table.parse` drops line 2 (no `role=`), so the summary note is truncated mid-sentence and the remedy sentence is lost; in the log, `fmt_host_records` routes line 2 to `info "ssh: …"`, i.e. it is reported as a remote ssh notice. This hits **the two branches the PR exists for**. The same bug garbles the bootstrap warning across two lines. The new tests cannot catch it: every stub does `echo <state>` and exits 0, which is the one exit code systemd never returns for these states.

### F2 — the whole Redis payload lives in `note=`/`info=`, which the stage-2 log formatter does not render; in the default (non-`--summary`) run the unsupervised Redis prints `[ OK ]`

`fmt_host_records` parses a fixed key list — **no `note=` / `info=`**. Until this PR that was harmless: `note=` appeared only on stage-1 records, which their own probes print via `ok/warn/bad`. Fed the two finding records through the real function: the unsupervised state tags **`[ OK ]`** with only the generic hand-started warning (misleading — it was the qserver launcher, not tmux), and the no-state-store case prints a bare `[DOWN]` with no explanation and **no `warn`/`bad` at all**, so it does not appear in the attention lines the log-reading operator scans. The live verification in the PR body is `--summary`-only, which is why this went unseen.

### F3 — the runbook's "start empty" instruction omits the pending queue (`qs_default_plan_queue`), and contradicts `fleet_map.md`

Verified against `bluesky_queueserver/manager/plan_queue_ops.py:85-96`. Queue items are what the console's queue panel (#764) fills. `docs/platform/fleet_map.md:267-269` already says the fallback dies "taking the queue and history with it", so the notes contradict the map. Concrete scenario: an operator with a loaded queue reads "only history is unrecoverable", starts empty, and silently loses the queued scans. (The permissions claim does hold — `--user-group-permissions-reload` defaults to `ON_STARTUP`.)

### F4 — the new `test_fleet_table.py` cases pin record shapes the shipped script never emits (fossils of the abandoned pid approach)

Because `managed` differs, the fossil record renders a *different row* than the real one: `notes()`'s `UNMANAGED → "no systemd unit"` line never fires and `runs_as()` returns `"?"` instead of `"unmanaged"` — so the test would still pass if that mapping regressed for the real record. Related: the script constructs `absent/no-unit`, but `glyph()` treats a leading `absent` as the "not deployed here" `·` — a finding row would render as the benign dot; and `state=down`, which `test_absent_redis_is_down` pins, is a string the Redis path never produces.

### F5 — the notes' own verification snippet is not executable as written (and an exact ownership check was available)

As a non-root user `ss -p` omits the owner of another account's socket — precisely the lesson the PR encodes — so the step yields nothing to compare and reproduces the false-absence it warns about. Better: `redis-cli info server` already returns `process_id` to any client on loopback, so ownership *can* be judged exactly, closing the residual hole in the "supervised" verdict (packaged unit active but not the thing on 6379 → the row currently reads `managed=systemd`, clean). The code comment "Deliberately NOT judged by pid" overstates: it was `ss -p` that hid the pid, not Redis.

### F6 — `.claude/skills/fleet-status/SKILL.md` was not updated with the new row

The skill is the agent-facing contract for reading these rows: it states stage 2 discovers services "two ways and deduplicated by pid" (now three paths, the third deliberately *not* pid-based), and its "warnings to act on" table has no Redis entries. Root `CLAUDE.md`: prefer updating the skill over re-deriving its workflow.

### F7 — minor
- `test_bootstrap_host_sh.py`'s docstring claims determinism "on any host", but only `systemctl` is stubbed; the run still consults `GEECS_CHECKOUT_ROOT=/home/geecs`. On a host where that clone exists, `test_redis_root_step_precedes_enabling_the_queueserver` raises `ValueError`.
- The printed root step omits `apt-get update`, which the runbook's own block has.
- A host with something on 6379 but no GEECS units emits a Redis record *and* `nounits`.
- The snippet extractor and its `bash -n` / no-bare-quote assertions are script-wide invariants, not Redis-specific; a future second snippet test will want them from a neutral file.

skbarber:
## Review disposition

All findings dispositioned; the same reviewer confirmed every fix and has
closed the review. Two review rounds, no undispositioned finding.

### Round 1 — seven findings, all **fixed** in `ef8057c8` (+ `1ceac90b`)

| # | Finding | Disposition |
|---|---|---|
| F1 | `systemctl` prints the state *and* exits non-zero, so `$(cmd \|\| echo default)` captured both and split the one-line record in two | **fixed** — first line only, default when empty, inside `{ …; } \|\| true` |
| F2 | `fmt_host_records` never rendered `note=`/`info=`, so the default log showed a bare `[ OK ]` for the very state this PR catches | **fixed** — a `warn` per note, an `info` per info |
| F3 | "start empty" omitted the pending queue and contradicted the fleet map | **fixed** — `qs_default_plan_queue` named; verified against the installed `bluesky_queueserver` |
| F4 | table tests pinned record shapes the script never emits | **fixed** — assertions now run through `fleet_table.parse` on real records |
| F5 | the notes' verification snippet was not executable as a non-root user | **fixed** — `redis-cli info server`'s `process_id`; also closed the residual hole in code |
| F6 | the `/fleet-status` skill was not updated | **fixed** — three discovery paths + four warnings with remedies |
| F7 | five minor items (`apt-get update`, spurious `nounits`, `absent/` glyph collision, host-dependent test, misplaced invariants) | **fixed** — all five |

F1's first fix was itself wrong: a bare pipeline propagated the non-zero exit
under `set -e -o pipefail` and aborted `bootstrap_host.sh` with rc 127 on a
host with no systemd. `1ceac90b` also corrected a test whose premise assumed
`systemctl` could be removed from PATH — it lives in `/usr/bin` beside the
tools the script needs, which is why CI caught it and the local run did not.

### Round 2 — two new problems from those fixes, **fixed** in `8a010213`

| # | Finding | Disposition |
|---|---|---|
| 1 | the pid-mismatch branch put the unit state in `managed=`, where `runs_as` falls through to `?` | **fixed** — `managed=UNMANAGED`, pinned by asserting `runs_as` |
| 2 | the UNMANAGED-suppression guard was unpinned | **fixed** — pinned, plus an info-path test |

### Waived

| # | Item | Reason |
|---|---|---|
| 3 | restart race between the `redis-cli` and `MainPID` reads | waived on the reviewer's own recommendation: self-clearing on re-run, a guard costs more than the transient |
| 4 | `redis-server.service` is Debian/Ubuntu-shaped (RHEL uses `redis.service`) | waived as pre-existing and out of scope; the runbook scopes itself to Ubuntu 22.04. Recorded for the other-facilities workstream |
| 5 | the unsupervised row prints an `[ OK ]` header above its `[WARN]` | accepted: `state=running*` is the rule every hand-started-process row shares, and the payload now prints immediately below. Changing it would alter unrelated rows |
| 6 | "no systemd unit" reads oddly beside "the unit is active" in the mismatch row's Notes | accepted on the reviewer's advice: defensible (the row describes the server on 6379, which has no unit) and strictly better than the `?` it replaced; special-casing `notes()` is not worth it |

### Verification

Every fix is mutation-tested: reintroducing F1, F2, the `absent/` glyph
collision, the `managed=` field, the suppression guard or the `info=` case
each fails a named test and passes again on revert. The reviewer independently
re-ran its own reproductions and checked the `fmt_host_records` change against
non-Redis records (Data Portal drift rows render unchanged).

CI green on `8a010213`: `unit-tests`, `pre-commit`, `console-windows`.
Root suite 81 passed / 63 deselected. Live: `✓ Redis systemd 6.0.16
loopback 6379, enabled`.

PR #803, opened 2026-09-06
Analysis config v2 + web config editor, old implementations deleted: GEECS-Schemas 0.20.0, ImageAnalysis 2.2.0, ScanAnalysis 1.21.0, portal 0.21.2, Bluesky 0.76.3

skbarber:
## What & why

The analysis-config overhaul that a **web config editor** motivated, landed as one version so the pieces stay consistent (v2 code refuses v1 files and vice versa, so schema, consumers, converter and editor cannot ship separately). Already **deployed on the worker host (2026-09-06)** and used against real scans; this PR promotes that state to `master`.

**Size, honestly: +10,525 / −11,042 over 174 files** — net negative. The old implementations go with the new schema (owner rule, 2026-09-08): the Qt `ConfigFileGUI` (6,406 lines), the ImageAnalysis config re-export shims, the ScanAnalysis alias module and the one-shot v1 converter are **deleted**, so a newcomer finds one schema, one editor, one import path. Of the additions, 3,254 are generated schema artifacts (`docs/geecs_schemas/`, the repo's existing convention), 2,089 the hand-written models (replacing ~1,700 deleted model lines, now typed where dicts were), 1,578 tests, 1,430 the web editor.

**1. Schema v2 — the analysis-config documents live in GEECS-Schemas** (GEECS-Schemas 0.19.0 → 0.20.0; ~3,170 LOC incl. docgen artifacts under `docs/geecs_schemas/`)
`geecs_schemas.analysis`: `AnalysisDiagnostic` (`schema_version: 2`) is **one file per diagnostic** — `analyzer:` is a closed `kind`-discriminated union (15 kinds, typed per-analyzer parameters; the class path leaves YAML), `image:` is the typed camera / line processing config, `scan:` is typed in-document with typed `renderer:` options — plus `AnalysisGroup`. Closes the three untyped holes (`scan`, `image.analysis`, `image_analyzer.kwargs`) that made a form editor impossible and let `HTU/U_FROG_Beam.yaml` carry seven silently-ignored FROG keys under a BeamAnalyzer. Cross-section checks (image kind vs analyzer, renderer side, LineStitcher output label ≠ master folder) pinned by tests. JSON Schema + Markdown reference regenerated by docgen.

**2. No runtime lift — the corpus is canonical** (converter ~310 LOC + fixtures)
Per owner ruling, the configs repo holds 100% of analyzers, so the model **refuses v1**. The corpus was regenerated once, with a one-shot converter that has since been **deleted** (0.20.0) — the configs repo branch `analysis-config-v2` holds the result (50 diagnostics + 11 groups in canonical v2; 34 `UNCLASSIFIED` legacy flats deleted; the FROG keys dropped; one temporary v1 live-test file removed). The corpus walk test covers every namespace.

**3. Consumers** (ImageAnalysis 2.0.0 → 2.2.0, ScanAnalysis 1.19.0, MCP 0.8.7, Bluesky 0.76.3 (master's 0.76.1 Redis notes merged in; this branch's entries renumbered); ImageAnalysis −2,469/+870 LOC is the models moving out — **no re-export shims**: every consumer imports the models from `geecs_schemas.analysis`, the 1D ones under their `Line*` names)
ImageAnalysis reads the documents: a `kind → class path` registry (lazy import), every analyzer takes `spec=` (typed), `Data1DLoading ↔ Data1DConfig` mirror pinned. ScanAnalysis reads `diag.scan` typed; MCP/Bluesky follow the renamed types and relock. **LineStitcher** keeps an explicit `output_label` (the earlier review's blocker: dropping the v1 `name` kwarg would have overwritten raw HTT inputs) with load- and run-time guards.

**4. The web config editor** (ScanAnalysis 1.20.0 → 1.21.0, ~1,430 LOC + tests; portal 0.21.0 → 0.21.2, ~350 LOC)
Three layers so the portal is one host, not the only one (owner requirement):
- `scan_analysis.config_store.ConfigStore` — plain Python over a `scan_analysis_configs` tree: list (validity + summary), read + etag, validate (pydantic locations + canonical YAML), atomic optimistic save, delete, pending (git) changes, JSON Schema. **Writes only inside the configs tree** (pinned).
- `scan_analysis.config_editor` — mountable FastAPI router + a hand-written schema-driven form (`editor.js`, no build chain, no library — the portal's doctrine), `scan-config-editor` standalone script; `editor` extra.
- **GEECS-DataPortal `--config-editor`** mounts it at `/configs` and the Analysis tab's per-analyzer **edit** button opens it in a **drawer over the scan page** with a **preview of the unsaved document on the drawer's device + shot** through `image_analysis.ephemeral.render_document_ephemeral` (the analyzer's own figure; preview on demand, `auto` per edit) — dial in an ROI without saving per iteration or switching windows. Charter amendment: a second explicit-opt-in write verb, confined to the configs tree. Unit template carries `--config-editor`.

**5. Deleted** (commit 8f6e7e03): `ScanAnalysis/ConfigFileGUI` and its tests/screenshots/package entry (`LiveWatchGUI` stays); `image_analysis.config.{array2d_processing,array1d_processing,diagnostic}` and the model re-exports from `image_analysis.processing.*`; `scan_analysis.config.diagnostic_models` + the `ScanRuntimeConfig` / `AnalysisGroupConfig` / `DiagnosticAnalysisConfig` aliases; `geecs_schemas.convert.analysis_diagnostics` + fixtures. `image_analysis.config` and `scan_analysis.config` export only what they own (loader, factory, registry; loader, factory, `ResolvedDiagnosticConfig`).

**Judgment calls, cheap to veto**
- `U_FROG_Beam` on the configs branch keeps its `metadata` block but gained the median-3 filtering step an uncommitted GUI save on the share had added.
- Preview `auto` re-render defaults **off** (owner asked for a button to limit load); a first render on open stays.
- ImageAnalysis 2.1.1 adds `vmin`/`vmax` passthrough to the ephemeral renderers so the preview can honour a document's `scan.renderer` limits.

**Not in this PR (owed follow-ups)**
- **Docs notebooks still use the v1 constructors** (HTT-magspec, HasoLift, pulse_duration_jitter, basic_usage); the analysis tutorial's authoring steps and the overviews were updated for the web editor, the generated `docs/geecs_schemas` reference is current.
- **Mirror pairs** still exist (schema `RendererOptions` ↔ ScanAnalysis renderer configs; `Data1DLoading` ↔ `Data1DConfig`, pinned) and three walkers of the configs tree — the "one definition per concept" follow-up.
- Enabling a processing section (e.g. `roi`) without adding its pipeline step validates but does nothing — a hint or auto-append is a small editor follow-up.
- HASO (WaveKit) and Bluesky `optimize`-extra paths are code-complete, not exercised (Windows / heavy deps).

## Deploy coupling (read before merging)

The configs repo branch **`analysis-config-v2` must be merged and checked out on the share together with this code** — v1 code cannot read v2 files and vice versa. The worker host is already on both (portal checkout on this branch; share clone on the configs branch). Anything still on `master` code that reads `scan_analysis_configs` from the share — the Windows LiveWatch, ScanAnalysis task runners, the qserver's optimize evaluators — will refuse the v2 files until it moves.

Adversarial review: the first four commits were reviewed mid-arc (blocker fixed in eff86d68); a fresh-context review of the whole PR diff, a second of the deletion pass, and the owner's Codex review (3 findings, fixed in ca59c5db: invalid JSON in a mapping textarea now blocks Save; docs examples in the v2 shape; `upload_to_scanlog` spelling) are posted below with dispositions. `origin/master` merged in at dc0d8959.

## Checklist

- [x] `poetry version patch|minor` run for every package whose code changed
- [x] `CHANGELOG.md` entry added under the new version (Keep a Changelog)
- [x] Tests: exact counts reported below (not "tests pass")
- [x] Base branch is `master` — master merges are maintainer-only
- [x] Public-repo hygiene: no lab accounts/hostnames/home paths (diff grepped; the only lab word added is `analyzers/Undulator/` class paths in the kind registry — the per-experiment code mechanism)
- [ ] Adversarial review posted and every finding dispositioned

## Tests

After the review fixes and the deletion pass (commit 8f6e7e03):
- ImageAnalysis + ScanAnalysis + GEECS-Schemas: **916 passed**, 14 skipped (root env)
- GEECS-DataPortal: **247 passed** (new: `test_config_editor_mount.py`, 7)
- GeecsBluesky optimization: **132 passed** (env with the `optimize` extra)
- Earlier `./scripts/check.sh --all` (before the deletion pass): lint clean; GEECS-MCP 125, GEECS-Data-Utils 325, root 59 passed; GeecsCAGateway / GeecsPvaGateway / GEECS-Console / GEECS-Core / GEECS-LogTriage untouched by the diff and run by CI.

## Hardware verification

N/A for devices/scan execution — nothing here touches the gateway, the RunEngine or shot control. **Live-verified on the worker host 2026-09-06** (portal 0.21.x; the host is on the branch tip and restarts on the maintainer's go, share on the configs branch, 61 documents valid): drawer on a real 2026-09-04 stretcher scan (UC_Stretcher_MI), ROI dialled in with the preview re-rendering from real frames; save/delete paths exercised on the local demo only.

https://claude.ai/code/session_015FZduf9Kf1aAroAxby2xb7


skbarber:

Branch `feat/analysis-config-v2` (13 commits, `origin/master...HEAD`), reviewed read-only in the worktree. Effort concentrated on the converter, the runtime-lift removal, `config_store.py`, `config_editor/`, the ephemeral/rendering changes and the portal mount; the earlier consumer commits were sanity-checked. Suites run (via `GEECS-DataPortal/.venv`, which carries ScanAnalysis + ImageAnalysis in develop mode): GEECS-Schemas 466 passed incl. the corpus walk over the sibling `analysis-config-v2` checkout (50 diagnostics, 11 groups); ScanAnalysis config/store/editor 74 passed; portal editor-mount/analysis 47 passed; ImageAnalysis config/ephemeral 86 passed. The two findings tagged MAJOR were reproduced, not inferred.

### Findings

**1. [MAJOR] One malformed YAML file anywhere in the tree takes the whole editor down (500), including the portal drawer.**
`ScanAnalysis/scan_analysis/config_store.py:261` — `raw = self._read_raw(path)` sits *outside* the `try` at `:262`, so a `yaml.YAMLError` (tab-indented file) or the `DocumentInvalid` raised for a non-mapping document (`:305`) escapes `list()`. Reproduced: with `analyzers/HTU/Tabbed.yaml` containing a tab, `ConfigStore.list("analyzer")` raises `ScannerError`; `GET /api/list` → **500**, `GET /api/analyzers/Tabbed` → **500** (the router at `config_editor/app.py:133-142` catches only `NotFound`/`ConflictError`). The same 500 with a top-level YAML list. Every editor surface depends on `/api/list` (standalone sidebar, the drawer's `ceFillDiagnostics` in `run.html`), so the editor is unusable until someone fixes the file by hand — the case an editor exists for. This contradicts the inline comment at `:271` ("a broken YAML is an entry, not a crash") and the router docstring (`app.py:20-23`, "never a 500 for a config problem"). The tests don't catch it because `Broken.yaml` in `tests/test_config_store.py:33` is *valid YAML* that fails pydantic. Fix: move `_read_raw` inside the `try` and catch `yaml.YAMLError`/`DocumentInvalid` there; map `DocumentInvalid` in the `read` route to 422; add a tab-indented fixture to the store test.

**2. [MAJOR] Opening a file that does not validate on disk silently rebuilds it as a hollow document, reports "valid", and offers Save — a data-loss path for the stray v1 files this branch itself ran into.**
`static/editor.js`: `open()` (`:392-398`) passes the raw document to `buildForm`; the `object` renderer (`:229-259`) walks *schema* properties only, so unknown keys (`image_analyzer`, `image.analysis`, `pipeline: {steps: [...]}`) are dropped; the `union` renderer (`:283, :288`) maps a missing/unknown `kind` to `variants[0]` (`StandardAnalyzerSpec`, `analyzers.py:424`) and `get()` emits it (`:299`). `buildForm` then calls `validate()` (`:436`), which on success **clears the load-time error box** and writes "valid" (`:448`). Scenario: a v1 camera diagnostic (what the still-shipped Qt `ConfigFileGUI` writes — `ScanAnalysis/CLAUDE.md` says so — or the "stray v1 file" named in commit 28eab25e) is opened from the drawer → form shows `analyzer: {kind: standard}` with an empty pipeline → right pane says "valid" → Save rewrites the file with every analyzer parameter gone; on the share checkout those files are uncommitted (no git undo). The `.bad` class in the sidebar is the only hint. Fix: when `loaded.valid` is false, show the on-disk YAML + error and keep Save disabled until the user edits (or require an explicit "reconstruct from schema" click); never let a successful `validate()` erase a load-time error banner. Test: TestClient can't drive the JS, so pin it at the store/router level too — e.g. `PUT` of a document whose `etag` came from an invalid file requires a `force`-style acknowledgement — or add a minimal JS unit harness.

**3. [MINOR] Store discovery disagrees with both runtime loaders: a store-approved save can break `discover_analyzers` for every LiveWatch/task-queue load.**
`config_store.py:226` globs `*/*` (namespace depth only) while `ImageAnalysis/image_analysis/config/loader.py:296` and `ScanAnalysis/scan_analysis/config/analysis_group_loader.py:191` use `rglob`. Reproduced: `analyzers/Stray.yaml` → store `known_ids()` = `['Good','Tabbed']`, loader `list_diagnostics()` = `['Good','Stray','Tabbed']`. So `save("analyzer","HTU","Stray",…, etag=None)` passes the store's cross-namespace uniqueness check (`:376-385`) and then `discover_analyzers` raises "Duplicate diagnostic ID" for *every* group load. Group cross-checks (`:342-348`) likewise flag a valid nested ref as unknown. The corpus is flat today (verified: all 50 files at depth 2), so this is latent — but it is also the **third** walker of the same tree (`_discover_analyzers`, `discover_analyzers`, `_files`; the portal's `_processing_names` at `app.py:966-988` reuses the ImageAnalysis one). Redundancy + placement: give the tree one discovery helper (natural home: next to `_iter_yaml_files` in `analysis_group_loader.py`, or in ImageAnalysis' loader since ScanAnalysis already imports it) and have the store use it, with `rglob` semantics.

**4. [MINOR] Converter CLI default still skips `UNCLASSIFIED`, contradicting the corpus-walk change it shipped next to.**
`convert/analysis_diagnostics.py:244` `skip_namespaces=frozenset({"UNCLASSIFIED"})` remains the `regenerate_tree`/`main` default, while commit 28eab25e made the corpus test walk `UNCLASSIFIED` precisely because "a stray v1 file hid behind the skip". Two real diagnostics now live there (`HTT-C-ASSERTHighR.yaml`, `UC_Amp4_IR_input.yaml`, configs @a4f3e40). Scenario: the next stray v1 file lands in `UNCLASSIFIED`, `--dry-run` prints `skip`, `--write` leaves it v1, the loader refuses it at run time. Drop the default (keep the parameter) or add `--skip-namespace` to the CLI.

**5. [MINOR] "duplicate as" in the drawer copies `scan.device`, `output_name` and `output_label` verbatim, so the copy is not "a variant config for THIS camera" as the comment claims.**
`run.html` `ceDuplicate()` patches only `name` (`CE.duplicate(cur.namespace, id, { name: device })`). With `scan.device` set (2/50 corpus files) the new diagnostic still analyzes the *old* device's folder (`scan.device or name` is what `_DiagInfo.from_diagnostic` and the wrapper read); with `output_name` set (2/50) the copy writes into the original's output folder and s-file columns — two diagnostics in one group overwriting each other. Either clear `scan.device`/`output_name`/`analyzer.output_label` in the duplicate patch, or say in the tooltip that they are copied.

**6. [MINOR] Contract files contradict the shipped unit: docs say opt-in/off-by-default, the template turns the write verb on for every rendered site.**
`GEECS-DataPortal/DEPLOYMENT.md:178` ("Off by default. Adding `--config-editor` to the ExecStart…"), `GEECS-DataPortal/CLAUDE.md:14-15` ("explicit opt-ins"), `app.py` docstring ("explicit opt-in") vs `deploy/geecs-data-portal.service:47`, where `--config-editor` is baked into `ExecStart`. `render_units.sh` has no hole for it, so a read-only viewer site must hand-edit the rendered unit — the pattern the unit's own header forbids. The endpoint has no auth (`DELETE /configs/api/analyzers/{id}` from any lab-network browser deletes a config), so the default matters. Pick one: keep the flag out of the template and add a `${GEECS_PORTAL_EXTRA_ARGS}`-style site.env hole, or rewrite the three docs to say the template enables it.

**7. [NIT] Stale editor JS after a ScanAnalysis-only change.**
`run.html` cache-busts `/configs/static/editor.js?v={{ portal_version }}`, but the asset ships with ScanAnalysis and the router's `FileResponse` (`app.py:105-109`) sets no `Cache-Control`. An editor.js fix without a portal version bump serves the old file from browser cache. Add `Cache-Control: no-cache` on the static route (cheap, the file is 34 KB) or version by ScanAnalysis' version.

**8. [NIT] `pending_changes()` shells out to `git status` on every `/api/list`.**
`config_store.py:431-444`, 5 s timeout, on a share checkout — and memory records that checkout as CRLF-normalised, where `git status` reports every file modified, so the badge would read "50 uncommitted changes" permanently. Consider caching for a few seconds or hiding the badge when the count equals the file count.

**9. [NIT] Converter: two small contradictions with its own "fails loudly" docstring.**
`_image()` at `analysis_diagnostics.py:141-142` replaces `analysis` with `analysis["magspec"]`, silently dropping any sibling keys (no historical corpus instance — checked `8cbbc21^`), and a non-mapping `analysis:` value reaches `dict(analysis)` as a `TypeError`, not `SchemaConversionError`. One-shot tool, so low stakes; noting for the record. Also `regenerate_tree.emit` re-implements `config_store.dump_yaml` ("the one YAML serialisation every writer uses") — unavoidable given GEECS-Schemas' no-yaml rule, but the store docstring overclaims.

### Checked and refuted

- **Scan-folder invariant** — the store creates at most the namespace folder (`config_store.py:398`, `mkdir(exist_ok=True)`, no `parents`); the converter's `mkdir(parents=True)` (`analysis_diagnostics.py:261`) is on the configs tree, never `scans/`; the preview goes through `render_document_ephemeral` with `_refuse_file_path` (`ephemeral.py:257-258`); pinned by `GEECS-DataPortal/tests/test_config_editor_mount.py::test_nothing_touches_the_scan_folder`.
- **Public-repo hygiene** — grep of added lines for `/Users/`, `/home/`, `.lbl.gov`, hostnames: clean (the `/Volumes/hdna2` literals in `density_from_phase_analysis.py` pre-date the branch).
- **Runtime v1 lift removal** — `AnalysisDiagnostic._refuse_v1_layout` (`diagnostic.py:110-146`) refuses `image_analyzer` and `schema_version: 1`; no lift code survives in ImageAnalysis/ScanAnalysis/portal/Bluesky/MCP (grep); `loader.py:51` keeps `image_analyzer` only as a marker so a stale file gets the converter hint. Corpus walk over the sibling checkout passes with `KNOWN_INVALID = {}`.
- **Converter mapping** — class-path table covers every `ANALYZER_SPECS` kind (test `test_every_v1_class_path_maps_to_a_registered_kind`); HASO `mask_*`→`mask`, LineStitcher `name`→`output_label` and the master-folder guard, `renderer_kwargs`→`renderer`, 1D background renames, `enabled` stripping (no v2 processing section has an `enabled` field — grep of `geecs_schemas/analysis/*.py` finds it only on `AnalyzerRef`), input not mutated.
- **`canonical_document`** — `exclude_unset` + default-`None` pruning + `schema_version` first; the JS form's default-dropping (`editor.js:245-251`) is consistent with it.
- **Ephemeral denylist by kind** — `{"haso","frog_retrieval"}` matches the registry's classes for the two former class paths (`registry.py:31,35`); `_refuse_file_path` applied on both the name and document paths.
- **`render_result_figure` vmin/vmax** — applied after `window_limits`, so explicit limits win (`rendering.py:432-436`); `RendererOptions` carries `cmap`/`vmin`/`vmax` for both kinds (`renderer.py:39-45`); 1D results ignore them.
- **Portal preview ladder** — `HTTPException` 404→`LookupError`→404, else `ValueError`→400; blocking analyzer work runs in a sync route (threadpool), same as the Images tab; `config_editor_enabled` is assigned before any request is served (closure late-binding is fine).
- **Router** — every URL mount-relative; `/api/schema/{kind}` and `/api/list` are registered before the `/api/{plural}/{doc_id}` catch-all; `read_only` guards PUT/DELETE; `_check_name` regex rejects `..`/`/` (tested).
- **Etag/atomic write** — temp+`os.replace` in the target directory; `.yml` targets handled; create refuses existing; stale etag on save and delete. (The check-then-write window between two saves carrying the same etag is theoretical; not raised.)
- **Dependency graph / placement** — `config_store.py` is plain Python below the web layer (correct); `config_editor` behind the `editor` extra with `scan_analysis/__init__.py` importing neither; portal guards the import (`app.py:1868-1873`); portal's own fastapi/jinja2/uvicorn satisfy the extra as `DEPLOYMENT.md` says; root `CLAUDE.md` graph updated for ImageAnalysis→Schemas and ScanAnalysis→Schemas (+editor extra).
- **Consumer test rewrites** (portal, MCP, Bluesky) are honest v1→v2 translations; the MCP `windows_only` case now uses the real `haso` kind whose module hard-imports WaveKit — a stronger test than the fake class path.
- **Versions/CHANGELOGs** — pyproject and top CHANGELOG heading agree for all six packages; `docs/geecs_schemas/*` freshness pinned by `tests/test_schema_reference.py` / `test_schema_export.py` (passing).
- **Environment note (not a finding)** — the worktree's `ScanAnalysis/.venv` and `ImageAnalysis/.venv` lack `pyyaml`/`pytest`; `/check` from those packages will need `/env-doctor` first.

_Disposition of each finding follows in a separate comment._

skbarber:

Fixes in **d0601e5d** (findings 1–7) and the follow-up residual commit; each fix went back to the same fresh-context reviewer, who confirmed it (its verification report is summarised per item). Versions: ScanAnalysis 1.20.2, GEECS-DataPortal 0.21.2, GEECS-Schemas 0.19.1.

| # | Finding | Disposition |
|---|---|---|
| 1 | Malformed YAML 500s the editor | **Fixed** — `ConfigStore.list` reads inside the `try`; `read()` returns an invalid `Loaded` (empty document, parse error, file text) for `YAMLError` / non-mapping / non-UTF-8 files; router 200s. Store + router tests pin tab-indented, list-shaped and Latin-1 files. Reviewer: confirmed (the non-UTF-8 residual it raised is in the follow-up commit). |
| 2 | Invalid-on-disk file reconstructed as "valid", Save armed | **Fixed** — `editor.js` keeps Save disabled while the loaded file does not validate, shows the on-disk YAML + error instead of "valid", skips the preview; the first edit arms Save under a "will REPLACE an invalid file" banner; save clears it. No JS harness — pinned at the store/router level. Reviewer: confirmed for the data-loss path; its display residual (stale YAML pane in the error branch) is in the follow-up commit. |
| 3 | Store walk (`*/*`) vs loaders' `rglob` | **Fixed** — `_files` is `rglob`; namespace = folder relative to the kind root everywhere `parent.name` was; nested files are listed, id-checked and readable. `namespaces()` offers only top-level folders as save targets (the reviewer's residual: nested paths were offered but unsaveable). **Waived (deferred):** unifying the three tree walkers into one helper — a cross-package refactor (ImageAnalysis loader ↔ ScanAnalysis), tracked as a follow-up with the mirror-pair cleanup. |
| 4 | Converter skips `UNCLASSIFIED` by default | **Fixed** — default `frozenset()`; dry-run over the corpus: 61 files, 0 skipped. Reviewer: confirmed. |
| 5 | "duplicate as" copies identity fields | **Fixed** — `duplicate()` deep-copies and drops `output_name`, `scan.device`, `analyzer.output_label`; tooltip says so. Reviewer: confirmed. |
| 6 | Unit template bakes `--config-editor`; docs say opt-in | **Fixed** — `ExecStart … $GEECS_PORTAL_EXTRA_ARGS` from `site.env`; `site.env.example` sets `--config-editor` (reference deployment); `DEPLOYMENT.md`, the unit comment and `docs/platform/site_profile.md` agree; an old `site.env` without the key renders a read-only portal. Reviewer: confirmed. **Host note:** the worker's rendered unit still has the flag baked; add `GEECS_PORTAL_EXTRA_ARGS=--config-editor` to its `/etc/geecs/site.env` before the next unit render (sudo). |
| 7 | Stale `editor.js` from browser cache | **Fixed** — `/static/*` served `Cache-Control: no-cache`, pinned. Reviewer: confirmed. |
| 8 | `git status` on every listing; CRLF badge | **Waived** — verified `git status --short -- scan_analysis_configs` is clean on the share checkout (autocrlf=true, 61 files); a millisecond subprocess per listing. Reviewer: accepted. |
| 9 | Converter edge cases / `dump_yaml` docstring overclaim | **Waived** — one-shot tool whose corpus run is complete and re-verified (0 failed); slated for deletion in the cleanup pass. Reviewer: accepted. |

No undispositioned findings remain.

skbarber:
## Adversarial review — deletion pass (fresh-context subagent, commit 8f6e7e03)

Read-only sweep of `git diff a6154f7f..8f6e7e03` against the whole worktree (`.py`, `.ipynb`, `.md`, `.yml`, `.toml`, `.sh`, `.html`, `.js`; `extras/`, `.claude/`, venvs excluded), plus `mkdocs.yml`, the package `pyproject`s and LiveWatchGUI.

### Findings

**1. [MAJOR] Three docs notebooks still import the deleted `image_analysis.config` re-exports, and one was rewritten to names that do not exist in `geecs_schemas.analysis.processing_1d`.**
- `docs/image_analysis/examples/basic_usage_image_analyzer.ipynb:400` `from image_analysis.config import CircularMaskConfig`; `:449` `TransformConfig`; `:502` `BackgroundConfig, CameraConfig, ROIConfig`; `:568-572` (the "Mode 1 building blocks" prose block) lists nine model names from `image_analysis.config`. All nine left `config/__init__.py` in this commit → `ImportError` at the cell.
- `docs/image_analysis/examples/HTT_magspec_analysis_tests.ipynb:80` `from image_analysis.config import BackgroundConfig` — same.
- `docs/image_analysis/examples/pulse_duration_jitter_analysis.ipynb:40-45` was changed *by this commit* to `from geecs_schemas.analysis.processing_1d import (Data1DConfig, Line1DConfig, ROI1DConfig, PipelineConfig)`. That module has `Data1DLoading`, `Line1DConfig`, `LineROIConfig`; `Data1DConfig` and `ROI1DConfig` existed only in the deleted shim, so the rewrite newly breaks two names (`PipelineConfig` never existed — pre-existing). `:136` `roi=ROI1DConfig(...)` follows.
`mkdocs.yml:165-166` has `execute: false`, so the site builds; the failure is a reader running the example. Fix: `from geecs_schemas.analysis import …` in the first two; `Data1DLoading, Line1DConfig, LineROIConfig` (drop `PipelineConfig`) in the third. (`basic_usage_image_analyzer.ipynb:503` also passes `CameraConfig(name=…)`, a field removed in #412 — pre-existing, noting only because the cell is being touched anyway.)

**2. [MINOR] Two overview pages still describe the deleted modules and the v1 shape.**
`docs/image_analysis/overview.md:100-102` — the package tree lists `config/array2d_processing.py`, `array1d_processing.py`, `diagnostic.py` (all deleted here). `docs/scan_analysis/overview.md:72` — "bundling … a `name` + `image_analyzer` class path" is the v1 layout the loader refuses; `:100-101` "the\nthe **config editor**" typo introduced by this commit.

**3. [NIT] Stale module docstrings that now name deleted things.**
`GEECS-Schemas/geecs_schemas/convert/__init__.py:4-6` still documents `analysis_diagnostics`. `geecs_schemas/analysis/canonical.py:3` lists "the converter" among the writers. `processing_1d.py:10-11` and `processing_2d.py:12` assert the shims this commit deleted.

### Checked and refuted

- **Surviving importers, code:** repo-wide grep for `ConfigFileGUI`, `scan_config_io`, `config_io`, `image_analysis.config.{array1d,array2d}_processing|diagnostic`, `diagnostic_models`, `ScanRuntimeConfig`, `AnalysisGroupConfig`, `DiagnosticAnalysisConfig`, `convert.analysis_diagnostics`, `convert_v1_diagnostic`, `V1_CLASS_PATH_TO_KIND`, `regenerate_tree`, model imports from `image_analysis.processing.array1d|array2d` — no hits in any package, root `tests/`, `scripts/`, `.github`, `pyproject` scripts. Only CHANGELOGs, `REFACTOR_RECON.md` and `Planning/data_portal/04_analysis_run_design.md` mention them, as history. Every `from image_analysis.config import …` / `from scan_analysis.config import …` in GeecsBluesky, GEECS-MCP, GEECS-DataPortal, GEECS-Console names only symbols still exported.
- **LiveWatchGUI:** imports nothing from ConfigFileGUI; `ScanAnalysis/pyproject.toml` drops only the `ConfigFileGUI` include; PyQt5 (win32) and `LiveWatchGUI` stay.
- **mkdocs:** nav has no ConfigFileGUI page; no `strict:`; no reference to the deleted `configgui_*.png`; the remaining tutorial asset exists.
- **1D rename meaning:** every `processing/array1d/*.py` and every 1D analyzer imports `Line*`; no file using the 1D pipeline also imports a 2D `BackgroundConfig|ThresholdingConfig|FilteringConfig|ROIConfig`; `single_device_scan_analyzer.py:671`'s `processing_2d.BackgroundMethod` is the camera background it rewrites — correct.
- **Converter-only implementations:** canonical writer survives (`canonical_document` + `config_store.dump_yaml`); kind table is `image_analysis.config.registry.ANALYZER_CLASS_PATHS` (pinned against `ANALYZER_SPECS`); `V1_CLASS_PATH_TO_KIND` had no other importer; `SchemaConversionError`/`load_legacy` remain for the other converters; `schema_reference.md` regenerated.
- **Moves:** `ResolvedDiagnosticConfig` lives with its producer and is still exported from `scan_analysis.config`; no import cycle; `to_data1d_config` in `data_1d_utils.py` removes the former config↔data-utils edge.
- **Docs links:** no dangling `assets/` references; cross-page anchors unchanged.

### Disposition

| # | Disposition |
|---|---|
| 1 | **Fixed** (7e1a2145 + 9fc5df76) — the three notebooks import the models from `geecs_schemas.analysis` (1D: `Data1DLoading`, `Line1DConfig`, `LineROIConfig`; `pipeline=["roi"]`). The reviewer's re-check found the package root exported only `CameraConfig`, so 9fc5df76 exports every processing sub-model from `geecs_schemas.analysis` (the documented import style; 64 names) — verified at runtime that every notebook import resolves. The v1-era kwargs those notebooks still pass (`name=`, `analysis={}`, `CameraConfig(name=…)`) are pre-existing and tracked as the notebooks follow-up in the PR body. |
| 2 | **Fixed** (7e1a2145) — both overview pages. |
| 3 | **Fixed** (7e1a2145) — the four docstrings. |

Reviewer re-check (same subagent) at 7e1a2145: (2) and (3) confirmed fixed, (1) residual — the package root exported only `CameraConfig`. At 9fc5df76: (1) **confirmed fixed**; a runtime sweep of every `from geecs_schemas.analysis import …` across `docs/**/*.ipynb` finds no missing name. No residuals.

skbarber:
## Disposition of the owner's Codex review (head 9fc5df76)

| # | Finding | Disposition |
|---|---|---|
| P2 | `editor.js` treats invalid JSON in a free-mapping textarea as `undefined`, so Save validates and writes the document with the field omitted — a malformed edit to `metadata` / `auxiliary_columns` silently drops the mapping | **Fixed** (ca59c5db) — the textarea's getter throws `FormParseError`; `currentDoc()` catches it, shows the message and returns `null`; `validate()`, `save()`, the preview button, `refreshPreview()` and `duplicate()` all stop on `null`. Empty textarea → unset is unchanged. Verifier (fresh-context subagent): confirmed — the only remaining `state.get()` call is inside `currentDoc()`, all five former callers go through it, nothing but `FormParseError` is swallowed. |
| P2 | Published Image Analysis docs still teach the removed v1 shape (`image_analyzer:`, `pipeline.steps`; HASO kwargs via `image_analyzer:`) | **Fixed** (ca59c5db + the follow-up docs commit) — `overview.md` "How a diagnostic is described" and "Pipeline is the source of truth", and `analyzer_index.md` "What each analyzer needs", show `schema_version: 2`, `analyzer: {kind}` and the flat `pipeline` list. Verifier: confirmed after the follow-up; no `image_analyzer:` / `analysis:` / `pipeline.steps` residue in `docs/**/*.md` (notebooks remain the tracked follow-up). |
| P2 | Group examples use `upload_to_scan-log`; `AnalysisGroup` accepts only `upload_to_scanlog` (`extra="forbid"`) | **Fixed** (ca59c5db) — tutorial and overview. Verifier: confirmed, 10 occurrences across docs/schemas/ScanAnalysis all spelled `upload_to_scanlog`, no variant survives. |
| — | PR marked CONFLICTING against master | **Fixed** (dc0d8959) — `origin/master` merged; the one conflict was GeecsBluesky's version (master shipped 0.76.1 with the Redis notes): master's entry kept, this branch's renumbered 0.76.2 / 0.76.3. Post-merge: ImageAnalysis + ScanAnalysis + Schemas 916 passed, GeecsBluesky 558 passed, portal 247 passed. |

Issue #807, opened 2026-09-09
GeecsBluesky as a thin layer on stock Bluesky: device namespace, stock plans + GEECS per-step, md-keyed preamble preprocessor, native pause; retire the funnel and named plans


skbarber:
## Decision

GeecsBluesky becomes a **thin facility layer on stock Bluesky** instead of a custom plan framework. Devices are long-lived nouns in the worker namespace, plans are the stock `bluesky.plans` verbs, and everything GEECS-specific lives in exactly four native extension points: a **per-step function**, a **RunEngine preprocessor**, the **`Pausable` device protocol**, and **run metadata**. The `ScanRequest` document stops being a plan argument and becomes the client-side template + provenance record that expands into a stock plan call. The funnel (`geecs_scan_request_plan`) and the three named plans (0.73.0) are retired together once the stock verbs run on hardware.

Agreed with @skbarber 2026-09-08. Guiding rule from Sam, to be applied at every step: **wherever we are tempted to patch with custom code and a native mechanism plausibly exists, stop and check the native one first**; and the result must not close off adding further stock plans or Bluesky features later.

Companion decision: #806 (image writer as an areaDetector-shaped gateway plugin + `StandardDetector`) — **separate track**, not in this branch, except that the device namespace built here is what its detectors will drop into.

## Why (short)

The current `geecs_step_scan` / `geecs_free_run_step_scan` hand-roll the step loop (~730 lines) that `bp.list_grid_scan` / `bp.count` / `bp.fly` already provide, including move-only-changed-axes, checkpoints, rewind, standard run md and hints. The pause module re-derives what `RE.request_pause` + `Pausable` give for free. The genuinely GEECS parts — preamble (validate/resolve/connect/claim/ScanInfo), strict single-shot acquisition (arm waiters → fire → wait → read, with bounded refire), compiled actions, save windowing, finalize chain, telemetry — all fit the stock hooks unchanged. Same pattern as #806: the native shape was known and deferred; this is the correction.

## Target architecture

**Devices: nouns in the namespace, lazily connected.** At `environment open` the startup profile builds one ophyd-async object per enabled GEECS device for the experiment from the DB — unconnected, so thousands cost nothing. A preprocessor connects a device on its first `stage` (and may drop/keep-warm on `unstage`). Queue items name devices by string as they name plans; `allowed_devices` regexes gate them. Adding a device to the DB = environment reload.

[...]

PR #808, opened 2026-09-09
GeecsBluesky 0.77.0: device namespace composed from the ca layer + connect-on-demand; type rule to GEECS-Core (#807 phase 1)

skbarber:
Phase 1 of #807 (plan of record: GeecsBluesky as a thin layer on stock Bluesky). Base: `feature/native-bluesky-plans` (integration branch for the phase PRs, per Sam; the feature→master PR is the maintainer merge).

## What

Devices become **long-lived nouns in the queue-server namespace**, connected on first use, so stock `bluesky.plans` verbs can be given devices by name (`count([UC_Amp4_IR_input])`, `scan([...], U_S1H.Current, -1, 1, 5)`). Additive — the per-scan construction path and every existing plan are untouched.

The namespace **owns no device behaviour**: it composes the existing `devices/ca` layer (`Planning/native_bluesky/01a_device_layer_audit.md` — the inventory that says the layer is right and should be kept):

- acquirer → `CaGenericDetector`; everything else → `CaSnapshotReadable`; each served settable attached as a `CaMotor` (DB tolerance) / `CaSettable` child.
- which children exist → the gateway's served set (`db_runtime.GeecsDbServedSetProvider`); what `read()` returns → the DB subscribed list (`db_runtime.GeecsDbScalarPolicy`); every variable's type → `geecs_core.db.variable_types.effective_vartype`.
- triggerable classification (`looks_triggerable`): trigger-named devicetype variables minus trigger-source devicetypes; a DB `acq_timestamp` row / `DeviceRoster.triggered` override. Verified against all 105 Undulator devices.
- `connect_on_demand` RunEngine preprocessor (message-level `ensure_connected`; `declare_stream` args included; installed outermost).
- startup profile exports the namespace (`QS_DEVICE_NAMESPACE=off` skips it); operators get `allowed_devices: ":?.*:depth=3"`.

## Why the cross-package changes

- **GEECS-Core 0.5.0** — `effective_vartype` + tables move **unchanged** from `geecs_ca_gateway.config` into `geecs_core.db.variable_types`: one home for the CA gateway, the PVA gateway (which imported a gateway's config module) and GeecsBluesky, so a client's declared type always matches the served PV. No behaviour change. **DB sweep item (Sam, 2026-09-09):** 18 Undulator rows carry `variabletype='numeric'` with a filter-wheel option list and should be `choice` — fixing the DB flips gateway and clients together (changing the rule alone would break today's `float`-declaring scan path; review #2 P1).
- **GeecsCAGateway 0.20.3** — imports the rule; re-exports under the old private names; no behaviour change. PV_CONTRACT §4 names the rule's home and records the DB defect.
- **GeecsPvaGateway 0.6.1** — imports the rule from GEECS-Core; the dead `geecs-ca-gateway` path dependency is dropped (relocked); root `CLAUDE.md` graph updated.
- **GeecsBluesky 0.77.0** — namespace, preprocessor, `datatypes=` on the readable constructors, startup export, permissions, hardware test. Event keys follow `EVENT_SCHEMA.md` (`safe_name`); the namespace binding keeps GEECS spelling (`U_S1H.current`).

## Per-concern LOC (approx.)

| concern | added | removed |
|---|---|---|
| namespace + preprocessor (GeecsBluesky) | ~330 code | — |
| readable constructors `datatypes=` | ~40 | — |
| variable-type rule → GEECS-Core | ~90 (moved) + tests | ~50 (gateway copy) |
| startup / permissions | ~25 | — |
| tests (unit) | ~420 | — |
| Planning notes | 3 files | — |

## Reuse ledger

| new symbol | reuses | replaces | new because |
|---|---|---|---|
| `namespace.GeecsNamespace` | `CaGenericDetector`, `CaSnapshotReadable`, `CaMotor`, `CaSettable`, the three `db_runtime` providers, `effective_vartype`, `safe_name` | (phase 3) per-scan device factories in `session.py`, `_DeferredConnectFactories`, `_build_request_detectors` | devices as nouns did not exist |
| `namespace.looks_triggerable` | — | — | the DB cannot answer it yet (LabVIEW-internal `acq_timestamp`) |
| `preprocessors.connect_on_demand` | `ensure_connected` | (phase 2) `_connect_in_batches` | connection was never lazy |
| `geecs_core.db.variable_types` | moved from `geecs_ca_gateway.config` | the gateway copy (deleted; PVA re-pointed) | three packages need it |
| `datatypes=` on readables | — | — | the served set is not all floats |

Deleted: the first attempt's parallel `devices/geecs_device.py` + tests (it duplicated `CaGenericDetector`/`CaSnapshotReadable` with fewer capabilities).

- Every served settable is attached to the noun (not only catalog scan variables); tree clutter handled by `:depth=`.
- Motor vs settable from a **positive** DB tolerance (every catalog `kind: motor` target present in the DB has one; 0/NULL → plain setpoint); catalog `kind: motor` opt-in, `confirm`/`pseudo` become nouns in phase 3.
- Keep-connected after first use; no TTL.
- Facility literals: `TRIGGER_SOURCE_DEVICETYPES` (DG645, DG535, Highland, TDK-Lambda) is a devicetype list in code — the agreed shortcut until the DB carries `acq_timestamp`; flagged for the DB sweep.

## Tests

- GeecsBluesky: **760 passed, 5 deselected** (full suite, `-m 'not hardware and not integration'`), including the rewritten `test_namespace.py` (16), `test_connect_on_demand.py` (6), `test_qserver_startup.py` (+1 namespace export test).
- GEECS-Core 95 passed, GeecsCAGateway 137 passed, GeecsPvaGateway 29 passed (`./scripts/check.sh GEECS-Core GeecsCAGateway GeecsPvaGateway`).

## Hardware verification

**Accepted 2026-09-09** (first form, before the re-cut) on the worker box against the live gateway, `tests/test_namespace_hardware.py`: namespace of 105 devices (53 triggerable); HTU-NoGas armed/disarmed via the existing `ShotController`; stock `bp.count` — 3 shots on `UC_Amp4_IR_input`, `acq_timestamp` at exactly 1 Hz, the 9 subscribed columns + shot stamp; stock `bp.list_scan` over `U_S1H.Current` −1 → +1 A / 0.5 A — readbacks −0.99984, −0.49966, 0.00008, 0.50008, 0.99989 A; setpoint restored.

**Re-run on the composed form 2026-09-09 (`1162341e`, and again on `07920347` after the review-2 fixes with the `safe_name` keys): passed with identical numbers** — namespace 105 devices / 53 triggerable, camera `CaGenericDetector UC_Amp4_IR_input` with 51 children, count 3 shots at 1 Hz, scan readbacks −0.99984, −0.49966, 0.00008, 0.50008, 0.99989 A, setpoint restored to 8e-05. Documents: `/tmp/807-phase1-composed.json` on the worker.

Review: adversarial reports and dispositions follow as comments (per `/land`).


skbarber:
## Adversarial review #1 (correctness-focused brief) — report + dispositions

Reviewer: fresh-context subagent on the **first-form** diff (before the re-cut that deleted `GeecsDevice`). 74/74 unit tests passed at review time; every finding was reproduced by the reviewer. Dispositions below are against this PR's HEAD `1162341e`.

|---|---|---|---|
| 1 | P1 | `install_connect_on_demand` must be outermost; nothing enforced it — `SupplementalData` appended later injected messages it never saw | **Fixed** `1162341e`: install removes any existing instance and re-appends last; `test_install_puts_connect_on_demand_outermost` + `test_baseline_of_an_unconnected_device_connects_when_installed_last`. The latter also exposed that `declare_stream` names devices in `msg.args` and the RE *describes* them there — now connected too. |
| 2 | P2 | A failed connect masked by `unstage` of never-connected signals under `stage_wrapper`'s finalize | **Waived (tracked)**: the custom stage/unstage it targeted is gone; the composed `CaGenericDetector`/`CaSnapshotReadable` use `StandardReadable`'s. The masking can still occur for *any* ophyd device whose connect fails inside a stock plan's stage — pre-existing behaviour of the ca layer, not introduced here. Phase 2's preamble connects explicitly with a clear error before staging; noted in `01_device_namespace.md`. |
| 3 | P2 | stage/unstage asymmetry when `configure` changes the selection mid-run | **Dissolved**: selection/`configure` removed (no Undulator save set uses an explicit scalar list; `read()` = the DB subscribed list via `GeecsDbScalarPolicy`). |
| 4 | P2 | `configure(variables=[..., "acq_timestamp"])` raised `KeyError` | **Dissolved**: `configure` removed. |
| 5 | P2 | hardware test carried the `integration` marker (arms the machine trigger under `-m integration`); restored to the *readback* not the setpoint | **Fixed** `1162341e`: `hardware` marker; restore target is the `:SP` readback. |
| 6 | P3 | variables named `connected_status` / `devicetype` clobbered instance attributes | **Fixed** `1162341e`: `_attribute_for` checks the detector class *and* the instance (`hasattr(dev, attr)`). |
| 7 | P3 | case-different variable names (`Trigger` / `trigger`) silently merged lookups | **Fixed** `1162341e`: collision check is case-insensitive across GEECS names and attributes; raises `GeecsConfigurationError`. |
| 8 | P3 | `looks_triggerable` with an empty devicetype; exclusion was exact-match | **Fixed (substring match)** `1162341e`; the DG645 fixture now has its type and is asserted non-triggerable. **Waived**: no warning on an empty devicetype — a device with no devicetype and a trigger variable stays triggerable (no source could be detected); the DB sweep Sam is planning adds `acq_timestamp` and retires the heuristic. |
| 9 | P3 | `GeecsTriggeredDevice.disconnect()` never reached `GeecsDevice.disconnect` | **Dissolved**: class deleted; `CaGenericDetector`'s existing teardown applies. |
| 10 | P3 | served-set rule drifted from `GeecsDbServedSetProvider` (`systimestamp`, case) | **Fixed** `1162341e`: the namespace *calls* `GeecsDbServedSetProvider` (no copy); the explicit-roster fallback matches case-insensitively and excludes the gateway-synthesised names. |
| 11 | P3 | dead `"describe"` in `TOUCH_COMMANDS` | **Fixed** `1162341e`. |
| 12 | P3 | docs overstated (superseded typing rules; startup export + permissions not on branch) | **Fixed** `1162341e`: `01_device_namespace.md` rewritten to the composed design; startup exports the namespace, `allowed_devices: ":?.*:depth=3"`. |
| 13 | — | `configure` selection was process-global state on a long-lived device | **Dissolved**: `configure` removed. |

Net: 7 fixed, 5 dissolved by the re-cut, 1 waived-with-reason (2) plus a partial waiver on 8. The reviewer's "what is solid" (mixin MRO preservation, `is_connected` mock reasoning, loud namespace failure modes) stands.

A second review with the repo's verbatim three-lens brief (correctness / redundancy / placement) runs on this HEAD — report follows as a separate comment.

skbarber:
## Adversarial review #1 — reviewer re-verification of the dispositions (same reviewer, HEAD `1162341e`) and follow-up fixes (`0b8f39dd`)

| # | Reviewer verdict | Follow-up |
|---|---|---|
| 1, 5, 6, 8, 10, 11, 12 | **CONFIRMED FIXED** | — (#5 docstring nit fixed in `0b8f39dd`) |
| 3, 4, 9, 13 | **CONFIRMED DISSOLVED** | — |
| 2 | **WAIVER ACCEPTABLE** (pre-existing ca-layer behaviour under `stage_wrapper`; tracked for phase 2) | — |
| 8 (empty devicetype warning) | **WAIVER ACCEPTABLE** | — |
| 7 | **DENIED** — `by_lower` collapsed `Trigger`/`trigger` before the collision check, so the readable was silently dropped and `variable("UC_X","Trigger")` returned the settable | **Fixed** `0b8f39dd`: two served variables differing only by case raise `GeecsConfigurationError`; `test_case_different_variable_names_are_refused`. |

**New findings from the re-cut**

| | Finding | Disposition |
|---|---|---|
| N1 (P3) | the `hasattr(dev, attr)` escape added for #6 renamed a settable over an existing *readable child* and hid the readable | **Fixed** `0b8f39dd`: a Device-valued attribute is a real collision and raises; `test_settable_colliding_with_a_readable_child_is_refused`. |
| N2 (P3) | design note / changelog carried first-form hardware numbers; composed form not yet re-run | **Fixed** `0b8f39dd` + hardware: the composed form (`1162341e`) was re-run on the worker 2026-09-09 and passed with identical numbers (camera `CaGenericDetector UC_Amp4_IR_input`, 51 children; sweep readbacks −0.99984 … 0.99989 A; setpoint restored). Note and changelog say so. |
| N3 (nit) | `DeviceRoster.served_for` restated "subscribed ∪ settable" for explicit rosters | **Fixed** `0b8f39dd`: the fallback runs `GeecsDbServedSetProvider` over the roster through a `_RosterDb` adapter — one copy of the rule. |
| #10 nit | `get_experiment_device_variables` issued twice at startup (once directly, once inside the provider) | **Waived**: one extra DB round trip at environment open; the provider has no metadata-injection seam and growing one for this is not worth it. |

No finding is undispositioned. The second (canonical-brief) review runs against the composed diff; its report and dispositions follow.

skbarber:
## Adversarial review #2 (the `/land` three-lens brief, fresh context) — report + dispositions

Reviewer: fresh-context subagent on the composed diff (`1162341e`). It ran GEECS-Core, both gateways' and GeecsBluesky's device-layer + new tests (all passing at review time). Dispositions against `07920347`.

|---|---|---|---|
| 1 | **P1** | The option-list-is-enum rule (carried by the move to GEECS-Core) makes the gateway serve 18 Undulator variables as `DBR_ENUM`, while the **existing** per-scan factories still declare `float`; ophyd-async refuses `float` on an enum PV, so any `ScanRequest` touching one of those variables would fail at connect after a gateway deploy. | **Fixed** `07920347`: the rule moved to GEECS-Core **unchanged** (no served-type change in this PR); GeecsCAGateway is a patch (0.20.3) again; PV_CONTRACT §4 and its pinned test restored to the served behaviour. Sam's decision that those lists are enums is recorded as a **DB fix** (`variabletype='choice'` on the 18 rows) in `variable_types.py`, the GEECS-Core changelog and PV_CONTRACT — correcting the DB flips gateway and every client together, which is the invariant this PR is built on. |
| 2 | P2 | Every numeric settable with a non-NULL tolerance became a convergence `CaMotor` (the fixture pinned `exposure` with tolerance 0 → motor at 0.005); a setpoint whose readback does not echo within tolerance would time out after 30 s where a plain put succeeds. | **Fixed** `07920347`: `CaMotor` only for a **positive** DB tolerance; 0/NULL → `CaSettable`; the duplicated default-tolerance constant removed; fixture and test updated (`exposure` → `CaSettable`). Catalog `kind: motor` opt-in arrives with the axes in phase 3. |
| 3 | P2 | Event keys kept GEECS case (`U_S1H-Current-position`) against `EVENT_SCHEMA.md`'s `safe_name` rule; the settable child's `_column_headers` was not visible to `run_wrapper`, which merges headers from top-level detectors only. | **Fixed** `07920347`: ophyd device names, child attributes and therefore event keys are `safe_name` (lowercase) — `u_s1h-current-position`, `uc_amp4_ir_input-meancounts` — only the namespace *binding* keeps GEECS spelling (`U_S1H`); each attached child's headers are aggregated onto the parent (pinned). No EVENT_SCHEMA change needed: the diff now conforms to it. Hardware re-run on `07920347` passed with the renamed keys. |
| 4 | P3 | GeecsPvaGateway's `geecs-ca-gateway` dependency is dead but still declared; root `CLAUDE.md` still describes the edge. | **Fixed** `07920347`: dependency dropped, `poetry.lock` regenerated (no `geecs-ca-gateway` entry), root `CLAUDE.md` graph updated, PVA changelog says so. |
| 5 | P3 | `shot_monitor.py` cited the deleted `geecs_device` module and had no second host after the re-cut. | **Fixed** `07920347`: extraction folded back — `triggerable.py` is master's file plus the `datatypes=` amendment; `shot_monitor.py` deleted. |
| 6 | P3 | `DeviceRoster.served_for` restated the served-set rule. | **Fixed** (already in `0b8f39dd`, reviewed HEAD predates it): the fallback runs `GeecsDbServedSetProvider` over the roster via a `_RosterDb` adapter. The reviewer's alternative (a `from_metadata` constructor on the provider) is equivalent; the adapter keeps the provider's API untouched. |
| 7 | P3 | Hygiene: stale "0.20.3" comment (now correct again), "moved verbatim" wording, unreachable `int` mapping, `CLAUDE.md` layout, facility literals as hardware-test defaults, duplicated `_Docs`. | **Fixed** `07920347` except one **waiver**: the hardware test keeps `Undulator` / `UC_Amp4_IR_input` / `HTU-NoGas` as **env-overridable defaults** — the same shape as the existing `test_scan_request_hardware.py` (`Undulator`, `Amp4In`, `HTU-LaserOFF`); flagged here per the facility-literal rule. |

Net: 6 fixed (one already in the previous commit), 1 partial waiver. The reviewer's "what is solid" (connect_on_demand subtree/retry reasoning, outermost install, the reuse of the `db_runtime` providers and `Ca*` classes, the deletion of the parallel device) stands. Fixes sent back to the same reviewer for confirm/deny.

skbarber:
## Adversarial review #2 — reviewer re-verification of the dispositions (same reviewer, HEAD `07920347`)

All seven findings **CONFIRMED FIXED** (7 fixed / waiver accepted on the hardware-test facility defaults, which match the existing `test_scan_request_hardware.py` convention and are flagged in the PR body). The reviewer re-ran GeecsBluesky 70, GEECS-Core 5, GeecsCAGateway 23, GeecsPvaGateway 29 targeted tests, verified `triggerable.py` is master +15/−4 (the `datatypes=` amendment only), the PVA lock carries no `geecs-ca-gateway` and no worktree paths, and that the namespace's declared types again match the served PVs.

**New findings from the fixes: none surviving.** The one seam probed — a case-only device-name pair (`U_S1H`/`u_s1h`) giving distinct bindings but one `_by_geecs_name` entry — cannot occur: GEECS device names are unique case-insensitively in the DB collation.

No undispositioned finding remains on this PR. Ready for maintainer review.

skbarber:
## External review (codex) — report + dispositions

**[P2] `namespace.py`: only case-only collisions were rejected; `safe_name` is also lossy for punctuation/whitespace.** **Confirmed and fixed** (`83d2f6d8`).

Reproduced exactly as described, and narrowed while confirming: `safe_name("A.B") == safe_name("A B") == "a_b"`, and for a **readable/readable** pair the device constructor's `setattr` silently overwrote one child while both lookups mapped to the survivor:

```
READABLE+READABLE: NO RAISE. children: ['a_b'] | lookups A.B -> u_x-a_b | A B -> u_x-a_b
SETTABLE+SETTABLE  raised: ... settable 'A B' collides with the child already bound as 'a_b'
READABLE+SETTABLE  raised: ... settable 'A B' collides with the child already bound as 'a_b'
```

So the two settable combinations were already caught by the attachment-time guard; the readable/readable pair was the live hole. Fix: the build-time guard now keys on `safe_name(var)` instead of `var.lower()`, which **subsumes** the case check — one guard for both classes. The citation of the gateway precedent is accurate and is now quoted in the code comment: `GeecsCAGateway/geecs_ca_gateway/gateway.py:379` ("PV name collision: … maps to both …") and `GeecsPvaGateway/geecs_pva_gateway/server.py:292` ("PV name collision after normalization").

```
device namespace: U_X: variables 'A.B' and 'A B' both normalise to 'a_b'
```

Tests: the collision test is now parametrized over case-only (`Trigger`/`trigger`), punctuation-vs-whitespace (`Position.Axis 1`/`Position Axis 1`) and an already-mangled spelling (`Wavelength (nm)`/`wavelength_nm`). The attachment-time guard kept its own test, repointed at the case only *it* can catch — a settable named like a child the device class itself creates (`connected_status` on a triggered device); the earlier fixture for it is now caught by the build-time guard instead. 72 passed across `test_namespace.py`, `test_connect_on_demand.py`, `test_qserver_startup.py`, `test_ca_devices.py`.

**Note — `QS_DEVICE_NAMESPACE=off` exports only the five `geecs_*` plans, not stock `count`/`scan`.** Correct, and non-blocking as you judged: stock-plan registration is phase 3 (`Planning/native_bluesky/00_overview.md`). Phase 1 deliberately ships only the device nouns and the lazy connect; the stock verbs reach `plans_allowed` with the per-step + registration table. Worth adding that with the namespace **on** (the deployed default) the export is the devices *plus* those five plans — the `off` switch exists for hermetic tests and boxes without DB reach.

Thanks — this was the one collision class both in-house reviews missed.

Issue #806, opened 2026-09-09
Image writing: replace the capture daemon with an areaDetector-shaped file plugin in GeecsPvaGateway + stock ophyd-async detectors


skbarber:
## Decision

GEECS-Plugins owns the image file writer, and the writer runs **on the camera servers, inside the GeecsPvaGateway process**, as a sibling consumer of the same in-memory frame the gateway already publishes over PVA. On the worker, each capture-eligible camera is a stock **ophyd-async `StandardDetector`** driving that writer over the standard areaDetector file-plugin PV contract, so the run's documents reference the files and Tiled reads them with no custom adapter. LabVIEW keeps acquiring; Python owns the format.

This supersedes the capture-daemon design (`geecs_bluesky/capture/`, #694/#699/#700) and the "asset-doc design wrinkle" left open in `Planning/data_capture/01_central_pva_capture_scope.md`. Agreed with @skbarber 2026-09-06 → 2026-09-08.

## Why

The daemon got the hard parts right (lossless PVA intake, timestamp dedupe, stale filter, single-thread HDF5 writer, format, diff tool, read side — all lab-verified). It got the **seam** wrong: it was modelled on `sfile_callback.py`, a *consumer* of the run, instead of on a detector, a *participant* in it. Everything awkward about it follows from standing outside the run:

- it learns the scan folder from the start document (a convention, not a contract);
- it needs discovery because it does not know the cameras;
- it needs a heartbeat preflight because the plan cannot see it;
- writers are created lazily because it raced the plan's `mkdir`;
- **nothing in the run's documents references the HDF5 stacks** — from Bluesky's point of view they do not exist.

The reference EPICS/Bluesky shape keeps driver and file writer as *separate components in the same process on the camera host*: an areaDetector IOC's HDF5 plugin sits beside its PVA plugin, both reading one in-memory `NDArray`; the ophyd-async detector on the worker is only a remote control (path, count, capture on, wait for frames-captured) plus the paperwork (stream-resource / stream-datum). The format is chosen by which file plugin runs. GEECS fused driver and writer inside the LabVIEW image server, which is why the format has been frozen at per-shot PNG. The gateway already un-fuses the *streaming* half; this issue un-fuses the *writing* half the same way.

[...]

PR #809, opened 2026-09-10
[ON HOLD] GeecsBluesky 0.79.0: the preamble as a RunEngine preprocessor (#807 B1) — superseded by #810

skbarber:
Phase **B1** of #807. Base: `feature/native-bluesky-plans` (phase 1 / #808 merged).
GeecsBluesky **0.79.0**. Adversarially reviewed; every finding fixed, none
waived — the review and the dispositions are in the comments.

## What

A stock `bluesky.plans` verb now runs a full GEECS scan in **strict** shot control:

```python
RE(bp.count([cam], num=5, md={"geecs": request}))
```

That claims a scan number, writes `ScanInfo`, configures native saving, runs setup actions, arms, **fires every shot itself**, records, and on the way out stops saving, disarms, runs closeout and disconnects — with no GEECS code in the plan.

It is the funnel's own code with two callers, never a copy:

- `plans/preamble.py` — `resolve_request` + `prepare_step_scan`, extracted verbatim from `geecs_scan_request_plan` (which keeps working unchanged and loses ~300 lines).
- `GeecsSession.configure_claimed_scan` — the post-claim tail (ScanInfo → role wiring → saving), split out of `build_claimed_scan_plan` because the preprocessor must not build an inner plan: the stock plan **is** the inner plan.
- `plans/run_wrapper.claimed_scan_metadata` — the run-metadata assembly (and the capture-dir creation it implies), split out of `geecs_run_wrapper` so both doors cannot drift from what the downstream readers require.

## The three mechanics worth reviewing

**The `open_run` seam.** `bps.open_run` puts `md` into `msg.kwargs`, and the RunEngine reads it only when it processes the message — so a `plan_mutator` can run the entire preamble and *then* forward a metadata-injected `open_run`. That is what preserves "the claim happens before the run opens", and it is why the claimed `scan_number`/`scan_folder` reach the start document. Re-wrapping needs a `forwarded`-id guard (otherwise the `ChainMap` recurses); injection is `inject_md_wrapper`'s technique.

**The fire.** `trigger_and_read` issues `trigger` for every detector under one group and then waits on that group; strict needs the shot to land in exactly that gap (the detectors have baselined `acq_timestamp` and are waiting). That is the gap `bps.trigger_and_read` cannot express and the reason `plans/single_shot.py` exists. The preprocessor inserts the fire there, so **a stock plan needs no `per_shot` hook**.

**The device source.** A stock plan is handed namespace nouns; the preamble used to build its own devices from the save set. Two objects for one device would double the connections and configure saving on something the plan never reads, so `prepare_step_scan` grows a `namespace=` seam and the save set **selects** rather than constructs (`namespace_detectors`). Roles are asserted against the namespace's triggerable classification, and a save set recording scalars the device does not read is refused rather than silently logging different columns.

## What the review changed

The reviewer named the pattern better than the original design note did:
**the funnel guaranteed by construction what this door can only enforce or
refuse.** One list of devices was both the read set and the save set; each
axis was built from its catalog entry's own topology; the devices were
discarded at the end of the scan. None of that survives a door where the
caller brings the plan and the devices are long-lived nouns.

So three things are now shared rather than re-derived — the fire with its
bounded refire and device-down gating (`single_shot.fire_and_await_shot`),
the eager save-off for capture-owned cameras
(`run_wrapper.save_control_only_off_plan`), and the GEECS execution keys the
Tiled catalog and schema read (`step_scan.geecs_execution_md`) — and three
things are refused before anything is claimed:

- a plan that does not read every device the save set records (saving files
  with no `acq_timestamp` row to join them to is the orphan-frame failure
  from the other side);
- a plan whose shape contradicts the request it is recorded as, values as
  well as counts;
- a scan axis needing a topology the namespace cannot build yet — pseudo,
  `confirm:`, or `kind: motor` with no positive DB tolerance. Degrading one
  of those to a fire-and-forget setpoint would take every row before the
  readback converged, with nothing in the data saying so.

Long-lived devices are also reset between runs now
(`devices/reset_support.py` chains the mixins), because a namespace noun
keeps whatever a run configured — stale saving state made an unrelated later
plan emit asset documents pointing into the previous scan's folder.

**The parity test the design note specified is in the suite**, and it earned
its place immediately: one request through both doors, every GEECS-owned
start-document key compared exactly, identical `ScanInfo`, event columns as
an exact set. It found a missing `acq_timestamp` s-file header, phantom
headers for children the device never reads, a raw request blob riding into
every start document, and six execution keys only the funnel emitted.

## Fixed here

The first cut armed the controller and *then* enabled saving. Arming drives to SCAN (external rising edges), so saving switched on while edges were already passing and every shot before the first recorded one wrote an **orphan frame**. Correct order is `arm_single_shot` (ARMED halts the free run, quiescence confirmed) → save-on; save-off stays the innermost finalize so it runs before disarm releases edges. A test asserts the ordered shot-control writes rather than inferring the bracket.

## Scope

`geecs_preamble` is **strict-only**: free-run is refused before anything is claimed (it is retired — see the plan restructuring on #807), as is strict without a trigger profile. `NonScalarSaveSupport.configure_saving_mode` makes the native-save mode per-run, which a long-lived namespace device needs and the per-scan classes got from their constructor.

## Reuse ledger

| new symbol | reuses | replaces | new because |
|---|---|---|---|
| `plans/preamble.py` | the funnel's body, moved verbatim | the same code in `scan_request_plan.py` (deleted there) | one implementation, two callers |
| `preprocessors.geecs_preamble` | those functions, `plan_mutator`, the `ChainMap` md technique | (phase C) the funnel's role as the only preamble door | stock plans had no GEECS preamble |
| `install_geecs_preamble` | phase 1's `install_connect_on_demand` shape | — | ordering: preamble inside, connect-on-demand outermost |
| `preamble.namespace_detectors` | `CaGenericDetector`/`CaSnapshotReadable`, the namespace | (phase C) `_build_request_detectors` for the stock door | object identity with the plan's detectors |
| `session.configure_claimed_scan` | `_write_scan_info`, `_configure_saving` | the same block inline in `build_claimed_scan_plan` | the preprocessor needs the tail without an inner plan |
| `run_wrapper.claimed_scan_metadata` | the md block, moved | the same block inline in `geecs_run_wrapper` | both doors must build identical md |
| `NonScalarSaveSupport.configure_saving_mode` | the device's own served `save`/`localsavingpath` children | — | per-run mode on a long-lived device |
| `GeecsNamespace.variable_names` | `_attrs` | — | the device object does not carry it |

Nothing here re-solves validation, resolution, claiming, ScanInfo, saving, actions or shot control.

## Known gap, owned by phase 3

Stock plans emit no `ScanContext` per-row columns (`bin_number`,
`scan_event_index`, `shot_index_in_bin`). Those arrive with the per-step
function, which is also when `ScanContext` retires; inferring step
boundaries from `set` messages here would be exactly the custom-code-over-a-
native-seam move this effort exists to avoid. The parity test asserts them
as the **only** column difference, so the gap cannot widen unnoticed.

- Strict-only, now, rather than carrying free-run through a phase we intend to delete.
- The fire lives in the preprocessor rather than a `per_shot` the caller supplies, so stock plans stay stock. The alternative needs the caller to build the controller, duplicating the preamble.
- The funnel stays alive and green; it retires in phase C.

## Tests

- 8 new in `tests/test_preamble_preprocessor.py`: install order, no-op without the key, the full preamble on a stock `count` (claim, ScanInfo, metadata, saving, events, one fire per shot), the Gate-2 bracket ordering, free-run refused pre-claim, the pre-claim invariant, and both namespace-source refusals. The harness drives the **real** `ShotController` through its existing `setter_factory` seam, so the real profile is replayed and nothing in the engine is stubbed.
- 11 more from the review, one per finding, including the parity test above.
- Full suite **783 passed, 0 failed** in 5m17s.
- **CI green** on the head commit: `unit-tests` 3m33s and 3m41s,
  `console-windows`, `pre-commit`.

## Hardware verification

**Accepted 2026-09-09 on the Undulator machine**, `tests/test_preamble_hardware.py` (hardware-marked), subscribing what the worker startup subscribes so the runs are downstream-indistinguishable:

- **Scan 63** — stock `bp.count` as a noscan: 3 shots, each fired by the preprocessor.
- **Scan 64** — stock `bp.list_scan` as a step scan: the catalog axis `S1H` resolved to `U_S1H:Current` and to the same object the plan moved; swept −1 → +1 A in 0.5 A steps; readbacks −0.99984, −0.49966, 0.00008, 0.50008, 0.99989 A; setpoint restored.
- Both wrote `ScanInfoScanNNN.ini` **and** `ScanDataScanNNN.txt`, and **both appear in the Tiled catalog** (`list_runs("Undulator", 2026-09-09)`), i.e. they are visible in the data portal beside the operators' scans.
- Event columns carry the full detector surface — `shot_id`, `shot_offset`, `nonscalar_save_path`, the image asset — so the composed namespace device keeps every capability the per-scan classes had.

Scans 56–62 that day are earlier test artifacts from a run with Tiled disabled (folder + ScanInfo only, no catalog entry); disposable.

### DIFF
--- a/GeecsBluesky/CHANGELOG.md
+++ b/GeecsBluesky/CHANGELOG.md
@@ -4,6 +4,39 @@ All notable changes to `geecs-bluesky` are documented here.
 
 Format follows [Keep a Changelog](https://keepachangelog.com/en/1.1.0/).
 
+## [0.78.0] - 2026-09-09
+
+### Added
+
+- **`preprocessors.geecs_preamble`** (GEECS-Plugins#807 phase 2) — the GEECS
+  scan preamble and finalize chain as one RunEngine preprocessor, keyed on
+  `md["geecs"]`. A stock `bluesky.plans` verb whose run metadata carries a
+  ScanRequest now gets, before its run opens: validation, name resolution,
+  the shot controller, the unserved/CONNECTED preflights, action slots, the
+  capture toggle, the preamble's own connects, the **scan-number claim**,
+  the ScanInfo write, native-save configuration, setup actions, arm and
+  save-on — and the GEECS run metadata injected into its start document.
+  On the way out, in the funnel's nesting order: save-off, disarm, closeout,
+  disconnect.
+
+  ```python
+  RE(bp.count([cam], num=5, md={"geecs": request}))   # a full GEECS scan
+  ```
+
+  It is the funnel's own code (`plans/preamble.py`), not a second copy; the
+  preprocessor supplies only the seam and the ordering. `install_geecs_preamble`
+  keeps `connect_on_demand` outermost.
+
+  The request must ride in the **plan's** `md=`: RunEngine per-call metadata
+  (`RE(plan, geecs=...)`) never enters a message, so a preprocessor cannot
+  see it. Pinned by a test.
+- `plans/run_wrapper.claimed_scan_metadata()` — the GEECS run-metadata
+  assembly (and the capture-dir creation it implies) split out of
+  `geecs_run_wrapper` so the funnel door and the stock-plan door cannot
+  drift from what the downstream readers require.
+- `GeecsNamespace.variable_names()` — the served scalar names a namespace
+  device has children for.
+
 ## [0.77.1] - 2026-09-09
 
 ### Changed
--- a/GeecsBluesky/CHANGELOG.md
+++ b/GeecsBluesky/CHANGELOG.md
@@ -4,6 +4,29 @@ All notable changes to `geecs-bluesky` are documented here.
 
 Format follows [Keep a Changelog](https://keepachangelog.com/en/1.1.0/).
 
+## [0.78.1] - 2026-09-09
+
+### Fixed
+
+- **Save windowing in `geecs_preamble`** (Gate-2): it armed the controller
+  and *then* enabled saving, so saving switched on while external edges were
+  already passing and every shot until the first recorded one wrote an orphan
+  frame. The order is now `arm_single_shot` (ARMED — single-shot source, free
+  run halted, quiescence confirmed) → save-on, with save-off still the
+  innermost finalize so saving stops before disarm releases edges. Pinned by
+  asserting the ordered shot-control writes.
+
+### Changed
+
+- `geecs_preamble` is **strict-only**: a free-run request is refused before
+  anything is claimed (free-run is retired, GEECS-Plugins#807), as is a
+  strict request with no trigger profile.
+- **The shot is fired by the preprocessor**, between the detectors' `trigger`
+  messages and the `wait` on their group — the gap `bps.trigger_and_read`
+  cannot express and the reason `plans/single_shot.py` exists. A **stock**
+  plan therefore needs no `per_shot` hook: `RE(bp.count([cam], num=5,
+  md={"geecs": request}))` fires every shot itself.
+
 ## [0.78.0] - 2026-09-09
 
 ### Added
--- a/GeecsBluesky/CHANGELOG.md
+++ b/GeecsBluesky/CHANGELOG.md
@@ -18,6 +18,16 @@ Format follows [Keep a Changelog](https://keepachangelog.com/en/1.1.0/).
 
 ### Changed
 
+- **Hardware-accepted 2026-09-09** on the Undulator machine: a stock
+  `bp.count` ran as a noscan (scan 61 — 3 shots, the preprocessor firing
+  each one) and a stock `bp.list_scan` ran as a step scan (scan 62 — the
+  catalog axis `S1H` swept −1 → +1 A in 0.5 A steps, readbacks within
+  0.4 mA, setpoint restored). Both claimed real scan numbers, wrote their
+  `ScanInfo` ini, and recorded the full detector surface — `shot_id`,
+  `shot_offset`, `nonscalar_save_path`, the image asset — so the composed
+  namespace device keeps every capability the per-scan classes had. The
+  catalog axis resolved to `U_S1H:Current` and to the same object the plan
+  moved.
 - `geecs_preamble` is **strict-only**: a free-run request is refused before
   anything is claimed (free-run is retired, GEECS-Plugins#807), as is a
   strict request with no trigger profile.
--- a/GeecsBluesky/CHANGELOG.md
+++ b/GeecsBluesky/CHANGELOG.md
@@ -4,6 +4,25 @@ All notable changes to `geecs-bluesky` are documented here.
 
 Format follows [Keep a Changelog](https://keepachangelog.com/en/1.1.0/).
 
+## [0.77.1] - 2026-09-09
+
+### Changed
+
+- **Scan preamble extracted to `plans/preamble.py`** (GEECS-Plugins#807
+  phase 2, step A) — a pure move, no behaviour change. `resolve_request`
+  (validation + trigger profile → `ShotController` + strict decision, shared
+  with optimize) and `prepare_step_scan` (save sets, preflights, action
+  slots, axis resolution, capture toggle, deferred construction, the in-plan
+  connect batches, the fail-fast gates, the metadata spec) now have **one
+  implementation with two callers**: `geecs_scan_request_plan`, unchanged in
+  behaviour, and — next — the RunEngine preprocessor that gives stock
+  `bluesky.plans` verbs the same preamble. `_DeferredConnectFactories`, the
+  connect/disconnect plan stubs and their two constants moved with the code.
+  `PreparedScan.as_claimed_kwargs()` keeps the funnel's
+  `build_claimed_scan_plan` call a one-liner.
+  Tests that patch `make_scalar_policy` / `ShotController` by module path now
+  target `geecs_bluesky.plans.preamble`.
+
 ## [0.77.0] - 2026-09-09
 
 ### Added
--- a/GeecsBluesky/CHANGELOG.md
+++ b/GeecsBluesky/CHANGELOG.md
@@ -22,6 +22,20 @@ Format follows [Keep a Changelog](https://keepachangelog.com/en/1.1.0/).
   `build_claimed_scan_plan` call a one-liner.
   Tests that patch `make_scalar_policy` / `ShotController` by module path now
   target `geecs_bluesky.plans.preamble`.
+- **Device-source seam** on `prepare_step_scan(..., namespace=...)`
+  (#807 phase 2): `None` keeps the funnel's per-scan construction; a
+  `GeecsNamespace` makes the save set **select** namespace devices instead
+  (`plans/preamble.namespace_detectors`), so a stock plan's detectors and
+  the preamble's are the same objects — two objects for one device would
+  double the connections and configure saving on something the plan never
+  reads. Save-set roles are asserted against the namespace's triggerable
+  classification, and a save set recording scalars the device does not read
+  is refused rather than silently logging different columns.
+- `NonScalarSaveSupport.configure_saving_mode()` — native-save mode as a
+  **per-run** setting. The per-scan classes fix it at construction, which a
+  long-lived namespace device cannot do (the same camera saves natively in
+  one scan and is capture-owned in the next). It only sets the flags; the
+  two control children are the device's own served settables.
 - `GeecsSession.configure_claimed_scan()` — the post-claim tail (ScanInfo
   write → role wiring → native-save configuration) split out of
   `build_claimed_scan_plan`, which now calls it. The preamble preprocessor
--- a/GeecsBluesky/CHANGELOG.md
+++ b/GeecsBluesky/CHANGELOG.md
@@ -22,6 +22,11 @@ Format follows [Keep a Changelog](https://keepachangelog.com/en/1.1.0/).
   `build_claimed_scan_plan` call a one-liner.
   Tests that patch `make_scalar_policy` / `ShotController` by module path now
   target `geecs_bluesky.plans.preamble`.
+- `GeecsSession.configure_claimed_scan()` — the post-claim tail (ScanInfo
+  write → role wiring → native-save configuration) split out of
+  `build_claimed_scan_plan`, which now calls it. The preamble preprocessor
+  needs that tail without building an inner plan (the stock plan *is* the
+  inner plan), so this keeps one implementation rather than a copy.
 
 ## [0.77.0] - 2026-09-09
 
--- a/GeecsBluesky/CHANGELOG.md
+++ b/GeecsBluesky/CHANGELOG.md
@@ -4,6 +4,127 @@ All notable changes to `geecs-bluesky` are documented here.
 
 Format follows [Keep a Changelog](https://keepachangelog.com/en/1.1.0/).
 
+## [0.79.0] - 2026-09-09
+
+Adversarial review of the phase-2 door (GEECS-Plugins#807, PR #809).  Every
+finding was about the same thing: what the ScanRequest funnel guaranteed **by
+construction** — one list of devices, one topology per axis, disposable
+per-scan objects — is, on the stock-plan door, a precondition on the caller.
+Each is now either enforced or refused, and the acceptance test the plan of
+record specified is in the suite.
+
+### Fixed
+
+- **Per-run device state leaked between runs.** A namespace device is a
+  long-lived noun, so the saving mode, save path, asset definitions and
+  shot-ID origin one run configured stayed set on it. An unrelated later
+  plan emitted a `nonscalar_save_path` column — and external asset
+  documents — resolving into the *previous* scan's folder. Every mixin that
+  holds per-run state now has a cooperative `reset_run_configuration`
+  (`devices/reset_support.py` chains them along the MRO), and
+  `GeecsNamespace.reset_run_configuration()` clears the whole namespace both
+  before a run configures it and on the way out.
+- **The `acq_timestamp` s-file header went missing on the stock door.** The
+  detector constructors add it when their save flags are fixed at build
+  time; a namespace device gets its mode per-run, so the header now follows
+  `configure_saving_mode`. Without it a saved scan's images have no join
+  key back to its rows.
+- **A stock scan wrote no `scans/ScanNNN/scan.log`.** The preamble now
+  attaches the per-scan log handler around the forwarded run, as the funnel
+  does, so `/triage` can read these scans.
+- **Capture-owned cameras were not driven `save="off"`.** Extracted as
+  `run_wrapper.save_control_only_off_plan` and called from both doors.
+- **The scan motors were missing from `geecs_scalar_headers`**, so the
+  Tiled→s-file exporter could not put the legacy `Device Variable` header
+  back on the axis column. The namespace also stopped contributing headers
+  for settable children it does not read — phantom column names.
+- **The raw ScanRequest rode into every start document** as a `geecs` key
+  no consumer declares and the funnel never emits. Dropped; the resolved
+  picture was already there.
+- **`install_geecs_preamble` re-derived `connect_on_demand`'s
+  configuration** instead of carrying it, so re-installing after a mock
+  worker's `install_connect_on_demand(RE, mock=True)` would have sent that
+  worker at real Channel Access.
+
+### Changed
+
+- **The stock door reuses `geecs_single_shot`'s refire.** The message-level
+  fire had the arm→fire→wait seam but none of the recovery around it: a
+  dropped frame (~1% per shot, live-measured) propagated a `FailedStatus`
+  into a stock plan that has no handler, aborting the run — at that rate a
+  100-shot scan aborts more often than not. The retry loop, the bounded
+  refire and the gateway-liveness gating are now one implementation,
+  `plans/single_shot.fire_and_await_shot`, called by `geecs_single_shot`
+  and by the preprocessor (which hands it the plan's pending `wait`).
+- **Background telemetry reaches the event again.** It was prepared,
+  connected and advertised, but a stock plan reads only its own detectors,
+  so no telemetry column was ever emitted. The preprocessor now inserts the
+  reads between the plan's last `read` and its `save`: still one row, still
+  the `primary` stream, no schema change.
+- **The GEECS execution keys are shared.** `acquisition_mode`,
+  `geecs_event_schema`, `fires_own_shots`, `motor`, `positions` and
+  `shots_per_step` — read by `tiled_catalog` and `tiled_schema` — come from
+  one `plans/step_scan.geecs_execution_md`, so a scan cannot be recorded
+  differently depending on which door ran it.
+- **Save-set selection moved to the namespace** as
+  `GeecsNamespace.select(devices_config)`, replacing
+  `plans/preamble.namespace_detectors`: it emits no plan messages, and as a
+  method it uses the namespace's own `_SYNTHESIZED` set and triggerable
+  classification instead of re-deriving them, which also lets
+  `variable_names` go back to being private.
+- The worker-wide default session moved to **`geecs_bluesky/plan_session.py`**
+  (`set_plan_session` / new `get_plan_session`). It has two consumers now,
+  and the preprocessor should not import a private name from the funnel
+  module that phase C deletes. Re-exported from `plans/scan_request_plan`.
+- **The queueserver installs the preamble** (`qserver/startup/startup.py`),
+  so the door is reachable from the worker rather than hand-driven sessions
+  only. A plan without `md["geecs"]` is untouched.
+
+### Added — refusals, where the funnel's guarantee cannot be kept
+
+- A plan that does **not read every device the save set records** is refused
+  before the claim. Saving files for a device with no `acq_timestamp` row to
+  join them to is the orphan-frame failure the Gate-2 windowing exists to
+  prevent, arriving from the other side.
+- A plan whose **shape contradicts the request it is recorded as** is
+  refused: `bp.count(num=2)` under a request declaring `shots_per_step: 5`
+  would write ScanInfo and a catalog row its own data contradicts. Equal
+  point counts do not hide a different set of points either — each
+  commanded position is checked against the declared list.
+- **Scan-axis topologies the namespace cannot build** are refused rather
+  than degraded: a pseudo/composite variable, a `confirm:` target, and a
+  `kind: motor` variable with no positive DB tolerance. Silently handing
+  back a fire-and-forget setpoint would let a stock `list_scan` step to the
+  next point before the readback converged — every row at the wrong
+  position, with nothing in the data saying so. They arrive with the
+  catalog axes in phase 3.
+
+### Testing
+
+- 11 new tests (19 in `tests/test_preamble_preprocessor.py`), one per
+  finding above, including the **document-parity test the plan of record
+  specified**: one request through both doors, every GEECS-owned start-doc
+  key compared exactly, identical `ScanInfo`, and the event columns compared
+  as an exact set — the funnel's three `ScanContext` columns (`bin_number`,
+  `scan_event_index`, `shot_index_in_bin`) are asserted as the *only*
+  difference, so any other divergence fails. That test found the missing
+  `acq_timestamp` header, the phantom settable headers, the stray `geecs`
+  key and the six missing execution keys.
+- `Docs` and `_sweep_points` were copies of `tests/ca_mock_helpers`'
+  `DocCollector` and of each other; the shared helpers gained the `start`
+  property and `sweep_points`, and the copies are gone.
+- The Gate-2 ordering test now records the save writes into the **same**
+  ordered log as the shot-control writes, so swapping `arm_single_shot` and
+  `save_enable_plan` fails it. Previously only the trigger states were
+  pinned and the reorder passed.
+
+### Known gap
+
+- Stock plans do not yet emit the `ScanContext` per-row columns
+  (`bin_number`, `scan_event_index`, `shot_index_in_bin`). Those come from
+  the per-step function in phase 3, which is also when `ScanContext`
+  retires; the parity test pins the difference so it cannot widen.
+
 ## [0.78.1] - 2026-09-09
 
 ### Fixed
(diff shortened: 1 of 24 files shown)


skbarber:
## Adversarial review — fresh-context subagent, `/land` three-lens brief

Posted verbatim; dispositions follow in a separate comment.

---

Reviewed against the whole repo (root + `GeecsBluesky/CLAUDE.md`, `Planning/native_bluesky/`, the existing funnel/orchestration/single-shot modules, `tests/ca_mock_helpers.py`). The extraction into `plans/preamble.py` is a genuine move, and `claimed_scan_metadata` really is the one md assembly (`grep bluesky_backend` finds exactly one). The findings below are all about the **new second door**, where the funnel's guarantees are carried by construction and the preprocessor's are carried by convention.

Three findings were reproduced with runnable probes against the branch (`poetry run python`, mock session, the PR's own fixtures); the transcripts are quoted inline. `tests/test_preamble_preprocessor.py` passes 8/8 here.

---

### P1 — Per-scan state is written onto long-lived namespace devices and never reset

`geecs_preamble`'s `_cleanup` (`preprocessors.py:232-243`) undoes saving, arming, closeout and connects — but nothing undoes the *device configuration* the preamble applied. `namespace_detectors` calls `configure_saving_mode()` (`plans/preamble.py:328-331`), and `configure_claimed_scan` calls `configure_nonscalar_file_logging()` / `configure_external_asset_logging()` (`session.py:670-718` → `_configure_saving`, `session.py:1697`). On the funnel these are set on per-scan objects that are thrown away; on the namespace door they are set on nouns that live for the whole worker session.

Reproduced — one GEECS scan (claim 7), then a *plain* `bp.count([cam], num=1)` with no `md["geecs"]`:

```
=== after the GEECS scan (claimed scan 7) ===
  _save_nonscalar_data : True
  _nonscalar_save_path : .../Scan007/U_Cam
=== the NEXT plain bp.count (no md['geecs']) ===
  start has scan_number: False
  LEAKED COLUMN u_cam-nonscalar_save_path = .../Scan007/U_Cam
```

An unrelated run emits a `nonscalar_save_path` column pointing at the previous scan's folder. With the DB reachable, `_asset_scan_number` and `_asset_definitions` survive too, so `_asset_datakeys()` (`devices/nonscalar_save.py:117-133`) describes `external: "OLD:"` columns and the run emits StreamResource/Datum documents resolving into **scan 7's** files. The same leak fires between two GEECS scans whenever save set B drops a device save set A configured (`namespace_detectors` only touches devices named by the current save set).

`test_a_plan_without_the_geecs_key_is_untouched` cannot catch this: it uses a fresh `namespace` fixture that has never seen a GEECS run.

Fix shape: `_cleanup` must reset what the preamble configured (a `reset_saving_mode()` / `configure_nonscalar_file_logging(None)` pass over `prepared.detectors`), and it must reset *every* namespace device the preamble ever touched, not just this scan's save set.

---

### P1 — Nothing checks that the plan's detector list covers the save set; the mismatch is silent and exits `success`

The funnel's read set *is* the save set (`_build_request_detectors` builds it and hands the same list to `build_claimed_scan_plan`). The preprocessor decouples them: `prepare_step_scan(namespace=…)` selects the save set's devices and turns their native saving on, while the plan reads whatever the caller passed to `bp.count`/`bp.list_scan`. Nothing compares the two — `tests/test_preamble_hardware.py:110-112` builds them by hand from the same save set, which is exactly the tell.

Reproduced — save set `UC_Test` = `[U_Cam, U_Cam2, U_Slow]`, plan handed only `U_Cam2`:

```
start nonscalar_save_paths: {'U_Cam': '.../Scan007/U_Cam'}
U_Cam localsavingpath was written to: .../Scan007/U_Cam
fires: 2
event columns: ['u_cam2-acq_timestamp', 'u_cam2-shot_id', ..., 'u_cam2-val', 'u_cam2-valid']
U_Cam in the event row?  False
exit: success
```

Scan 7 claims a real number, writes ScanInfo, enables native saving on `U_Cam`, fires two shots at the machine, and records **no `U_Cam` row and no `U_Slow` row**. Every frame `U_Cam` writes for those shots is an orphan with no `acq_timestamp` to join on — precisely the failure class the Gate-2 fix in this same PR exists to prevent, arriving through the other side. The start document advertises `nonscalar_save_paths` for files nothing can join, and the capture daemon's directory was mkdir'd for them (`run_wrapper.py:196-203`).

The preamble knows both lists at `open_run`; refusing (or at minimum warning loudly) when `set(prepared.detectors) ⊄ set(plan detectors)` is a five-line guard in `_run_preamble`.

---

### P1 — The message-level fire drops `geecs_single_shot`'s refire and device-down recovery

`_fire_before_wait` (`preprocessors.py:245-275`) re-implements the arm→fire→wait seam, but only the seam. `plans/single_shot.py:90-180` — described in `GeecsBluesky/CLAUDE.md` as the strict acquisition primitive, and used by `plans/step_scan.py:365` and `plans/optimize.py:172` — carries three behaviours the preprocessor path does not:

- **Bounded refire** (`max_refires=2`, fresh group per attempt) on a `FailedStatus` no-frame wait. Its own docstring quantifies the need: *"known camera frame-drop intermittency, ~1% observed"*, live-verified in `CHANGELOG.md` 0.20.0.
- **Gateway-liveness gating** — a frameless device whose `CONNECTED` PV reads `Disconnected` raises `GeecsDeviceDownError` instead of burning refires (`single_shot.py:56-88`).
- **Re-baselining/orphan drain** on the retry trigger, which is what keeps "one recorded row = one physical shot" true across a refire.

In the preprocessor path the `FailedStatus` from `bps.wait` propagates straight out of `_fire_then_wait` through `trigger_and_read` into a stock `bp.count`, which has no handler: the run aborts. At the stated ~1% per-shot drop rate a 100-shot stock scan aborts with ~63% probability, where the funnel would have recovered. Scan 63/64 (3 and 5 shots) is far too short to have exercised this.

This is also the review's clearest redundancy hit: the fire seam is now implemented twice, and the second copy is the weaker one. The insertable-plan shape `plan_mutator` gives you can call the existing primitive's retry loop rather than re-deriving it.

---

### P2 — `install_geecs_preamble` silently rewrites `connect_on_demand`'s `mock` and `timeout`

`preprocessors.py:390-399` strips the existing `connect_on_demand` partial and re-appends via `install_connect_on_demand(run_engine, mock=mock)` — its own `mock` (default `False`) and the module default timeout, discarding whatever the caller had configured. Reproduced:

```
after connect install:  [('connect_on_demand', {'mock': True, 'timeout': 45.0})]
after preamble install: [('geecs_preamble', {...}), ('connect_on_demand', {'mock': False, 'timeout': 20.0})]
```

`qserver/startup/startup.py:282` is `install_connect_on_demand(RE, mock=session._mock)`. The moment phase 2 is wired into startup in the documented order (preamble, then connect outermost — or a later `install_geecs_preamble` call), a mock worker's `connect_on_demand` starts attempting real CA connects. `test_install_keeps_connect_on_demand_outermost` compares only `[getattr(p, "func", p) …]`, so it passes with the kwargs wrong. Carry the existing partial's `keywords` forward instead of re-deriving them.

---

### P2 — Telemetry is prepared, announced, and then never read

`prepare_step_scan` builds and connects the Tier-2 group and returns `all_detectors = detectors + telemetry_readables` (`plans/preamble.py:539-547`, `:590`). The preprocessor feeds that list to `configure_claimed_scan` and `claimed_scan_metadata` (`preprocessors.py:322`, `:335`) — but the *plan* reads only the devices it was handed, so no telemetry column ever reaches an event. Meanwhile `build_step_scan_spec(telemetry_selected=…)` puts the selection in the start document and `_collect_scalar_headers` puts its headers in `geecs_scalar_headers`, so the run advertises columns that do not exist.

`Planning/native_bluesky/02_preamble_preprocessor.md` (correction 1, added by this PR) argues at length that telemetry must stay in the `primary` read set rather than move to a `baseline` stream — and then this door drops it entirely, which is a bigger schema change than the one the correction rejects. Either hand the group to the plan (a `SupplementalData`-style injection, or refuse `background_telemetry` on the stock door) or stop advertising it.

---

### P2 — `geecs_scalar_headers` loses the scan axis on the stock-plan door

The funnel passes `devices=scalar_devices = detectors + normalize_motors(motor)` (`orchestration.py:186,193`), so the movable's own `_column_headers` (`devices/ca/settable.py:85`) reaches the map. The preprocessor passes `devices=prepared.detectors` only (`preprocessors.py:335`) — the movables are in `prepared.motor_arg`, never in the list.

The namespace *parent* aggregates its settable children's headers (`namespace.py:358-364`), so this is invisible whenever the swept device is also in the save set — which is exactly the hardware case (`U_S1H` swept, `Amp4In` recorded → `U_S1H` is not a save-set device, so its header is already missing from scan 64). Consequence: the Tiled→s-file exporter cannot rename the axis column back to `U_S1H Current`, so the s-file for a stock scan carries the mangled `u_s1h-…` key where the funnel's carries the legacy header. `test_preamble_hardware.py:81` asserts only `"geecs_scalar_headers" in start`.

---

### P2 — Stock-plan scans get no `scans/ScanNNN/scan.log`, and no eager save-off for capture-owned cameras

Two side effects of the funnel's post-claim path have no counterpart in the preprocessor:

- **`scan_log`** — `scan_request_plan.py:477` wraps the run in `with scan_log(scan_number, scan_folder)`, attaching the per-scan log handler and flushing the buffered pre-claim records. The preprocessor never does. Stock-plan scans land in the data tree with no `scan.log`, which is what the repo's `/triage` skill reads. `02_preamble_preprocessor.md`'s own destination table assigns this to the preprocessor ("`scan_log` attach | **preprocessor**, around the forwarded run") — it did not travel with the code.
- **`save_control_only` save-off** — `geecs_run_wrapper` actively drives `save="off"` for capture-owned cameras before the run (`run_wrapper.py:271-280`), with the stated reason "a save flag left on out-of-band must never keep writing native files to a stale path during a toggle-off scan." `claimed_scan_metadata` took the md but left this behind, so the reuse-ledger claim that "both doors must build identical md" holds while the doors themselves diverge. `namespace_detectors` sets `_save_control_only` on the namespace devices (`plans/preamble.py:328`), so the flag exists and simply goes unacted-on.

---

### P2 — ScanInfo and md are written from the request; nothing checks them against the plan

`configure_claimed_scan` is handed `positions=prepared.spec.positions` and `shots_per_step=prepared.request.capture.shots_per_step` (`preprocessors.py:322-329`), both derived from the ScanRequest, while the geometry actually executed comes from the stock plan's own arguments. `RE(bp.count([...], num=2, md={"geecs": request(shots_per_step=5)}))` writes `ScanInfoScanNNN.ini` and a start document claiming 5 shots per step and records 2; `bp.list_scan(..., other_points, md={"geecs": {"axes": [{"positions": {...}}]}})` writes the request's positions into ScanInfo while moving to different ones. `tiled_schema`/`tiled_catalog` read `shots_per_step`, `num_points` and the positions, so the divergence reaches the portal. Either derive these from the plan's md (the planning doc's own stated preference: "keys a stock plan cannot supply … come from the stock plan's own metadata, which is **better**") or refuse the mismatch. Note `_request(shots_per_step=2)` + `bp.count(num=2)` in the new tests happens to agree, so the divergence is untested.

---

### P2 — Scan axes on the namespace door lose `confirm:` and `kind: motor`, and pseudo axes crash with `AttributeError`

`plans/preamble.py:529-531`:

```python
movables = [namespace.variable(target.device, target.variable) for target in axis_resolved]
```

versus `build_movable` (`scan_request_runner.py:1237-1264`), which dispatches on `target.confirm` (→ `CaConfirmSettable`, topology C) then `target.kind == "motor"` (→ `CaMotor`).

- A catalog variable with a `confirm:` target silently becomes whatever `namespace._movable` picked, dropping the confirming poll — `set` completes immediately and `bp.list_scan` steps to the next point before the readback has converged.
- `namespace._movable` (`namespace.py:398-412`) only produces `CaMotor` when the *DB* tolerance is `> 0`; its own comment says "the catalog's `kind: motor` opt-in arrives with the axes in **phase 3**". This PR routes scan axes through it in phase 2, so a `kind: motor` variable with a NULL/0 DB tolerance becomes a fire-and-forget `CaSettable`.
- `PseudoMovableTarget` (`scan_request_runner.py:1006-1027`) has `variable_name`/`mode`/`components` and **no** `.device`/`.variable`, so a pseudo/composite axis raises a bare `AttributeError` out of the preamble rather than a `GeecsConfigurationError` saying pseudo axes are not supported on this door. (Consistently, `_restore_movables` — the pseudo baseline restore — is also absent from `_cleanup`.)

---

### P2 — Contract files did not travel with the behaviour

- `Planning/native_bluesky/02_preamble_preprocessor.md:3` lands reading **"Status: designed, not yet built."** in the same commit that builds it.
- That doc's step **C** ("Startup installs it, then re-installs `connect_on_demand` last … proven by `test_qserver_startup.py`") is not done: `grep install_geecs_preamble` finds it only in `preprocessors.py` and the two new tests. `qserver/startup/startup.py` is untouched, so the feature is unreachable from the queueserver — the shipped door is hand-driven sessions only. Neither the doc nor the PR body says so.
- The doc's central acceptance test — *"a parity test compares the whole start document minus `uid`/`time` … the existing `_start_essence` helper already does exactly this"* (`tests/test_scan_request_plan.py:256`) — is not in the diff. Every P2 above (telemetry columns, missing axis header, the raw `geecs` blob that now rides into the start doc because `msg.kwargs` keeps it) is exactly what that test was specified to catch.
- The doc's "tests that must be re-expressed (not deleted)" list (`test_devices_and_claim_happen_inside_the_running_plan` moving to the preprocessor's call sites) is unaddressed.
- `GeecsBluesky/CLAUDE.md` is not updated. Its `plans/` block (`:137-171`) enumerates every module individually and does not list the new 597-line `plans/preamble.py`; the `preprocessors.py` entry (`:54`) still describes only `connect_on_demand`. Phase 1 (#808) *did* add its `namespace.py` entry (`:55-56`), so this is a departure from the precedent this arc set one PR ago.

---

### P3 — smaller items

- **The Gate-2 test does not pin the half of the fix the changelog leads with.** `test_the_preamble_disarms_and_stops_saving_on_the_way_out` asserts only `Trigger.Source` values from the shot-controller write log. That log (`_RecordingSetter`) never sees the save-enable `mv`, so *"the order is now `arm_single_shot` → save-on"* is unpinned — reordering `arm_single_shot` and `save_enable_plan` at `preprocessors.py:351-353` leaves the test green. It does correctly pin ARMED-not-SCAN. Record `cam.save`'s write into the same ordered log to close it.
- **`forwarded: set[int]` holds `id(msg)`** (`preprocessors.py:229`, `:359`). The forwarded `Msg` is dropped once the RE processes it, and CPython reuses freed addresses for same-size objects — so in a multi-run plan a later `open_run` can collide and have its preamble silently skipped. Hold the `Msg` objects (a list/`WeakSet` won't work on a namedtuple; a strong ref list is fine) rather than their ids.
- **CHANGELOG names the wrong scans.** `CHANGELOG.md:21-23` records the acceptance as **scan 61** and **scan 62**; the PR body says 63 and 64 and explicitly calls "Scans 56–62 … earlier test artifacts … disposable."
- **Test-helper duplication.** `Docs` in both new test files is `tests/ca_mock_helpers.py:63 DocCollector` plus a `start` property — and `DocCollector` is what the phase-1 siblings `test_connect_on_demand.py:32` and `test_namespace_hardware.py:42` import. `_sweep_points` (`test_preamble_hardware.py:62-65`) is verbatim `test_namespace_hardware.py:48-51`. The free-run pacer in `test_a_plan_without_the_geecs_key_is_untouched` re-derives `ca_mock_helpers.start_pacer:35`.
- **`namespace_detectors` is in the wrong package layer** (`plans/preamble.py:249-346`). It emits no plan messages and knows nothing about plans; it is save-set → namespace-device selection. It reaches into namespace internals to do it — imports `ACQ_TIMESTAMP_VARIABLE` *inside the loop* (`:311`), re-derives `synthesized` as a local when `namespace._SYNTHESIZED` (`namespace.py:93`) is the same set, re-derives the triggerable classification as `isinstance(device, CaGenericDetector)` when `namespace.looks_triggerable` (`:112`) owns it, and needed a brand-new public `GeecsNamespace.variable_names()` accessor added solely to serve it. As `GeecsNamespace.select(devices_config)` all four go away and `variable_names` stays internal. Phase 3's registration table will want this selection for `list_grid_scan` and would otherwise have to import a plan module to get it.
- **Dead parameter and a false docstring.** `namespace_detectors(..., free_run=…)` ends in `_ = free_run` (`:345`) while its docstring describes free-run role assignment ("later ones contributors") that the function cannot perform — the namespace builds `CaGenericDetector`, which has no `set_reference` (`devices/contributor.py:70` is a different mixin). Drop the parameter.
- **`require_strict_single_shot()` runs twice** — `preprocessors.py:311` and again inside `prepare_step_scan` (`plans/preamble.py:564`), as does the `controller is None` refusal. The preprocessor's pre-checks are already subsumed except for the free-run refusal.
- **`_worker_session` is imported as a private global from the module phase 5 deletes** (`preprocessors.py:224`, `plans/scan_request_plan.py:206`). There is a `set_plan_session` setter but no getter. With a second consumer, the worker-default session wants a neutral home now rather than as part of the funnel's removal.

---

**Summary of the shape.** The extraction itself is clean and the reuse ledger is honest about what moved. What the diff does not carry across is everything the funnel guaranteed *by construction* rather than by code: the read set equals the save set, the movable is built from the catalog entry's topology, the motors are in the header map, the shot has a retry, the scan folder gets a log, and the devices are disposable. On the funnel those were free; on this door each one is now an unwritten precondition on the caller — and the two tests that would have caught most of them (the specified start-document parity test, and a save-write in the Gate-2 ordering log) are the two the plan of record called for and the PR does not contain.

skbarber:
## Dispositions — every finding fixed, none waived

Commit `50558b2` (GeecsBluesky **0.79.0**). CI green on it: `unit-tests` 3m33s / 3m41s, `console-windows`, `pre-commit`.

The review's closing paragraph named the pattern exactly, and it is worth restating because it is the design lesson of this phase: **the funnel guaranteed by construction what this door can only enforce or refuse.** One list of devices was both the read set and the save set; each axis was built from its catalog entry's own topology; the devices were thrown away at the end of the scan. None of that survives a door where the caller brings the plan and the devices are long-lived nouns. Where the guarantee could be kept by sharing code, it is now shared; where it could not, the request is refused before anything is claimed.

### P1 — per-run device state leaked onto the namespace nouns · **fixed**

Confirmed exactly as reported, including the second half (a save set that drops a device the previous one configured). Every mixin holding per-run state now exposes `reset_run_configuration`, chained along the MRO by the new `devices/reset_support.py` — `CaGenericDetector` mixes in two of them, so the first must not shadow the second. `GeecsNamespace.reset_run_configuration()` clears **every** device, and the preamble calls it both before it configures a run and in `_cleanup`, so a crash between the two cannot leak either. Pinned at the document level: a GEECS scan, then a plain `bp.count`, asserting no `nonscalar_save_path` column.

Fixing this surfaced a second bug of the same kind. The detector constructors add the `acq_timestamp` s-file header when their save flags are fixed at build time; a namespace device gets its mode per-run, so the header was silently absent — the join key that ties a scan's saved images back to its rows. It now follows `configure_saving_mode`.

### P1 — nothing checked that the plan reads the save set · **fixed (refused)**

Refused before the claim. Every stock `bluesky.plans` verb records `md["detectors"]`, so the check needs nothing new. The message names the missing devices and points at `resolve_save_set`.

### P1 — the fire dropped the refire and device-down recovery · **fixed (deduplicated)**

The right call, and the redundancy framing was the useful part: the seam was implemented twice and the second copy was weaker. The retry loop, the bounded refire and the gateway-liveness gating moved into `plans/single_shot.fire_and_await_shot`, which `geecs_single_shot` now calls for its own shot. The preprocessor passes the plan's pending `wait` as `pending_wait=`, so the first attempt fires into the gap `trigger_and_read` already opened and a refire falls back to a fresh group (which also re-baselines). One implementation, two callers. Tested by swallowing the first fire: the run recovers and records exactly one row, where before it aborted.

### P2 — `install_geecs_preamble` rewrote `connect_on_demand`'s kwargs · **fixed**

Carries the existing partial's `keywords` forward. The test now asserts `mock` and `timeout` survive, not just the function identity.

### P2 — telemetry prepared, announced, never read · **fixed**

Kept as `primary`-stream columns, per the correction the review cited. The preprocessor inserts the reads between the plan's last `read` and its `save`: one row, one stream, no schema change. Re-entrancy is a bounded flag rather than a growing set, since this fires once per recorded shot.

### P2 — `geecs_scalar_headers` lost the scan axis · **fixed**

The motors join the header map. The namespace also stopped contributing headers for settable children it does not read — those were phantom column names in front of the exporter, the mirror of the same bug.

### P2 — no `scan.log`, no eager save-off · **fixed**

The preamble attaches the per-scan log handler around the forwarded run. The save-off is now `run_wrapper.save_control_only_off_plan`, called by both doors.

### P2 — ScanInfo written from the request, unchecked against the plan · **fixed (refused)**

Both halves. The recorded-shot count is compared against the plan's `num_points` before the claim, with a message that says a stock step scan records one shot per point so `shots_per_step > 1` needs the per-step hook. And because equal counts hide unequal values, every commanded position is checked against the declared list as the plan moves.

### P2 — axis topologies lost `confirm:` / `kind: motor`, pseudo crashed · **fixed (refused)**

Refused, not degraded. `build_movable`'s three topologies are exactly what the namespace does not build yet, and handing back a fire-and-forget setpoint would take every row before the readback converged with nothing in the data saying so. Pseudo targets get a real error instead of an `AttributeError`; `kind: motor` with no positive DB tolerance names the DB row to fix.

### P2 — contract files did not travel · **fixed**

The planning doc no longer says "not yet built" and carries a section on the three places the shipped code differs from the design and why. Step C is done: the queueserver installs the preamble, and the startup test pins the order and the namespace. `CLAUDE.md` gains `plans/preamble.py`, `plan_session.py` and the two new preprocessor entries.

**The specified parity test now exists, and it earned the review's point about it immediately.** One request through both doors: every GEECS-owned start-document key compared exactly, identical `ScanInfo`, and the event columns compared as an exact set. It found four things nothing else caught — the missing `acq_timestamp` header, the phantom settable headers, a raw `geecs` blob riding into every start document (predicted in the review), and six execution keys `tiled_catalog`/`tiled_schema` read that only the funnel's inner plan emitted. Those six now come from one `plans/step_scan.geecs_execution_md`.

### P3 · all fixed

- The Gate-2 test records the save writes into the **same** ordered log as the shot-control writes and asserts positions, so swapping `arm_single_shot` and `save_enable_plan` now fails it.
- `forwarded` holds the `Msg` objects, not their ids.
- CHANGELOG corrected to scans 63 and 64.
- `Docs` and `_sweep_points` are gone; `tests/ca_mock_helpers` gained the `start` property and `sweep_points`, and the free-run pacer copy went with them.
- `namespace_detectors` became `GeecsNamespace.select`. As a method it uses the namespace's own `_SYNTHESIZED` and its triggerable classification instead of re-deriving them, the loop-body import is gone, and `variable_names` went back to private.
- The dead `free_run` parameter and its false docstring are gone.
- The duplicated `require_strict_single_shot` / controller-present checks are gone; only the free-run refusal, which is this door's own, remains.
- The worker session moved to `geecs_bluesky/plan_session.py` with a `get_plan_session`, re-exported from the funnel module so nothing else changed.

### One thing deliberately left

Stock plans still emit no `ScanContext` columns (`bin_number`, `scan_event_index`, `shot_index_in_bin`). Those come from the per-step function in phase 3, which is also when `ScanContext` retires; adding them here would mean inferring step boundaries from `set` messages, which is the custom-code-over-native-seam move this whole effort exists to avoid. The parity test asserts them as the **only** column difference, so the gap is pinned and cannot widen unnoticed.

### Test record

CI is the authority here and it is green. Locally the phase-2 file is 19/19 (11 new). The whole-suite run on this Mac deadlocks in a `zmq.Context.__del__` → `term()` during garbage collection — an uninterruptible C-level block, which is why `--timeout` cannot fire on it. It is a pre-existing local teardown hazard, not this change: the same suite completes on CI in 3m33s, and the affected files pass in isolation and together.

skbarber:
### Correction to the test record in the dispositions

My disposition comment says the whole-suite run deadlocks locally in a `zmq.Context.__del__` → `term()`. **That is wrong, and I am withdrawing it.** The full suite passes:

```
783 passed, 6 deselected, 7 warnings in 317.93s (0:05:17)
```

What actually happened: pytest block-buffers its progress output when stdout is a file, so I was reading a stale 4 KB buffer, concluding a healthy run had stalled, and killing it. The one genuine stall was a run competing with two other pytest processes I had left going on the same machine; the `zmq` teardown stack came from that run's timeout dump and is not a property of this branch.

The PR body has been corrected to the real number. Nothing about the code findings or their dispositions changes — CI was green independently at 3m33s and 3m41s.

skbarber:
## Holding this PR — do not merge pending a direction decision

Not a defect report. The code here works: hardware-accepted as scans 63 and 64, CI green, every review finding dispositioned.

The two review passes surfaced something the code cannot fix. Twenty-one findings, and **twelve of them have one cause**: a scan is described twice, once by the plan as detectors and points, and once by the ScanRequest as a save set and axes. This preprocessor's real job had become reconciling the two, and each of those twelve is a reconciliation rule that was missing. The parity test added here exists to detect divergence between the two doors, which is a useful test and also an admission that there are two doors describing one thing.

@skbarber's read, which is the right one: a save set is a read set, and Bluesky already calls that the `detectors` argument. A scan variable is a settable with an alias, which the device namespace already provides. Given that, this layer should not exist rather than be corrected.

The proposed direction is in **#810** (`Planning/native_bluesky/03_clean_room_rebuild.md`), with the full finding classification and the concept-by-concept mapping. It recommends **#806 before any rebuild**, and it recommends not merging this.

**Status of the two open P1s from the verification pass** (the refire recursing through the mutator and firing extra shots, and the residual per-run state leak on the pre-claim failure path): **not fixed, and deliberately so.** Nothing here is deployed — the worker is on master — so neither is a live hazard, and fixing them is only worthwhile if this ships.

Worth salvaging by hand if the rebuild goes ahead: the document-parity test, `GeecsNamespace.select`'s role assertions, `plan_session.py`, and `fire_and_await_shot` if per-shot triggering survives at all.

Leaving this open as the evidence. Close it whenever the direction is settled.

skbarber:
Closing without merge, as decided in the plan of record (`Planning/native_bluesky/03_clean_room_rebuild.md` §8, "On #809"): this branch is the evidence behind §3 (the scan described twice; per-run device state set from outside the device), and its two open P1s need no fix if it does not ship. What survived was rebuilt natively on `feature/native-bluesky-plans`: #808 (device namespace), #811 (phase 0, one camera as a StandardDetector + the strict take_reading), #816 (the deletions), #821 (the plan layer: registration table, claim preprocessor + path provider, callbacks, presets), #822 (hardware acceptance). Not reopening; the branch stays for reference.

PR #810, opened 2026-09-10
Planning: clean-room rebuild direction for native Bluesky (#807)

skbarber:
Docs only. A handoff written at the end of the #809 session, proposing a different starting point for #807. It supersedes `00_overview.md`'s phase plan; nothing is deleted or changed in code.

## Why

#809 moved the GEECS scan preamble into a RunEngine preprocessor so that a stock `bluesky.plans` verb could run a full GEECS scan. It works on hardware (scans 63 and 64) and CI is green. It was then reviewed adversarially twice: sixteen findings, then five more on the fixes.

Twelve of the twenty-one have a single cause. **A scan is described twice** — once by the plan, as detectors and points, and once by the ScanRequest, as a save set and an axis list. The preprocessor's real job had quietly become reconciling the two descriptions, and each of those twelve findings is a missing reconciliation rule: the plan need not read what the save set saves, the request's shot count need not match the plan's, the request's positions need not match the plan's, telemetry is advertised but never read, long-lived devices keep the previous run's saving state, axes lose their catalog topology. Fixing them individually produces a reconciliation engine, and the parity test #809 added exists precisely to detect divergence between the two doors, which is a useful test and also an admission.

Sam's response, which this document records:

> save sets are really just readbacks […] and the scan variables just the settables with aliases. If you encounter a weird hack or workaround to accommodate save sets we should think, "what is a save set doing and how does Bluesky solve the same problem?"

and then the framing for the rebuild:

> start with the assumption that we *only* have our CA gateway, PVA gateway and bluesky/ophyd devices […] don't try to hold on to *anything* if it doesn't slot in completely cleanly

with the DB as source of truth and the s-files kept.

## What the document contains

- **The evidence** (§3): all twenty-one findings classified by cause, twelve as artifacts of the duplication and nine as real defects independent of the architecture. This is the part worth reading; a fresh session given only a target architecture will re-litigate the design.
- **The target** (§4) and **the mapping** (§5): every GEECS concept to its native replacement. Save set to the `detectors` argument. Trigger profile to a `FlyerController`. Strict and free-run to one fly implementation with different counts. Gate-2 windowing to the detector's open/close. `acq_timestamp` joins to StreamDatum indices. Scan folder to a `PathProvider`. ScanInfo to a start-document callback.
- **What stays GEECS** (§6): the DB roster, day-scoped scan numbering, the s-file format, ScanInfo, and PV naming. None of them is in the scan path. `ScanRequest` survives as the client-side record of intent the GUI needs.
- **Verified versus assumed** (§7): the API surface checked against the installed ophyd-async **0.19.3** and bluesky **1.15.0**, including two names that have moved (`FlyerController`, not `TriggerLogic`; no `DetectorWriter` under that name). Everything resting on #806 is marked assumed, because the fly design depends on the PVA gateway delivering indexed edge-triggered frames.
- **Sequencing** (§8): recommends **#806 before the rebuild**, which reverses the current ordering, and recommends **not merging #809**.

## What this implies for the open work

- **#808 survives unconditionally.** The device namespace is exactly the foundation the target needs, and it is already merged.
- **#809 should not merge.** It is scaffolding between two descriptions of one scan. Its two open P1 defects need no fix if it does not ship, and nothing from it is deployed — the worker is on master. Salvageable by hand: the document-parity test, `GeecsNamespace.select`'s role assertions, `plan_session.py`.
- **#806 moves to the front.**

The one thing the document flags as needing an answer before any deletion lands: operators scan daily through the funnel and the Console, so the rebuild has to run beside them rather than replace them in place.

### DIFF
--- a/Planning/native_bluesky/00_overview.md
+++ b/Planning/native_bluesky/00_overview.md
@@ -1,5 +1,10 @@
 # GeecsBluesky as a thin layer on stock Bluesky
 
+> **Superseded 2026-09-09.** The phase plan below was replaced by
+> `03_clean_room_rebuild.md` (the document of record; its §8 has the
+> phases, its §11–12 the hardware facts). Kept for the rule in the next
+> section and the history; do not plan against it.
+
 **Plan of record: GEECS-Plugins issue #807** (companion: #806, the image
 writer). This directory holds the per-phase design notes; the issue holds
 the decision log and the running status. Read the issue first.
--- a/Planning/native_bluesky/03_clean_room_rebuild.md
+++ b/Planning/native_bluesky/03_clean_room_rebuild.md
@@ -60,7 +60,7 @@ is right in five years, not the one that is reachable in small steps.
 
 | thing | state |
 |---|---|
-| `feature/native-bluesky-plans` | integration branch off master; **#808 merged** into it 2026-09-09 (device namespace, phase 1) |
+| `feature/native-bluesky-plans` | integration branch off master; **#808 merged** into it 2026-09-09 (device namespace, phase 1); **phase 0 hardware-accepted 2026-09-09** (`phase/00-one-camera-detector`, Scan 065 — `04_phase0_measurements.md` M2), PR pending |
 | #809 `phase/02-preamble-preprocessor` | **OPEN, on hold, will not merge** (13 commits, GeecsBluesky 0.79.0, CI green). The evidence behind §3; close with a pointer here once this amendment lands (§8) |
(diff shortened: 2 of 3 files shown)


skbarber:
**Codex P3 — the header said "phase 0 is next" while the table said phase 0 was accepted.** Fixed in 3ad43251: the status line now reads current (phase 0 hardware-accepted, #811 awaiting merge; phase-1 decisions recorded in §10; #812 first). The staleness rule in the same header is the reason this should not recur: direction or status changes edit §2/§8/§10 in the same PR.

PR #811, opened 2026-09-10
GeecsBluesky 0.78.0: phase 0 — one camera as a stock StandardDetector, ShotControl device, strict take_reading (#807)

skbarber:
## What

Phase 0 of the native-Bluesky rebuild (#807; plan of record `Planning/native_bluesky/03_clean_room_rebuild.md` §8, on #810): **one GEECS camera as a stock ophyd-async `StandardDetector`**, the trigger box as a device, and strict single-shot acquisition as the stock `take_reading` hook — so `bp.count` and every N-d scan plan from `bluesky.plans` run a strict GEECS scan with no preamble and no preprocessor beyond `connect_on_demand`.

- `devices/detector.py` — `GeecsDetector`, composed of the three 0.19.3 logics:
  - `GeecsTriggerLogic`: external edges only (the DG645 fires, LabVIEW acquires; nothing to program). Its one config signal is the calibrated **drain offset** (§11.4), which `get_deadtime` returns.
  - `GeecsAcquireLogic`: a shot **is** `acq_timestamp` advancing. The baseline is taken **synchronously in `trigger()`** so the plan's fire — the very next message — can never land in a blind window. The wait lives in `wait_for_idle` (a per-event data provider has no count for `_wait_for_index` to watch).
  - `ScalarsDataLogic`: the device's own DB-subscribed scalars (+ the stamp) as event columns.
  - `LvNativeFileDataLogic`: LabVIEW native saving driven from a `PathProvider` — `localsavingpath` + `save=on` at `prepare`, `save=off` at `stage` (eager, clears a stale flag) and `unstage`. The device directory is created with `mkdir(exist_ok=True)` **only inside an existing scan folder**; a missing parent raises. There is no write-complete readback (§10.1), so the provider is a per-event reading (the directory); files join to rows by stamp.
  - A plain `bp.count([cam])` is refused at prepare (`INTERNAL not supported`): a GEECS camera cannot self-trigger.
- `devices/shot_control.py` — `ShotControl`: `Movable` over the profile's named states (ordered writes, composed over the existing `ShotController.from_writes`, not copied), `Pausable` (no-op in ARMED — the plan just stops firing; SCAN → OFF and `resume` → SCAN, §10.3), the standing state as a config signal.
- `plans/strict.py` — `geecs_take_reading`: `bps.trigger_and_read` with the one GEECS difference (the SINGLESHOT fire between the triggers and the wait) plus the bounded refire; `geecs_per_shot` / `geecs_per_step` bind it into the stock `one_shot` / `one_nd_step` hooks.
- `plans/single_shot.py` — the arm → fire → await → refire seam is now `fire_and_await_shot`, called by `geecs_single_shot` (the funnel) **and** `geecs_take_reading`: one implementation, two callers (#809's review finding, applied).

## Why

#809's twenty-one findings had two structural causes (§3): a scan described twice, and per-run device state set from *outside* the device. This PR removes the second at the device layer: every per-run fact (save path, saving on/off, the shot wait) is set through the detector's own `stage → prepare → trigger → unstage`. The first is the plan layer's job (phase 1). Sequencing per Sam's decision 2026-09-09 (§8, option 1½): this is the seam #806 swaps one logic into, and it lets the plan-layer rebuild and #806 proceed in parallel.

## Scope

One concern. LOC: `detector.py` 463, `shot_control.py` 155, `strict.py` 117, `single_shot.py` +67/−19 (extraction), tests 909, CHANGELOG 36.

**Judgment calls, cheap to veto:**
- `GeecsDetector.stage()` stages the scalar signals so reads come from the monitor cache (the 0.7 s/row regression in `GeecsBluesky/CLAUDE.md`); ophyd-async's `StandardDetector` does not do this for children by itself.
- The refire stays a `GeecsDeviceDownError` on a confirmed-dead device (abort), not `bps.pause()` as §4.B first said — a pause inside `take_reading` replays the stashed trigger/fire messages on resume (rewind), which would double-fire. Doc amended to match.
- `default_trigger_info()` returns the edge-triggered `TriggerInfo` (truthful hardware state), so `OPHYD_ASYNC_PRESERVE_DETECTOR_STATE=YES` environments work too; `take_reading` prepares explicitly regardless.
- Not touched: the namespace still builds `CaGenericDetector` nouns; switching it to `GeecsDetector` (and deleting `CaGenericDetector`, `CaTriggerable`, `nonscalar_save.py`, `shot_id.py`) is the plan-layer PR, with the funnel that uses them.

**Facility literals:** `tests/test_phase0_hardware.py` defaults (`Undulator`, `UC_Amp4_IR_input`, `HTU-NoGas`, `U_S1H:Current`) are env-overridable test parameters, same shape as `test_namespace_hardware.py`; none in package code.

## Tests

- `tests/test_geecs_detector.py` (10): children/headers; edge-only trigger + livetime refusal; the shot wait; **the synchronous-baseline race** (fire before the trigger task's first step); timeout attribution; stale-update drain; the saving lifecycle (stage → off, prepare → on, unstage → off); **the scan-folder invariant** (missing parent raises, nothing created); drain offset in `read_configuration`; the GEECS error on `FailedStatus.__cause__`.
- `tests/test_shot_control_device.py` (8): ordered writes; enum/lowercase; SINGLESHOT momentary; undefined state refused; standing state in configuration; pause no-op in ARMED; pause → OFF / resume → SCAN; **the RE pauses the box it `set`**.
- `tests/test_strict_plans.py` (8): stock `count` fires once per shot; stock `scan` moves then fires; two cameras share one fire; **dropped frame → whole-event refire, rows stay one shot**; refires bounded; dead device not refired; saving brackets the run; plain `count` refused.
- Review-driven additions (79f24fba): `test_shot_control_device.py` +5 (the stop-from-paused leak, STANDBY quiesce, no-OFF profile never raises, unknown state names, one standing-state field), `test_geecs_detector.py` +1 (`native_save` clears a stale flag without a path).
- Affected suites after the review fixes (new files + `test_ca_devices`, `test_pause_semantics`, `test_scan_request_runner`, `test_shot_control*`, `test_single_shot_plan`, `test_connect_on_demand`, `test_preflight_connected`, `test_free_run_plan`): **213 passed**; plan-level (`test_namespace`, `test_scan_request_plan`, `test_grid_and_action_plans`): **70 passed**. Lint clean.
- Full GeecsBluesky suite on this Mac (`--durations` run): **792 passed, 1 timed out** — `tests/test_tiled_integration.py::test_subscribe_tiled_reachable_server_still_subscribes` at 180.59 s; the next-slowest test is 9 s. The same suite on the base branch also times out one test (a different one, 764 passed), so the slowness pre-dates this PR and is a test-hygiene issue of its own — filed separately. CI runs the suite green in ~3.5 min.

## Hardware verification — done twice (Scan 065 of 26_0909 = M2; re-run on the review-fixed code, Scan 001 of 26_0910 = M3; `04_phase0_measurements.md`)

`tests/test_phase0_hardware.py` on the worker host: `GeecsDetector("UC_Amp4_IR_input")` with 8 scalars + native saving into a claimed scan folder, `ShotControl.from_profile("HTU-NoGas")`, `bps.mv(shot_control, "ARMED")` → stock `bp.count(3, per_shot=geecs_per_shot)` → stock `bp.list_scan(U_S1H.Current, [-1, -0.5, 0, 0.5, 1], per_step=geecs_per_step)` → STANDBY + restore.

- 3 + 5 events, `success`; every row carries the stamp, the 8 scalars, `save_path`, the magnet readback within tolerance (−0.9998 … 0.9999 A). Setpoint restored (8e-05 A); STANDBY restored; `save` reads `off`.
- `Scan065/UC_Amp4_IR_input/`: exactly 8 PNGs, one per row, named by the row's stamp; still 8 two minutes later (no orphans in STANDBY). `localsavingpath` written as the Windows share path via the config.ini mapping.
- Cadence: **1 Hz strict** on the count (0.95 / 0.99 s); 2.0 s/step on the scan — move + the ~200 ms fire put exceed the ~550 ms budget measured in M1 (the single shot fires on the *next* external edge), so the shot lands on the next-next edge. Recoverable in the plan layer; recorded in §7.
- M3 (79f24fba): `1 passed in 27.7 s` end to end — same shape as M2 (3 + 5 events, 8 files, 1 Hz count, 2.0 s/step, setpoint restored).

## Adversarial review

The reviewer's full report and the dispositions are comments on this PR. The original reviewer was cut off by a rate limit before confirming the fixes; a fresh-context verifier given only the findings and the fix commits did the confirm/deny (the `/land` substitution rule), noted in the disposition comment.

### DIFF
--- a/GeecsBluesky/CHANGELOG.md
+++ b/GeecsBluesky/CHANGELOG.md
@@ -39,6 +39,19 @@ Format follows [Keep a Changelog](https://keepachangelog.com/en/1.1.0/).
 - `plans/single_shot.py` — the arm → fire → await → refire seam is
   `fire_and_await_shot`, called by `geecs_single_shot` (the funnel) and by
   `geecs_take_reading`: one implementation, two callers.
+- `devices/ca/triggerable.py` — `CaAcqTimestampReadable` / `CaTriggerable`
+  compose `GeecsAcquireLogic` for the stamp monitor, the synchronous
+  baseline and the shot wait instead of carrying their own copy (review of
+  #811); `_last_acq` / `_shot_queue` / `_monitoring` / `_trigger_timeout`
+  remain as views for the funnel-era callers.
+- `trigger_writes_from_profile` (TriggerProfile → `ShotControlWrites`) now
+  lives in `devices/shot_control.py` next to the device that consumes it;
+  `scan_request_runner` imports it from there.  `QUIESCE_FROM` (the
+  standing states a pause must quiesce from) has its one home in
+  `models/shot_control.py`; the device and `plans/pause_semantics.py` both
+  import it.
+- `ShotController._record_state` → `record_state` (public; the device
+  records through it so `last_state` is the one standing-state field).
 
 ## [0.77.0] - 2026-09-09
 
--- a/GeecsBluesky/CHANGELOG.md
+++ b/GeecsBluesky/CHANGELOG.md
@@ -41,6 +41,10 @@ Format follows [Keep a Changelog](https://keepachangelog.com/en/1.1.0/).
 - `plans/single_shot.py` — the arm → fire → await → refire seam is
   `fire_and_await_shot`, called by `geecs_single_shot` (the funnel) and by
   `geecs_take_reading`: one implementation, two callers.
+- The refire is gated on the failed status's cause being a detector's
+  `GeecsTriggerTimeoutError` (a dropped frame); a failed **fire** (the
+  SINGLESHOT put refused) or any other failed status re-raises untouched,
+  so a refire can never issue an extra physical shot (Codex review of #811).
 - `devices/ca/triggerable.py` — `CaAcqTimestampReadable` / `CaTriggerable`
   compose `GeecsAcquireLogic` for the stamp monitor, the synchronous
   baseline and the shot wait instead of carrying their own copy (review of
--- a/GeecsBluesky/CHANGELOG.md
+++ b/GeecsBluesky/CHANGELOG.md
@@ -4,6 +4,42 @@ All notable changes to `geecs-bluesky` are documented here.
 
 Format follows [Keep a Changelog](https://keepachangelog.com/en/1.1.0/).
 
+## [0.78.0] - 2026-09-09
+
+### Added
+
+- **Phase 0 of the native-Bluesky rebuild (#807, plan of record
+  `Planning/native_bluesky/03_clean_room_rebuild.md`):** a GEECS acquirer as
+  a stock ophyd-async `StandardDetector`.
+  - `devices/detector.py` — `GeecsDetector` composed of the three 0.19
+    logics: `GeecsTriggerLogic` (external edges only; the calibrated drain
+    offset is its one config signal and `get_deadtime`), `GeecsAcquireLogic`
+    (a shot is `acq_timestamp` advancing; the baseline is taken
+    synchronously in `trigger()` so the plan's fire can never land in a
+    blind window), `ScalarsDataLogic` (the device's own scalars as event
+    columns) and `LvNativeFileDataLogic` (LabVIEW native saving driven from
+    a `PathProvider`: `save=on` at prepare, `save=off` at stage and
+    unstage; the device directory is created only inside an existing scan
+    folder — a missing parent raises, never a `mkdir`).  A plain
+    `bp.count([cam])` is refused at prepare: a GEECS camera cannot
+    self-trigger.
+  - `devices/shot_control.py` — `ShotControl`, the trigger box as a
+    `Movable` over the profile's named states (ordered writes via the
+    existing `ShotController.from_writes`), `Pausable` (a no-op in ARMED;
+    SCAN → OFF and back), with the standing state as a config signal.
+  - `plans/strict.py` — `geecs_take_reading`, `bps.trigger_and_read` with
+    the one GEECS difference (the `SINGLESHOT` fire between the triggers
+    and the wait) plus the bounded refire; `geecs_per_shot` /
+    `geecs_per_step` bind it into the stock `one_shot` / `one_nd_step`
+    hooks, so `bp.count` and every N-d scan plan run strict GEECS scans
+    unchanged.
+
+### Changed
+
+- `plans/single_shot.py` — the arm → fire → await → refire seam is
+  `fire_and_await_shot`, called by `geecs_single_shot` (the funnel) and by
+  `geecs_take_reading`: one implementation, two callers.
+
 ## [0.77.0] - 2026-09-09
 
 ### Added
--- a/GeecsBluesky/CHANGELOG.md
+++ b/GeecsBluesky/CHANGELOG.md
@@ -21,8 +21,10 @@ Format follows [Keep a Changelog](https://keepachangelog.com/en/1.1.0/).
     a `PathProvider`: `save=on` at prepare, `save=off` at stage and
     unstage; the device directory is created only inside an existing scan
     folder — a missing parent raises, never a `mkdir`).  A plain
-    `bp.count([cam])` is refused at prepare: a GEECS camera cannot
-    self-trigger.
+    `bp.count([cam])` is refused at prepare — a GEECS camera cannot
+    self-trigger — unless `OPHYD_ASYNC_PRESERVE_DETECTOR_STATE=YES`, where
+    the implicit prepare takes the edge-triggered default and the shot
+    times out waiting for a fire nobody sends.
   - `devices/shot_control.py` — `ShotControl`, the trigger box as a
     `Movable` over the profile's named states (ordered writes via the
     existing `ShotController.from_writes`), `Pausable` (a no-op in ARMED;
--- a/GeecsBluesky/geecs_bluesky/devices/ca/triggerable.py
+++ b/GeecsBluesky/geecs_bluesky/devices/ca/triggerable.py
@@ -1,24 +1,15 @@
 """CA acq_timestamp-monitored readables: the shot-aware CA device bases.
 
 The GEECS shot signal is the device's ``acq_timestamp`` advancing once per
-shot.  A **persistent CA monitor** on that PV (started at ``connect()``,
-stopped by ``disconnect()``) feeds a local cache and a bounded drop-oldest
-event queue:
-
-* :class:`CaAcqTimestampReadable` — readable signals plus the monitor/cache
-  (``_last_acq``); free-run *contributors* build on this (no blocking trigger).
-* :class:`CaTriggerable` — adds ``trigger()``, which blocks until the queue
-  delivers a value different from the baseline.
-
-The stale-frame drain and baseline capture happen **synchronously inside**
-``trigger()`` so an immediately-fired shot cannot be missed — see
-:meth:`CaTriggerable.trigger`.  Design rationale: ``GeecsBluesky/CLAUDE.md``
-(Device Layer).
+acquisition.  The shot semantics — the persistent monitor, the synchronous
+baseline, the wait with the attributable timeout — live in **one** place,
+:class:`~geecs_bluesky.devices.detector.GeecsAcquireLogic`; these classes
+compose it so the funnel's readables keep working until the plan layer
+deletes them (#807 phase 1), and that deletion is purely subtractive.
 """
 
 from __future__ import annotations
 
-import asyncio
 import logging
 from collections.abc import Mapping
 from typing import Any
@@ -27,7 +18,7 @@ from ophyd_async.core import AsyncStatus, StandardReadable
 from ophyd_async.epics.core import epics_signal_r
 
 from geecs_bluesky.devices.ca._pv import ca_pv
-from geecs_bluesky.exceptions import GeecsTriggerTimeoutError
+from geecs_bluesky.devices.detector import GeecsAcquireLogic
 from geecs_bluesky.utils import safe_name
 
 logger = logging.getLogger(__name__)
@@ -37,10 +28,9 @@ class CaAcqTimestampReadable(StandardReadable):
     """Readable GEECS device over gateway PVs with a persistent shot monitor.
 
     One ``epics_signal_r`` child is created per data variable, plus an
-    ``acq_timestamp`` child that carries the shot stamp.  A monitor
-    subscription (started at ``connect()``, stopped by ``disconnect()``) keeps
-    ``_last_acq`` (latest value) and ``_shot_queue`` (bounded update stream)
-    current.
+    ``acq_timestamp`` child that carries the shot stamp; the stamp monitor
+    (started at ``connect()``, stopped by ``disconnect()``) is the composed
+    :class:`GeecsAcquireLogic`.
 
     A non-readable ``connected_status`` child reads the gateway's per-device
     liveness PV (``[Experiment:]Device:CONNECTED``): the authoritative
@@ -74,18 +64,12 @@ class CaAcqTimestampReadable(StandardReadable):
     _acq_timestamp_variable : str
         GEECS variable that advances per shot.  Default ``"acq_timestamp"``.
     _shot_queue_maxsize : int
-        Bound on the shot-update queue (default 128 — the worst case is
-        rep-rate × trigger-timeout updates between baseline and the awaited
-        get, i.e. 15 at the 5 Hz system limit, with a wide margin so the
-        bound never needs revisiting below ~40 Hz).  Only ``trigger()``
-        drains the queue, so an unbounded queue would grow one float per
-        machine shot on idle devices; overflow is drop-oldest and
-        correctness-preserving (any surviving post-baseline update passes
-        the ``!= t0`` shot test).
+        Bound on the shot-update queue (see :class:`GeecsAcquireLogic`).
     """
 
     _acq_timestamp_variable: str = "acq_timestamp"
     _shot_queue_maxsize: int = 128
+    _trigger_timeout_default: float = 3.0
 
     def __init__(
         self,
@@ -126,13 +110,12 @@ class CaAcqTimestampReadable(StandardReadable):
             str, ca_pv(experiment, device, "CONNECTED")
         )
         super().__init__(name=name)
-        # Persistent-monitor state (populated by _on_acq_timestamp): the latest
-        # seen value, and a bounded queue of updates trigger() waits on.
-        self._last_acq: float | None = None
-        self._shot_queue: asyncio.Queue[float] = asyncio.Queue(
-            maxsize=self._shot_queue_maxsize
+        self._acquire = GeecsAcquireLogic(
+            self.acq_timestamp,
+            device,
+            shot_timeout=self._trigger_timeout_default,
+            queue_maxsize=self._shot_queue_maxsize,
         )
-        self._monitoring = False
 
     async def connect(
         self,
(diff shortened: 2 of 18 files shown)


skbarber:
## Adversarial review — PR #811 (`phase/00-one-camera-detector` → `feature/native-bluesky-plans`)

Reviewed the full diff against the repo (root and `GeecsBluesky/CLAUDE.md`, `plans/pause_semantics.py`, `devices/ca/triggerable.py`, `devices/nonscalar_save.py`, `plans/run_wrapper.py`, `GEECS-Data-Utils/geecs_data_utils/tiled_schema.py`/`tiled_export.py`, `EVENT_SCHEMA.md`) and the installed sources (`bluesky/run_engine.py` 1.15.0, `ophyd_async/core/_detector.py`, `_signal.py`, `_device.py`, `_path_providers.py` 0.19.3). Findings ranked by severity; each includes what I tried to refute it with.

---

### P1 — `ShotControl._resume_to` survives a stop/abort/halt-from-paused and drives **SCAN** on a later, unrelated resume

`GeecsBluesky/geecs_bluesky/devices/shot_control.py:137-149`

`pause()` records `_resume_to = SCAN`; the only thing that clears it is `resume()`. The RunEngine notifies `obj.resume()` **only** from `RunEngine.resume()` (`run_engine.py:1020-1022`) and the `Msg('resume')` handler (`:2433-2435`). `stop()`/`abort()`/`halt()` from the paused state set `_exception` and go through `__interrupter_helper → _resume_task()` (`:1415-1437`, `:1052`), which never touches Pausables. So:

1. Gated run: `mv(sc, "SCAN")` → operator pauses → `pause()` drives OFF, `_resume_to = SCAN`.
2. Operator `RE.stop()`. `finalize` drives STANDBY (`_standing = "STANDBY"`); `_resume_to` is still `SCAN`.
3. Next run on the same long-lived device (the hardware test and the namespace both build it once per session): `mv(sc, "ARMED")`, strict shots. Operator pauses — `pause()` is a no-op (standing is ARMED) — then `RE.resume()` → `resume()` sees `_resume_to is not None` → **drives SCAN**: full-amplitude free-running edges (gas jet on, per profile semantics) in the middle of a strict scan. Every subsequent `wait_for_idle` returns on the first free-run stamp, so rows are no longer the plan's fired shot and nothing errors.

Refutation attempts: looked for anything resetting `_resume_to` — `set()`/`_drive()` do not; `bp.count` stages only detectors, so no stage/unstage hook fires on the box; `_stop_movable_objects` only calls `stop()` on `Stoppable`s (`StandardReadable` is not one). The existing `ShotControlPauseQuiescer` has the same `_reassert` field but is constructed per plan (`plans/orchestration.py:232`), so its state dies with the run — `ShotControl` has no such scoping. `test_shot_control_device.py::test_pause_stops_edges_in_gated_mode` only exercises pause → resume on one device in one sequence.

Fix: clear `_resume_to` in `_drive()` whenever a standing state is set by the plan (a `set` supersedes the pause bookkeeping), and add a test: SCAN → pause → `RE.stop()` → ARMED → pause → resume ⇒ standing stays ARMED.

---

### P2 — A second pause owner, with rules that diverge from the one `CLAUDE.md` names as sole

`devices/shot_control.py:137-149` vs `plans/pause_semantics.py:80-215` (`ShotControlPauseQuiescer`)

`GeecsBluesky/CLAUDE.md` ("The one pause owner") and the quiescer's docstring say the quiescer "is now the sole pause-quiesce mechanism". The new device reimplements pause/resume with three differences:

- **Quiesce set.** Quiescer: `QUIESCE_FROM = {SCAN, STANDBY}` — STANDBY keeps passing external edges, so a pause there saves orphan frames for its whole duration (the Gate-2 failure mode the quiescer docstring cites). `ShotControl.pause()` quiesces from **SCAN only** (`:139`). With saving on and the box in STANDBY (the between-steps state any gated plan will use), a pause leaks frames.
- **Exception policy.** The quiescer guards `defines_state(OFF)` and swallows/logs every failure because "an exception out of the pause/resume notification would take down the run" (pinned by `test_pause_semantics.py::test_profile_without_off_warns_and_writes_nothing`). `ShotControl.pause()` calls `_drive(OFF)`, which raises `GeecsConfigurationError` when the profile defines no OFF writes. In `RunEngine._run` that call sits at `:1530-1534` inside the outer `try` but outside the per-message handler; only `NoReplayAllowed` is caught, so the error lands in `except Exception` at `:1758` → **"Run aborted"**: a profile without OFF turns an operator *pause* into an abort. A raising `resume()` is worse: `RE.resume()` has already `_rewind()`-ed and pushed the plan (`:1013-1016`) when `fut.result()` raises at `:1022`.
- **Source of truth for the standing state.** Two fields (`ShotControl._standing`, `ShotController.last_state`) can disagree: `ShotControl._drive` writes through `state_setters` directly, never `_record_state`, so `last_state` on the composed controller stays `None`. Anything that later consults the controller (the quiescer, `free_run_step_scan.py:284`) sees a box that was "never driven".

Refutation attempts: checked whether phase 0's strict flow can reach STANDBY-with-save-on inside a run — it cannot (ARMED for the whole run, STANDBY only in cleanup after unstage), so the first bullet is latent until the gated plan lands; the second is live today for any profile lacking OFF. Checked whether the RE's `Pausable` dispatch makes both objects fire if both are present — it does (`isinstance(obj, Pausable)` over `_objs_seen`), so running the new device under the existing funnel would double-quiesce.

Suggestion: keep `ShotControl` as the Pausable the RE sees (that part of the design is right — being the `set` target is enough), but make it *compose* the quiescer's rules rather than re-derive them: `QUIESCE_FROM`, the `defines_state(OFF)` guard, log-never-raise, and `self._controller.last_state` as the single standing-state field (drop `_standing`/`state` duplication or derive the soft signal from `last_state`).

---

### P2 — Second copies of the shot monitor/queue/wait and of the save bracketing, with the deletion deferred

`devices/detector.py:137-215` duplicates `devices/ca/triggerable.py:129-263` (`CaAcqTimestampReadable` + `CaTriggerable`) essentially line for line — persistent `subscribe_reading`, 128-deep drop-oldest queue, `value <= 0` filter, synchronous baseline + drain, `wait_for_idle` loop with the attributable timeout. `LvNativeFileDataLogic` (`:261-324`) is the third home of `localsavingpath`/`save` bracketing alongside `plans/run_wrapper.py:83-130` (`save_enable_plan` / `_save_cleanup_plan`) and `devices/nonscalar_save.py`.

The PR acknowledges this ("Not touched… is the plan-layer PR") and cites Sam's sequencing decision. I accept that `CaGenericDetector` must keep working until phase 1. What the sequencing does **not** require is two copies of the queue logic: `CaAcqTimestampReadable` could hold a `GeecsAcquireLogic` today (`connect → attach`, `disconnect → detach`, `_last_acq → logic.last_acq_timestamp`) and `CaTriggerable.trigger()` become `logic.baseline(); return AsyncStatus(logic.wait_for_idle())`. That makes the phase-1 deletion purely subtractive, lets the existing `test_ca_devices.py` / `test_single_shot_plan.py` pin the shared implementation, and removes the drift window (the new copy already dropped the queue-bound rationale and the cold-cache `t0 is None` note that `triggerable.py:223-231` carries). This is the repo's stated rule ("modernization PRs delete the old path in the same change; back-compat is the 10% exception") applied to the part that can be folded now.

Refutation attempt: checked whether `GeecsAcquireLogic` must differ (the `DetectorAcquireLogic` ABC needs `start_acquiring`/`ensure_stopped` no-ops) — those are additive; the shared core is identical.

---

### P3 — The save-path column renames an event-schema contract without the contract files traveling

`devices/detector.py:289` (`datakey_suffix = "-save_path"`)

`EVENT_SCHEMA.md:131` specifies `<dev>-nonscalar_save_path`; `GEECS-Data-Utils/geecs_data_utils/tiled_schema.py:54` lists `-nonscalar_save_path` in `COMPANION_SUFFIXES` (hidden from Y-column pick lists and drift analysis in the console/portal); `assets/tiled_readback.py:250` builds `f"{safe_name(device_name)}-nonscalar_save_path"` to find a run's save directory. A `GeecsDetector` row carries `<name>-save_path` instead: not in any of those places, so the string column shows up as a plottable scalar in the browsers and the asset readback finds no save dir. Not live yet (the namespace still builds `CaGenericDetector`), but this is exactly the "contract files travel with behavior" check: either keep the `-nonscalar_save_path` suffix (zero cost) or update `EVENT_SCHEMA.md`, `tiled_schema.py`, `tiled_readback.py` in the PR that flips the namespace. I checked `tiled_export.py:35-48`: the s-file exporter is driven by `geecs_scalar_headers`, so that path is unaffected.

---

### P3 — A scalars-only `GeecsDetector` has no `save` control, so the `save_control_only` mode (stale `save=on` clearing) has no successor

`devices/detector.py:392-399` creates `save` only when a `path_provider` is given. `run_wrapper.py:235-242` exists because a capture-owned camera (`native_image_save` off) must still be commanded `save="off"` at scan start — "a save flag left on out-of-band would otherwise write files to a stale path" (found live 26_0828; `capture/FORMAT.md:92`). Scenario: previous run crashed with `save=on` and `localsavingpath` pointing at yesterday's scan; today the camera is built as `GeecsDetector(..., path_provider=None)` and fires shots → PNGs land in the old scan folder. Phase 1 has to either give the detector a `save` control independent of the path provider or consciously drop the mode — per the repo's "ask before preserving features" rule, flagging it for a decision now rather than discovering it in the namespace flip.

---

### P3 — "A plain `bp.count([cam])` is refused at prepare" is only true with `OPHYD_ASYNC_PRESERVE_DETECTOR_STATE` unset

`devices/detector.py:22-25, 108-110`, `CHANGELOG.md` 0.78.0

`StandardDetector.trigger()` (`_detector.py`, the `preserve_state` branch) calls `default_trigger_info()` when the env var is `YES`; `GeecsTriggerLogic.default_trigger_info` returns `STRICT_TRIGGER_INFO`, so the implicit prepare **succeeds** and the camera waits `shot_timeout` per shot for a fire nobody sends (or, if the box happens to be in SCAN, silently records free-run shots). `test_plain_count_is_refused_clearly` fails under that env. The PR body is careful ("works too"); the module docstring and CHANGELOG overclaim. One sentence fixes it.

---

### P3 — `ShotControl.defines()` / `set()` raise `ValueError` for an unknown state name

`devices/shot_control.py:108-110, 152-155`: `_state("foo")` → `TriggerState("FOO")` → `ValueError`. `defines()` is a predicate (the composed `ShotController.defines_state` returns `False` for unknown names) and `set()` elsewhere in the module promises `GeecsConfigurationError`. Small, but the hardware test's `assert shot_control.defines(...)` pattern invites typos.

---

### P3 — Placement: a `devices/` module importing the request runner for a 30-line pure adapter

`devices/shot_control.py:97` defers `from geecs_bluesky.scan_request_runner import trigger_writes_from_profile`. The adapter (`scan_request_runner.py:126-160`) is pure `TriggerProfile → ShotControlWrites` and its natural home is next to `ShotControlWrites` (`models/shot_control.py`, whose docstring already points at it) — the deferral avoids the cycle textually but inverts the layer direction (device layer → runner). The PR comment says phase 1 relocates it; moving it now is smaller than the comment explaining why it wasn't.

---

### Checked and dropped

- **Scan-folder invariant** (`detector.py:309-314`): `directory.parent.is_dir()` then `mkdir(exist_ok=True)` without `parents` — correct, and `StaticPathProvider`'s `create_dir_depth` is ignored by this logic (`_path_providers.py:175-197`), so a caller cannot widen it. An SMB blip becomes a loud `FileNotFoundError`, which is the documented desired behavior.
- **Synchronous baseline** (`detector.py:454-463`): `WatchableAsyncStatus.wrap` creates the task at call time, so `baseline()` before `super().trigger()` is the right place; the queue preserves a stamp that lands between baseline and `wait_for_idle`. `test_baseline_is_taken_synchronously_in_trigger` and `test_stale_updates_before_trigger_are_not_a_shot` both fail if the baseline moved into the coroutine or the drain were dropped.
- **Hard-pause rewind double-fire**: `all_safe_rewind` is `True` for ophyd-async devices, so a hard pause inside `geecs_take_reading` replays prepare/trigger/fire on resume — identical to stock `trigger_and_read` and to the legacy funnel (`geecs_single_shot` has no `rewindable_wrapper` at all), and documented in `CLAUDE.md` as the accepted floor.
- **`disconnect()` without `super()`**: `ophyd_async.core.Device` has no `disconnect` in 0.19.3; the override is the package's own convention (same as `CaAcqTimestampReadable`).
- **Per-shot `prepare`** in `strict.py:69-76`: `_update_prepare_context` reuses providers when `collections_per_event` is unchanged, so `save` is not re-toggled per shot; `prepare_edge` is a no-op. Cost is a few soft-signal awaits, as claimed.
- **Staging the scalar children** in `GeecsDetector.stage`: `StandardDetector.stage` indeed does not stage child signals; the coherence argument in `CLAUDE.md` ("Read path") carries over unchanged because `acq_timestamp` shares the same listener-refcounted cache.
- **Facility literals** in `test_phase0_hardware.py:97-99`: same env-overridable defaults as `test_namespace_hardware.py:71-73`; none in package code. Disclosed in the PR; not new debt.
- **`pytestmark` + per-test `hardware` marker** duplicated in the hardware test — harmless.

skbarber:
## Dispositions of the adversarial review

Every finding from the review comment above is dispositioned below. **Process note:** the original reviewer was cut off by an API rate limit before it could confirm the fixes, so the confirm/deny was done by a **fresh-context verifier** given only the findings text and the fix commits (`79f24fba`, `7e672837`), per the `/land` substitution rule. Its report follows the table. Verifier nits were then fixed in `2bea7f38`.

| # | Finding | Disposition |
|---|---|---|
| P1 | `_resume_to` survives a stop/abort from a paused gated run and drives SCAN on a later resume | **Fixed** (79f24fba): a plan-driven standing state clears the pause bookkeeping (`ShotControl.set`). Pinned by `test_stop_from_a_paused_gated_run_does_not_leak_into_the_next` — the verifier traced that the test fails with the clearing line removed. |
| P2 | Second pause owner with diverging rules (quiesce set, raise policy, two standing-state fields) | **Fixed** (79f24fba): `pause()`/`resume()` never raise (logged); a profile without OFF logs a warning and leaves the box; `QUIESCE_FROM = {SCAN, STANDBY}` with its one home in `models/shot_control.py`, imported by the device and by `plans/pause_semantics.py`; the standing state is `ShotController.last_state` (via the now-public `record_state`) and the `state` config signal mirrors it. Residual noted by the verifier and accepted: a device and the quiescer coexisting in one *funnel* run would double-quiesce — phase 0 never runs that combination and phase 1 deletes the quiescer. |
| P2 | `GeecsAcquireLogic` duplicates `CaAcqTimestampReadable`'s monitor/queue/wait; `LvNativeFileDataLogic` is a third home of save bracketing | **Fixed** for the shot logic (79f24fba): `CaAcqTimestampReadable`/`CaTriggerable` now compose `GeecsAcquireLogic`; `_last_acq`/`_shot_queue`/`_monitoring`/`_trigger_timeout` stay as property views for the funnel-era callers and go with the funnel. **Waived** for the save bracketing: `run_wrapper.save_enable_plan` / `NonScalarSaveSupport` are plan stubs on the funnel's device, `LvNativeFileDataLogic` is awaited inside the ophyd-async lifecycle — no seam to compose; deleted with the funnel in phase 1. Verifier: waiver holds. |
| P3 | `-save_path` renames an event-schema contract column | **Fixed** (79f24fba): `-nonscalar_save_path`, matching `EVENT_SCHEMA.md`, `tiled_schema.COMPANION_SUFFIXES`, `tiled_readback`. |
| P3 | A scalars-only detector cannot clear a stale `save=on` | **Fixed** (79f24fba): `GeecsDetector(native_save=True)` owns the flag without a path provider; `stage` clears it, no column is produced. Verifier's reminder carried to phase 1: capture-owned cameras in the namespace must be built with `native_save=True`. |
| P3 | "Plain `bp.count` is refused" overclaims under `OPHYD_ASYNC_PRESERVE_DETECTOR_STATE=YES` | **Fixed**: module docstring (79f24fba) and CHANGELOG (2bea7f38). |
| P3 | `ValueError` on unknown state names | **Fixed** (79f24fba): `GeecsConfigurationError` from `set`, `False` from `defines`. |
| P3 | `devices/` importing the request runner for the profile→writes adapter | **Fixed** (79f24fba): adapter lives in `devices/shot_control.py`; all import sites point there (2bea7f38); no import cycle (fresh-interpreter check). |

Verifier nits (all fixed in `2bea7f38`): CHANGELOG caveat; three import sites still reaching the adapter through the runner's re-export; the runner docstring; the quiescer's mid-file import.

Hardware re-acceptance on the fixed code: Scan 001 of 26_0910, `1 passed in 27.7 s` (M3 in `Planning/native_bluesky/04_phase0_measurements.md`).

---

### Verifier report (fresh context; given the findings and the fix commits only)

| # | Finding | Verdict |
|---|---|---|
| 1 | P1 `_resume_to` leak | **CONFIRMED FIXED.** `ShotControl.set()` clears `_resume_to` after any non-SINGLESHOT drive; `pause()` goes through `_drive` so it does not self-clear. Traced the new test with the clearing line removed: the `== "ARMED"` assert fails — the test is load-bearing. |
| 2 | P2 pause-owner divergence | **CONFIRMED FIXED.** `pause()` body entirely inside try/except → `logger.exception`; a no-OFF profile logs a warning and returns before setting `_resume_to`; `resume()` likewise. `QUIESCE_FROM` has one home (`models/shot_control.py`), imported by the device and `plans/pause_semantics.py`. Standing state is `ShotController.last_state` via public `record_state`; SINGLESHOT never leaks into the `state` signal. Residual, not a regression: device + quiescer in one funnel run would double-quiesce — phase 0 never runs that; phase 1 deletes the quiescer. |
| 3 | P2 duplicate shot monitor | **CONFIRMED FIXED** for the shot logic: `triggerable.py` carries no queue/monitor/wait of its own; the property views satisfy `preflight.py`, `generic_detector.py`, `timestamped_readable.py` and `tests/test_ca_devices.py` (instance assignment of `_trigger_timeout` + class-attr `_shot_queue_maxsize`). **Save-bracketing waiver holds**: `LvNativeFileDataLogic` is awaited inside the ophyd-async lifecycle, the funnel's are plan stubs / a mixin — no seam to compose; goes with the funnel. |
| 4 | P3 contract column | **CONFIRMED FIXED.** |
| 5 | P3 stale save flag | **CONFIRMED FIXED.** Reminder for phase 1: every capture-owned camera in the namespace must be built with `native_save=True`. |
| 6 | P3 env-var overclaim | **PARTIALLY FIXED** at the time of verification — docstring yes, CHANGELOG still unqualified. → fixed in 2bea7f38. |
| 7 | P3 unknown state names | **CONFIRMED FIXED.** |
| 8 | P3 placement | **CONFIRMED FIXED.** No import cycle (fresh-interpreter imports of `plans.pause_semantics`, `devices.shot_control`, `scan_request_runner`, `session`, `devices.ca.triggerable`, `devices.detector`, `shot_controller`). |

New findings: CHANGELOG overclaim (above); three import sites still via the runner's re-export + the runner docstring; the quiescer's mid-file `# noqa: E402` import. No behavioural defect introduced. `ruff check` / `ruff format --check` pass.

Tests: `pytest tests/test_shot_control_device.py tests/test_geecs_detector.py tests/test_ca_devices.py tests/test_pause_semantics.py` → `81 passed in 5.26s`.

skbarber:
## Codex review — disposition

**P1 — `fire_and_await_shot` treated every `FailedStatus` (trigger, fire, wait) as a no-frame timeout and refired; a refused or ambiguous SINGLESHOT put could have issued extra physical shots.** Confirmed and **fixed in `71fbc40b`**: the refire is gated on `exc.__cause__` being a detector's `GeecsTriggerTimeoutError`; any other failed status re-raises untouched (the `_no_frame_device` "unknown device" fallback is gone with it). Pinned by `tests/test_strict_plans.py::test_a_failed_fire_is_not_a_dropped_frame` — a box whose SINGLESHOT put raises: the run fails with that error as the cause and exactly one fire was attempted. Single-shot suites: 43 passed. Thanks — that was a real hazard on the shared seam, and it also applied to the funnel's `geecs_single_shot`.

PR #813, opened 2026-09-10
GeecsBluesky 0.78.1: stop the RunEngine loop threads tests leave behind (#812)

### LINKED ISSUE #812: GeecsBluesky test suite: a 180 s Tiled test and leaked RunEngine loops make local runs 2–4× slower than CI
## What

The full GeecsBluesky unit suite (`-m "not integration and not fake_server"`, ~800 tests) takes 6–12 minutes on a developer Mac over VPN and intermittently times out one test, while CI runs it green in ~3.5 minutes. Profiled 2026-09-10 with `pytest --durations=30 --timeout=180` (on `phase/00-one-camera-detector`, but the base branch shows the same shape — 764 passed, 1 timed out):

- **`tests/test_tiled_integration.py::test_subscribe_tiled_reachable_server_still_subscribes` took 180.59 s** — the per-test timeout — while the next-slowest test is 9 s. Its monkeypatches (`socket`, `tiled.client.from_uri`) should make it instant, so something in `subscribe_tiled` still reaches out (or spins) when a real Tiled URI is in `config.ini`.
- In earlier runs without `--durations`, the timeout landed on other tests in `tests/test_scan_request_plan.py` (`test_optimize_persists_history_to_optimization_json`, `test_trigger_profile_builds_the_controller_worker_side`) that pass alone in ~5 s; by ~50 % of the run the process carries ~40 live RunEngine loop threads (every `RunEngine()` in a test spawns one that is never closed), which is the leading suspect for the order dependence.
- `scripts/check.sh` pipes pytest output, so a slow test looks like a hang; run with `PYTHONUNBUFFERED=1` and `-v`/`--durations` to see it.

## Proposed

1. Make the Tiled test hermetic (assert the monkeypatched seam is the only one used; add a hard `timeout` marker).
2. A shared, session-scoped `RunEngine` fixture (or an autouse teardown that closes each test's loop) so threads do not accumulate.
3. `--durations=10` in `scripts/check.sh` output so regressions show up.

Context: found while landing #811 (phase 0 of #807); not caused by it.

skbarber:
## What

An autouse `tests/conftest.py` fixture that stops and joins every RunEngine event-loop thread a test leaves behind, plus `--durations=10` on the GeecsBluesky line of `scripts/check.sh`. Refs #812 (the leak is the bulk of it; a residual one-test stall is documented there and stays open).

## Why

Every `RunEngine()` a test constructs starts a daemon thread running its own asyncio loop forever (`_ensure_event_loop_running`), and nothing in bluesky stops it. The suite has 57 inline constructions across 20 files; by the midpoint of a run ~150 live loops sat idling in `select` and the whole process crawled on macOS. The symptom was misleading: a 1.4 s test (`test_tiled_integration.py::test_subscribe_tiled_reachable_server_still_subscribes`) took the full 180 s per-test timeout, and on other runs a different test did — whichever was running when the process stalled. CI on Linux ran the same suite green in 3.5 min, so it looked like "something slow on the Mac".

bluesky records loop → thread in `_ensure_event_loop_running.loop_to_thread`; the fixture snapshots it before each test and, after, stops and joins the loops that appeared. Loops created outside a test (module-level engines) are left alone.

## Measured (same Mac, same suite, `--timeout=180 --durations`)

| | before | after |
|---|---|---|
| wall | 6:24 – 12 min | **3:25** |
| result | 764–792 passed, **1 timed out** | **740 passed, 0 timeouts** |
| live threads at midpoint | ~150 | **18** |
| slowest test | 180.59 s (the timeout) | 8.96 s (a funnel parity test) |

(The "after" row was measured on `master`'s test set before this PR was retargeted to `feature/native-bluesky-plans`; the branch now carries the fix as 0.78.1 on that line.)

**Residual, kept open on #812:** with the review's drain step added and the machine's memory pressure removed (a Docker VM had the Mac at ~15 MB free pages and 8 M swap-outs), one full run on the feature branch still stalled a single funnel-parity test (`test_scan_request_plan.py::test_named_noscan_plan_matches_the_funnel`, 9 s alone) for the full 180 s: **798 passed, 1 timed out, 6:33**. Its RunEngine thread was not being scheduled at all (it did not even process `loop.stop` within 5 s — the new warning fired), and memory, Tiled, Channel Access over the VPN (8.88 s with CA at localhost vs 8.93 s over the VPN) and config-file size are each ruled out by measurement. The thread-leak fix here is independent of that residual and verified on its own (150 → 18 live threads; a clean 740-pass run in 3:25). Phase 1 of #807 deletes the funnel-parity tests wholesale.

## Scope

One concern; +34/−1 in `conftest.py`, one flag in `check.sh`, changelog + patch bump 0.78.0 → 0.78.1 (test-only change → patch, per repo convention). No facility literals. No hardware surface.

### DIFF
--- a/.github/workflows/unit-tests.yml
+++ b/.github/workflows/unit-tests.yml
@@ -77,7 +77,7 @@ jobs:
 
       - name: Run GeecsBluesky pure unit tests
         working-directory: GeecsBluesky
-        run: poetry run pytest tests -m "not integration and not fake_server" --tb=short -q
+        run: poetry run pytest tests -m "not integration and not fake_server" --tb=short -q --durations=10
 
       - name: Install GEECS-MCP dependencies
         working-directory: GEECS-MCP
--- a/GeecsBluesky/CHANGELOG.md
+++ b/GeecsBluesky/CHANGELOG.md
@@ -4,6 +4,20 @@ All notable changes to `geecs-bluesky` are documented here.
 
 Format follows [Keep a Changelog](https://keepachangelog.com/en/1.1.0/).
 
+## [0.78.1] - 2026-09-10
+
+### Fixed
+
+- **Test suite: RunEngine loop threads no longer accumulate across the run**
+  (#812). Every `RunEngine()` a test constructs starts a daemon thread running
+  its own asyncio loop forever and nothing stopped it, so by the midpoint of
+  the suite ~150 live loops were idling in `select` and the process crawled —
+  on a developer Mac a 1.4 s test took the full 180 s per-test timeout and
+  whichever test ran at that point was blamed (CI on Linux ran the same
+  suite in 3.5 min). An autouse `conftest.py` fixture now stops and joins
+  every loop a test left behind. `scripts/check.sh` prints the ten slowest
+  tests for the GeecsBluesky suite so a regression shows up as a number.
+
 ## [0.78.0] - 2026-09-09
 
 ### Added
--- a/GeecsBluesky/pyproject.toml
+++ b/GeecsBluesky/pyproject.toml
@@ -1,6 +1,6 @@
 [tool.poetry]
 name = "geecs-bluesky"
-version = "0.78.0"
+version = "0.78.1"
 description = "Bluesky / ophyd-async bridge for the GEECS control system"
 authors = ["Sam Barber <sbarber@lbl.gov>"]
 readme = "README.md"
--- a/GeecsBluesky/tests/conftest.py
+++ b/GeecsBluesky/tests/conftest.py
@@ -16,18 +16,26 @@ def pytest_collection_modifyitems(items: list[pytest.Item]) -> None:
 
 
 @pytest.fixture(autouse=True)
-def _stop_run_engine_loops_left_by_the_test():
+def _stop_run_engine_loops_left_by_the_test(request: pytest.FixtureRequest):
     """Stop the event-loop threads a test's ``RunEngine()`` instances leave behind.
 
     Every ``RunEngine()`` starts a daemon thread running its own asyncio loop
     forever; nothing in bluesky stops it, so a suite with ~60 inline
-    constructions accumulates 100+ live loops and the whole process crawls
-    (#812: a 1.4 s test taking the full 180 s timeout at the midpoint of the
-    run, on macOS).  bluesky records loop → thread in
-    ``_ensure_event_loop_running.loop_to_thread``; this stops and joins every
-    loop that appeared during the test.  Session-wide loops created before
-    the test (module-level engines) are left alone.
+    constructions (plus one per ``GeecsSession``) accumulates 100+ live loops
+    and the whole process crawls (#812: a 1.4 s test taking the full 180 s
+    timeout at the midpoint of the run, on macOS).  bluesky records
+    loop → thread in ``_ensure_event_loop_running.loop_to_thread``; this
+    stops every loop that appeared during the test, drains what it still had
+    pending (a cancelled pacer, a paused RunEngine) the way
+    ``asyncio.Runner.close`` does — so nothing is destroyed while pending and
+    no stray traceback lands after the pytest summary — then closes it.
+    Loops created before the test (module- or session-level engines) are
+    left alone.  A loop whose thread will not stop within 5 s is left
+    running, as before this fixture, and reported so the leak is visible.
     """
+    import asyncio
+    import warnings
+
     from bluesky.run_engine import _ensure_event_loop_running
 
     registry = _ensure_event_loop_running.loop_to_thread
@@ -41,5 +49,23 @@ def _stop_run_engine_loops_left_by_the_test():
             loop.call_soon_threadsafe(loop.stop)
         if thread is not None and thread.is_alive():
             thread.join(timeout=5.0)
-        if not loop.is_running() and not loop.is_closed():
-            loop.close()
+        if loop.is_running():
+            warnings.warn(
+                f"{request.node.nodeid}: a RunEngine loop thread did not stop "
+                "within 5 s and was left running (a blocking callback on the "
+                "loop?)",
+                RuntimeWarning,
+                stacklevel=1,
+            )
+            continue
+        if loop.is_closed():
+            continue
+        # Drain before closing: cancel what is still pending and let it finish
+        # cancelling (the test thread may drive a stopped loop).
+        pending = asyncio.all_tasks(loop)
+        for task in pending:
+            task.cancel()
+        if pending:
+            loop.run_until_complete(asyncio.gather(*pending, return_exceptions=True))
+        loop.run_until_complete(loop.shutdown_asyncgens())
+        loop.close()
--- a/GeecsBluesky/tests/conftest.py
+++ b/GeecsBluesky/tests/conftest.py
@@ -13,3 +13,33 @@ def pytest_collection_modifyitems(items: list[pytest.Item]) -> None:
             "timeout"
         ):
             item.add_marker(timeout)
+
+
+@pytest.fixture(autouse=True)
+def _stop_run_engine_loops_left_by_the_test():
+    """Stop the event-loop threads a test's ``RunEngine()`` instances leave behind.
+
+    Every ``RunEngine()`` starts a daemon thread running its own asyncio loop
+    forever; nothing in bluesky stops it, so a suite with ~60 inline
+    constructions accumulates 100+ live loops and the whole process crawls
+    (#812: a 1.4 s test taking the full 180 s timeout at the midpoint of the
+    run, on macOS).  bluesky records loop → thread in
+    ``_ensure_event_loop_running.loop_to_thread``; this stops and joins every
+    loop that appeared during the test.  Session-wide loops created before
+    the test (module-level engines) are left alone.
+    """
+    from bluesky.run_engine import _ensure_event_loop_running
+
+    registry = _ensure_event_loop_running.loop_to_thread
+    before = set(registry.keys())
+    yield
+    for loop in list(registry.keys()):
+        if loop in before:
+            continue
+        thread = registry.get(loop)
+        if loop.is_running():
+            loop.call_soon_threadsafe(loop.stop)
+        if thread is not None and thread.is_alive():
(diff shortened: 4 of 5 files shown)


skbarber:
## Dispositions

All three findings fixed in `acb20079` (none was a correctness failure; the reviewer found no test broken and no hang):

| Finding | Disposition |
|---|---|
| P3 — `loop.close()` destroyed pending tasks (a cancelled pacer, a paused RunEngine), printing "Task was destroyed but it is pending" after the pytest summary | **Fixed**: after the join, the fixture cancels `asyncio.all_tasks(loop)`, drives the stopped loop to completion with `gather(..., return_exceptions=True)`, runs `shutdown_asyncgens()`, then closes — the `asyncio.Runner.close` drain the reviewer pointed at. The reviewer's reproduction set (`test_pause_semantics`, `test_telemetry_group_and_t0check`, `test_pause_checkpoints`, `test_grid_and_action_plans`, `test_free_run_plan`) now runs with **0** destroyed-task lines (57 passed with the phase-0 files added). |
| P3 — `thread.join(timeout=5.0)` timing out was silent | **Fixed**: a loop whose thread is still alive after 5 s is left running (the pre-PR behaviour — the safe fail) and reported as a `RuntimeWarning` naming the test, so the leak shows up in the run output. |
| P3 — `--durations=10` made `check.sh` diverge from the CI line it mirrors | **Fixed**: the same flag on the CI step (`unit-tests.yml`), so the slowest-10 table is in the CI logs too and the "keep in sync" comment stays true. |

Reviewer's full report below.

---
# Adversarial review — PR #813 `fix/test-suite-speed` (GeecsBluesky 0.76.4)

Scope reviewed: `GeecsBluesky/tests/conftest.py` (the autouse fixture), `scripts/check.sh` (`--durations=10`), changelog + version bump. Verified against bluesky 1.15.0 (`run_engine.py:406-444` `RunEngine.__init__`, `:2839-2854` `_ensure_event_loop_running` + `loop_to_thread` WeakKeyDictionary) and ophyd-async 0.19.3 (`core/_device.py:411-425`).

The central claim holds: every `RunEngine()` registers its loop → thread in `_ensure_event_loop_running.loop_to_thread`, nothing in bluesky ever stops it, and the fixture's before/after snapshot correctly isolates loops created during a test. I found no correctness failure that breaks a test or hangs the run. Three findings survive, all P3.

## P3 — `loop.close()` destroys pending tasks; the fixture's stop races a test's own `pacer.cancel()` and any RE left non-idle

**Anchor:** `GeecsBluesky/tests/conftest.py:40-45` — `call_soon_threadsafe(loop.stop)` → `join` → `close()` with no drain step in between.

**Scenario 1 (reproduced on this branch):** `tests/ca_mock_helpers.py:60` returns a `concurrent.futures.Future` chained to the loop-side `pace()` task. A test calls `pacer.cancel()` in its `finally`; `_chain_future` schedules the loop-side `task.cancel()` via `call_soon_threadsafe`, but the fixture's `loop.stop` lands before the task processes the `CancelledError`. `loop.close()` then clears `_ready`, the task is GC'd in `cancelling` state, and asyncio logs:

```
Task was destroyed but it is pending!
task: <Task cancelling name='Task-1557' coro=<start_pacer.<locals>.pace() running at .../tests/ca_mock_helpers.py:58> ...>
```

Running `tests/test_pause_semantics.py tests/test_telemetry_group_and_t0check.py tests/test_pause_checkpoints.py tests/test_grid_and_action_plans.py tests/test_free_run_plan.py`: with the conftest → 37 passed + 1 such line on stderr at exit; with `--noconftest` → 37 passed, 0 such lines. So the message is introduced by this PR.

**Scenario 2 (probe test):** a test that returns with its RE paused (e.g. asserts `RE.state == "paused"` and never resumes/stops — the kind of test `test_pause_semantics.py` is full of) leaves `RunEngine._run()` pending. Same mechanism → `Task was destroyed but it is pending! ... coro=<RunEngine._run() ...>` plus a `TransitionError` traceback printed at GC time. The following test still passes; no hang.

**Why it matters (modestly):** the text lands on stderr at interpreter exit, after the pytest summary, where it reads like a new bug in whatever ran last — exactly the misattribution shape #812 was about. It does not fail any test (asyncio logs via `logging`, not `warnings`, so `-W error` wouldn't catch it either), which is why it is P3.

**What bluesky does instead:** its own `tests/conftest.py::RE` fixture (installed at `GeecsBluesky/.venv/.../bluesky/tests/conftest.py:16-32`) calls `RE.halt()` when the state isn't idle/panicked *before* `loop.stop`/`_th.join()`/`loop.close()`. This fixture only has loops, not engines, so the equivalent is the `asyncio.Runner.close()` drain: after the join and before `close()`, cancel `asyncio.all_tasks(loop)` and `loop.run_until_complete(gather(*tasks, return_exceptions=True))` from the test thread (legal: the loop is stopped, not closed), then `run_until_complete(loop.shutdown_asyncgens())`. ~6 lines; removes both scenarios.

**Tried to refute:** checked every `start_pacer(` site cancels in a `finally` (all 17 do, incl. `test_telemetry_group_and_t0check.py:156` → `:177`), so the race is the only path to scenario 1 — and it still fires. Checked that no test currently leaves an RE paused (no `RunEngine._run` destroyed-task lines in the real-suite run), so scenario 2 is latent, not present.

## P3 — `thread.join(timeout=5.0)` timing out is silent

**Anchor:** `GeecsBluesky/tests/conftest.py:42-45`.

If a loop's current callback blocks (a sync call made on the RE thread, a stuck mock), `loop.stop` never runs, the join burns 5 s, `loop.is_running()` stays True, and the `is_running()` guard correctly skips `close()` — so no exception, and the loop leaks exactly as pre-PR. That is the right fail-safe, but nothing reports it: a regression of the #812 kind would surface only as the per-test `--durations` creeping up, and a reader of the fixture has no signal that the 5 s branch was ever taken. A `warnings.warn(...)`/`logging.warning` when `thread.is_alive()` after the join (naming the test via `request.node.nodeid`) makes the leak visible in the run output, matching the PR's own goal of "a regression shows up as a number".

**Tried to refute:** confirmed no current test leaves a blocking callback on the loop (full-suite timing in the PR body: slowest 8.96 s, no 5 s-teardown cluster). Latent only.

## P3 — `--durations=10` makes `check.sh` diverge from the CI line it is declared to mirror

**Anchor:** `scripts/check.sh:235` vs `.github/workflows/unit-tests.yml:80` (`poetry run pytest tests -m "not integration and not fake_server" --tb=short -q`), under the header comment at `check.sh:7-11, 35` — "CI mapping (keep in sync with .github/workflows/unit-tests.yml) … when CI changes, change the runner table here in the same PR."

Outcome-neutral (display only), so this is a documentation consistency nit, not a behaviour bug: the next "keep in sync" pass has to guess whether the flag was meant to be local-only or is a CI omission. Either add it to the CI step too (a per-run slowest-10 table in CI logs is cheap and is where the Linux-vs-Mac comparison in the PR body came from) or leave a one-line comment on `check.sh:235` saying the divergence is intentional.

---

## Probed and refuted (no finding)

- **Shared/wider-scoped RunEngines.** No module-level `RunEngine(` assignments anywhere in the repo; no `scope=` fixtures in `GeecsBluesky/tests`; the only RE fixture is function-scoped (`test_action_compiler.py:103`), and the `return RunEngine(), ...` helpers (`test_shot_control.py:76`, `test_pause_checkpoints.py:218`, `test_shot_controller_writes.py:67`) are plain functions called inside tests. pytest runs autouse fixtures before other same-scope fixtures, so `before` is snapshotted before `re_and_factory` builds its RE, and higher-scoped fixtures (if any ever appear) would be set up before the snapshot and left alone, as the docstring says.
- **`GeecsSession`** constructs an RE (`geecs_bluesky/session.py:221`) — ~30 more loops per run the PR's "57 inline constructions" undercounts, but all 28 test sites are function-scoped fixtures or inline; the `runpy.run_path` startup-profile tests (`test_qserver_startup.py`) build theirs inside the test body. All correctly reaped; the docstring count ("~60") is merely low.
- **`run_coroutine_threadsafe(..., RE._loop)` after stop.** All 15 test sites and all 7 production sites (`session.py`) use an RE created in the same test. `devices/ca/oneshot.py:40` runs its own `geecs-ca-oneshot` loop outside bluesky's registry — untouched.
- **`set_bluesky_event_loop` global left pointing at a closed loop.** Only consumer in the installed stack is ophyd-async's *sync* `init_devices.__exit__` (`core/_device.py:411-420`); the repo never uses the sync form (`connect_mock`/`async with` only) and has no `get_bluesky_event_loop`/`call_in_bluesky_event_loop` callers. A future sync `with init_devices(mock=True)` without an RE in the same test would now fail deterministically (`NotConnectedError`) instead of order-dependently riding a previous test's loop — an improvement.
- **pytest-asyncio (`asyncio_mode = "auto"`).** Its loops are not in bluesky's registry; unaffected.
- **"Closes #812" with the Tiled test untouched.** `test_tiled_integration.py:140-170` patches `socket.create_connection`, `tiled.client.from_uri`, and `TiledWriter`, and uses a stub `_Engine`, so it is hermetic; the 180 s was a process stall, as the PR body says. Issue item 1 is correctly dispositioned as a symptom.
- **Background threads orphaned by the stop.** `test_operator_abort.py:64` and `test_pause_semantics.py:511` threads are daemon and the RE calls they interrupt return within the test.
- **Redundancy.** No other `loop.stop`/thread-reaping helper exists in any test tree (repo-wide grep, excluding `.venv`/`extras`); other conftests' autouse fixtures (`ImageAnalysis`, `GEECS-DataPortal`, root `tests/`, `GEECS-Console`) are config/QSettings isolation, unrelated. bluesky's own `RE` fixture is a per-fixture pattern inside its test tree, not an importable plugin — a reference (see P3 #1), not a duplicate.
- **Placement.** `GEECS-Console/tests` and `GEECS-MCP/tests` construct no `RunEngine`/`GeecsSession` (grep), so there is no second consumer to justify a shared `geecs_bluesky.testing` plugin today. The MCP CI job that runs `GeecsBluesky/tests/qs_client` (`unit-tests.yml:145`) picks this conftest up via directory ancestry; harmless (no REs there, bluesky importable in that env). No facility literals; no scan-folder surface; changelog + patch bump present.

PR #816, opened 2026-09-10
GeecsBluesky 0.79.0: phase 1 PR 1 — delete the funnel, free-run, GeecsSession and the funnel-only devices; stock plans over the namespace (#807)

### LINKED ISSUE #812: GeecsBluesky test suite: a 180 s Tiled test and leaked RunEngine loops make local runs 2–4× slower than CI

skbarber:
Phase 1 PR 1 of the native-Bluesky rebuild (#807; plan of record `Planning/native_bluesky/03_clean_room_rebuild.md` §8, §10.5, amended in this PR). **Deletion-led:** everything that existed only to serve the `ScanRequest` funnel goes, and the worker registers the stock `bluesky.plans` verbs over the device namespace. Lands into `feature/native-bluesky-plans`; the deployed worker stays on `master` until PR 3's hardware acceptance flips it.

## What goes

- **The funnel and its plans** — `plans/scan_request_plan.py`, `step_scan.py`, `free_run_step_scan.py`, `named_plans.py`, `pause_semantics.py`, `t0_sync.py`, `orchestration.py`, `plans/liveness.py`, `plans/optimize.py`; `session.py` (`GeecsSession`), `scan_request_runner.py`, `preflight.py`, `optimize.py`, `shot_controller.py`.
- **The funnel-only devices** — `CaGenericDetector`, `CaTriggerable` / `CaAcqTimestampReadable`, `CaTelemetryReadable`, `CaTimestampedReadable`, the shot-id / contributor / nonscalar-save mixins, `ScanContext`; `ShotControlConfig` / `ShotControlState` and the legacy YAML loader; the four exceptions only they raised.
- **The `function_execute` verbs** (`geecs_move_variable`, `geecs_describe_action`) — a manual move is a stock `mv` queue item; a preview is client-side resolver work (Sam, 2026-09-10).
- **Optimization glue, option B** (Sam, 2026-09-10) — `optimization/session_bridge.py`, `worker_loader.py`, the loader hook. The Xopt core stays importable with its tests; **optimization is not runnable** until re-glued in its own phase.
- The two funnel-era scripts and **36 test files** (~12.6k lines).

## What changes

- **Namespace builds `GeecsDetector`** for every `looks_triggerable` device; `native_save` iff the DB lists both `save` and `localsavingpath` (§10.5 rule) — the detector then owns those two controls (never scan-settable children) and a `PathProvider` given at build points its files at the run. Subscribed settables' readbacks join the detector's columns via the new `GeecsDetector.add_readables`.
- **`run_engine.make_run_engine`** replaces the session: RE + `connect_on_demand` outermost + opt-in Tiled / s-file callbacks. No scan API.
- **Startup profile** exports the namespace and `count, scan, rel_scan, list_scan, rel_list_scan, grid_scan, rel_grid_scan, list_grid_scan, rel_list_grid_scan, mv` (`plan_names.GEECS_PLAN_NAMES`, asserted by `geecs-qserver-ensure-ready`; the operator group in `user_group_permissions.yaml` allows exactly those). PR 2 rebinds the same names with the strict `take_reading`.
- **`ShotControl` absorbs `ShotController`'s write machinery** (`CaPutSetter`, cached setters, ordered replay, the standing state). `plans/strict.py` absorbs `fire_and_await_shot` + the CONNECTED refire gate; `plans/claim_scan.py` holds the claim alone.
- **`qs_client.run_submit_preflight`** keeps only engine-free checks: every save set resolves, `worker_ready`, gateway liveness, free-run staleness. Its submit verbs still name the retired funnel plan, so `worker_ready` refuses against this worker — correct; the client seam is rewired with PR 2. **GEECS-Console and GEECS-MCP import unchanged** (Console's lazy imports of `pause_semantics` / `merge_save_sets` have pinned fallbacks).
- Docs: `CLAUDE.md`, `README.md`, `EVENT_SCHEMA.md` (the native shape; v1 is history), `TILED_SETUP.md`, both `qserver/` READMEs, the root package table, the plan of record (§2/§8/§10 — the second-round decisions: preset = device group + plan call, option B, verbs dropped, deployment at the freeze).

## Also fixed

- **#812** — the residual one-test stall: the in-process startup-profile test built the 0MQ `Publisher`; a connected-but-peerless PUB socket blocks the zmq context's teardown for the *next* test's whole timeout. The hermetic startup tests now run with `QS_DOC_PUBLISH_ADDR=OFF`. Suite: 3.5 min → 12 s.
- **`tests/test_phase0_hardware.py` is now gated on `GEECS_HW=1`.** ⚠️ Incident, disclosed: the `hardware` marker alone did not protect it — CI's and `check.sh`'s explicit `-m` overrides the `addopts` deselect — and while building this PR on a laptop on the lab VPN, ordinary suite runs executed it: **2–3 three-shot `bp.count` bursts on `UC_Amp4_IR_input` under HTU-NoGas (ARMED → SINGLESHOT×3 → STANDBY), no motor move, no save, setpoints untouched.** The trigger was restored to STANDBY by the test's finalize each time.

## LOC (per concern)

| Concern | Files | Lines |
|---|---|---|
| Deletions (source) | 26 | −9,480 |
| Deletions (tests) | 36 | −12,567 |
| Trims / folds (`strict`, `claim_scan`, `shot_control`, models, exceptions, `plan_names`, `scanner_configs`, `submit_preflight`) | 9 | net −900 |
| Namespace + `GeecsDetector.add_readables` + `run_engine.py` + startup profile + permissions | 6 | +290 / −120 |
| Test rewrites (13 files) | 13 | net −650 |
| Docs (CLAUDE.md rewrite, README, EVENT_SCHEMA, qserver READMEs, plan of record, root table) | 9 | net −700 |

Total: 123 files, +1,359 / −27,339.

- `plan_names` keeps `SCAN_REQUEST_PLAN` / `RUN_ACTION_PLAN` as the names the client's legacy submit verbs still spell, so Console/MCP import and their tests stay green. Deleting them cascades into the Console (`submission.py`) and MCP (`control_tools.py`) before their rewire.
- The profile exports the **raw** stock plans now (scalar-only scans work; a `GeecsDetector` under a bare stock plan is refused at prepare). The alternative — export nothing — makes `geecs-qserver-ensure-ready` report `plans_empty` forever.
- `capture/` and `assets/` are untouched (they go with #806, per Sam).
- `GeecsUnservedVariablesError` deleted with its last raiser rather than kept for PR 2's preflight.

## Tests

- GeecsBluesky: **369 passed, 1 skipped (the gated hardware test), 1 deselected in 11.8 s** (`-m "not integration and not fake_server"`, `--timeout=120`).
- GEECS-MCP: **103 passed, 1 skipped**.
- GEECS-Console: not runnable here (no pytest in the local Console env); CI's Windows job covers it. The `qs_client` tests it also runs: **50 passed** inside the GeecsBluesky count.
- `pre-commit` clean on every changed file.

## Hardware verification

**OWED — by design, PR 3.** This PR runs nothing on hardware. Headless acceptance (HTU-NoGas, `U_S1H:Current` −1 → +1 A in 0.5 A steps, amp4in, setpoint restored) comes with the plan layer; the worker is not flipped by this PR.

Closes #812.


skbarber:
# Adversarial review (fresh context)

Diffed `origin/feature/native-bluesky-plans...phase/01-foundation` (123 files, +1373/−27339 — the local `feature/native-bluesky-plans` ref is stale, so use the origin ref). All API claims checked against the installed ophyd-async 0.19.3 / bluesky 1.15.0 sources in `GeecsBluesky/.venv`. Targeted suite run here: 185 passed (detector, namespace, shot-control, strict, startup, gateway-put, qs_client, readiness, connect-on-demand, ca-devices). Hardware test not run.

## 1. `native_save` keys on *served*, but `:SP` exists only for *settable* — a get-only `save`/`localsavingpath` row makes the whole camera unconnectable

- **Severity:** P2 · **Lens:** correctness
- **File:** `GeecsBluesky/geecs_bluesky/namespace.py:329` (`native_save = triggered and NATIVE_SAVE_VARIABLES <= {n.lower() for n in typed}`), `devices/detector.py:451-454`; premise repeated in `Planning/native_bluesky/03_clean_room_rebuild.md:596-597` ("so the gateway serves their `:SP`").
- **Scenario:** `typed` is the served set (`served_for` = subscribed ∪ settable, `namespace.py:160-162`). A device whose DB rows list `save` and `localsavingpath` with `get='yes'` but not settable is served *without* `:SP` — `GeecsCAGateway/PV_CONTRACT.md:104-105`: "Non-settable variables … have **no** `:SP` PV." Reproduced with the test roster (`row("save", settable=False)`, `row("localsavingpath", settable=False)`, both subscribed): `cam.native_save is True`, `cam.save.source == ca://testexp:uc_testcam:save` with an `:SP` twin. `connect_on_demand` connects the whole device, so the missing `:SP` fails the detector's connect — the camera is unusable even for a scalars-only `count`, where before this PR those rows were plain readables. The only test (`test_native_save_iff_the_db_lists_both_saving_controls`) builds both rows with `settable=True`.
- **Fix:** require `row.get("settable")` for both in `_build` (and say "settable" in §10.5); add the `settable=False` case to the test. Whether this bites HTU today depends on the DB flags; the rule is wrong against the contract file regardless.

## 2. `GeecsDetector.add_readables` — the one new detector method — has no test that would fail if it were wrong

- **Severity:** P2 · **Lens:** correctness (tests)
- **File:** `GeecsBluesky/geecs_bluesky/devices/detector.py:475-485`; `tests/test_namespace.py:175`, `tests/test_geecs_detector.py`.
- **Scenario:** The "subscribed AND settable → readback logged" assertion at `test_namespace.py:175` is on `U_S1H`, a `CaSnapshotReadable` (stock `StandardReadable.add_readables`). The roster's camera has no subscribed settable, and `test_geecs_detector.py` never calls `add_readables`. Drop the `self._scalars_logic.add(*signals)` line and every test still passes. I verified the method works under mock (a `CaSettable` child added this way yields `uc_x-gain-readback` in `describe()`/`read()` after `prepare`), so this is a coverage gap, not a bug — but it is the PR's own claim ("Subscribed settables' readbacks join the detector's columns via the new `GeecsDetector.add_readables`") with nothing pinning it. Also the annotation says `Sequence[SignalR]` while the namespace passes a `Device`; it works by duck typing (`read/describe/stage/unstage`), so either widen the hint or document the contract.
- **Fix:** give `UC_TestCam` a subscribed settable in `ROSTER` and assert its `-readback` key in the `cam.read()` set at line 170.

## 3. The incident's root cause is still in place: `check.sh`/CI `-m` still drops `not hardware`

- **Severity:** P2 · **Lens:** correctness (process)
- **File:** `scripts/check.sh:235`, `.github/workflows/unit-tests.yml:80` (both `-m "not integration and not fake_server"`, unchanged vs base); `GeecsBluesky/pyproject.toml:115` `addopts = "-m 'not hardware and not integration'"`.
- **Scenario:** The PR body correctly diagnoses that an explicit `-m` overrides `addopts`, then fixes only the one file (`GEECS_HW=1` module-level skip, `allow_module_level=True` present — verified). The override that fired real shots is untouched, so the next `@pytest.mark.hardware` test anyone adds under `GeecsBluesky/tests` runs under `/check` and CI again. `test_phase0_hardware.py` is currently the only hardware-marked test in GeecsBluesky/Console/MCP, so nothing is exposed *today*.
- **Fix:** `-m "not integration and not fake_server and not hardware"` in both places (one line each). Keep the env gate too — belt and braces for a test that arms the machine.

## 4. "GEECS-Console imports unchanged (pinned fallbacks)" is only true for `pause_semantics`; `merge_save_sets` degrades to an error string in the UI

- **Severity:** P3 · **Lens:** correctness / placement
- **File:** `GEECS-Console/geecs_console/services/configs.py:234-240`.
- **Scenario:** `from geecs_bluesky.scan_request_runner import merge_save_sets` sits inside `except Exception as exc: return UnionPreview(hint=str(exc))`. Against this branch the save-set union preview reads "No module named 'geecs_bluesky.scan_request_runner'" instead of a device count — swallowed, not pinned. `scan_monitor.py:46-58` is a genuine pinned fallback (and `tests/test_scan_monitor.py:57` returns on `ImportError`), so the Console *imports*; it just loses that preview.
- **Placement note:** `merge_save_sets` was pure logic over `SaveSet` (role-conflict `ValueError`, entry union). Its natural home is `geecs_schemas.save_set` — schema knowledge in its one module — where it would have survived this deletion and served both Console and the PR-2 client expansion. Either move it there now or list the Console preview as owed for PR 2.

## 5. Stale cross-references the docs pass missed (contract files now name deleted code)

- **Severity:** P3 · **Lens:** correctness (docs)
- `GeecsBluesky/TILED_SETUP.md:94,141` — still points at `tests/test_scan_request_hardware.py`, deleted by this PR (the file was edited in this PR).
- `GeecsCAGateway/PV_CONTRACT.md:306`, `geecs_ca_gateway/gateway.py:702`, `tests/test_gateway.py:853` — the shot-id publication rationale cites `CaTriggerable.trigger()`, deleted here; the behaviour is now `GeecsAcquireLogic.wait_for_idle`.
- `GEECS-Data-Utils/geecs_data_utils/native_files.py:28` — names `geecs_bluesky.optimization.session_bridge` as the waiter side (deleted here).
- `GEECS-Console/CLAUDE.md:203,245,487-491,522-529` — describes `UnservedVariablesCheck`, `geecs_move_variable`/`geecs_describe_action`, and `optimization.worker_loader` "registered at qserver startup", all deleted by this PR while the root `CLAUDE.md` row was updated.
- Kept modules still narrate the deleted runner: `devices/ca/settable.py:88`, `snapshot.py:85`, `action_signals.py:34,153` ("the runner's `session.disconnect` cleanup"), `capture/daemon.py:16` (`geecs_run_wrapper`), `plans/claim_scan.py:26` ("outside the GUI's own `ScanDataManager`" — the legacy GUI is gone since 2026-08-20).

## 6. Dead funnel-only path survives a deletion-led PR: `CaSnapshotReadable(save_control_only=…)`

- **Severity:** P3 · **Lens:** redundancy
- **File:** `GeecsBluesky/geecs_bluesky/devices/ca/snapshot.py:74-82`.
- **Scenario:** No caller of `save_control_only` remains in GeecsBluesky, Console or MCP (its only caller was the runner's snapshot-role branch). It also hard-codes `"save"` — the variable name `namespace.NATIVE_SAVE_VARIABLES` now owns — and its purpose (clear a stale `save=on` on a camera whose frames are not wanted) is exactly the `native_save=True`-without-`path_provider` `GeecsDetector` case `LvNativeFileDataLogic`'s docstring describes. Delete it with the funnel, or say why it stays.

## 7. `CaPutSetter` lives in the trigger-box module but is the generic gateway put convention

- **Severity:** P3 · **Lens:** placement
- **File:** `GeecsBluesky/geecs_bluesky/devices/shot_control.py:56-69`; its base and the module that already refers to it: `devices/ca/gateway_put.py:7`; pinned by `tests/test_gateway_put.py:22,107`.
- **Scenario:** It is a two-line subclass of `GatewaySetpointPut` (`coerce=str`, 10 s), tested in the gateway-put test file. PR 2's action-plan setters or any device wanting the stringified-wire convention would import it from `ShotControl`'s module or move it. Beside its base in `gateway_put.py` costs nothing now and avoids the move later.

## 8. Unlogged semantic change in the client `free_run_staleness` check

- **Severity:** P3 · **Lens:** correctness
- **File:** `GeecsBluesky/geecs_bluesky/qs_client/submit_preflight.py:108-121`, `:348-352`.
- **Scenario:** The old `save_set_to_devices_config` ordered devices by role (`reference` first), so the staleness probe sampled the reference device. The new `_resolve_devices_config` keeps file order, so it samples the first non-snapshot entry — the docstring says so ("first save-set device") but the CHANGELOG entry does not. Low stakes since free-run is being retired, but it is a client-visible behaviour change.

---

**Examined and found clean:** `devices/detector.py` against ophyd-async 0.19.3 (`_supported_triggers` derived from `prepare_edge` → bare `bp.count` refused at prepare as claimed; per-shot `bps.prepare` reuses providers via `_update_prepare_context`, so `LvNativeFileDataLogic.prepare_single` runs once per stage; `wait_for_idle` reached from `_wait_for_index`; `mkdir(exist_ok=True)` only inside an existing parent — scan-folder invariant honoured); `plans/strict.py` (byte-parity with the deleted `single_shot.fire_and_await_shot` bar docstring/rename; `FailedStatus.__cause__` gating intact); `ShotControl` fold (per-target setter cache, ordered replay, SINGLESHOT never recorded as standing — matches deleted `ShotController`); `plan_names` ↔ `startup.py` ↔ `user_group_permissions.yaml` ↔ `_qserver_startup_probe.py` ↔ `qserver_ready` agreement; `submit_preflight` resolver/schema API use (`resolve_save_set`, `SaveSetEntry.device/role`); `run_engine.py` (only `RunEngine(` factory in non-test code — no duplicate); `plans/__init__` import weight (base already pulled the ca-backed devices); MCP's `snapshot_images`/`unserved_variables` names in `control_tools.py:62-63` are an allow-list, not a branch; hygiene grep of added lines (no home paths, hostnames, accounts); CHANGELOG 0.79.0 + `pyproject.toml` bump; `.github`/`check.sh` unchanged vs base; `capture/`/`assets/` untouched per scope.

skbarber:
# Disposition of the adversarial review

Fix commit 5a0c089f (+ the doc leftovers in the follow-up commit). Each fix went back to the same reviewer for confirmation; its verdicts are quoted.

| # | Finding | Disposition |
|---|---|---|
| 1 | `native_save` keyed on served, not settable | **Fixed** — `_build` requires settable `save` + `localsavingpath`; §10.5 says "as settable"; get-only case pinned in `test_namespace`. Reviewer: confirmed fixed. |
| 2 | `GeecsDetector.add_readables` unpinned | **Fixed** — the camera roster subscribes settable `exposure`; `uc_testcam-exposure-readback` asserted in `read()`/`describe()`; hint widened to the duck-typed contract. Reviewer: confirmed fixed. |
| 3 | CI/`check.sh` `-m` still selected hardware tests | **Fixed** — both say `not hardware`. Reviewer: confirmed fixed. |
| 4 | Console save-set union preview degrades to a hint; `merge_save_sets` placement | **Waived** — the Console is rewired once, when the foundation is stable; save sets are replaced by presets in PR 2, so `merge_save_sets` is not moved into `geecs_schemas`. Recorded in the CHANGELOG. Reviewer: "waived, accurately recorded … no objection". |
| 5 | Stale cross-references | **Fixed in-package** (TILED_SETUP.md, `settable.py`, `snapshot.py`, `action_signals.py`, `capture/daemon.py`, `capture/FORMAT.md`, `claim_scan.py`, `forward_expr.py`). **Waived cross-package**: GeecsCAGateway's shot-id note (PV_CONTRACT.md / `gateway.py` / `test_gateway.py`), GEECS-Data-Utils `native_files.py`, GEECS-Console/CLAUDE.md travel with their own packages' PRs (the Console rewire; the gateway note with #806 or whichever PR next touches it). Reviewer: confirmed in-package; waivers reasonable. |
| 6 | Dead `CaSnapshotReadable(save_control_only=…)` | **Fixed** — deleted with its imports; FORMAT.md sentence historical; CHANGELOG entry. Reviewer: confirmed fixed. |
| 7 | `CaPutSetter` placement | **Fixed** — lives beside `GatewaySetpointPut` in `devices/ca/gateway_put.py`. Reviewer: confirmed fixed. |
| 8 | Unlogged staleness-sample order change | **Fixed** — CHANGELOG records it. Reviewer: confirmed fixed. |

Suite after the fixes: 369 passed, 1 skipped, in 12 s. No undispositioned finding remains.

skbarber:
# Disposition of the Codex review

| # | Finding | Disposition |
|---|---|---|
| P2 | Console `union_preview()` loses the save-set union counts / reference hints / role-conflict preview (its `merge_save_sets` import fails and the `ImportError` becomes the hint) | **Waived** — same as finding 4 of the fresh-context review. The feature branch is not deployed (the lab's Console and worker run `master`); the Console is rewired once, when the foundation is stable; and save sets are replaced by presets in PR 2, so re-homing `merge_save_sets` into `geecs_schemas.save_set` would move code into a module PR 2 deletes. Recorded in the CHANGELOG. Cheap to veto: say the word and it moves into `geecs_schemas.save_set` with a Console import edit and a service-level test. |
| P3 | `GeecsBluesky/CLAUDE.md` said the scan path runs "with the strict `take_reading`" while the worker exports raw stock plans | **Fixed** — the paragraph now says the strict `take_reading` exists and is tested but is not registered in the worker's plan names until PR 2's table (and what a `GeecsDetector` under a bare stock plan does). |

skbarber:
# Disposition of the Codex second pass

Commit 19df6b00.

| # | Finding | Disposition |
|---|---|---|
| P1/P2 | `qs_client` still calls the dropped worker functions `geecs_move_variable` / `geecs_describe_action`; the Console's Movable panel and action previews fail against this worker | **Waived — by decision.** The function_execute verbs were dropped with `GeecsSession` (Sam, 2026-09-10; plan of record §10.5): a manual move is a stock `mv` queue item, a preview is client-side resolver work. The Console and MCP are rewired once, when the foundation is stable; the deployed worker stays on `master` until PR 3, so no operator meets this. The PR body says so under "Judgment calls". |
| P2 | Console `configs.py` imports the deleted `merge_save_sets`; the union preview degrades to a hint | **Waived** — same as the first-pass finding and fresh-context finding 4; confirmed by the maintainer. Recorded in the CHANGELOG. |
| P2/P3 | `GeecsDetector.disconnect()` detached the stamp monitor without clearing the connect cache, so `connect_on_demand.is_connected()` stayed true and the monitor was never re-attached | **Fixed by deletion** — verified against ophyd-async 0.19.3 `Device.connect` (the cached `_connect_task` is reused unless `force_reconnect`). The `disconnect()` hook was the deleted runner's per-scan `session.disconnect` contract and has no caller; it is removed from every device (`GeecsDetector`, `CaSettable`/`CaMotor`, `CaSnapshotReadable`, `CaConfirmSettable`, `CaPseudoMovable`, `CaActionSignalFactory`) along with `GeecsAcquireLogic.detach` and the two tests that pinned the hook. Namespace devices live as long as the RunEngine; nothing disconnects them mid-process. CHANGELOG entry added. |
| P3 | MCP `validate_scan_request` description advertises engine validation and the unserved-variables check | **Fixed** — the tool description now lists what `run_submit_preflight` checks (save sets resolve, worker ready, liveness, staleness) and says the resolver checks return with the plan layer. GEECS-MCP 0.8.8, CHANGELOG entry. |

Suites after the change: GeecsBluesky 367 passed in 12 s; GEECS-MCP 103 passed.

PR #818, opened 2026-09-10
GeecsBluesky 0.78.0: single-shot refires only on an actual missing frame (#817)

### LINKED ISSUE #817: geecs_single_shot mislabels a failed fire() as a camera frame drop, then burns the refire budget on it
`geecs_single_shot` reports every attempt failure as a camera frame drop, including a failed `fire()`.

## Observed (Undulator, 2026-09-10, Scan033)

```
12:02:48.398 INFO    geecs_bluesky.shot_controller  scan=Scan033 - Shot controller → SINGLESHOT
12:02:49.156 INFO    geecs_bluesky.devices.ca.motor scan=Scan033 - u_modeimageresp_position_axis_1: moving Position.Axis 1 → -10.5 (tol=0.005)
12:02:50.558 WARNING geecs_bluesky.plans.single_shot scan=Scan033 - single-shot attempt 1 of 3: no frame from unknown device — re-firing (known camera frame-drop intermittency, ~1% observed)
12:02:50.752 WARNING geecs_bluesky.plans.single_shot scan=Scan033 - single-shot attempt 2 of 3: no frame from unknown device — re-firing (known camera frame-drop intermittency, ~1% observed)
12:02:51.396 INFO    geecs_bluesky.sfile_callback   scan=Scan033 - Skipping legacy scalar file export for scan 33 (uid=...): exit_status='fail'
```

The scan died reporting a camera frame drop. It was not one.

## Why the log is wrong

Three things in that excerpt are inconsistent with a missing frame:

1. **"unknown device."** A real frame miss can only come from
   `CaTriggerable._wait_for_shot`, which always raises
   `GeecsTriggerTimeoutError(device_name, ...)`. `_no_frame_device`
   (`plans/single_shot.py`) falls back to `"unknown device"` only when
   `FailedStatus.__cause__` is something else.
2. **The timing.** `CaTriggerable._trigger_timeout` is 3.0 s with no
   overrides. Attempt 1 → attempt 2 is **194 ms**; all three attempts
   finished in 838 ms. Nothing waited for a frame.
3. **No fire happened.** `ShotController.set_state` logs
   `Shot controller → SINGLESHOT` only *after* its writes complete. There
   is no such line for any of the three attempts.

`geecs_single_shot`'s `try` spans the whole attempt — `bps.trigger`,
`fire()`, and `bps.wait` — but its `except FailedStatus` treats everything
that lands there as a camera frame drop. A `fire()` whose gateway `:SP`
write was refused (`aioca.caput(..., wait=True)` raises `aioca.CANothing`)
therefore gets: the frame-drop wording, the `~1%` reassurance, the device
identity thrown away, and two more refires against a write that never
landed — before propagating.

## Reproduced

Simulating a `SINGLESHOT` put that errors after 200 ms, with cameras
delivering frames normally, gives the Scan033 shape exactly:

```
 1163ms  u_modeimageresp_position_axis_1: moving Position.Axis 1 -> -10.5 (tol=0.005)
 2770ms  single-shot attempt 1 of 3: no frame from unknown device — re-firing …
 2971ms  single-shot attempt 2 of 3: no frame from unknown device — re-firing …
 3178ms  Run aborted
```

For contrast, a failing *motor* status surfaces at `bps.mv`
(`step_scan.py:124`) inside `move_changed_axes` and never reaches
`single_shot` — `CaMotor` is not `Triggerable`, so the shot loop never
touches it. Both stage failure modes are named and loud
(`GeecsMotorTimeoutError` at `move_timeout` = 30 s; a rejected setpoint at
`motor.py:106`), and under `failed_move_policy="pause"` there would also be
a `FAILED_MOVE` ERROR line. None of that is in the record.

## Cost

The operator-visible failure named the wrong subsystem and the wrong
device. On this scan the shot-control device shares a GEECS device server
with `u_modeimageresp`, so the actual signal — one box taking out both the
trigger write and the stage — was hidden behind a message about camera
intermittency.

## Fix

Classify before retrying: only a `GeecsTriggerTimeoutError` cause is a
missing frame. Anything else logs at ERROR with its real cause (`CANothing`
already carries the PV name and the CA message) and propagates on the first
attempt instead of burning the refire budget. Frame-drop refire, the
`CONNECTED` device-down gate, and the strict one-row-per-shot semantics
stay as they are.

skbarber:
Closes #817.

## What

`geecs_single_shot`'s `try` spans the whole attempt — `bps.trigger`,
`fire()`, and `bps.wait` — but its `except FailedStatus` treated everything
landing there as a camera frame drop. A `fire()` whose gateway `:SP` write
was refused (`aioca.caput(..., wait=True)` raises `aioca.CANothing`) got the
frame-drop wording, the `~1% observed` reassurance, its device identity
discarded as `"unknown device"`, and two more refires against a write that
never landed — before propagating.

`_no_frame_device` becomes `_no_frame_timeout`: it returns the attributed
`GeecsTriggerTimeoutError` or `None`. Only the former is a missing frame.
Anything else is logged at ERROR with its real cause (`CANothing` carries
the PV name and the CA message) and propagates on the first attempt.

Unchanged: frame-drop bounded refire, the `CONNECTED` device-down gate, the
strict one-row-per-shot semantics, and the no-cancellation reasoning for
abandoned statuses (a straggler `GeecsTriggerTimeoutError` still consumes
one refire rather than aborting the scan).

## Why

Undulator Scan033, 2026-09-10:

```
12:02:48.398 INFO    shot_controller   Shot controller → SINGLESHOT
12:02:49.156 INFO    devices.ca.motor  u_modeimageresp_position_axis_1: moving Position.Axis 1 → -10.5 (tol=0.005)
12:02:50.558 WARNING plans.single_shot single-shot attempt 1 of 3: no frame from unknown device — re-firing (known camera frame-drop intermittency, ~1% observed)
12:02:50.752 WARNING plans.single_shot single-shot attempt 2 of 3: no frame from unknown device — re-firing (known camera frame-drop intermittency, ~1% observed)
12:02:51.396 INFO    sfile_callback    Skipping legacy scalar file export for scan 33: exit_status='fail'
```

Three tells that this is not a frame drop: `"unknown device"` (a real miss
always carries `GeecsTriggerTimeoutError.device_name`); attempts 194 ms
apart against a 3.0 s `_trigger_timeout` with no overrides; and no
`Shot controller → SINGLESHOT` line for any of the three fires, which
`set_state` logs only after its writes complete.

The operator-visible failure named the wrong subsystem and no device. The
shot-control device shares a GEECS device server with `u_modeimageresp`, so
the real signal — one box taking out both the trigger write and the stage —
was hidden behind a message about camera intermittency.

## Scope

**Two concerns, bundled deliberately** — the same idiom, found by the same
investigation. Cheap to veto: the second is one line of source plus its pin.

| Concern | Files | LOC |
|---|---|---|
| 1. A failed `fire()` reported and retried as a camera frame drop | `plans/single_shot.py` (+73/-9), `tests/test_single_shot_plan.py` (+98/-1) | ~171 |
| 2. `exc.__cause__ or exc` discards a falsy `CANothing` — **pre-existing** on the `FAILED MOVE` line | `plans/step_scan.py` (+8/-1), `tests/test_pause_semantics.py` (+20/-1) | ~28 |
| Bookkeeping | `CHANGELOG.md` (+42), `pyproject.toml`, `Planning/data_capture/…` (+8/-4) | ~52 |

Total 233 insertions / 18 deletions across 7 files.

**The judgement call to veto if you disagree**: concern 2 fixes a bug this PR
did not introduce, on a different plan. Splitting it would leave a known
operator-facing defect in the `FAILED MOVE` line — the one your #820 work
touches — for the sake of a purer diff. I bundled it; say the word and I'll
lift it into its own PR.

The `Planning/` edit replaces stale `single_shot.py` line numbers this branch
invalidated with named landmarks. No facility literals added; no
`PV_CONTRACT.md` / `EVENT_SCHEMA.md` surface touched.

## Tests

- `tests/test_single_shot_plan.py` — **12 passed** (11 before + the new
  `test_single_shot_failed_fire_propagates_without_refire`).
- Verified the new test fails without the fix: on `master`'s
  `single_shot.py` it fires 3 times and emits 2 refire warnings.
- `./scripts/check.sh` (GeecsBluesky, own env, the way CI runs it) —
  **872 passed, 1 skipped, 4 deselected** in 253 s; lint clean.
- `./scripts/check.sh --all` — see the comment below.

## Hardware verification

**Done** — 2026-09-10, against the live CA gateway (0.20.2, 95 devices), run
from this branch's env. The queueserver was not involved and still runs
`master`.

Method, chosen to write nothing: `fire()` puts to a `:SP` PV that does not
exist, so no trigger can receive it; the only real hardware contact is a
read-only CA monitor on `UC_ModeImager` — the camera from the Scan033 incident.
No shot fired, no scan number claimed, no device variable written.

```
ERROR geecs_bluesky.plans.single_shot - single-shot attempt 1 of 3 failed,
  but not from a missing frame — re-firing cannot help, so the failure
  propagates: CANothing: undulator:u_dg645_shotcontrol:verify818:SP:
  User specified timeout on IO operation expired

fires      : 1   (no refire against a write that never landed)
event rows : 0
propagated : FailedStatus / cause CANothing
```

### The hardware run found a bug in the fix

The first run printed `FailedStatus: <AsyncStatus …, done>` — no PV, no CA
message. `aioca.CANothing.__bool__` is `errorcode == ECA_NORMAL`, so a
**failed** put is falsy and `exc.__cause__ or exc` fell through to the
`FailedStatus`. (`done` rather than `errored` for the same reason —
`AsyncStatusBase.__repr__` also tests the exception for truthiness.)

Fixed with `is not None` in both plans. `step_scan.py:171` carried the
identical idiom on the **`FAILED MOVE`** line, so any move failing at the CA
layer would have reported the same useless repr; `GeecsMotorTimeoutError` is
truthy, which is why the tolerance path never exposed it.

No mock could have caught this — a plain exception is truthy. The stand-in now
mirrors `CANothing`'s falsiness and fails against the old idiom.

## Note on the Scan033 root cause

The scan-side root cause is being addressed separately in #820 (per-axis move
tolerance from the DB). The Bluesky-side tolerance was tighter than the LV
device's, so the stage is marginal at −10.5: Scan034 shows two genuine
`FAILED MOVE` timeouts at that exact position, 30 s each.

That is **not** what killed Scan033 — 1.402 s to failure and zero ERROR lines
in the file — so what actually failed there remains unidentified. This PR does
not claim to fix it. It makes the next occurrence self-describing: one ERROR
line naming the PV and the CA message, instead of "no frame from unknown
device — known camera frame-drop intermittency".

**Merge order**: this lands *after* #820. Both are numbered 0.77.0; I'll merge
master in and renumber once #820 is in.

### DIFF
--- a/GeecsBluesky/CHANGELOG.md
+++ b/GeecsBluesky/CHANGELOG.md
@@ -4,6 +4,28 @@ All notable changes to `geecs-bluesky` are documented here.
 
 Format follows [Keep a Changelog](https://keepachangelog.com/en/1.1.0/).
 
+## [0.77.0] - 2026-09-10
+
+### Fixed
+
+- `geecs_single_shot` no longer reports every attempt failure as a camera
+  frame drop. Its `try` spans the whole attempt (trigger + `fire()` +
+  wait), so a `fire()` whose gateway `:SP` write was refused also arrives
+  as `FailedStatus` — and was logged as "no frame from unknown device
+  (known camera frame-drop intermittency)", retried twice more against a
+  write that never landed, and propagated with the cameras blamed.
+  Observed live 2026-09-10 (Undulator Scan033): three attempts in 838 ms
+  against a 3.0 s trigger timeout, no device named, and no
+  `Shot controller -> SINGLESHOT` line for any of them.
+
+  Only a `GeecsTriggerTimeoutError` cause is a missing frame now. Anything
+  else is logged at ERROR with its real cause — `aioca.CANothing` carries
+  the PV name and the CA message — and propagates on the first attempt
+  instead of burning the refire budget. Frame-drop refire, the
+  `CONNECTED` device-down gate, and the strict one-row-per-shot semantics
+  are unchanged.
+
+
 ## [0.76.3] - 2026-09-08
 
 ### Changed
--- a/GeecsBluesky/CHANGELOG.md
+++ b/GeecsBluesky/CHANGELOG.md
@@ -16,7 +16,7 @@ Format follows [Keep a Changelog](https://keepachangelog.com/en/1.1.0/).
   write that never landed, and propagated with the cameras blamed.
   Observed live 2026-09-10 (Undulator Scan033): three attempts in 838 ms
   against a 3.0 s trigger timeout, no device named, and no
-  `Shot controller -> SINGLESHOT` line for any of them.
+  `Shot controller → SINGLESHOT` line for any of them.
 
   Only a `GeecsTriggerTimeoutError` cause is a missing frame now. Anything
   else is logged at ERROR with its real cause — `aioca.CANothing` carries
--- a/GeecsBluesky/CHANGELOG.md
+++ b/GeecsBluesky/CHANGELOG.md
@@ -25,6 +25,26 @@ Format follows [Keep a Changelog](https://keepachangelog.com/en/1.1.0/).
   `CONNECTED` device-down gate, and the strict one-row-per-shot semantics
   are unchanged.
 
+  Hardware-verified 2026-09-10 against the live gateway (a `:SP` put to a
+  nonexistent PV, `UC_ModeImager` armed read-only): one fire, no refire,
+  no event row, and the ERROR line naming the PV and the CA message.
+
+- Both plan-side error handlers selected the cause with
+  `exc.__cause__ or exc`, which discards exactly the object they exist to
+  report: `aioca.CANothing.__bool__` is `errorcode == ECA_NORMAL`, so a
+  *failed* CA put is falsy and `or` silently fell through to the
+  `FailedStatus`, whose text is only the status repr. The operator got
+  `<AsyncStatus …, done>` — `done` because `AsyncStatusBase.__repr__` also
+  tests the exception for truthiness. Now `is not None` in both
+  `plans/single_shot.py` and `plans/step_scan.py` (the `FAILED MOVE`
+  line, where the same bug hid any CA-layer move failure;
+  `GeecsMotorTimeoutError` is truthy, which is why the tolerance path
+  never exposed it).
+
+  Caught on hardware, not in tests: the mock stand-in was a plain
+  exception and therefore truthy. It now mirrors `CANothing`'s falsiness
+  and repr/str split.
+
 
 ## [0.76.3] - 2026-09-08
 
--- a/GeecsBluesky/geecs_bluesky/plans/single_shot.py
+++ b/GeecsBluesky/geecs_bluesky/plans/single_shot.py
@@ -39,18 +39,24 @@ from geecs_bluesky.exceptions import (
 logger = logging.getLogger(__name__)
 
 
-def _no_frame_device(exc: BaseException) -> str:
-    """Name the device that produced no frame, from a wait failure.
+def _no_frame_timeout(exc: BaseException) -> GeecsTriggerTimeoutError | None:
+    """The no-frame timeout behind a wait failure — or ``None`` if it is not one.
 
-    The RunEngine wraps the status's own error in
+    The RunEngine wraps the failing status's own error in
     :exc:`~bluesky.utils.FailedStatus` (``raise FailedStatus(ret) from exc``),
-    so the device-attributed :exc:`GeecsTriggerTimeoutError` rides on
+    so a device-attributed :exc:`GeecsTriggerTimeoutError` rides on
     ``__cause__``.
+
+    An attempt creates two kinds of status and only one of them is a frame
+    drop: the detectors' ``trigger()`` waits (which fail *only* as a named
+    :exc:`GeecsTriggerTimeoutError`) and ``fire()``'s own gateway ``:SP``
+    put (a rejected write surfaces as ``aioca.CANothing``, whose text
+    carries the PV and the CA message).  Returning ``None`` for the latter
+    is what keeps a failed *fire* from being reported — and retried — as a
+    camera frame drop.
     """
     cause = exc.__cause__
-    if isinstance(cause, GeecsTriggerTimeoutError):
-        return cause.device_name
-    return "unknown device"
+    return cause if isinstance(cause, GeecsTriggerTimeoutError) else None
 
 
 def _confirm_device_down(devices: Sequence[Any], device_name: str):
@@ -105,8 +111,16 @@ def geecs_single_shot(
     the recovery because a missed pulse never yields a frame (live-verified;
     numbers in ``GeecsBluesky/CHANGELOG.md`` 0.20.0).
 
-    Refire is gated on gateway liveness: a frameless device whose
-    ``CONNECTED`` PV reads Disconnected went down mid-scan, so
+    Refire is gated twice, because only one kind of failure is a frame drop.
+    First on **attribution**: the ``try`` spans the whole attempt (trigger +
+    ``fire()`` + wait), so a failed ``fire()`` — a rejected gateway ``:SP``
+    put — also arrives as ``FailedStatus``.  Only a
+    :exc:`~geecs_bluesky.exceptions.GeecsTriggerTimeoutError` cause is a
+    missing frame; anything else is logged with its real cause and
+    propagates on the first attempt rather than burning the budget against
+    a fault refire cannot fix (see :func:`_no_frame_timeout`).  Then on
+    **gateway liveness**: a frameless device whose ``CONNECTED`` PV reads
+    Disconnected went down mid-scan, so
     :exc:`~geecs_bluesky.exceptions.GeecsDeviceDownError` is raised instead
     of burning refires; a live or unreadable status (fail-open) keeps the
     bounded-refire behavior.
@@ -158,8 +172,28 @@ def geecs_single_shot(
             # so the stash is consumed right here at the wait; a straggler
             # that lands inside the next attempt is caught by this same try
             # (it wraps the whole attempt: trigger + fire + wait) and merely
-            # consumes one refire instead of aborting the scan.
-            device_name = _no_frame_device(exc)
+            # consumes one refire instead of aborting the scan.  That the try
+            # spans the whole attempt is also why the cause must be
+            # classified before anything is retried: a fire that never fired
+            # lands here too.
+            timeout = _no_frame_timeout(exc)
+            if timeout is None:
+                # Not a missing frame, so a refire cannot help and would only
+                # burn the budget against the real fault (live 2026-09-10,
+                # Scan033: a rejected SINGLESHOT put failed all three attempts
+                # in under a second and the scan died blaming the cameras).
+                # Name the actual cause - it is the only record of which PV
+                # failed and why.
+                logger.error(
+                    "single-shot attempt %d of %d failed, but not from a "
+                    "missing frame - re-firing cannot help, so the failure "
+                    "propagates: %r",
+                    attempt,
+                    attempts,
+                    exc.__cause__ or exc,
+                )
+                raise
+            device_name = timeout.device_name
             down = yield from _confirm_device_down(devices, device_name)
             if down:
                 raise GeecsDeviceDownError(
--- a/GeecsBluesky/geecs_bluesky/plans/single_shot.py
+++ b/GeecsBluesky/geecs_bluesky/plans/single_shot.py
@@ -192,7 +192,14 @@ def geecs_single_shot(
                 # the bare ECA errorcode).  Hence ``%s`` on the cause, with
                 # its type spelled out separately so an exception with an
                 # empty message still names itself.
-                cause = exc.__cause__ or exc
+                #
+                # ``is not None``, never ``or``: ``CANothing.__bool__`` is
+                # ``errorcode == ECA_NORMAL``, so the *failed* put whose
+                # identity this line exists to report is falsy, and
+                # ``exc.__cause__ or exc`` silently selects the useless
+                # ``FailedStatus`` instead (caught on hardware 2026-09-10 —
+                # the mock stand-in was truthy and could not see it).
+                cause = exc.__cause__ if exc.__cause__ is not None else exc
                 logger.error(
                     "single-shot attempt %d of %d failed, but not from a "
                     "missing frame — re-firing cannot help, so the failure "
--- a/GeecsBluesky/geecs_bluesky/plans/single_shot.py
+++ b/GeecsBluesky/geecs_bluesky/plans/single_shot.py
@@ -122,9 +122,8 @@ def geecs_single_shot(
     line is deliberately the record: the ``FailedStatus`` propagates
     unwrapped (the run's stop-document reason is the status repr), so the
     scan log is where the failing PV and its CA message are written down.
-    Then on
-    **gateway liveness**: a frameless device whose ``CONNECTED`` PV reads
-    Disconnected went down mid-scan, so
+    Then on **gateway liveness**: a frameless device whose ``CONNECTED`` PV
+    reads Disconnected went down mid-scan, so
     :exc:`~geecs_bluesky.exceptions.GeecsDeviceDownError` is raised instead
     of burning refires; a live or unreadable status (fail-open) keeps the
     bounded-refire behavior.
--- a/GeecsBluesky/geecs_bluesky/plans/single_shot.py
+++ b/GeecsBluesky/geecs_bluesky/plans/single_shot.py
@@ -182,11 +182,11 @@ def geecs_single_shot(
                 # burn the budget against the real fault (live 2026-09-10,
                 # Scan033: a rejected SINGLESHOT put failed all three attempts
                 # in under a second and the scan died blaming the cameras).
-                # Name the actual cause - it is the only record of which PV
+                # Name the actual cause — it is the only record of which PV
                 # failed and why.
                 logger.error(
                     "single-shot attempt %d of %d failed, but not from a "
-                    "missing frame - re-firing cannot help, so the failure "
+                    "missing frame — re-firing cannot help, so the failure "
                     "propagates: %r",
                     attempt,
                     attempts,
--- a/GeecsBluesky/geecs_bluesky/plans/single_shot.py
+++ b/GeecsBluesky/geecs_bluesky/plans/single_shot.py
@@ -118,7 +118,11 @@ def geecs_single_shot(
     :exc:`~geecs_bluesky.exceptions.GeecsTriggerTimeoutError` cause is a
     missing frame; anything else is logged with its real cause and
     propagates on the first attempt rather than burning the budget against
-    a fault refire cannot fix (see :func:`_no_frame_timeout`).  Then on
+    a fault refire cannot fix (see :func:`_no_frame_timeout`).  That ERROR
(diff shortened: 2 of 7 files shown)


skbarber:
## Local test results (`./scripts/check.sh`, the way CI runs each suite)

| Suite | Result |
|---|---|
| root-tests | 66 passed, 1 skipped, 63 deselected |
| ImageAnalysis | 276 passed, 2 skipped, 10 deselected |
| ScanAnalysis | 192 passed, 2 deselected |
| GEECS-Data-Utils | 325 passed, 3 skipped |
| GEECS-Schemas | 440 passed, 8 deselected |
| **GeecsBluesky** (the changed package) | **872 passed, 1 skipped, 4 deselected** (253 s) |

Lint (`pre-commit`, all files): clean.

**Not run locally**: GeecsCAGateway, GeecsPvaGateway, GEECS-Console,
GEECS-Core, GEECS-DataPortal, GEECS-LogTriage, GEECS-MCP. Those run from
their own per-package Poetry envs, which a fresh worktree does not have —
`--all` reported them as env-shaped collection errors
(`ModuleNotFoundError`), not test failures. CI's `unit-tests` job covers
them and passed on this branch.

The blast radius supports that: `grep` across every other package finds no
reference to `geecs_bluesky.plans.single_shot`. The two packages that depend
on GeecsBluesky (GEECS-MCP, GEECS-Console) are queueserver *clients* and
never import plan internals.

### The new test is a real pin

`test_single_shot_failed_fire_propagates_without_refire` fails on `master`'s
`single_shot.py` — 3 fires and 2 `no frame from unknown device` warnings
instead of 1 fire and 1 ERROR.

skbarber:
## Adversarial review (fresh-context subagent, three lenses)

Reviewed at `315f23f6`. Posted in full; disposition follows in the next comment.

---

### Verification the reviewer performed

Checked the classifier logic against the real exception types (`aioca.CANothing`, `bluesky.RunEngine._status_object_completed`, `CaTriggerable._wait_for_shot`), and confirmed the new test is a genuine regression pin: reverting `single_shot.py` to the merge base (`c26bb781`) makes `test_single_shot_failed_fire_propagates_without_refire` fail and the other 11 pass.

> The core change is right. `CaTriggerable.trigger()` is the only `Triggerable` in play (`CaTelemetryReadable`, `CaSnapshotReadable`, `CaTimestampedReadable` all lack `trigger`, and `StandardReadable` doesn't define one), and its status can only fail as `GeecsTriggerTimeoutError` — so the `isinstance` gate is exhaustive, not a heuristic. It also silently fixes a worse case the PR body doesn't claim: in the generalized `ShotController.set_state` path (ordered multi-write `SINGLESHOT`), a refire after a *partial* write list re-executed the writes that had already succeeded — a second physical fire attempt against half-applied trigger state.

### 1. `%r` on the cause drops the CA message the fix promises — and the test cannot catch it (moderate)

> `single_shot.py:190` logs the cause with `%r`. For the real object:
>
> ```
> repr(CANothing('undulator:u_dg645:trigger.executesingleshot:SP', 80))
>   -> CANothing('undulator:u_dg645:trigger.executesingleshot:SP', 80)
> str(...)
>   -> undulator:u_dg645:trigger.executesingleshot:SP: User specified timeout on IO operation expired
> ```
>
> `CANothing.__repr__` emits the bare ECA errorcode; only `__str__` renders `ca_message(errorcode)`. So the operator gets `80` where the docstring (`single_shot.py:53-54`), the CHANGELOG (`0.77.0`, "carries the PV name and the CA message") and the PR body all promise the CA message. And it is not recoverable elsewhere: the propagated `FailedStatus`'s own text is `<AsyncStatus … errored: CANothing('…:SP', 80)>` — also a `repr`. With `%r`, the CA message text appears **nowhere** in scan.log, the console tail, or the stop-doc reason.
>
> The sibling handler for the same shape — `step_scan.py:171-178`, `cause = exc.__cause__ or exc` logged with `%s` — already gets this right. Changing `%r` → `%s` matches it and restores the promised content.
>
> The new test cannot discriminate: `_PutRejected` is a plain `RuntimeError`, whose `repr()` embeds the entire message, so `assert "Trigger:SP" in message` passes identically under `%r` and `%s`. To make it a real pin, the stand-in needs `CANothing`'s repr/str split (`__repr__` → `_PutRejected('…:SP', 80)`, `__str__` → `…:SP: <message>`) and an assertion on the message text, not just the PV substring. As written, the *only* assertion the fake weakens is the one the docstring rests on.

### 2. Stale line-number cross-references into `single_shot.py` (low)

> `Planning/data_capture/01_central_pva_capture_scope.md:375-379` (Phase 5) cites `plans/single_shot.py:146-182` and `` `_confirm_device_down` `:56` `` as the pattern to copy. This diff moves `_confirm_device_down` from line 56 to 62 and the `bps.wait`→`create` region from 146-182 to ~160-216, so both references now point at the wrong lines.
>
> Worth more than a line-number bump: that Phase-5 note is specifically about needing to *"distinguish 'no frame, refire helps' from 'writer stalled, refire won't'"* — exactly the classification this PR adds. `_no_frame_timeout` is now the second (and closer) precedent for the pattern the plan describes and should be named there.

### 3. The propagated failure stays unattributed, unlike the branch 15 lines below (low, judgement call)

> The non-frame path re-raises the raw `FailedStatus`; the frame-drop-plus-device-down path immediately below raises `GeecsDeviceDownError` with an operator-facing message and `device_name`. `GeecsBluesky/exceptions.py` treats the `GeecsError` tree as the operator-facing surface.
>
> Concretely, in the PR's own hypothesised Scan033 mechanism — one GEECS device server taking out both the shot-control write and `u_modeimageresp` — the fire now fails *first*, so the plan never reaches the `_confirm_device_down` branch and the operator never gets the "went down mid-scan" diagnosis for the camera on the same box; they get `<AsyncStatus … errored: CANothing('…:SP', 80)>` as the stop-doc reason and one ERROR line in scan.log. That is strictly better than blaming the cameras, so this is not a regression — but the good record lives only in a log line, not in the exception, the stop document, Tiled, or the queue item report.
>
> Refutable in part: the reason string does carry the PV name, so the operator is not left with nothing. If you keep the bare `raise`, a sentence in the docstring saying the ERROR line is the record (and where it lands) would make the deliberateness legible; wrapping in an attributed `GeecsError … from exc` is the alternative.

### Redundancy — none found

> Searched for an existing `FailedStatus.__cause__` classifier before concluding `_no_frame_timeout` is new: `__cause__` appears at exactly two sites repo-wide (`plans/step_scan.py:171`, `plans/single_shot.py`) across `GeecsBluesky/geecs_bluesky`, `GEECS-Console`, `GEECS-MCP`; `plans/liveness.py`, `devices/ca/liveness.py`, `plans/pause_semantics.py` and `exceptions.py` carry no cause-unwrapping helper. `step_scan.py::move_or_pause` is the nearest twin but classifies nothing — it logs and pauses. The overlap is the `%s`-vs-`%r` consistency point in finding 1, not duplicated logic.

### Placement — no issues

> `_no_frame_timeout` is pure exception logic with one consumer, private to the module that owns the refire contract; nothing in it is reusable enough to earn a shared home, and nothing crosses a package boundary. Version bump (`0.76.3` → `0.77.0` minor for a behaviour change), CHANGELOG entry, and blank-line convention all match the repo rules; the `Undulator Scan033` reference in the CHANGELOG has precedent at line 2743 and is a historical incident record, not a facility value baked into code.

skbarber:
## Review disposition

All three findings **fixed**, each confirmed by the same reviewer against the
actual commits (not the summaries). Two rounds, because the first fix for
finding 2 was itself wrong.

| # | Finding | Disposition |
|---|---|---|
| 1 | `%r` on the cause drops the CA message; the test's `RuntimeError` stand-in cannot catch it | **Fixed** — `4318abc1` |
| 2 | Stale `single_shot.py` line-number cross-references in the Planning Phase-5 note | **Fixed** — `4318abc1`, corrected again in `92ee2439` |
| 3 | Propagated failure stays unattributed, unlike the branch below it | **Fixed** (as the judgement call it was framed as) — `4318abc1` |

### 1 — the log line

The ERROR line now binds `cause = exc.__cause__ or exc` and logs
`"%s: %s"` with `type(cause).__name__` and `cause`. Reviewer verified against
a real `CANothing`:

```
… the failure propagates: CANothing: undulator:u_dg645:trigger.executesingleshot:SP:
User specified timeout on IO operation expired
```

The reviewer **withdrew its own "match `step_scan.py`'s bare `%s`" framing**:
an empty-message exception renders as `Empty:` under the new form and as a
blank tail under `step_scan.py`'s, so spelling the type separately is the
stronger of the two.

The test stand-in `_PutRejected` now mirrors `CANothing`'s repr/str split
(`__repr__` → `_PutRejected('…:SP', 80)`, `__str__` → `'…:SP: <message>'`) and
asserts on the CA message text. Reviewer independently re-ran it with the log
line reverted to `%r`: **fails**, other 11 pass.

### 2 — two rounds

`4318abc1` re-derived the range as `160-216`. That was correct at `315f23f6`
but the same commit's docstring and comment additions moved the landmarks
again — the range was stale on arrival, for the second time in one PR. The two
*symbol-name* citations stayed correct through every commit because they sit
above every edit.

So `92ee2439` drops the range: the bullet now names `bps.wait(group=grp)` and
`bps.create(name)`. The reflow in that same commit shifted both by one line,
which is a live demonstration that naming was the right call.

### 3 — kept the bare `raise`

Plus a docstring sentence saying the ERROR line is deliberately the record and
why. Reviewer confirmed the sentence is literally true: `str(FailedStatus)` is
the `AsyncStatus` repr, which nests the cause's repr, so the scan log really is
the only place the CA message lands.

### Noted, not fixed (pre-existing, out of scope)

The same Phase-5 bullet cites `plans/liveness.py:29`, but `rd_confirmed_down`
is at line 26. It was already off by three at the merge base (`c26bb781`), so
this PR neither introduced nor worsened it. Flagging for whoever picks up
Phase 5 rather than widening this diff.

### Reviewer's closing line

> Nothing further. All three findings are resolved, and finding 1's fix is
> stronger than what I originally proposed.

skbarber:
## Review disposition, rounds 3–4 (after hardware verification)

The hardware run produced new material, so the reviewer took two more passes.
Both findings **fixed**, both confirmed by the same reviewer running the
reverts itself rather than taking my word for it.

| # | Finding | Disposition |
|---|---|---|
| 4 | `exc.__cause__ or exc` discards a falsy `CANothing` — the fix's own log line was broken | **Fixed** — `c48aea76` |
| 5 | The `step_scan.py` half of that fix had no regression pin | **Fixed** — `736659b6` |

### The reviewer retracted a round-1 claim

Its first report cited `step_scan.py:171` as the sibling that "already gets
this right" and recommended matching it. In its own words:

> I checked the format specifier and stopped there, never testing the `or`
> guard against `CANothing.__bool__` — and then cited that site as the model
> to copy, which is how the defect got carried into `single_shot.py`.

Worth recording, because it is why the bug reached hardware rather than being
caught in review.

### 4 — found by hardware, not by tests

The first live run printed `FailedStatus: <AsyncStatus …, done>`: no PV, no CA
message. `aioca.CANothing.__bool__` is `errorcode == ECA_NORMAL`, so a
**failed** put is falsy and `or` fell through to the `FailedStatus`. The
reviewer confirmed the second-order cause at
`ophyd_async/core/_status.py:103` — `if e := self.exception():` — which is why
the status renders `done` rather than `errored`. That is the same bug
upstream and not fixable from here, which is what makes the ERROR line the
only usable record for CA-layer failures.

Repo-wide sweep for a third site: none. The `GeecsError` tree defines no
`__bool__`, no CA return value is tested for truthiness, nothing tests a
`Status` for truthiness.

**How urgent this was**, from the reviewer, and the most useful sentence in
the whole review:

> The classification decision was never at risk; only the logging was. A
> falsy cause never caused a wrong refire, only an unreadable log line.

### 5 — the gap that let it happen twice

`single_shot`'s pin got `_PutRejected.__bool__`; `step_scan`'s did not.
`_FlakyMotor` raised a truthy `RuntimeError`, so `test_pause_semantics` passed
14/14 under **either** idiom — a correct fix on the operator-facing
`FAILED MOVE` line that any refactor could silently revert.

Fixed by making that fixture falsy. No new assertion; the existing
`assert "simulated move failure" in reasons[0]` became discriminating.
Reviewer-verified both ways:

```
736659b6 as-is                          -> 14 passed
revert only plans/step_scan.py to `or`  -> 1 failed, 13 passed
     test_failed_move_pauses_with_reason_and_resume_retries
```

### Reviewer's closing state

> Nothing further. All findings across four rounds are resolved, and two of
> them — the `%r` rendering and the missing `step_scan` pin — produced fixes
> stronger than what I originally proposed. The one substantive defect I got
> wrong (asserting `step_scan.py` was already correct) was caught by hardware,
> and the test that would have caught it now exists.

skbarber:
## Rebased onto master (#820 landed) — and a narrative correction

### Version renumbered

#820 merged as GeecsBluesky **0.77.0**, which this branch had also claimed.
Merged `origin/master` in (`bde2740e`), renumbered this to **0.78.0**, and
re-anchored the CHANGELOG entry above #820's. No code conflicts — the two
diffs are disjoint (`single_shot.py`/`step_scan.py` here vs
`motor.py`/`session.py`/`confirm.py` there); `CHANGELOG.md` and
`pyproject.toml` were the only overlap.

Post-merge: **888 passed, 1 skipped, 4 deselected**; lint clean. Both fixes
verified present in the merged tree.

### The falsy-cause finding is confirmed, and sharper than stated

`CANothing.__bool__` returns the *success* status, so it is falsy **exactly
when it represents a failure**:

| status | `bool(exc)` | `exc.__cause__ or exc` |
|---|---|---|
| `ECA_NORMAL` (success) | True | cause kept |
| a write-failure code | **False** | **cause discarded** |

Worst possible polarity: `or` drops the cause only in the case where there is
one. Independently reproduced on hardware during the #820 investigation — a
refused Aerotech `:SP` put raises exactly

```
CANothing: undulator:u_compaerotech:position_axis1:SP: Channel write request failed
```

which is the only object naming the PV and the CA message. On the `FAILED
MOVE` path that would have left the operator a bare `FailedStatus` repr. This
PR's reading of why it stayed hidden is right: #820's own failure raised
`GeecsMotorTimeoutError`, which is truthy.

### Correction to the Scan033 narrative

The PR body infers "one box taking out both the trigger write and the stage".
That inference is weaker than it looked, and the body should not be read as
having closed the root cause.

`u_modeimageresp/Position.Axis 1 -> -10.5 (tol=0.005)` in the Scan033 log is
the *same* move, target and tolerance that failed in Scan034 eight minutes
later — but for an unrelated reason that #820 has now fixed (the stage landed
at -10.505, exactly one tolerance out, and lost the comparison to binary
floating point by 8e-16). So the stage's appearance in this incident has its
own explanation and is not evidence of a shared-device-server fault.

Note also that in Scan033 the shot was attempted **1.4 s** after the move
started, so that move had already completed — it was not hanging or failing.

**The Scan033 `fire()` failure therefore remains unexplained.** That is an
argument for merging this, not against it: this PR does not fix that failure,
it makes the next occurrence report its real cause instead of blaming the
cameras. As it stands the same misleading message would appear again.

Issue #819, opened 2026-09-10
GEECS-Core: %.12f set formatting sends binary representation error to LabVIEW devices


skbarber:
## Summary

`GeecsUdpClient.set` formats every float set with `%.12f`
(`GEECS-Core/geecs_core/transport/udp_client.py:334`):

```python
elif isinstance(value, float):
    cmd = f"set{variable}>>{value:.12f}"
```

At 12 decimal places this stops being "extra digits" and starts exposing the
binary floating-point representation of the value — we send digits that are
not the number the caller asked for:

| value | what we transmit |
|---|---|
| `40854.24625` | `40854.246249999997` |
| `40865.8875` | `40865.887499999997` |
| `40966.0` | `40966.000000000000` |

Sampling 20 000 five-decimal values in `[40000, 41000]` (the working range of
`U_CompAeroTech/Position.Axis1`, in µm): **86.3%** format with a garbage tail
under `%.12f`.

## Impact

Observed on `U_CompAeroTech` as a fast set rejection — the device complains
the value is not a number and the caput fails in ~0.18 s, well before any
motion could occur. This is distinct from the check-values race also seen on
that device (fixed device-side by raising `DelayAfterMove` to 800 ms); the two
were interleaved during debugging, which made both harder to read.

Blast radius is the whole facility: this is the single code path for every
float set to every GEECS device, via the CA gateway and every client above it.

## Why the obvious fix is wrong

Truncating to one decimal handles the Aerotech (µm units, 1.0 µm DB tolerance)
but silently destroys other devices:

| value | source | `%.1f` |
|---|---|---|
| `0.001` | `U_ModeImagerESP/Position.Axis 2` DB tolerance | `0.0` |
| `1e-05` | `U_CompAeroTech/Speed.Axis2` DB minimum | `0.0` |

Silent truncation to zero on a position or speed set is a worse failure than
the one being fixed.

## Suggested direction

Bounded decimals with trailing zeros stripped — short, exact, never scientific
notation (which LabVIEW may not parse), no representation tail:

```python
f"{value:.6f}".rstrip("0").rstrip(".")
```

| value | today | proposed |
|---|---|---|
| `40854.24625` | `40854.246249999997` | `40854.24625` |
| `40966.0` | `40966.000000000000` | `40966` |
| `0.001` | `0.001000000000` | `0.001` |
| `1e-05` | `0.000010000000` | `0.00001` |

[...]

PR #820, opened 2026-09-10
GeecsBluesky 0.77.0: per-axis move tolerance from the DB, and stop losing on-boundary arrivals

skbarber:
## What

Two defects in `CaMotor`'s readback-confirmation poll (layer 2), both surfaced
by a live scan pausing for an operator on a move that had actually converged:

```
2026-09-10 12:10:28 ERROR geecs_bluesky.plans.step_scan scan=Scan034 -
FAILED MOVE - pausing for operator: commanded u_modeimageresp_position_axis_1
-> -10.5, one axis failed - see cause for which:
U_ModeImagerESP/Position.Axis 1: position -10.505 did not reach -10.5 within 30.0s
```

### 1. Exactly-on-tolerance arrivals failed (bug)

The stage landed exactly one tolerance from target and lost the comparison to
binary floating point:

```
|-10.505 - -10.5| = 0.005000000000000782
             <= 0.005 ?  False    (loses by 7.8e-16)
```

`motor.py` polled the full `move_timeout` and raised `GeecsMotorTimeoutError`
over 8e-16 mm. The arrival comparison now carries a relative epsilon.

This is **representation slack, not a widened tolerance** — a second test pins
that a miss of twice the tolerance still times out. Note `math.isclose` does
not help here: with `abs_tol=0.005` it reduces to the same comparison.

### 2. Every axis on every device used a hardcoded 0.005 (behaviour change)

`CaMotor.__init__` and `GeecsSession.motor()` both hardcoded `tolerance=0.005`,
and the scan path (`scan_request_runner.py:1263`) never passes one — so every
axis in every scan got that number. The GEECS DB already records a
per-variable `tolerance`, and the hardcoded value was wrong in **both**
directions:

| device / variable | DB tolerance | old hardcoded |
|---|---|---|
| `U_ModeImagerESP` / `Position.Axis 1` | 0.015 mm | 0.005 — failed converged moves |
| `U_ModeImagerESP` / `Position.Axis 2`, `Axis 3` | 0.001 mm | 0.005 — accepted 5x outside spec |
| `U_CompAeroTech` / `Position.Axis1` | 1.0 µm | 0.005 |

`GeecsSession.motor()` now resolves it through the new
`GeecsSession.variable_tolerance()`. Passing `tolerance=` still overrides.

The lookup is best-effort and cached per device: an unreachable DB, a missing
device/variable row, and the DB's "unset" spellings (NULL, `0.0`) all fall
back to `DEFAULT_TOLERANCE`. The `0.0` guard matters — `Speed.Axis1` carries
`0.0`, and a zero tolerance would demand bit-exact equality of a float
readback and never converge, turning every move into a `move_timeout`. Mock
sessions never query the DB, so hermetic tests stay hermetic.

## Heads-up for the first scan after this lands

This is **tighter** on `U_ModeImagerESP` axes 2 and 3 (0.001 vs 0.005), not
only looser on axis 1. Those axes have been silently accepting positions 5x
outside spec, so moves that used to pass may now legitimately time out. That
is the correct behaviour, but it will read as a new failure.

## Scope

One concern: the motor arrival check. LOC is ~55 in `session.py` (the resolver
+ cache), ~15 in `motor.py` (the epsilon + constants), the rest tests and
CHANGELOG. Three adjacent problems found during the same debugging session are
deliberately **not** here:

- **`U_CompAeroTech` short-move failures** — device-side, not ours. The
  LabVIEW `check values` step read the pre-move position verbatim (reported
  `actual value= 40965.810000` for a move that started at exactly that
  position), racing its own ~0.6-1.0 s position refresh. Fixed on the device
  by raising `DelayAfterMove` 200 ms -> 800 ms; 10 µm moves now clean. Layer 1
  failed there, so layer 2 never ran — this PR could not have fixed it.
- **`udp_client.py:334` `%.12f` set formatting** — filed as #819. Sends
  `40854.246249999997` for `40854.24625`; 86% of 5-decimal positions in the
  Aerotech's working range get a garbage tail. Whole-facility blast radius,
  needs its own PR and hardware verification.
- **Readback freshness gate** — `CaMotor` still reads the readback with zero
  delay after the put, into a lag measured at 0.6-1.0 s. Raising axis 1's
  tolerance to 0.015 makes a stale-read false arrival more reachable when the
  scan step is under 15 µm. Layer 1 blocks on device convergence so this does
  not corrupt data for honest devices, but it does quietly defeat layer 2 for
  exactly the early-ack devices (#779) layer 2 exists to catch. Judgment call
  to defer — cheap to veto if you want it bundled.

No facility literals added; no scan-folder paths touched; no gateway
PV-contract surface changed.

## Tests

`GeecsBluesky`: **847 passed, 2 skipped, 4 deselected** (deselected are
`integration`-marked — lab network, out of scope by design). Full
`./scripts/check.sh --all` results posted as a comment.

New pinning tests:

- `test_motor_arrival_exactly_on_tolerance_counts_as_arrived` — verified it
  reproduces the exact production error message when the epsilon is removed,
  and passes with it.
- `test_motor_beyond_tolerance_still_times_out` — the epsilon is slack, not a
  wider tolerance.
- `test_variable_tolerance_reads_the_db_value` — both directions (0.015 and
  the tighter 0.001).
- `test_variable_tolerance_falls_back_when_db_has_nothing_usable` —
  parametrized over `0.0`, NULL, and missing.
- `test_variable_tolerance_survives_an_unreachable_db`
- `test_variable_tolerance_is_cached_per_device` — one round trip per device.
- `test_mock_session_never_touches_the_db`

## Hardware verification

**OWED.** All findings above were measured on live hardware
(`U_CompAeroTech` move sweeps through the gateway, `U_ModeImagerESP` DB
tolerances read from the production DB), but the *fix itself* has only been
exercised against mock backends.

To verify after merge:

1. A step scan on `U_ModeImagerESP/Position.Axis 1` with steps that previously
   landed on the tolerance boundary — expect no `FAILED MOVE - pausing for
   operator`, where Scan034 paused.
2. Confirm `variable_tolerance("U_ModeImagerESP", "Position.Axis 1")` returns
   `0.015` against the live DB (not the `0.005` fallback) — a silent fallback
   would look like success while changing nothing.
3. A scan touching axes 2 or 3, which now run at the tighter 0.001 — this is
   where a new timeout would legitimately appear.

Co-Authored-By: Claude Opus 5 (1M context) <noreply@anthropic.com>

### DIFF
--- a/GeecsBluesky/CHANGELOG.md
+++ b/GeecsBluesky/CHANGELOG.md
@@ -30,9 +30,13 @@ Format follows [Keep a Changelog](https://keepachangelog.com/en/1.1.0/).
   per device rather than one per move) — an unreachable DB, a missing row, or
   the DB's "unset" spellings (NULL, `0.0`) fall back to `DEFAULT_TOLERANCE`,
   since a `0.0` tolerance would demand bit-exact float equality and never
-  converge. A DB tolerance more than 10x the default is served but logged as
-  suspect: that is the units-mismatch shape, and it would otherwise silently
-  confirm every move on the first poll. Mock sessions never query the DB.
+  converge. A DB tolerance larger than 1% of the variable's own `min`/`max`
+  travel span is served but logged as suspect: that is the units-mismatch
+  shape, and it would otherwise silently confirm every move on the first
+  poll. The comparison is span-relative rather than absolute because
+  tolerances carry each variable's own units (µm on `U_CompAeroTech`, mm on
+  the ESPs), so any fixed threshold flags correctly-configured axes for their
+  unit choice alone. Mock sessions never query the DB.
 
 - **The same on-boundary fix applied to `CaConfirmSettable`'s confirming
   poll** (`confirm.py`), which `build_movable` dispatches to *before*
--- a/GeecsBluesky/CHANGELOG.md
+++ b/GeecsBluesky/CHANGELOG.md
@@ -4,6 +4,33 @@ All notable changes to `geecs-bluesky` are documented here.
 
 Format follows [Keep a Changelog](https://keepachangelog.com/en/1.1.0/).
 
+## [0.77.0] - 2026-09-10
+
+### Fixed
+
+- **`CaMotor` no longer fails a converged move that lands exactly on its
+  tolerance.** `abs(-10.505 - -10.5)` is `0.005000000000000782` in binary
+  floating point, so a stage that arrived exactly one tolerance from target
+  failed `<= 0.005` by 8e-16, polled the full `move_timeout`, and paused the
+  scan for an operator (`U_ModeImagerESP/Position.Axis 1`, Scan034). The
+  arrival comparison now carries a relative epsilon — representation slack,
+  not a widened tolerance: a miss of twice the tolerance still times out.
+
+### Changed
+
+- **`GeecsSession.motor()` resolves the move tolerance from the GEECS DB**
+  (new `GeecsSession.variable_tolerance()`) instead of hardcoding `0.005` for
+  every axis on every device. The DB's per-variable `tolerance` is the
+  facility's own statement of what "arrived" means, and the hardcoded value
+  was wrong in both directions: `U_ModeImagerESP/Position.Axis 1` is
+  0.015 mm (where 0.005 failed converged moves) while axes 2 and 3 are
+  0.001 mm (where 0.005 silently accepted a position 5x outside spec).
+  Passing `tolerance=` still overrides. The lookup is best-effort and cached
+  per device — an unreachable DB, a missing row, or the DB's "unset"
+  spellings (NULL, `0.0`) fall back to `DEFAULT_TOLERANCE`, since a `0.0`
+  tolerance would demand bit-exact float equality and never converge. Mock
+  sessions never query the DB.
+
 ## [0.76.3] - 2026-09-08
 
 ### Changed
--- a/GeecsBluesky/CHANGELOG.md
+++ b/GeecsBluesky/CHANGELOG.md
@@ -19,17 +19,27 @@ Format follows [Keep a Changelog](https://keepachangelog.com/en/1.1.0/).
 ### Changed
 
 - **`GeecsSession.motor()` resolves the move tolerance from the GEECS DB**
-  (new `GeecsSession.variable_tolerance()`) instead of hardcoding `0.005` for
+  (new `GeecsSession.move_tolerance()`) instead of hardcoding `0.005` for
   every axis on every device. The DB's per-variable `tolerance` is the
   facility's own statement of what "arrived" means, and the hardcoded value
   was wrong in both directions: `U_ModeImagerESP/Position.Axis 1` is
   0.015 mm (where 0.005 failed converged moves) while axes 2 and 3 are
   0.001 mm (where 0.005 silently accepted a position 5x outside spec).
   Passing `tolerance=` still overrides. The lookup is best-effort and cached
-  per device — an unreachable DB, a missing row, or the DB's "unset"
-  spellings (NULL, `0.0`) fall back to `DEFAULT_TOLERANCE`, since a `0.0`
-  tolerance would demand bit-exact float equality and never converge. Mock
-  sessions never query the DB.
+  per device (failures too, so an off-network worker pays one connect timeout
+  per device rather than one per move) — an unreachable DB, a missing row, or
+  the DB's "unset" spellings (NULL, `0.0`) fall back to `DEFAULT_TOLERANCE`,
+  since a `0.0` tolerance would demand bit-exact float equality and never
+  converge. A DB tolerance more than 10x the default is served but logged as
+  suspect: that is the units-mismatch shape, and it would otherwise silently
+  confirm every move on the first poll. Mock sessions never query the DB.
+
+- **The same on-boundary fix applied to `CaConfirmSettable`'s confirming
+  poll** (`confirm.py`), which `build_movable` dispatches to *before*
+  `CaMotor` when a scan variable declares `confirm` — so for topology-C axes
+  it, not `CaMotor`, is the arrival check. `abs(1.20 - 1.15)` is
+  `0.050000000000000044`, so an EMQ set landing exactly on the 0.05 default
+  raised `GeecsConfirmTimeoutError` on a converged set.
 
 ## [0.76.3] - 2026-09-08
 
--- a/GeecsBluesky/geecs_bluesky/devices/ca/confirm.py
+++ b/GeecsBluesky/geecs_bluesky/devices/ca/confirm.py
@@ -27,6 +27,7 @@ from ophyd_async.core import AsyncStatus
 from ophyd_async.epics.core import epics_signal_r
 
 from geecs_bluesky.devices.ca._pv import ca_pv
+from geecs_bluesky.devices.ca.motor import _ULP_SLACK
 from geecs_bluesky.devices.ca.settable import CaSettable
 from geecs_bluesky.exceptions import GeecsConfirmTimeoutError
 
@@ -193,4 +194,9 @@ def _matches(
     """
     if datatype is str:
         return current == target
-    return abs(float(current) - float(target)) <= tolerance
+    # Same ULP slack as CaMotor's arrival check, and for the same reason: an
+    # exactly-on-tolerance match otherwise loses to representation error
+    # (|1.20 - 1.15| == 0.050000000000000044 > 0.05). This path takes
+    # precedence over kind: motor for topology-C axes, so it needs the fix too.
+    slack = _ULP_SLACK * max(abs(float(current)), abs(float(target)))
+    return abs(float(current) - float(target)) <= tolerance + slack
--- a/GeecsBluesky/geecs_bluesky/devices/ca/confirm.py
+++ b/GeecsBluesky/geecs_bluesky/devices/ca/confirm.py
@@ -27,7 +27,7 @@ from ophyd_async.core import AsyncStatus
 from ophyd_async.epics.core import epics_signal_r
 
 from geecs_bluesky.devices.ca._pv import ca_pv
-from geecs_bluesky.devices.ca.motor import _ULP_SLACK
+from geecs_bluesky.devices.ca.motor import ULP_SLACK
 from geecs_bluesky.devices.ca.settable import CaSettable
 from geecs_bluesky.exceptions import GeecsConfirmTimeoutError
 
@@ -198,5 +198,5 @@ def _matches(
     # exactly-on-tolerance match otherwise loses to representation error
     # (|1.20 - 1.15| == 0.050000000000000044 > 0.05). This path takes
     # precedence over kind: motor for topology-C axes, so it needs the fix too.
-    slack = _ULP_SLACK * max(abs(float(current)), abs(float(target)))
+    slack = ULP_SLACK * max(abs(float(current)), abs(float(target)))
     return abs(float(current) - float(target)) <= tolerance + slack
--- a/GeecsBluesky/geecs_bluesky/devices/ca/motor.py
+++ b/GeecsBluesky/geecs_bluesky/devices/ca/motor.py
@@ -37,10 +37,12 @@ _DEFAULT_MOVE_TIMEOUT = 30.0  # seconds
 #: resolves the per-axis DB value; this is what it returns when it cannot.
 DEFAULT_TOLERANCE = 0.005
 
-#: Layer-2 tolerances above this multiple of :data:`DEFAULT_TOLERANCE` are
-#: logged as suspect — almost certainly a units mismatch or an uncurated DB
-#: row rather than a genuinely coarse axis.
-_TOLERANCE_SANITY_FACTOR = 10.0
+#: A DB tolerance larger than this fraction of the variable's own travel span
+#: is logged as suspect.  Span-relative, not absolute: tolerances are in each
+#: variable's own units (µm on U_CompAeroTech, mm on the ESPs), so any fixed
+#: threshold flags correctly-configured axes purely for their unit choice —
+#: exactly the unit-blindness it is meant to detect.
+TOLERANCE_SPAN_FRACTION = 0.01
 
 # Binary floating point puts an exactly-on-tolerance arrival a few ULPs *over*
 # the limit: |-10.505 - -10.5| evaluates to 0.005000000000000782, not 0.005.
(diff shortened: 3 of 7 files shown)


skbarber:
## Local `./scripts/check.sh --all`

| suite | result |
|---|---|
| lint (pre-commit, all files) | pass |
| root-tests | 81 passed, 63 deselected |
| ImageAnalysis | 277 passed, 19 deselected |
| ScanAnalysis | 182 passed, 1 skipped, 2 deselected |
| GEECS-Data-Utils | 328 passed |
| GEECS-Schemas | 440 passed |
| **GeecsBluesky** | **847 passed, 2 skipped, 4 deselected** |
| GeecsCAGateway | 136 passed |
| GeecsPvaGateway | 29 passed |
| GEECS-Console | 895 passed |
| GEECS-Core | 90 passed, 5 deselected |
| GEECS-DataPortal | **did not run — not covered** (see below) |
| GEECS-LogTriage | 59 passed |
| GEECS-MCP | 103 passed, 1 skipped |

`integration`-marked tests are deselected throughout — they need the lab
network and are out of scope for this script by design.

### GEECS-DataPortal did not run

Collection errored on this machine with `ModuleNotFoundError: No module named
'fastapi'` and `... 'pandas'` — its local Poetry env is missing its own
dependencies, an environment gap rather than a code failure (`/env-doctor`
territory, tracked separately from this PR).

Recording it as **not covered** rather than passing. It is unaffected by this
change on two independent grounds: GEECS-DataPortal does not import
`geecs_bluesky` anywhere in its package or its `pyproject.toml`, and CI —
which builds its own environments — passed all six checks on this branch
(`pre-commit`, `unit-tests`, `console-windows`, two runs each).

skbarber:
## Adversarial review (fresh-context reviewer, full report)

Reviewed through the correctness / redundancy / placement lenses against the
whole repo. Posted in full; dispositions follow in a separate comment.

---

### 1. (High, correctness/design) Layer 2's threshold now equals layer 1's own convergence criterion — the backstop becomes a coincident re-check that can fail moves GEECS itself declared complete

The repo's own definition of the DB column, in two contract files:

- `GeecsCAGateway/PV_CONTRACT.md:481` — "The DB `tolerance` field is a **set-convergence** criterion, never a monitor deadband"
- `GeecsCAGateway/CLAUDE.md:135`, and the `deadband=0.0` comment at `GeecsCAGateway/geecs_ca_gateway/config.py:283`

And `motor.py`'s own module docstring says layer 2 is "belt-and-suspenders … It only adds information when the readback is an independent measurement."

So after this PR, layer 2 polls an independently-sampled readback against *exactly the number layer 1 already stopped on*. Scenario: `U_ModeImagerESP/Position.Axis 2`, DB tolerance 0.001 mm. The GEECS blocking set returns as soon as its own check sees |err| <= 0.001 (say 0.00095). `CaMotor._set_and_wait` then reads the streamed readback — a *later* sample, through the 0.6-1.0 s lag the PR body itself documents, with no freshness gate — and sees 0.0011: 30 s of polling, `GeecsMotorTimeoutError`, `FAILED MOVE - pausing for operator`, mid-scan. One micron of servo dither or backlash on an ESP stage is not exotic. The old hardcoded 0.005 gave 5x headroom that absorbed exactly this on axes 2/3.

The PR body anticipates these timeouts and calls them "the correct behaviour". On the repo's own definition of the column they are not: the device met the facility's stated spec, and only a differently-timed sample of the same axis disagreed. A backstop whose threshold is identical to the thing it backs up adds no information for honest devices and invents a scan-pausing failure for marginal ones. Suggest `tolerance = k * db_tolerance` with `k` documented (2-3), keeping the DB as the *source* while leaving layer 2 strictly looser. This is the one finding I'd want dispositioned before merge, because its failure mode is a paused scan on live beam.

### 2. (Medium, placement + redundancy) The DB read belongs in `db_runtime.py` — and the same rows are already fetched there and thrown away

`geecs_bluesky/db_runtime.py:206-234` (`GeecsDbServedSetProvider.served_by_device`) already calls `GeecsDb.get_experiment_device_variables(experiment)` — one batched whole-experiment query whose rows are `_variable_row_to_meta` dicts *including* `tolerance` (`GEECS-Core/geecs_core/db/geecs_db.py:186`) — and keeps only the names. That provider is constructed on every scan path (`scan_request_runner.py:1546`, `:1612`, `qs_client/submit_preflight.py:344`). The new code adds a *second*, per-device query plus a hand-rolled cache inside `session.py`, while `GeecsBluesky/CLAUDE.md:809-811` still states db_runtime is "the one place touching `GeecsDb`" for the scan runtime.

db_runtime's providers also take an injectable `db=` seam, which is why their tests never monkeypatch a global; the new tests must patch `geecs_core.db.geecs_db.GeecsDb.get_device_variables` module-wide. A `tolerances_by_device()` on that provider (or a sibling provider in the same module) is the natural home — a future per-component pseudo tolerance, a confirm tolerance, or a preflight "tolerance sanity" check would all have to move this code there.

Honest counterweight: `session.py:1751` (`_configure_assets` -> `GeecsDb.get_device_type`) and `capture/discovery.py:21` already touch `GeecsDb` outside db_runtime, and the provider's lifetime is per-scan while `motor()` also serves manual moves. So it's a placement preference plus a genuinely duplicated round trip, not a rule violation.

### 3. (Medium, correctness) No test pins the behaviour change itself

Every new session test calls `variable_tolerance()` directly; every new motor test constructs `CaMotor` directly with an explicit `tolerance=`. Nothing asserts that `GeecsSession.motor()` actually passes the resolved value into the device. Delete the two new lines in `motor()` (`if tolerance is None: tolerance = self.variable_tolerance(...)`) and restore the old `tolerance: float = 0.005` default, and the whole suite still goes green — the headline change is untested. `motor()` is also bound onto the `_DeferredConnectFactories` facade at plan time (`plans/scan_request_plan.py:283-314`, `:696`), which routes `self.variable_tolerance` back to the real session via `__getattr__` — I verified that works, but nothing pins it either. Fix is cheap: non-mock session with `GeecsSession._connect` monkeypatched to identity, assert `s.motor("U_ModeImagerESP", "Position.Axis 1")._tolerance == 0.015`.

### 4. (Medium, correctness) The identical on-boundary bug is left unfixed in `confirm.py`, on the path that *wins over* `kind: motor`

`geecs_bluesky/devices/ca/confirm.py:196`:

```python
return abs(float(current) - float(target)) <= tolerance
```

`build_movable` (`scan_request_runner.py:1255-1262`) dispatches to `confirm_settable` *before* `motor` whenever the scan variable declares `confirm`, so for topology-C axes this is the arrival check, not `CaMotor`'s. Concrete reproduction with that module's own default `_DEFAULT_TOLERANCE = 0.05` and EMQ-scale currents: target 1.15, readback 1.20 -> `abs(1.20 - 1.15) == 0.050000000000000044 > 0.05` -> `GeecsConfirmTimeoutError` on a converged set. The PR's deferred list names three adjacent problems but not this one, and it's the same defect two files over, fixable by sharing the same constant. Either bundle it or say explicitly why not.

### 5. (Low-Medium, correctness) The DB value is floored but not ceilinged — a mis-unit or uncurated row silently disables layer 2

The only guard is `tolerance > 0`. The facility mixes units across devices (the ESP reads ~10 in mm, the Aerotech ~4e4 in its units), the wholesale-inheritance rule means a devicetype default can land on an instance whose units differ (`geecs_db.py:_merge_variable_rows` docstring), and DB row curation is a known open problem. A stray `1.0` on a mm axis makes every readback "arrive" on the first poll, logged at `debug` only — reinstating precisely the silent false-arrival mode (#779) layer 2 exists to catch, and doing it invisibly. The old hardcoded 0.005 bounded that. A ceiling, or a `WARNING` when the DB value exceeds `DEFAULT_TOLERANCE` by more than ~10x, keeps the fix from having a blind side.

### 6. (Low, correctness) No negative caching: a DB outage costs a fresh 10 s connect on every `motor()` call, on the RunEngine loop

`_variable_tolerances` caches only success; the exception path in `variable_tolerance` re-queries on every call. The two existing providers deliberately do the opposite — `GeecsDbScalarPolicy` caches `{}` on failure (`db_runtime.py:104-113`) and `GeecsDbServedSetProvider` carries an explicit `_attempted` flag. Scenario: worker off-network, or MySQL host-blocked (error 1129, which this fleet has hit); a 3-axis scan pays 3 x `CONNECT_TIMEOUT_S` = 30 s (`GEECS-Core/geecs_core/db/geecs_db.py:88`), and because `motor()` runs bound to the in-plan facade, those synchronous connects block the RE's event loop rather than a setup thread. Repeated failed connects also feed MySQL's `max_connect_errors`. One line: cache `{}` on failure.

### 7. (Low, correctness) The epsilon scales with the tolerance, not with the operand magnitude — it under-covers exactly the large-coordinate axes

The representation error in `abs(current - value)` is ~`ULP(|position|)` ~= `|x| * 2.2e-16`; the slack granted is `tol * 1e-9`. That covers the case only while `|x|/tol <~ 4.5e6`. Measured: at the Aerotech's ~4.1e4-scale readback with the 0.005 fallback, the worst excess found over 20 k samples was 4.66e-12 against a 5.0e-12 threshold — 7% margin. Push the ratio further and it inverts: target `42317.5335`, readback `42317.5345`, tolerance `0.001` gives `|diff| = 0.0010000000038417056 > 0.001 * (1 + 1e-9)` — the identical timeout the PR is fixing. Robust form: `abs(current - value) <= tolerance + 4 * eps * max(abs(current), abs(value))`, or add an absolute floor term to the slack. Not reachable with today's device/tolerance pairs, hence low — but the comment claims the boundary case is solved, and it is solved only for small coordinates.

### 8. (Low, docs) Package contract file didn't travel

`GeecsBluesky/CLAUDE.md` is untouched: line 809-811 still names db_runtime as "the one place touching `GeecsDb`", line 470 documents `CaMotor`'s tolerance poll with no mention of DB resolution, and the new public `GeecsSession.variable_tolerance` is absent. Separately, the cache justification in the new docstring — "a gateway restart is already required for DB edits to take effect" — is inaccurate for this particular value: the gateway deliberately never consumes `tolerance` (`config.py:283`, `PV_CONTRACT.md:481`), so no restart was involved before; after this PR a DB tolerance edit needs a *worker/session* restart, i.e. closing and reopening the qserver environment. That matters because editing the DB is the operator's likely first response to the new axis-2/3 timeouts the PR predicts.

### Checked and clean

Scan-folder invariant (nothing path-shaped); facility literals (device names and tolerances appear only as docstring examples, which the root CLAUDE.md allows); version bump + CHANGELOG present and matching; gateway PV contract surface untouched, correctly; no import cycle from `session.py -> devices/ca/motor.py`; the `row.get("tolerance")` shape assumption is correct against the real `_variable_row_to_meta` (`float | None`, never `Decimal`/`str`); and no pre-existing tolerance resolver anywhere in the repo — this is the first consumer of that DB column, so the resolver itself is not duplicated.

skbarber:
## Review dispositions

Two rounds. The reviewer confirmed round 1 and found a defect that round 1
introduced; round 2 fixes it. Verdicts below are the reviewer's, not mine.

| # | Finding | Disposition | Reviewer verdict |
|---|---|---|---|
| 1 | Headroom multiple on the DB tolerance | **Waived** | waiver acceptable |
| 2 | Lookup belongs in `db_runtime` | **Waived** | waiver acceptable for a hotfix |
| 3 | No test pinned the behaviour change | **Fixed** (`32d7a003`) | confirmed fixed |
| 4 | Same on-boundary bug in `confirm.py` | **Fixed** (`32d7a003`) | confirmed fixed |
| 5 | DB tolerance floored but not ceilinged | **Fixed** (`32d7a003`, corrected `7f3aa363`) | round 1 newly broken; round 2 addresses it |
| 6 | No negative caching on DB failure | **Fixed** (`32d7a003`) | confirmed fixed |
| 7 | Epsilon scaled with tolerance, not magnitude | **Fixed** (`32d7a003`) | confirmed fixed |
| 8 | Package CLAUDE.md didn't travel | **Partially fixed** (`32d7a003`) | not a blocker |

### 1 — waived, owner's call

Use `k x db_tolerance` rather than the DB value itself, so layer 2 stays
strictly looser than layer 1's own convergence criterion.

Waived: the DB tolerances are set per axis from operational experience, so
they are the facility's considered statement of what "arrived" means, and no
scaling in code would be better informed than they are. Measured on live
hardware this session, axes converge well inside their tolerance rather than
at the edge, which is where the finding's dither scenario needs them:

| device / variable | DB tolerance | observed error | % of tolerance |
|---|---|---|---|
| `U_CompAeroTech/Position.Axis1` | 1.0 µm | 0.04-0.11 µm | 4-11% |
| `U_ModeImagerESP/Position.Axis 1` | 0.015 mm | 0.005 mm | 33% |

The residual risk (an axis parked at the very edge failing layer 2 on dither
alone) and the revisit condition are recorded in `move_tolerance`'s docstring.

### 2 — waived for this hotfix

Real, and the duplicated round trip is real: `db_runtime` already fetches
these rows and discards the `tolerance`. Waived because it is a refactor of a
shared provider rather than a fix, and `session.motor()` also serves manual
moves outside the per-scan provider's lifetime. Worth doing on its own.

### 5 — round 1 was wrong, round 2 fixes it

Round 1 used an absolute threshold (`10 x DEFAULT_TOLERANCE` = 0.05) against
values whose units vary per device — as unit-blind as the problem it was
meant to detect. Against the real production rows it fired on
`U_CompAeroTech/Position.Axis1`, a correctly configured 1.0 µm tolerance on
115000 µm of travel, on **every move**, claiming "moves will confirm on almost
any readback" when it is 0.0009% of range. The round-1 test enshrined the bug
by labelling the Aerotech's genuine value implausible.

`7f3aa363` compares against the variable's own `min`/`max` span instead —
already in the row — warning above 1% of travel. Aerotech stays silent; a
mis-unit'd 1.0 on a 25 mm axis is 4% and flags. Rows with no usable span are
not judged, and the check moved to cache construction so it logs once per
device rather than per `motor()` call. Tests pin both directions.

Also from that pass: `motor()`'s docstring still promised "layer-2 headroom
applied" after the headroom was removed, and `_ULP_SLACK` became `ULP_SLACK`
now that three modules share it.

### 8 — accepted as-is

`GeecsBluesky/CLAUDE.md` is not updated. Its line 809 claim that `db_runtime`
is "the one place touching `GeecsDb`" was already stale before this PR
(`session.py:_configure_assets`, `capture/discovery.py`), so this worsens an
existing inaccuracy rather than creating one. Belongs with the finding-2
refactor.

### Tests after both rounds

GeecsBluesky **854 passed, 2 skipped, 4 deselected**; `check.sh --lint` clean.

PR #821, opened 2026-09-11
GeecsBluesky 0.80.0: phase 1 PR 2 — the plan layer (#807)

skbarber:
Phase 1 PR 2 of the native-Bluesky rebuild (#807; plan of record `Planning/native_bluesky/03_clean_room_rebuild.md` §4.B–§4.D, §8, §10.7). Base: `feature/native-bluesky-plans` (after #816).

## What

A queue item naming a stock plan and namespace devices runs a **complete strict GEECS scan**: claimed scan number, native files in `ScanNNN/<device>/`, `ScanInfoScanNNN.ini`, the s-file (`ScanDataScanNNN.txt` + `analysis/sNNN.txt`), `scan.log`, and a `baseline` telemetry stream — hermetically proven end to end on mock CA backends (`tests/test_callbacks.py::test_a_strict_scan_leaves_every_legacy_file`).

Built in the agreed order:

1. **Registration table** — `plans/registry.py`. Every `bluesky.plans` verb with a `per_step`/`per_shot` hook that a queue item can express (18: `count`, `scan`, `list_scan`, `log_scan`, `grid_scan`, `list_grid_scan`, the spirals, `x2x_scan`, their `rel_*` twins — not `scan_nd` (Cycler arg), not the deprecated aliases) registered under its own name with the strict `take_reading` pre-bound; stock parameters kept minus the hook (vararg annotations dropped — the manager validates against them and `list_scan`'s says tuples), plus two keyword-only GEECS parameters, `trigger_profile` and `shots_per_step`. Each bound plan brackets the run ARMED → STANDBY through the profile's `ShotControl`; `TriggerProfiles` loads one device per configs-repo profile (experiment default from `experiment_defaults.yaml`). `plan_names.GEECS_PLAN_NAMES` pins the table; the operator permissions regex follows it.
2. **`claim_scan` preprocessor + `GeecsScanPathProvider`** — every run claims a day-scoped number on `open_run` (`scan_number`/`scan_id`/`scan_folder`/`experiment`/`scan_tag` into the start doc; a failed claim refuses the run); the one `PathProvider` the native-saving detectors share points at `ScanNNN/<GEECS device>/` for the open run and refuses outside one. Plus `preprocessors.scalar_headers` (staged devices' `Device Variable` headers → `geecs_scalar_headers`).
3. **Callbacks** — `callbacks.py`: `ScanInfoCallback` (the legacy `[Scan Info]` keys downstream parses, derived from the stock metadata; `ScanEndInfo` filled at the stop), `SFileCallback` (from the run's **own** primary events at the stop document — no Tiled round trip; `bin_number` from the GEECS `per_step`), `ScanLogCallback`. All write into the claimed folder, never creating it.
4. **Telemetry** — `SupplementalData(baseline=namespace.telemetry())`: every scalar-only device and every detector's scalar *signals* (never a detector — it would wait in ARMED for a shot that never comes), read at open and close.
5. **Preset v1** — `geecs_schemas.preset.Preset` (`devices[{device, save_images}]` + `plan{name, args, kwargs}` + `trigger_profile`, `background`, `description`; `plan` optional = a device group). `save_set.py`, both converters, their fixtures/golden files and the Console's save-set editor/store are deleted in the same change; the corpus is regenerated on the configs-repo branch **`presets-v1`** (83 presets across Undulator/Thomson/DataLogging; `save_devices/`, `scan_presets/`, `multiscan_presets/` removed there).
6. **Client seam** — `qs_client.submit_plan(name, args, kwargs)` / `submit_preset(preset)` with `presets.expand_preset` (device bindings, `Device:Variable`/catalog names → `U_S1H.current`, `X.scalars` for `save_images: false`, provenance `md`); preflight = expand + worker lists the plan + preset devices' `CONNECTED`. Readiness expects `GEECS_PLAN_NAMES` (unchanged mechanism, new list).

Also: `GeecsDetector.scalars` (the scalars-only view, `Triggerable` through the parent, no files), `BinCounter`, `make_run_engine(experiment, claim=, path_provider=, telemetry=)`, the startup profile wiring, `utils.identifier_name`/`device_reference` shared by namespace and client, `geecs_data_utils.write_scalar_files` (the exporter split so the callback can feed it a DataFrame).

## Decisions taken (the four open questions were left unfilled in the brief — defaults per the memory recommendations, recorded in §10.7; cheap to veto)

- **Telemetry** = baseline at open/close, not per-event everything (§10.5's wording): per-event reads of every device are the funnel's unstaged-read regression path again.
- **Claim** = every run; no md opt-out (md is provenance), hermetic switch `make_run_engine(claim=False)`.
- **Preset v1 fields** as above; `plan` optional so the 46 Undulator save elements regenerate as device groups without an invented plan call. `essential` is phase 2.
- **Two GEECS kwargs on the bound plans** — the one deliberate deviation from "stock signature preserved": both are facts of the scan and belong in its description.
- **s-file from the documents, written for any exit status with rows** (an aborted 500-shot scan's 300 rows are data, as the legacy scanner left them). Judgment call.
- **Corpus on a configs branch, not main** (the brief said direct to main): the deployed master worker still reads `save_devices/`; deleting them on main would break the lab's scans today. Merge `presets-v1` with the worker flip (PR 3).
- Settled from the code: the s-file headers ride in the **start document** (`geecs_scalar_headers`); the ScanInfo keys downstream parses are `Scan Parameter`, `Start`/`End`/`Step size`/`Shots per step`/`ScanMode`/`ScanStartInfo`, `Background`.

## Not in this PR (by the brief)

Console/MCP rewiring (they import unchanged; their submit paths call the removed funnel verbs and refuse at runtime; the Console's R2 save-set list reads empty), the worker flip (deployment untouched), #806, gated batch, the calibration plan, the cadence fix (PR 3 measures it), pseudo scan variables (`expand_preset` refuses them until phase 3).

## LOC breakdown (87 files, +3462 / −6088)

| Concern | Files | LOC |
|---|---|---|
| GeecsBluesky — registry, claim/provider, headers, callbacks, telemetry, detector view, client seam, tests, docs | 37 | +2830 / −1726 |
| GEECS-Schemas — `Preset` v1; `SaveSet` + converters + fixtures deleted; docgen/reference | 28 | +353 / −1829 |
| GEECS-Console — save-set editor/store deleted, listing degraded, tests | 13 | +36 / −2427 |
| GEECS-Data-Utils — `write_scalar_files` split | 5 | +92 / −24 |
| Planning §2/§4/§8/§10.7 + two docs status notes | 4 | +151 / −82 |

## Tests

- GeecsBluesky: **377 passed**, 1 skipped, 1 deselected (12 s) — was 367 (+18 new in `test_plan_registry.py`, `test_claim_scan.py`, `test_callbacks.py`, `qs_client/test_presets.py`; `test_sfile_callback.py` gone, `test_scan_log.py` rewritten, qs_client/listing/db_runtime tests rewritten for presets).
- GEECS-Schemas: **419 passed**, 7 deselected (incl. the schema-reference no-drift guard; corpus walk against `presets-v1` green).
- GEECS-Console: **814 passed** (offscreen).
- GEECS-MCP: **103 passed**, 1 skipped (imports unchanged).
- GEECS-Data-Utils `test_tiled_export.py`: **9 passed** (+2).
- `tests/test_phase0_hardware.py` not run (fires shots; `GEECS_HW=1` only).

## Hardware verification

**OWED — PR 3 (headless acceptance on HTU):** `submit_preset` of a regenerated preset (or `scan([UC_Amp4_IR_input], U_S1H.current, -1, 1, 5, shots_per_step=…, trigger_profile="HTU-NoGas")`) through the RE Manager → expect a claimed `ScanNNN/` with `UC_Amp4_IR_input/` PNGs named by the rows' stamps, `ScanInfoScanNNN.ini` with `Scan Parameter = "U_S1H Current"`, `Start/End = -1.0/1.0`, `Shots per step` as submitted, `analysis/sNNN.txt` with `Bin #` = the step, a `baseline` stream of ~2 rows in Tiled, ARMED before the first shot and STANDBY after the last, 1 Hz on a count and the every-other-edge cadence on the motor scan (M2) to be measured. Nothing here has touched hardware.

## Facility literals

None added outside docstrings/examples (`HTU-Normal`/`HTU-NoGas`/`Undulator` appear in docstrings, the docgen example and test fixtures only).


skbarber:
## Adversarial review (fresh-context reviewer, /land step 7) — report, verbatim

Checked against the installed libraries in `GeecsBluesky/.venv` (bluesky 1.15.0, ophyd-async 0.19.3, bluesky-queueserver 0.0.25), the regenerated corpus on the configs repo branch `presets-v1`, and a run of the GeecsBluesky unit suite (377 passed, matches the PR body).

### Findings

**1. P1 — `expand_preset` turns every `save_images: false` device into `X.scalars`, but only `GeecsDetector` has a `.scalars` child; 9 of the 13 submittable presets on `presets-v1` fail at the worker after cycling the trigger box.**
`GeecsBluesky/geecs_bluesky/qs_client/presets.py:99-104` emits `device_reference(d.device) + ".scalars"` for any device whose `save_images` is off. `GeecsBluesky/geecs_bluesky/devices/detector.py:551` is the only place `.scalars` is created; a `CaSnapshotReadable`/settable parent (magnet, stage, gauge, hall probe) has none. The regenerated corpus (`presets-v1`) carries `save_images: false` on scalar-only devices in `EMQ1_1DScan`, `EMQ2_1DScan`, `HP Compressor - E beam`, `HP-1000 shot noscan-E beam`, `HP-AllDiagnostics`, `HP-BasicEBeamNoScan`, `HasoBackground` (`U_ESP302_01/02`, `U_StretchterXMCC`, `U_ESP_JetXYZ`, `U_Hexapod`, `U_ModeImagerESP`, `U_VacuumGauge`, `U_BCaveHallProbe`, `U_BCaveMagSpecPS`, …) and on every magnet in `SteeringMagCurrents`. Reproduced hermetically: `expand_preset` of a preset with `U_S1H, save_images: false` yields `['UC_Cam', 'U_S1H.scalars']`; the manager's `validate_plan` passes it and `prepare_plan` hands the string through unchanged; the bound `count` then puts `DG Trigger.Source = single` (ARMED), raises `AttributeError: 'str' object has no attribute 'name'` inside `bp.count`'s md build, and puts STANDBY. So the hardware acceptance the PR body plans ("`submit_preset` of a regenerated preset") fails for most Undulator presets with a physical trigger-box cycle and an opaque error. This also contradicts the schema's own contract at `GEECS-Schemas/geecs_schemas/preset.py:52-60` ("Meaningless for a scalar-only device"). Tried to refute by checking whether the manager would refuse `U_S1H.scalars` at submit (it does not, see #2) and whether those devices could be `GeecsDetector`s (`looks_triggerable` needs a `trig*` variable — stages/magnets/gauges have none). Fix candidates: expand to `.scalars` only when the device is a detector (the manager's `devices_allowed` tree lists `scalars` under `components`, so the client can tell), or regenerate the corpus with `save_images: false` only on cameras and make the schema say so.

**2. P2 — "The typo fails at submit, not at queue-front" is false for this manager: unknown device names pass validation and reach the RunEngine as strings.** Claimed in `qs_client/presets.py:18-21`, `submit_preflight.py:3-4` and the PR body. Verified with `load_allowed_plans_and_devices` + `validate_plan` + `prepare_plan` exactly as the manager runs them: `count([["UC_Typo"]], num=1)` validates `True` and `prepare_plan` returns `[['UC_Typo'], 1]`; same for `U_S1H.scalars` and a bad motor `U_S1H.nope` in `scan`. The `Sequence[Readable]` annotation becomes `__READABLE__` but is not enforced against `allowed_devices`. The preflight that replaced the deleted unserved-variables check (`submit_preflight.py:145-150`) only reads `CONNECTED` for the preset's device names, so a typo (no PV → fail-open) and #1 both sail through; the failure is then the AttributeError above, after ARMED. Suggest a name check against `client.devices_allowed()` (the client already fetches `plans_allowed`) as a `validate` refusal, and rewording the two docstrings.

**3. P2 — Baseline telemetry is all-or-nothing and connects *after* the claim: one unconnectable device burns a scan number on every run.** `run_engine.py:120-131` installs `claim_scan_preprocessor` innermost and `SupplementalData(baseline=namespace.telemetry())` + `connect_on_demand` outside it, so on `open_run` the number is claimed (folder created, `ScanInfoScanNNN.ini` and `scan.log` written) before the baseline `declare_stream` triggers `ensure_connected(*all telemetry)`, which raises `NotConnectedError` if any one device fails. Reproduced hermetically with one never-served `CaSnapshotReadable` in the baseline: two `bp.count` attempts → both raise, `scans/Scan001` and `scans/Scan002` exist with the ini inside. Real trigger: a device added to the DB after the gateway started (the gateway roster is startup-only), or a worker restarted while the gateway lags — every scan then fails and each failure orphans a numbered folder that analysis code must never create or recreate. The funnel's per-run telemetry was scoped by the save set; here it is every subscribed device of the experiment (~99 on Undulator), so the blast radius of one bad PV is the whole experiment. Tried to refute via `ensure_connected` semantics (`wait_for_connection` aggregates and raises) and the preprocessor order in `RunEngine.__call__` (first-appended = innermost, confirmed). Suggest connecting the telemetry set once at worker start (loud, like the namespace build) and dropping unconnectable members with a warning, so a run's baseline can only contain devices already known to connect.

**4. P3 — The recorded `trigger_profile` is `ShotControl.profile_name` (the YAML `name` field), not the key the plan resolved.** `plans/registry.py:227` writes `md["trigger_profile"] = shot_control.profile_name`; `TriggerProfiles.resolve` keyed by the resolver's file stems. The PR's own tests show the drift (registered `"HTU-Test"`, start doc and ScanInfo record `"test"`, `test_plan_registry.py:187`). Legacy-converted profiles get `name=stem` so the corpus is safe today, but a new-schema profile whose `name` differs from its stem records a value `profiles.resolve()` would refuse — ScanInfo's `Trigger profile` and the provenance stop being replayable. Record the resolved key (the registry has it in hand).

**5. P3 — Test-coverage claim overstated: `test_a_strict_scan_leaves_every_legacy_file` does not exercise native saving through the run provider.** PR body and `GeecsBluesky/CLAUDE.md` cite it as proving "native files in `ScanNNN/<device>/`"; the test builds `_camera(RE, box, "UC_Cam")` with no path provider (`tests/test_strict_plans.py:99-116`), and `GeecsScanPathProvider` is only tested standalone (`tests/test_claim_scan.py`). I ran claim → `GeecsScanPathProvider` → `LvNativeFileDataLogic` through the bound `count` by hand and it works (`Scan001/UC_Cam/` created, `localsavingpath` put with the folder), so this is a gap in the pinning, not a bug — add a native-saving camera to that end-to-end test so the claim holds.

**6. P3 (redundancy) — `callbacks._RunCallback` re-implements `bluesky.callbacks.core.CallbackBase`'s name dispatch.** `callbacks.py:47-79` vs `CallbackBase.__call__` (dispatches `start/descriptor/event/stop`). The never-raise wrapper and the per-run start cache are legitimate additions but could sit on top of `CallbackBase` rather than beside it; minor.

Checked and not flagged: the failed-claim path (a `msg_mutator` raise is thrown back into the plan — `finalize_wrapper` sends STANDBY and `stage_wrapper` unstages, verified in the RE); `scan_id` override survives `_open_run`; `first_axis` against the stock `plan_pattern_args` shapes; `write_scalar_files` never creates `scans/ScanNNN/` (only `analysis/` beside `scans/`, which is allowed); no lab paths or account names in the diff; leftover `submit_scan`/`describe_action`/`move_variable` callers in GEECS-MCP and the Console are as the PR states (their preflight refuses first); version bumps and CHANGELOGs present for all four packages.

### Verdict

The plan layer itself is sound — the claim/provider/callback chain, the strict binding and the ARMED→STANDBY bracket all behave as documented under the real RunEngine, and the hermetic suite passes. What does not hold is the client seam's contract with the corpus and the manager: `save_images: false` on a scalar-only device (most of the regenerated Undulator presets) produces a device reference nothing refuses until the worker has already touched the trigger box, and the "fails at submit" premise the preflight was slimmed down to is not true of bluesky-queueserver 0.0.25. Together with the baseline telemetry burning a scan number per unconnectable device, these are the three things I would fix (or at least the first two) before PR 3 points a regenerated preset at HTU; the rest are accuracy and coverage nits.

skbarber:
## Verification of the fixes (fresh-context verifier — the original reviewer became unreachable (API rate limit) before confirming, so a separate agent was given only the finding text and the fix commit, per /land)

Environment: clone `GEECS-Plugins-native-bluesky`, branch `phase/02-plan-layer` @ `9cc8248c`, GeecsBluesky `.venv` (bluesky 1.15.0, ophyd-async 0.19.3, bluesky-queueserver 0.0.25). No files in the clone were modified; `tests/test_phase0_hardware.py` was not run. Baseline: the eight touched/related test files — **105 passed**. Scratch experiments were written outside the clone and run so they import the repo's own test helpers.

### 1. P1 — `X.scalars` on a scalar-only device: **CONFIRMED FIXED for the reported scenario — but the adjacent scenario (the scanned device listed with `save_images: false`) still fails after ARM**

- Every namespace device carries the view, rooted at itself: `dev.scalars` is a `ScalarsView` (a `GeecsDetectorScalars` iff the device is a `GeecsDetector`), `view.parent is dev`, `root_ancestor(view) is dev`, `view.name == f"{dev.name}-scalars"`, `children()` empty. So `stage_wrapper` stages the owner, and `scalar_headers`' descendant walk reaches the owner's `_column_headers`. `connect_on_demand`'s `is_namespace_object` walks `.parent`, so the view is recognised.
- The manager's tree lists it (`existing_plans_and_devices_from_nspace` → `U_S1H.scalars`, `UC_TestCam.scalars`) for `primary`, `operator` (depth 3) and `admin`.
- The original mechanism reproduced: `prepare_plan` passes `["U_Typo.scalars"]` through as a `str`; `["U_S1H.scalars"]` resolves to the exact view object. The bound `count` runs it: 2 fires, 2 events, keys include `u_s1h-voltage`, `u_s1h-current-position`, `uc_testcam-meancounts`.

**Break attempt that succeeded — collision matrix (`scan([detector], motor, -1, 1, 3)`, mock mode):**

| detector list | motor | result |
|---|---|---|
| `U_S1H` (whole) | `U_S1H.current` | OK — `separate_devices` drops the child |
| **`U_S1H.scalars`** | **`U_S1H.current`** | **FAILS after ARM, 1 shot fired, `exit_status: fail`**: `ValueError: Data keys (field names) from <CaMotor> collide with those from <ScalarsView>. The colliding keys are {'u_s1h-current-position'}` |
| `UC_TestCam.scalars` | `UC_TestCam.exposure` | FAILS the same way — pre-existing for `GeecsDetectorScalars` |
| `U_S1H.scalars` | `UC_TestCam.exposure` | OK |

`separate_devices` only removes ancestor/descendant redundancy; `X.scalars` and `X.current` are siblings, so both are read and the bundler refuses the duplicate key. This is the preset shape "list the scanned magnet as a device with `save_images: false`" (the optimizer-config converter emits `save_images=False` for every legacy device requirement). Not a regression relative to 2f19ecab, but the fix does not close the class.

### 2. P2 — unknown device names reach the plan as strings: **CONFIRMED FIXED for the preset path; three narrow gaps noted**

- Wrong-case detector `uc_cam1` and a scan of `U_S1H:Voltage` when the tree lists only `U_S1H.current` → refusal names both. `trigger_profile="HTU_Normal"` (identifier-shaped) is excluded by key; `md` is never walked. `flatten_device_tree` tolerates `None` entries.
- Gaps (none reachable from `expand_preset` output today): (1) nested lists / list-or-dict kwargs are not walked while the manager's `process_argument` does recurse; (2) a digit-first child segment (`U_S1H.2nd`) is accepted by the manager but not matched; (3) fail-open when `allowed_device_names()` itself raises on a ready manager (`worker_ready` recorded `skipped`).
- Doc nit: `utils.device_reference`'s docstring still says the manager refuses a collision at submission — the preflight does.

### 3. P2 — baseline telemetry all-or-nothing after the claim: **CONFIRMED FIXED; residual risk stated**

- Real CA mode, no lab network: an unservable `CaSnapshotReadable` is dropped with the warning and the installed baseline is `[]`; build returns instead of raising.
- Startup wiring: `qserver/startup/startup.py` → `namespace.telemetry()` → `make_run_engine(telemetry=…)` → `install_telemetry`; `connect_on_demand` re-appended last.
- Granularity is per device for scalar-only devices (one ghost DB row drops the device whole, loudly) and per signal for detectors.
- A member that connected at build can still fail the run after the claim: the CA backend's `get_reading` uses `caget(..., timeout=None)`, so a PV that disconnects after build makes the baseline read hang rather than fail. Inherent ophyd-async behaviour, applies to the primary stream too; the fix does not claim to address it, but the baseline set's blast radius is the whole experiment.

### 4. P3 — `trigger_profile` recorded as the resolved key: **CONFIRMED FIXED**

`strict_plan` records the argument or `profiles.default`, exactly what `TriggerProfiles.resolve` looks up; the tests discriminate (YAML `name="test"` under key `"HTU-Test"`, both the start doc and the ini now read `"HTU-Test"`).

### 5. P3 — native saving exercised through the run provider: **CONFIRMED FIXED**

Variant A (a fresh provider nobody points at the run): the run aborts with `GeecsConfigurationError('no scan claimed …')`, `Scan001/UC_Cam/` absent. Variant B (the pre-fix camera, no provider): the new assertions fail. Variant C (the run provider): `Scan001/UC_Cam/` exists, `save == "off"`, `localsavingpath` ends with it, `provider.folder is None` after `close_run`. The leaf directory is created by `LvNativeFileDataLogic.prepare_single`, not by the fake claim, so the assertion genuinely proves the provider was pointed at the run.

### 6. P3 (redundancy) — `_RunCallback` vs `CallbackBase`: **WAIVER OK**

Adopting `CallbackBase` (`event_model.DocumentRouter`) would still require overriding `__call__` (never-raise) and `start`/`stop` (the per-run start cache), so the net saving is roughly the four `elif` lines; the event-page repacking is moot on this path.

**Summary:** 2, 3, 4, 5 confirmed fixed (with the residuals listed under 2 and 3); 6 waiver OK; 1 fixed for the reported scenario, but `scan([U_S1H.scalars], U_S1H.current, …)` still fails after ARM with a data-key collision and is not caught by the new preflight. That one deserves a disposition before merge.

skbarber:
## Verifier round 2 (same fresh-context verifier), on a82a227d — report verbatim

Repo tests on `a82a227d`: `test_plan_registry.py`, `qs_client/test_presets.py`, `test_strict_plans.py`, `qs_client/test_submit_preflight.py`, `test_callbacks.py` — **54 passed**. Scratch experiments outside the clone, built on **namespace-constructed** devices (`GeecsNamespace(tests.test_namespace.ROSTER)`: the magnet's `current` is a subscribed `CaMotor`, `enable_output` a settable-but-not-subscribed `CaSettable`, the camera's `exposure` a subscribed `CaSettable`), i.e. the shapes the worker actually binds.

### (a) `ScalarsView.covers` + strict `take_reading` dedup — **CONFIRMED FIXED** (one contrived residual)

`covers` behaves as claimed on namespace devices: `magnet.scalars.covers(magnet.current)` True, `.covers(magnet.enable_output)` False, `.covers(magnet.voltage)` True, `cam.scalars.covers(cam.exposure)` and `.covers(cam.meancounts)` True, cross-device False.

Matrix, `scan([detectors], motor, -1, 1, 3)` on a mock RE, all after ARM:

| detectors | motor | result |
|---|---|---|
| `U_S1H.scalars` | `U_S1H.current` (subscribed child) | **OK** — 3 fires, positions `[-1, 0, 1]` via the view, `u_s1h-current-position` once |
| `UC_TestCam.scalars` | `UC_TestCam.exposure` (CaSettable child) | **OK** — 3 fires, `uc_testcam-exposure-readback` once |
| `U_S1H.scalars` | `UC_TestCam.exposure` | OK, both columns present |
| `UC_TestCam.scalars` | `U_S1H.current` | OK |
| `[U_S1H.scalars, U_S1H]` | `U_S1H.current` | OK — `separate_devices` drops the view |
| `[U_S1H.scalars, U_S1H.voltage]` | `U_S1H.current` | OK — the signal is covered, one `u_s1h-voltage` |
| `[U_S1H.scalars, UC_TestCam.scalars]` | `U_S1H.current` | OK |
| `count([U_S1H.scalars, U_S1H.enable_output], 2)` (child **not** subscribed) | — | OK — `u_s1h-enable_output-readback` stays in the row exactly once |
| `[UC_TestCam.scalars, UC_TestCam.acq_timestamp]` | `U_S1H.current` | **still collides**: `{'uc_testcam-acq_timestamp'}`, 1 fire, `exit_status: fail` |

The last row is the residual: for a `GeecsDetector` owner `covers` tests membership in `_scalars`, but `GeecsDetectorScalars.read()` reads `_scalar_signals()` = `_scalars` plus `acq_timestamp`. Reachable only from a hand-typed plan (`expand_preset` never names a signal) — a nit; a one-line fix is to test against `owner._scalar_signals()`.

### (b) `device_references` nested walk + digit-first segment — **CONFIRMED FIXED**; gap (3) **WAIVER OK**

Nested shapes are walked the way the manager's `process_argument` walks them (lists, tuples, dicts; a bare string at `args[0]` too); `md` and `trigger_profile` still excluded; regex accepts `U_S1H.2nd`, `U_S1H.current.setpoint`, rejects `HTU-Normal`, `U_S1H.`, `U_S1H:Current`, trailing-space spellings. Note for a future enum-valued list scan: an identifier-shaped literal string argument is treated as a reference and refused unless the tree lists it — no registered stock plan takes one today. Gap (3) waiver: reasonable — the same posture as an unanswered `status()`/`plans_allowed()`, the plan-name and environment gates have already passed, and the submit itself still surfaces a broken manager.

### (c) `utils.device_reference` docstring — **CONFIRMED FIXED**

**Summary:** (a) CONFIRMED FIXED for every asked scenario, with one contrived residual; (b) CONFIRMED FIXED, gap (3) WAIVER OK; (c) CONFIRMED FIXED. Nothing here needs to block the merge; the round-1 residuals under findings 2 and 3 (fail-open breadth, the baseline's `caget(timeout=None)` hang class) stand as previously stated.

skbarber:
## Dispositions (every finding, per /land step 7)

The original reviewer was terminated by an API rate limit before it could confirm the fixes; a **fresh-context verifier** was given only the finding text and the fix commits (substitution noted per the skill). Commits: 9cc8248c (findings 1–5), a82a227d (the verifier's adjacent case + preflight gaps), e7a27a96 (the round-2 residual).

| # | Finding | Disposition |
|---|---|---|
| 1 | P1 — `X.scalars` only existed on detectors; scalar-only presets failed after ARM | **Fixed** (9cc8248c): every namespace device carries `X.scalars` (`devices/ca/_view.py` `ScalarsView`). Verifier: confirmed, then found the adjacent collision `scan([U_S1H.scalars], U_S1H.current, …)` → **fixed** (a82a227d): a view yields the child's column to a listed child it already covers (`ScalarsView.covers`, strict `take_reading`). Verifier round 2: confirmed on the full matrix; residual `[cam.scalars, cam.acq_timestamp]` → **fixed** (e7a27a96): `covers` checks the detector's full column set. |
| 2 | P2 — unknown device names reach the plan as strings | **Fixed** (9cc8248c): `QueueClient.allowed_device_names()` + the preflight refuses references the manager's tree does not list; docstrings reworded. Verifier gaps (1) nested values and (2) digit-first segments → **fixed** (a82a227d). Gap (3) fail-open when `devices_allowed` itself raises on a ready manager → **waived**: consistent with the documented fail-open for an unanswered manager read; the plan-name and environment gates have already passed and the submit still surfaces a broken manager (verifier: waiver OK). |
| 3 | P2 — baseline telemetry connected after the claim, all-or-nothing | **Fixed** (9cc8248c): `install_telemetry` connects the set once at build and drops unconnectable members loudly. Verifier: confirmed. Residual noted by the verifier (a PV that disconnects *after* build makes an unbounded `caget` hang, inherent to ophyd-async and shared with the primary stream) → **waived** for this PR as out of the finding's scope; recorded as a PR 3 / phase-2 watch item (a bounded baseline read or per-run drop). |
| 4 | P3 — `trigger_profile` recorded as the YAML name, not the resolved key | **Fixed** (9cc8248c). Verifier: confirmed. |
| 5 | P3 — end-to-end test did not exercise native saving through the run provider | **Fixed** (9cc8248c): the camera runs on the run's `GeecsScanPathProvider`; the verifier showed the assertions fail under a fresh provider and under the pre-fix camera. Confirmed. |
| 6 | P3 — `_RunCallback` beside `CallbackBase` | **Waived**: adopting `CallbackBase` would still need `__call__` (never-raise) and `start`/`stop` (the per-run start cache) overridden; net saving is four `elif` lines. Verifier: waiver OK. |

Suite after the three fix commits: GeecsBluesky **385 passed**, 1 skipped, 1 deselected; CI green on a82a227d (pre-commit, unit-tests, console-windows); e7a27a96 running.

Judgment calls still open for the maintainer (from the PR body): telemetry as baseline rather than per-event; every run claims; `plan` optional on `Preset`; the two GEECS kwargs on the bound plans; s-file written for any exit status with rows; the corpus regeneration on the configs branch `presets-v1` rather than main.

skbarber:
## Codex review — dispositions (fb3c239d)

| Finding | Disposition |
|---|---|
| `presets.py` treated any identifier-looking string in nested plan args as a device reference, so `list_scan(…, ["on", "off"])` was refused as unknown devices `on`, `off` | **Fixed**: the string heuristic is gone. `expand_preset` records the references it *creates* — the detector bindings and every `Device:Variable` / catalog name it resolved — in `QueueItem.references`, and the preflight checks exactly those against the manager's device tree. A literal string argument is never a reference. Pinned by `tests/qs_client/test_presets.py::test_expansion_records_its_references_and_leaves_literal_strings_alone` and `test_submit_preflight.py::TestWorkerReady::test_literal_string_arguments_are_not_device_references`. |
| `plan.name == "mv"` passed because `mv` is in `GEECS_PLAN_NAMES`, expanding to the invalid `mv([detectors], motor, value)` | **Fixed**: `PRESET_PLAN_NAMES` = the scan verbs (`GEECS_PLAN_NAMES` minus `mv`); `expand_preset` refuses `mv` naming `submit_plan("mv", …)` as the manual move. Pinned by `test_presets.py::test_a_preset_cannot_name_mv`. |

GeecsBluesky: 386 passed, 1 skipped, 1 deselected.

Issue #738, opened 2026-09-01
Retire native per-shot image saving: consumer checklist + image analysis as a PV producer


skbarber:
## The design question

The capture daemon writes one HDF5 frame stack per device per scan
(`<device>.h5`, `geecs-capture/1`). Today every consumer reads a stack
**after** the scan closes it. Several use cases — one immediate, the rest
on the roadmap — need to read a stack **while the daemon is still
appending to it**, and we have no contract for that. This issue is the
home for that discussion; the optimize toggle below is the first forcing
use case, not the scope.

This matters beyond the capture arc: the services-box data-format plan
points at one-HDF5-per-device-per-scan as *the* format (the file-count
pathology fix), at which point the live-read contract stops being a
capture-daemon detail and becomes load-bearing for the whole fleet.

## Consumers that want live stack reads

1. **Optimize evaluators (per bin, mid-scan)** — the immediate case.
   Optimize mode currently *ignores* `capture.native_image_save=false`
   with a warning ("v0: evaluators read per-shot files"), pinned in both
   `run_scan_request` and `_scan_request_body`. The analysis side has
   since grown `data_format: device_hdf5`
   (`SingleDeviceScanAnalyzer`, PNG fallback, first production opt-in:
   `Amp4Input`, configs `e4083ce`), so the pin is conditionally
   liftable — but the evaluator reads *per bin while the stack is being
   written*, exactly the unspecified case. The PNG fallback silently
   masks any failure whenever PNGs exist, so this only gets honestly
   tested with native saving genuinely off.
2. **Live quick-look during a scan** (scan browser / data portal
   showing the current scan's images) — post-hoc today; loses its
   source when native files go away.
3. **Any future during-scan analysis hook** (per-shot/per-bin analysis
   beyond optimization — the deferred "event-builder service" territory).
4. **Post-scan analysis of an aborted scan** whose stack was never
   cleanly finalized — the degenerate live-read case.

## Option space (for the discussion)

- **(a) HDF5 SWMR** — the daemon's `FrameStackWriter` opens in
  single-writer-multiple-readers mode; readers poll/refresh. The classic
  answer; constrains libver/container choices (the writer is deliberately
  swappable behind `FrameStackWriter`), and readers over SMB add their
  own visibility caveats.
- **(b) Reopen-per-read with retry** — readers reopen the file each bin
  and treat any read failure as "not yet". Cheapest; racy without SWMR
  guarantees, and failure shapes are opaque.
[...]

PR #822, opened 2026-09-11
GeecsBluesky 0.80.1: phase 1 PR 3 — headless hardware acceptance of the plan layer (#807)

skbarber:
Phase 1 PR 3 of the native-Bluesky rebuild (#807; plan of record `Planning/native_bluesky/03_clean_room_rebuild.md` §8, §10.8). Base: `feature/native-bluesky-plans` (after #821).

## What

**Headless hardware acceptance of the plan layer on HTU, done.** `GeecsBluesky/tests/test_phase1_hardware.py` (`GEECS_HW=1`, fires shots; skips otherwise):

1. **In process** — the worker's own wiring (`make_run_engine(claim=True, path_provider, telemetry=namespace.telemetry())`, the namespace on the shared `GeecsScanPathProvider`, `TriggerProfiles` from the configs repo, the bound plans) runs a strict `count` (3 shots) and a strict `scan` of `U_S1H:Current` −1 → +1 A in 5 points × 2 shots, and asserts every GEECS output from disk: the claimed folder, the camera's native PNGs named by the rows' stamps in `ScanNNN/UC_Amp4_IR_input/`, `ScanInfoScanNNN.ini` (the keys downstream parses), the s-file with `Bin #` per step, `scan.log`, the `baseline` stream (419 columns, 2 rows per run), ARMED → STANDBY, setpoint restored. **Scans 104 and 105 of 26_0910, `1 passed in 72 s`.**
2. **Through the RE Manager** — a second manager on the host (own ports, own Redis prefix, document publisher off; runbook in `05_phase1_acceptance.md`): a `Preset` → `run_submit_preflight` (`validate`, `worker_ready` incl. the device tree, `gateway_liveness` all passed) → `submit_preset` → the history item `completed` → the same files. **Scan 108 (and 106 on the first run), `1 passed in 39 s`.**

`Planning/native_bluesky/05_phase1_acceptance.md` carries the runbook, the M4–M6 records and the flip record; §2/§8/§10.8 of the plan of record updated; GeecsBluesky 0.80.1 (test + docs).

## What the shots measured (M6, `05_phase1_acceptance.md`)

Per shot on this camera at 1 Hz: fire put ~108 ms, frame message ~886 ms after it, plan-layer work between frame arrival and the next fire ~7 ms — about **100 ms of margin** for anything else per shot. In process a `count` holds 1 Hz (`[2.0, 1.0, 1.0, 1.0, 1.0]`); through the manager's worker process the repeat shot lost an edge (2 s). A 0.5 A `U_S1H` move is a **1.3 s blocking set**, so a moved step lands on the third edge (`3.0 s`; M2 saw the second — a faster move that day). Conclusion recorded in §10.8: strict single-shot is not the 1 Hz mode, phase 2's gated batch is; the "cadence fix" is scoped accordingly.

## The worker flip — prepared, one command owed to the maintainer

Done as the service account on the host: the acceptance manager stopped; `~/qs-checkout` (the deployed worker's clone) checked out on `feature/native-bluesky-plans` at the #821 merge, `poetry install --extras "ca tiled qserver"` clean (optimize extra left in place), the plan layer importable in the production env. **The running `geecs-qserver` still holds master's code**: restarting the unit needs sudo, which the service account cannot do non-interactively.

```bash
sudo systemctl restart geecs-qserver                 # pulls geecs-qserver-ready along
journalctl -u geecs-qserver-ready -n 3 --no-pager    # expect "ready: 19 allowed plans, all 19 expected present"
sudo systemctl restart geecs-capture                 # optional: same clone's code
```

After that the deployed worker runs stock plans over presets; the master Console and MCP can no longer submit until their rewire (their preflight refuses first) — the accepted state of §10.5. The clients' configs stay on the share's `main`; the worker reads no presets, so `presets-v1` merges with the client rewire.

## Two lessons from the first manager run (fixed in the test)

- A status poll right after `queue_start` still reads idle — "done" is the item reaching the manager's history (`_wait_for_item`).
- The restore `mv` submitted while the scan item was still queued was refused by the #648 front-of-queue guard and left `U_S1H` at 1 A until I restored it by hand; the test now awaits its own item and restores with `clear_pending=True`. Nothing else was touched: every run restored the setpoint (0.0 A now).

## Tests

- GeecsBluesky: **386 passed**, 2 skipped (the two hardware modules), 1 deselected (52 s).
- Hardware: the two acceptance tests above, on the host.

## LOC

| Concern | Files | LOC |
|---|---|---|
| the acceptance test | 1 | +365 |
| `05_phase1_acceptance.md` + §2/§8/§10.8 + CLAUDE.md/README/CHANGELOG/version | 6 | ~+220 / −20 |

## Facility literals

`Undulator`, `UC_Amp4_IR_input`, `HTU-NoGas`, `U_S1H:Current` are the test's defaults, overridable by `GEECS_HW_*` environment (the same shape as `test_phase0_hardware.py`); host paths in the runbook are the service account's `~` and the site's share path in a docs example.

### DIFF
--- a/GeecsBluesky/CHANGELOG.md
+++ b/GeecsBluesky/CHANGELOG.md
@@ -4,6 +4,24 @@ All notable changes to `geecs-bluesky` are documented here.
 
 Format follows [Keep a Changelog](https://keepachangelog.com/en/1.1.0/).
 
+## [0.80.1] - 2026-09-10
+
+### Added
+
+- **Phase 1 PR 3 (#807) — headless hardware acceptance of the plan
+  layer.** `tests/test_phase1_hardware.py` (`GEECS_HW=1`; fires shots): the
+  worker's own wiring drives a strict `count` and a strict `scan` of
+  `U_S1H:Current` on HTU and every GEECS output is asserted from disk
+  (claimed folder, native files named by the rows' stamps, ScanInfo keys,
+  s-file bins, scan.log, the baseline stream, ARMED → STANDBY); and a
+  `Preset` goes through a second RE Manager (`run_submit_preflight` →
+  `submit_preset` → the history item) with the same assertions.  Accepted
+  2026-09-10 (Scans 104–108 of 26_0910; `Planning/native_bluesky/05_phase1_acceptance.md`
+  M4–M6, with the runbook for the second manager and the per-shot budget:
+  ~7 ms of plan-layer work per shot, ~100 ms of margin at 1 Hz on this
+  camera, a moved step on the third edge because a 0.5 A magnet move is a
+  1.3 s blocking set).
+
 ## [0.80.0] - 2026-09-10
 
 ### Added
--- a/GeecsBluesky/CHANGELOG.md
+++ b/GeecsBluesky/CHANGELOG.md
@@ -6,6 +6,18 @@ Format follows [Keep a Changelog](https://keepachangelog.com/en/1.1.0/).
 
 ## [0.80.1] - 2026-09-10
 
+### Fixed
+
+- `utils.settable_attribute` is the **one** rule for the attribute a
+  settable binds to (a frozen `RESERVED_DEVICE_ATTRIBUTES`, pinned to
+  `dir(GeecsDetector)` by a test): the namespace and the client seam now
+  agree, so `"UC_Amp4_IR_input:trigger"` in a preset spells
+  `UC_Amp4_IR_input.trigger_` instead of a reference the preflight refuses
+  (Codex review of #821).
+- The hardware acceptance test asserts its restore move *completed* and
+  reads the setpoint back, not merely that it queued (Codex review of
+  #822).
+
 ### Added
 
 - **Phase 1 PR 3 (#807) — headless hardware acceptance of the plan
--- a/GeecsBluesky/CLAUDE.md
+++ b/GeecsBluesky/CLAUDE.md
@@ -17,11 +17,15 @@ the s-file, `scan.log`, the baseline telemetry stream.  The worker
 registers the stock `bluesky.plans` verbs under their own names with the
 strict `take_reading` pre-bound (`plans/registry.py`); a client submits a
 stock plan item or a saved preset (`qs_client.submit_plan` /
-`submit_preset`).  Next (PR 3): headless hardware acceptance on HTU
-(`U_S1H:Current` −1 → +1 A, amp4in) and the worker flip.  **The deployed
-worker stays on `master` until then**; the Console and GEECS-MCP are
-rewired once, when the foundation is stable — not per step (their submit
-paths call the removed funnel verbs meanwhile).
+`submit_preset`).  Hardware-accepted 2026-09-10 (PR 3,
+`tests/test_phase1_hardware.py`, `Planning/native_bluesky/05_phase1_acceptance.md`):
+in process and through a second RE Manager, Scans 104–108 of 26_0910.  The
+deployed worker's checkout is on the feature branch; the restart that
+flips it is the maintainer's.  The Console and GEECS-MCP are rewired
+once, when the foundation is stable — not per step (their submit paths
+call the removed funnel verbs meanwhile).  Per-shot budget: ~7 ms of
+plan-layer work, ~100 ms of margin at 1 Hz on this camera — strict
+single-shot is not the 1 Hz mode, phase 2's gated batch is.
 
 ## The two rules
 
--- a/GeecsBluesky/CLAUDE.md
+++ b/GeecsBluesky/CLAUDE.md
@@ -19,9 +19,8 @@ strict `take_reading` pre-bound (`plans/registry.py`); a client submits a
 stock plan item or a saved preset (`qs_client.submit_plan` /
 `submit_preset`).  Hardware-accepted 2026-09-10 (PR 3,
 `tests/test_phase1_hardware.py`, `Planning/native_bluesky/05_phase1_acceptance.md`):
-in process and through a second RE Manager, Scans 104–108 of 26_0910.  The
-deployed worker's checkout is on the feature branch; the restart that
-flips it is the maintainer's.  The Console and GEECS-MCP are rewired
+in process and through a second RE Manager, Scans 104–108 of 26_0910 (the
+worker flip is recorded there).  The Console and GEECS-MCP are rewired
 once, when the foundation is stable — not per step (their submit paths
 call the removed funnel verbs meanwhile).  Per-shot budget: ~7 ms of
 plan-layer work, ~100 ms of margin at 1 Hz on this camera — strict
--- a/GeecsBluesky/README.md
+++ b/GeecsBluesky/README.md
@@ -45,9 +45,9 @@ deleted the `ScanRequest` funnel, the free-run mode, `GeecsSession` and
 the funnel-only devices; PR 2 added the plan layer — the stock plans
 registered strict under their own names, every run claiming a scan
 number, the ScanInfo / s-file / `scan.log` callbacks, the baseline
-telemetry stream, presets as the saved queue item.  Next: headless
-hardware acceptance and the worker flip.  Until then the deployed worker
-stays on `master`.
+telemetry stream, presets as the saved queue item.  PR 3 accepted it on
+HTU (Scans 104–108 of 26_0910, in process and through a second RE
+Manager); the worker flip is the maintainer's restart.
 
 ## Requirements
 
--- a/GeecsBluesky/README.md
+++ b/GeecsBluesky/README.md
@@ -47,7 +47,7 @@ registered strict under their own names, every run claiming a scan
 number, the ScanInfo / s-file / `scan.log` callbacks, the baseline
 telemetry stream, presets as the saved queue item.  PR 3 accepted it on
 HTU (Scans 104–108 of 26_0910, in process and through a second RE
-Manager); the worker flip is the maintainer's restart.
+Manager — `Planning/native_bluesky/05_phase1_acceptance.md`).
 
 ## Requirements
 
--- a/GeecsBluesky/geecs_bluesky/namespace.py
+++ b/GeecsBluesky/geecs_bluesky/namespace.py
@@ -69,7 +69,7 @@ from geecs_bluesky.devices.ca.settable import CaSettable
 from geecs_bluesky.devices.ca.snapshot import CaSnapshotReadable
 from geecs_bluesky.devices.detector import GeecsDetector
 from geecs_bluesky.exceptions import GeecsConfigurationError
-from geecs_bluesky.utils import identifier_name, safe_name
+from geecs_bluesky.utils import identifier_name, safe_name, settable_attribute
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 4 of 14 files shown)


skbarber:
## Adversarial review (fresh-context reviewer, /land step 7) — report, verbatim

Scope reviewed: the whole diff (`tests/test_phase1_hardware.py`, `05_phase1_acceptance.md`, `03_clean_room_rebuild.md` §2/§8/§10, `CLAUDE.md`/`README.md`/`CHANGELOG.md`/version). Reproduced: `386 passed, 2 skipped, 1 deselected` with the CI marker expression; `test_phase1_hardware.py` skips without `GEECS_HW=1`; CI (pre-commit, unit-tests, console-windows) green. Nothing was run against hardware.

### Findings

1. **P2 — the manager test's restore is refused in exactly the cases it exists for** — `GeecsBluesky/tests/test_phase1_hardware.py:341-353`.
   `ZmqQueueClient._submit_item` (`qs_client/client.py:613-677`) does `queue_get` → `item_add` → `queue_start`, and the manager's `_start_plan` raises `"RE Manager is busy."` whenever `_manager_state != IDLE` (`bluesky_queueserver/manager/manager.py:999`). Two concrete paths leave `U_S1H` at +1 A, the outcome lesson (2) was meant to close:
   (a) `_process_plan_report` writes the history entry (`set_processed_item_as_completed`) and only *then* schedules `create_task(_set_manager_state(IDLE))` (`manager.py:838-852`). `_wait_for_item` can therefore return with the item in history while the manager is still `executing_queue`; the `mv` is added, `queue_start` is refused, `_submit_item` removes the `mv` again, `restore.ok` is False, the `assert restore.ok` fires, and nothing restores the setpoint.
   (b) `_wait_for_item` times out at 300 s with the scan still *running* (a device hang — the realistic timeout case): same "busy" refusal. `clear_pending=True` only covers an item still *queued*, and an item still queued after 300 s means the `queue_start` inside `submit_preset` never took, which that call already reports.
   Refutation attempted: (a)'s window is one event-loop turn against a 1 s poll, and Scan 108 passed — but the failure mode is "magnet left moved", and (b) is not a window at all. Fix: gate the restore on a bounded wait for `client.status().manager_state == "idle"` (`QueueStatus.manager_state` already exists, `client.py:137`), and on a refused `mv` retry rather than assert — or fall back to the in-process restore test 1 uses (`RE(bps.mv(motor, initial))`), which does not depend on the manager's state.

2. **P3 — dead-or-wrong catalog branch, and a hand-rolled device reference** — `test_phase1_hardware.py:346`, `:362`.
   `scan_parameter=f"{device_name} {variable}" if catalog_entry is None else SWEEP`: the catalog is keyed by friendly name (`ScanVariables` docstring, `GEECS-Schemas/geecs_schemas/scan_variables.py:241`; the live Undulator catalog has 61 keys, none containing a colon), so `catalog.get("U_S1H:Current")` is always `None`. If it ever were not, the branch is wrong: `callbacks.scan_parameter` writes the `Device Variable` header from `geecs_scalar_headers` regardless of how the argument was spelled, never the colon form. Conversely the docstring allows only `Device:Variable` in `GEECS_HW_SCAN_VARIABLE`, so a catalog *name* there breaks `SWEEP.partition(":")` → `pv_name(EXPERIMENT, "<name>", "")` and the restore arg `"<name>."`. Either drop the branch or resolve `SWEEP` once through `scan_variable_reference(SWEEP, catalog)` (`qs_client/presets.py:62`), which also replaces the hand-rolled `f"{device_name}.{variable.lower()}"` — `device_reference` (`utils.py:37`) uses `safe_name`, which only coincides with `.lower()` for a variable like `Current`.

3. **P3 — preflight assertions are weaker than the PR body's claim** — `test_phase1_hardware.py:319-323`.
   Only `report.refusal is None` is asserted; `worker_ready` and `gateway_liveness` record `skipped` on any error (fail-open by design, `submit_preflight.py:98-103`), and `report.questions` are printed, not asserted. A regression that made `worker_ready` always-skip (or a DISCONNECTED device surfaced as a question) would leave this acceptance test green. Assert `all(r == "passed" for _, r, _ in report.outcomes)` and `not report.questions` — that is what "all passed" in the PR body means.

4. **P3 — "ARMED → STANDBY" is asserted from software bookkeeping, and ARMED not at all** — `test_phase1_hardware.py:257-258`, `05_phase1_acceptance.md` M4 table ("trigger box: ARMED before, STANDBY after").
   `ShotControl.standing_state` is "the last standing state driven" (`devices/shot_control.py:193-195`), not a readback; nothing in the test observes ARMED. It still catches a plan that skipped the STANDBY `set`, so keep it — but the docstring and M4 table read as a hardware observation. Refutation: `set()` awaits the gateway put, so "driven" is close to "written"; it is the wording, not the check, that overstates.

5. **P3 — duplicated test helpers, and a PNG-only file check** — `test_phase1_hardware.py:78-82`, `:85-101`.
   `_ini` is a verbatim copy of `tests/test_callbacks.py:43-47`; `_wait_for_files` is a verbatim copy of `tests/test_phase0_hardware.py:68-84` (timeout 10 → 15). Both hardware tests already import `tests/ca_mock_helpers.py` — the natural home. Both copies glob `*.png`: with `GEECS_HW_CAMERA_DEVICE` pointing at a camera whose native format is not PNG (the #738 retirement is under way) the check times out although the files landed. Count non-directory entries, or take the suffix from the detector's resource documents the collector already holds.

6. **P3 — a transient fleet fact written into durable package docs** — `GeecsBluesky/CLAUDE.md:20-25`, `README.md:48-50`.
   "The deployed worker's checkout is on the feature branch; the restart that flips it is the maintainer's" is false the moment the command in the PR body runs, and no PR is planned to update it (the feature branch merges to master later). The durable record is `05_phase1_acceptance.md` ("The worker flip"), and the fleet map (`docs/platform/fleet_map.md:25`) still states everything runs at `master`. Keep CLAUDE.md/README at "hardware-accepted 2026-09-10 (PR 3, see 05…)" and drop the checkout-state sentence — it is the same class of staleness the previous text ("stays on `master` until then") had.

### Checked and not raised
Facility literals (same shape as the merged phase-0 test, disclosed in the PR); public-repo hygiene (no account names, hostnames or home paths in the diff — `~`/`$HOME` and `192.168.6.14` only); §10.x cross-references (none point at the renumbered carry-over item; `§10.5/§10.7` targets unchanged); the flip's extras (`ca tiled qserver` matches `qserver/deploy/DEPLOYMENT.md:63`, and `geecs-qserver-ensure-ready` reaches the manager via `zmq_single_request` from the `qserver` extra, not `bluesky-queueserver-api`); runbook CLI flags exist in bluesky-queueserver 0.0.25 (`--keep-re` is deprecated and ignored — harmless) and `QSERVER_ZMQ_CONTROL_ADDRESS` is the `qserver` CLI's variable; s-file path `analysis/s{number}.txt` matches `write_scalar_files`; `Shots per step` for `count` = `num`; scan-folder invariant untouched (the test only reads; claiming stays in `claim_scan`).

### Verdict
No blocking finding. The acceptance test proves what the PR body says it proves, the docs are internally consistent, and the suite/CI reproduce. Finding 1 is worth fixing before this test is run on hardware again — the restore path is the one part of the test whose failure moves a magnet and leaves it there, and both the history-before-idle window and the plan-still-running timeout defeat it as written; findings 2–3 tighten the test so it would actually fail on the regressions it is meant to guard; 4–6 are wording, helper placement and doc staleness.

skbarber:
## Re-review of 7184e49f (same reviewer) — verdicts, verbatim

1. **P2 restore path — CONFIRMED FIXED.** `_wait_for_manager_idle` polls `QueueStatus.manager_state` (the field `_start_plan` gates on), so the history-before-idle window can no longer produce a "busy" refusal; a plan still running past 600 s gets five refused attempts and the loud `RESTORE NOT QUEUED — set U_S1H:Current back to 0.0 by hand` assertion. Also tried: a failed scan item returned to the queue front — `clear_pending=True` clears it and the restore queues; a disconnected manager loops the idle wait to its timeout and fails loudly. Residual, accepted: no CA-level fallback if the manager itself is wedged — disclosed by the message.
2. **P3 catalog branch / hand-rolled reference — CONFIRMED FIXED.** The restore argument is `scan_variable_reference(SWEEP, catalog)` → `U_S1H.current`, the same spelling `expand_preset` gave the scan. `GEECS_HW_SCAN_VARIABLE` remains `Device:Variable`-only, as the docstring says.
3. **P3 preflight assertions — CONFIRMED FIXED.** A fail-open `skipped` now fails the test; `not report.questions` closes the "question printed, scan submitted anyway" path.
4. **P3 ARMED/STANDBY wording — CONFIRMED FIXED.** The assertion is unchanged, which is right.
5. **P3 duplicated helpers / PNG-only glob — CONFIRMED FIXED.** One copy in `tests/ca_mock_helpers.py`; any-suffix count, polls a not-yet-created directory; a stray file could satisfy the count early but every s-file stamp must still appear in a filename.
6. **P3 transient fleet fact in durable docs — CONFIRMED FIXED.**

No new findings from the follow-up commit.

## Dispositions

| # | Disposition |
|---|---|
| 1 | **Fixed** (7184e49f); re-run on hardware after the fix: Scan001 of 26_0911, restore queued and ran, U_S1H at 0.0 A. Residual (no CA-level fallback when the manager itself is wedged) **waived**: the failure names the value to restore by hand. |
| 2 | **Fixed** (7184e49f). |
| 3 | **Fixed** (7184e49f). |
| 4 | **Fixed** (7184e49f): wording in the docstring and `05_phase1_acceptance.md`. |
| 5 | **Fixed** (7184e49f): `read_scan_info` / `wait_for_native_files` in `tests/ca_mock_helpers.py`, used by three test modules. |
| 6 | **Fixed** (7184e49f). |

GeecsBluesky: 386 passed, 2 skipped, 1 deselected.

skbarber:
Docs-only follow-up (M6 reworded after the maintainer's point that the fire request is asynchronous to the laser edges): the first shot's request-to-frame delay is uniform over one period, every later shot is phase-locked, and the camera's edge-to-message latency measured from the documents is ~0.8 s (0.70 s exposure + readout + push) — the ~100 ms per-shot margin stands, now derived from the right quantities. `05_phase1_acceptance.md` M6 and §10.8.

skbarber:
Docs-only: M7 added — the same runs at a 1 ms camera exposure (Scans 002/003 of 26_0911). Stamp-to-event latency 0.75 s → 0.05 s; the in-process count at 1 Hz from the first shot; the preset through the manager now repeats at 1 Hz and a moved step lands on the second edge (2 s, the magnet's 1.3 s move alone). Confirms M6's reading: the exposure sets the per-shot margin, the plan layer's ~7 ms is noise. Magnet restored, acceptance manager stopped.

skbarber:
## Codex review (maintainer-run) — dispositions, on 8e035090

| Finding | Disposition |
|---|---|
| P2 (#822) — the restore `mv` was only asserted *queued*; a restore that ran and failed would leave `U_S1H:Current` at the scan endpoint | **Fixed** (8e035090): the test captures the restore's history item and asserts `result.exit_status == "completed"`, then reads the setpoint PV back and asserts it equals the pre-scan value. Re-run on hardware: Scan004 of 26_0911, `restored U_S1H:Current: setpoint reads 0.0`, `1 passed in 29 s`. |
| P3 (#821, closed) — `device_reference` spelled `UC_TestCam.trigger` for a variable the namespace binds as `trigger_`, so a valid `Device:Variable` was unusable through presets | **Fixed here** rather than waived (8e035090): `utils.settable_attribute` is the one rule (a frozen `RESERVED_DEVICE_ATTRIBUTES`, pinned to `dir(GeecsDetector)` by `tests/test_namespace.py::test_reserved_device_attributes_pin_the_detector_class`); the namespace's `_attribute_for` and the client's `device_reference` both use it, so `scan_variable_reference("UC_Cam:trigger") == "UC_Cam.trigger_"` (pinned in `test_presets.py`). |

GeecsBluesky: 387 passed, 2 skipped, 1 deselected.

Issue #744, opened 2026-09-01
Image analysis as a PV producer: per-shot analyzer scalars published as PVs (scalar-only optimization, live diagnostics)


skbarber:
## Why

Optimization at BELLA is scalar in every other respect, except that the
evaluator's objective often needs an image reduced to a number — so today
the evaluator runs an ImageAnalysis analyzer over per-shot files mid-scan.
That ties optimization to a file format (the reason #738 existed) and
makes image-derived objectives unavailable to anything that is not the
in-plan evaluator (Badger, OSPREY, console live plots, gateway derived
channels).

Decision from the #738 discussion (2026-09-01): move image analysis
**upstream of the record**. A small node subscribes to a camera's live
frames, runs the same analyzer the offline diagnostic config describes,
and publishes the resulting scalars as PVs. Everything downstream then
treats them as ordinary device scalars.

This issue is **not** gated on PNG retirement or on #738. Live per-shot
scalars are useful in today's dual-write world. #738's "optimize
evaluators" checklist item depends on this issue, not the other way round.

## What it is

```
camera NTNDArray PV (GeecsPvaGateway)
   → analysis node: ImageAnalysis analyzer (same config as offline)
   → scalar PVs, each frame's scalars stamped with that frame's acq_timestamp
   → engine logs them as save-set telemetry  → evaluator `scalars:` path
                                             → console live plots
                                             → Badger / OSPREY (plain PVs)
```

Pixels never leave the live path; the record (native files / capture
stack) is untouched and remains the source for post-hoc reproduction.

## Contract — hard rules up front

1. **Shot identity.** Every published scalar carries the source frame's
   `acq_timestamp` (LabVIEW epoch, the raw value, same as the gateway's
   intrinsic timestamp variables — PV_CONTRACT.md §3). This is what lets
   the engine's `shot_id` machinery bin the value to a shot and lets
   per-bin evaluators filter on it. A scalar without it is unusable.
2. **Frame coherence.** Scalars from one frame are posted before that
   frame's timestamp PV, so a client latching on the timestamp observes a
   consistent set (the gateway's existing derived-channel ordering rule).
3. **One analyzer config.** The node instantiates its analyzer through the
   ScanAnalysis diagnostic config (`create_scan_analyzer` /
   `ResolvedDiagnosticConfig`) exactly as the offline path does. No second
   config vocabulary. Online/offline drift is the failure mode to test for,
   not to design around.
[...]

PR #823, opened 2026-09-11
#806: the PVA gateway's areaDetector-shaped file plugin + stock ADHDFDataLogic on the worker; partial-row strict semantics; capture daemon deleted

skbarber:
Phase 1's second leg (#806, plan of record `Planning/native_bluesky/03_clean_room_rebuild.md` §4.A/§8): the image writer moves into the PVA gateway on each camera server as an **areaDetector-shaped HDF5 file plugin**, and a camera on such a host becomes a `GeecsDetector` with the **stock** ophyd-async `ADHDFDataLogic` + `NDFileHDF5IO` as its data logic. The capture daemon is deleted. Design (every API claim read in the installed 0.19.3 / 1.15.0 / 0.2.9 source): `Planning/native_bluesky/06_pva_file_plugin.md`.

## What

- **GeecsPvaGateway 0.7.0** — `file_plugin.HdfFilePlugin`, one per served image variable: the full `NDFileHDF5IO` PV set under `<image PV>:hdf1:` (`geecs_core.pv_naming.hdf_plugin_prefix`) plus `Rewind`, `WriteStatus`, `WriteMessage`; lossless intake branching off `_on_frame` *before* the latest-wins slot; one writer thread owns every session state and the file handle. `Capture=1` retains the GEECS subscription like a client and completes only once a frame has been decoded (that is where the geometry the worker describes the stream with comes from — LabVIEW's idle re-push is enough); dedupe on `acq_timestamp`; stale watermark set at `Capture=1`, moved by `Rewind`; `NumCaptured_RBV` after each frame is on disk; `Capture=0` stamps reconciliation counters. NDFileHDF5 layout (`/entry/data/data` chunked one frame per chunk, `/entry/instrument/NDAttributes/acq_timestamp` + `recv_timestamp`, declared through `NDAttributesFile`). Never creates a directory. Never HDF5 SWMR across SMB (`SWMRMode` accepted and ignored; flush per frame; locking off). `geecs-pva-gateway diff` = the old `capture/diff.py`. `h5py` dependency ⇒ **re-bootstrap per camera server**; a box without it serves no plugin PVs.
- **GeecsBluesky 0.81.0** — `devices/hdf_plugin.py` (`GeecsHdfIO`, `PluginPathProvider`: the Windows directory for the plugin's `FilePath` and the worker's `file://` URI for Tiled, from one run folder — `PathInfo` carries both; `file_plugin_hosts`); `GeecsDetector(hdf_plugins=…)` wires the stock logic; the namespace rule (DB image variable + endpoint in `config.ini [pva] file_plugin_addr_list`, else `addr_list`; `DeviceRoster.endpoints`); `STRICT_TRIGGER_INFO.exposure_timeout = 3 s` with the count timeout translated into `GeecsTriggerTimeoutError`; `discard_uncollected()`; `StackCheckCallback` (frames on disk == rows' stamps, per stack, into `scan.log`); `p4p` under the `ca` extra; `capture/` deleted.
- **Strict shot semantics** (Sam, 2026-09-11): a missed frame keeps the row — every scalar the shot produced, the frameless device's columns `NaN` (`mask_missed_shot`), **no frames** — and the plan takes one more shot for the step (bounded by `max_refires`; exhaustion raises naming the devices, partial rows kept). Per-device trigger groups; the devices' `missed_shot` flags are the census.
- **GEECS-Core 0.5.1** (`hdf_plugin_prefix`), **GEECS-Data-Utils 0.27.1** (`scan_stack` on the areaDetector datasets; callers unchanged), **GEECS-DataPortal 0.21.3** (cache follows).

## Where the design departs from #806 / the plan doc, and why

1. **Bluesky's bundler forbids the "empty shot keeps B's frame" shape.** One same-width datum per external key per event, or none (`bundlers.py:611`, `:841-851`, `:938-943`); TiledWriter needs consecutive indices; the consolidator reads positionally. So the partial row carries no frames and every plugin-backed camera is rewound to its last referenced frame *before* the row is read — B's uncollected frame goes with A's late one (scalars stay). 03 §4.B's "orphan unreferenced" claim is corrected in the same PR. **Judgment call, cheap to veto:** parking discarded frames in a side dataset instead of truncating is ~15 lines if you want nothing lost.
2. **`Rewind` is the one non-areaDetector verb** — the late-frame guard; kept for the stock property.
3. **`Capture=1` acks on the first decoded frame** (≤ 1 s per run per camera; a camera pushing nothing fails `prepare` on the put timeout naming the PV).
4. **Two config keys**, both facility values in `config.ini`: `[Paths] geecs_pva_plugin_data_base_path` (the UNC data root the LocalSystem service can write — it cannot see `Z:`) and `[pva] file_plugin_addr_list` (the per-box rollout knob; goes when the fleet is rolled). Flagged per the facility-literal rule: no lab value is in code.
5. Q1–Q3 of #806: one frame per chunk; the engine keeps creating the device leaf (`CreateDirectory` ignored); prefix `<image PV>:hdf1:` per image variable. `Compression` ships `None` for a like-for-like parity diff (`zlib` = gzip-1+shuffle is a put away).

## LOC by concern

| Concern | Files | + | − |
|---|---|---|---|
| Gateway plugin + server hook + tests | 4 | 1146 | 6 |
| Worker detector, IO, path provider, namespace rule + tests | 7 | 731 | 21 |
| Strict plan semantics + stack check + tests | 4 | 464 | 110 |
| Read side (scan_stack, portal cache, fixtures) | 6 | 70 | 54 |
| Naming contract | 2 | 33 | 0 |
| Diff tool move | 2 | 12 | 12 |
| Capture daemon deletion | 14 | 0 | 2502 |
| Docs, planning, deploy, changelogs, locks | 26 | 1525 | 211 |

## Tests (`./scripts/check.sh --base origin/feature/native-bluesky-plans`, all seven suites)

GEECS-Core 96 passed; GEECS-Data-Utils 327 passed, 3 skipped; GEECS-DataPortal 209 passed, 32 skipped; GeecsBluesky 365 passed, 2 skipped; GeecsPvaGateway 43 passed (incl. `test_file_plugin.py`, which drives the real plugin with the stock `ADHDFDataLogic` over `pva://` — the PV contract pinned on both sides); ImageAnalysis 276 passed, 2 skipped; ScanAnalysis 192 passed.

## Hardware verification — OWED (design §8)

1. Share clone → feature branch (the plugin ships inert: no PVs without `h5py`, no detector uses it unless the host is listed).
2. Re-bootstrap **one** camera server; `pvget …:image:hdf1:Capture_RBV`; `FilePathExists_RBV` true for a UNC run folder — **decides whether LocalSystem's machine account can write `data` on the share** (known to read `software`); else the domain-account fallback in `DEPLOYMENT.md`.
3. Worker: add the host to `[pva] file_plugin_addr_list`, the UNC root to `[Paths]`, `HDF5_USE_FILE_LOCKING=FALSE` for Tiled; restart; a strict scan on that camera: stack frames == rows, stamps == rows' `acq_timestamp` (the stack check in `scan.log`), stream resource/datum documents, a Tiled read of the image array. Expected: 1 Hz cadence unchanged (the count wait adds one PVA monitor update per shot after the ~4 MB write over SMB).
4. Several dual-write scans; `geecs-pva-gateway diff` clean.
5. Roll the fleet box by box.

Related: #807, #738 (PNG retirement — `assets/` and the `geecs://` descriptor patch stay until then), #744.

https://claude.ai/code/session_01CQd7VdzpqYFKjDYrqRdEKB


skbarber:
## Adversarial review (fresh-context reviewer, three lenses) — report as returned

Read against the installed ophyd-async 0.19.3 / bluesky 1.15.0 sources and the whole repo; `tests/test_file_plugin.py` (6) and the four new/changed GeecsBluesky suites (53) pass locally. Findings ranked; each survived my attempt to refute it.

### P1 — `file_plugin_hosts` falls back to `[pva] addr_list`, so the "ships inert" rollout is false wherever the worker's config carries the fleet roster
`GeecsBluesky/geecs_bluesky/devices/hdf_plugin.py:97` — `for key in ("file_plugin_addr_list", "addr_list")`.

`[pva] addr_list` is the *deployed PVA fleet* (9 boxes; `fleet_map.md`, `fleet.py:118`, `gen_fleet_status.py`). With this fallback, a worker whose service-account `config.ini` has that key (the natural state — it is the key every PVA client tool reads) treats every camera on all 9 boxes as plugin-backed the moment this branch is pulled. None of those boxes has `h5py` yet, so no `:hdf1:` PVs exist; `connect_on_demand` connects the touched detector's subtree including `GeecsHdfIO`, and every scan on any camera fails at connect with NotConnected. Contradicts the PR body's "no detector uses it unless the host is listed", `GeecsPvaGateway/DEPLOYMENT.md:120-124`, and the namespace docstring. The worker host is on the feature branch today, so the merge itself is the exposure.

Fix: drop the fallback — absent key means no host; `test_file_plugin_hosts_reads_the_config_keys` pins the current behaviour and needs its first assertion flipped.

### P2 — `StackCheckCallback` reads the stack while the plugin still has it open, breaking the design's own rule 2
`GeecsBluesky/geecs_bluesky/callbacks.py:370-396`, `06_pva_file_plugin.md` §5 rule 2.

Every stock plan is `stage_wrapper(run_wrapper(...))`: the stop document (where this check runs, synchronously) precedes `unstage`, and `Capture=0` is `ADHDFDataLogic.stop()` at unstage. So the check opens `<device>.h5` on the CIFS mount with `h5py.File(path, "r")` (default HDF5 file locking) while the Windows writer's session is open and the file is un-finalized — exactly the read the design forbids, and with the lock the design disables for Tiled via `HDF5_USE_FILE_LOCKING=FALSE` (§5 rule 3) but which the `geecs-qserver` unit never gets. Over CIFS that `flock` is the known ENOLCK/hang mode; either way the failure is swallowed by `_RunCallback.__call__` into a warning every scan. The portal's `ShotDataCache` already gates on the `finalized` attr; the check does not. `test_callbacks._feed_stack_run` writes and closes the file before feeding documents, so it cannot see this.

Fix: make `scan_stack`'s readers open with `locking=False`, and either gate the check on `finalized` with a short bounded wait or run it after `Capture=0` — accepting that the verdict then lands after `ScanLogCallback` closed `scan.log`.

### P2 — The stack check flags a false mismatch on every partial row where the plugin-backed camera *delivered*
`callbacks.py:381-396`; scenario straight from `tests/test_strict_plans.py::test_missed_frame_on_plugin_cameras_rewinds_the_partial_row`.

A misses shot 2, B (plugin-backed) delivers. By design B's scalars stay — the test asserts `events[1]["uc_b-acq_timestamp"] == 1002.0` — while B's frame is rewound (`b_rewinds == [1]`, three datums). `_check` builds `expected` from every non-NaN `<det>-acq_timestamp`, so for B it sees 4 stamps against 3 frames and warns — on a healthy scan, and the *common* configuration during the box-by-box rollout.

Fix: derive the expected rows from the run's `stream_datum` documents (a datum's `seq_nums` say exactly which rows own a frame); the count check becomes frames == total datum width.

### P3 — `Capture=1` arriving during arming double-retains the subscription and can orphan a session
`GeecsPvaGateway/geecs_pva_gateway/file_plugin.py:526-532`. While `_capture_on` blocks in its arm loop, a second `("command","Capture",True)` falls through to `self._dispatch(item)` → nested `_capture_on`: `self._session` is still `None`, so it validates, calls `_retain` again, arms, assigns `self._session`; the outer loop then continues and either times out (releasing once, posting `Capture=False` over a live session) or overwrites `self._session` — dropping the inner session's open file without close. Fix: treat a `Capture=1` during arming as `op.done()` and continue.

### P3 — `mask_missed_shot` lets bool, numpy-integer and array columns carry the previous shot
`detector.py:282-289`. A bool passes through, `np.int64` is not an `int` subclass so it passes through, and waveform/array readbacks pass through. Fix: blank everything not `str` (`numbers.Real` → NaN, arrays → NaN-filled) — and a test with a bool column.

### P3 — The plugin's arming failure reason never reaches the worker
`ADHDFDataLogic.prepare_unbounded` uses `set_and_wait_for_value(capture, True, wait_for_set_completion=False)` and does not await the returned status: the plugin's `op.done(error="no frame from …")` is dropped and the worker's `prepare` fails 10 s later as a bare timeout on `Capture_RBV`; `WriteMessage` stays on the box. A wrap of `prepare` that reads `hdf.write_message` on failure and re-raises with it would make the PR body's "fails prepare naming the PV" also name the reason.

### P3 — Data-Utils 0.27.1 is a patch bump for a change that silently un-reads every stack already on the share
The capture daemon was deployed, so scans since late August carry `<device>.h5` in the `/frames` + `schema` layout; `is_stack_file` now returns False for them. Deleting the old path is the house rule, but it is a behaviour change: bump minor and add a `Removed` line naming the retired layout.

### P3 — Redundancy (four places), cited
- `data_paths.plugin_save_path` is `device_server_save_path` with one key swapped.
- `namespace.image_variables` is `geecs_pva_gateway.config.image_variables` line for line; the one home is `geecs_core.db.variable_types` beside `effective_vartype`.
- The NDFileHDF5 layout has three homes: `file_plugin.FRAMES_DATASET/ATTRIBUTES_GROUP`, `scan_stack.FRAMES_DATASET/TIMESTAMPS_DATASET`, `callbacks._FRAMES_DATASET`.
- `file_plugin_hosts` is a fourth hand-rolled `configparser` read of `config.ini` in GeecsBluesky.

### Checked and not flagged
- Rewind → retake: `trigger()` re-runs `_update_prepare_context`, which re-reads `collections_written`, so a rewound count does not strand the next count wait; `last_emitted` is public.
- Per-group re-wait loop: `_status_object_completed` sets `_exception` and the wait's group is popped, so a re-wait on a consumed group is a no-op; the census via `missed_shot` is sound.
- Fresh pre-scan frame during arming: unreachable through the registry, which drives `ARMED` before the stock plan.
- Scan-folder invariant: the plugin only `isdir`s; `PluginPathProvider` delegates to the existing scanner-side provider; the check reads only.
- Hygiene: nothing new. `h5py` removal from GeecsBluesky is safe; `subscribe_scan_outputs`'s 4-tuple has no unpacking caller.

skbarber:
## Dispositions — all fixed in 5a834051 (verifier confirm/deny to follow)

| Finding | Disposition |
|---|---|
| P1 `[pva] addr_list` fallback | **Fixed.** `file_plugin_hosts` reads `file_plugin_addr_list` only; absent → no host. Test flipped; `hdf_plugin.py`, `namespace.py`, `GeecsBluesky/CLAUDE.md` and design §6 say why the fleet list is never the plugin list. |
| P2 stack check reads an open file | **Fixed.** The check runs on a thread that waits (bounded, 15 s) for the plugin's `finalized` root attribute, reads through `scan_stack.open_stack` (every reader in that module now opens with `locking=False` — one home for the SMB rule), and appends its verdict to `scan.log` directly since the log callback has closed the file by then (also logged). A never-finalized stack is reported as "not checked". Tests feed an unfinalized file and assert the verdict. |
| P2 false mismatch on partial rows | **Fixed.** The rows a stack owns come from the run's `stream_datum` documents (`seq_nums`); expected stamps are those rows' only; the count check is frames == total datum width. Test: complete / missed (NaN, no datum) / delivered-but-rewound (stamp, no datum) / complete → pass. |
| P3 `Capture=1` during arming | **Fixed.** A repeated `Capture=1` while arming is acknowledged and ignored; test races two puts and asserts one subscription, released once. |
| P3 `mask_missed_shot` blind spots | **Fixed.** `numbers.Real` (bool, numpy scalars) → NaN, arrays → NaN-filled, `str` → `""`; test covers all of them. |
| P3 arming reason lost | **Fixed.** `GeecsDetector.prepare` wraps the stock prepare and, on failure of a plugin-backed camera, attaches each plugin's `WriteMessage` as an exception note (`add_note`, 3.11); test asserts the note. |
| P3 Data-Utils bump | **Fixed.** 0.28.0 with a `Removed` entry naming the retired `geecs-capture/1` layout and what happens to stacks already on the share. |
| P3 redundancy ×4 | **Fixed.** `data_paths._translate_to` + `read_config_entry(section, key)` back both translators and `file_plugin_hosts`; `geecs_core.db.variable_types.image_variables` is the camera test (the gateway re-exports it); `scan_stack.FRAMES_DATASET/ATTRIBUTES_GROUP/TIMESTAMPS_DATASET` are the layout's only definition (the plugin and the callback import them). |

check.sh after the fixes: GEECS-Core 97 passed; GEECS-Data-Utils 327 passed, 3 skipped; GEECS-DataPortal 209 passed, 32 skipped; GeecsBluesky 366 passed, 2 skipped; GeecsPvaGateway 43 passed; ImageAnalysis 276 passed, 2 skipped; ScanAnalysis 192 passed.

skbarber:
## Verifier's verdict on 5a834051 (same reviewer, continued)

Read `git diff 359d07f6..5a834051` in full; re-ran the suites after the fix: GeecsPvaGateway 43 passed, GeecsBluesky (detector/strict/callbacks/namespace) 54 passed, GEECS-Core 96 passed, Data-Utils `test_scan_stack` 9 passed, DataPortal `test_resources` 56 passed.

- **P1 `addr_list` fallback — confirmed fixed.** `file_plugin_hosts` reads only `[pva] file_plugin_addr_list`; absent (or empty) means no host; docs say opt-in. The scenario is closed.
- **P2 read-while-open — confirmed fixed, one doc nit.** Daemon thread from `on_stop`, bounded poll of the `finalized` root attribute (`OSError` retried), lock-free open, verdict appended to `scan.log`. The poll does open the file before `Capture=0`, but only the root attribute, lock-free, with errors retried — nothing it can corrupt. Nit: the `subscribe_scan_outputs` docstring still claimed an ordering constraint.
- **P2 false mismatch — confirmed fixed.** Verified the assumption: the bundler overwrites datum `seq_nums` from the same counter event-model uses for `event.seq_num` (`bundlers.py:852-853`), so they are the 1-based event numbers. The delivered-but-rewound case is in the test and passes.
- **P3 arming re-entry — confirmed fixed.** One session, one `retain`; the raced-puts test asserts release to 0.
- **P3 mask — confirmed fixed.** `numbers.Real` covers bool/int/float/numpy scalars; arrays NaN-filled. (An int-typed array column would clash with its descriptor dtype in TiledWriter, but `ScalarsDataLogic` never carries array columns — unreachable today.)
- **P3 arming reason — confirmed fixed, with a caveat.** When the stock logic refuses before any `Capture` put (`FilePathExists_RBV` false), the note carries whatever `WriteMessage` last held; clearing it on the `FilePath` put would keep the note current.
- **P3 Data-Utils version — confirmed fixed.**
- **P3 redundancy — confirmed fixed.**

No surviving refutations; the two nits are cosmetic.

### Disposition of the nits — fixed in 41444df4
`WriteStatus`/`WriteMessage` reset on every `FilePath` put (the first put of a prepare, so the note is always this run's); the `subscribe_scan_outputs` docstring no longer claims an ordering. Plugin and callback suites re-run green (6 and 12).

CI on 5a834051: pre-commit, unit-tests, console-windows all pass; the 41444df4 run follows.

skbarber:
## Codex review (the maintainer's) — two findings, both fixed in 28494847

**P1** `file_plugin.py` — during `Capture=1` arming, `_session` is set and `_on_frame()` runs before the put completes; if the first accepted frame fails in `_open_file()` (permissions, SMB write failure, existing file) the exception escapes to `_run()`'s generic logger, the put never gets `op.done(error=…)`, `_session` stays set and the retained subscription is never released — the worker's `prepare` fails as a bare timeout and the plugin can stay capturing.

**Fixed.** `_open_file` failures are caught inside `_on_frame` (a new `open_failures` counter, `WriteStatus`/`WriteMessage` set; the count never advances, so a shot fails loudly with the reason). On the arming path the outcome is checked: a failed or raising arming frame clears the session, releases the subscription, posts `Capture`/`Capture_RBV` false and completes the put with the real error. Test: a fresh arming frame with `_open_file` refusing → the put raises "share refused", `Capture_RBV` false, no session, connections back to 0, no file.

**P2** `GEECS-DataPortal/geecs_portal/cache.py` — the stack cache still opened plugin files with `h5py.File(...)` directly, bypassing `scan_stack.open_stack()` (locking off for SMB).

**Fixed.** The cache opens through `open_stack`.

Suites after the fix: GeecsPvaGateway 43 passed; GEECS-DataPortal `test_resources.py` 56 passed, 12 skipped.

skbarber:
## Hardware verification — DONE on 192.168.6.100 / `UC_Amp4_IR_input` (2026-09-11), record in `Planning/native_bluesky/07_806_acceptance.md`

Steps 1–4 of the rollout, run from this branch before the merge (staging clone in process; the production worker untouched):

- Box: h5py installed; the launcher predated GEECS-Core (would have crash-looped) and was replaced by the repo's. Share clone on the branch; `:restart` → `version` 0.7.0 in 24 s, `:hdf1:` PVs served.
- **The service (LocalSystem, machine account) sees and writes the share over UNC**: `FilePathExists_RBV` true for a run folder, false for a nonexistent one; the stack was created and finalized by the service.
- **First run (Scan006) failed at `prepare`** and found a rule: the camera's DB lists three image variables (`image`, `bakground image`, `processed image`); the secondary ones are pushed only on demand, so their plugins never armed. The plugin's reason reached the worker (`no frame from UC_Amp4_IR_input bakground image within 8 s`). Fixed in 69a9870a: the worker captures the primary image variable only. Both plugins and the camera's `save` were left clean.
- **Second run (Scan007) passed** (`tests/test_806_hardware.py`, 1 passed in 63.6 s): 5 rows; stack `(5, 600, 600)` uint16, chunks `(1, 600, 600)`, finalized; one stream resource + five contiguous datums; stack stamps == rows' `acq_timestamp` == the five PNG names; `scan.log`: `stack check: … 5 frame(s) … match the rows' stamps`; **pixel parity** `geecs-pva-gateway diff` → `pass (matched=5 identical=5 png_only=0 stack_only=0)`; **cadence `[1.0, 1.001, 1.0, 0.999]` s — 1 Hz holds** with the count wait in the loop; plugin counters `received 10, written 5, duplicates 1, stale 4, rewound 0`, no errors; `save` off after the run.
- **Tiled read: 500** — `Refusing to serve file://…/UC_Amp4_IR_input.h5 because it is outside the readable storage area for this server`. The Tiled server's `readable_storage` lists only its own directory; adding `/mnt/hdna2/data` (+ `HDF5_USE_FILE_LOCKING=FALSE` on the unit) is a deployment edit (sudo restart), owed.

Still owed after merge: the production worker restart (config keys already in place, backed up), the Tiled config, and the fleet roll (launcher fix + h5py per box).

PR #824, opened 2026-09-11
GeecsPvaGateway 0.7.1: fleet requirements — new external deps roll from the share, not by a per-box visit

skbarber:
Follow-up to #823 for the fleet roll. Today a third-party dependency added after a box was bootstrapped (h5py, 0.7.0) reaches that box only by a per-box visit, because `launch.bat` reinstalls with `--no-deps` by design (restarts need no internet). Every future writer dependency (zarr is the named one) would repeat the sweep.

**What**: `deploy/requirements-fleet.txt` (exact pins of post-bootstrap external deps), `deploy/stage_wheels.sh` (downloads their Windows/CPython-3.11 wheels into `<Active Version>/pva-wheels` beside the share clone, from any machine with PyPI reach), and one `launch.bat` step that installs the pins offline from that cache (`--no-index`) before the reinstall — best effort, like the reinstall. A new dependency becomes: pin, stage, pull the clone, `:restart`. `tests/test_deploy_files.py` pins the launcher's five-package reinstall list, the wheel step's placement, and that every fleet pin is a declared `pyproject` dependency. Runbook updated; launcher changes themselves remain a one-time per-box copy (the 0.4.4 fleet's launcher predates GEECS-Core and would crash-loop on the 0.7.x restart — found on the 2026-09-11 roll of 6.100).

Judgment call flagged: the cache lives on the share beside the clone, not in git (a 3 MB wheel has no place in the repo), resolved at restart as `<GEECS_PVA_SOURCE>\..\pva-wheels` outside the parenthesized block (cmd expands `%VAR%` inside a block at parse time — the one cmd trap in this diff).

**Tests**: GeecsPvaGateway 46 passed (3 new).

**Hardware verification — OWED, and this PR's purpose**: the eight remaining camera servers get the new launcher as their last hand copy and a `:restart`; the launcher's wheel step must install h5py and the box must come up at 0.7.1 serving `:hdf1:` PVs. The wheel is already staged (`pva-wheels/h5py-3.16.0-cp311-cp311-win_amd64.whl`).

https://claude.ai/code/session_01CQd7VdzpqYFKjDYrqRdEKB

### DIFF
--- a/GeecsPvaGateway/CHANGELOG.md
+++ b/GeecsPvaGateway/CHANGELOG.md
@@ -4,6 +4,21 @@ All notable changes to this package will be documented here.
 Format follows [Keep a Changelog](https://keepachangelog.com/en/1.1.0/).
 
 
+## [0.7.1] - 2026-09-11
+
+### Added
+
+- **Fleet requirements**: `deploy/requirements-fleet.txt` (exact pins of
+  external dependencies added after a box was bootstrapped — h5py first),
+  `deploy/stage_wheels.sh` (downloads their Windows/CPython-3.11 wheels
+  into `<Active Version>/pva-wheels` beside the share clone), and a
+  `launch.bat` step that installs the pins offline from that cache before
+  the reinstall.  A new dependency is a pin, a staged wheel and a
+  restart — no per-box visit.  `tests/test_deploy_files.py` pins the
+  launcher's package list and the requirements' consistency with
+  `pyproject.toml` (the 0.4.4 fleet's launcher predated GEECS-Core and
+  would have crash-looped on the 0.7.0 restart).
+
 ## [0.7.0] - 2026-09-11
 
 ### Added
--- a/GeecsPvaGateway/CHANGELOG.md
+++ b/GeecsPvaGateway/CHANGELOG.md
@@ -17,7 +17,12 @@ Format follows [Keep a Changelog](https://keepachangelog.com/en/1.1.0/).
   restart — no per-box visit.  `tests/test_deploy_files.py` pins the
   launcher's package list and the requirements' consistency with
   `pyproject.toml` (the 0.4.4 fleet's launcher predated GEECS-Core and
-  would have crash-looped on the 0.7.0 restart).
+  would have crash-looped on the 0.7.0 restart).  Review of #824: the pin
+  file is the closure (`--no-deps` on both sides, so one unstaged
+  transitive wheel cannot block every pin); a launcher copy happens with
+  the service **stopped** — cmd reads a batch file by byte offset, so a
+  copy over a running service followed by `:restart` resumes the new file
+  mid-way (`DEPLOYMENT.md`).
 
 ## [0.7.0] - 2026-09-11
 
--- a/GeecsPvaGateway/CLAUDE.md
+++ b/GeecsPvaGateway/CLAUDE.md
@@ -47,8 +47,13 @@ deploy/
   bootstrap.ps1   # one-time per-box setup (venv, firewall, NSSM service with
                   #   USERPROFILE override -> service-owned profile; installs
                   #   Python 3.11 itself when `py -3.11` is missing)
-  launch.bat      # pull-on-restart launcher (reinstall from the shared
+  launch.bat      # pull-on-restart launcher (fleet pins offline from the
+                  #   share's wheel cache, then reinstall from the shared
                   #   GEECS-Plugins clone; its checked-out commit = fleet pin)
+  requirements-fleet.txt # exact pins of external deps added after bootstrap
+                  #   (the closure: --no-deps on both sides); h5py first
+  stage_wheels.sh # downloads the pins' win_amd64/cp311 wheels into
+                  #   <Active Version>/pva-wheels beside the share clone
   gen_fleet_status.py # --experiment X -> fleet_status_<x>.bob from
                   #   fleet.py's roster (DB + [pva] addr_list); rerun + commit
                   #   when the fleet changes, never hand-edit
@@ -68,6 +73,8 @@ tests/
                   #   session semantics (arming, dedupe, stale, rewind,
                   #   counters, no directory creation, no empty file)
   test_diff.py    # the parity tool
+  test_deploy_files.py # the launcher's package list + wheel step, the pins'
+                  #   consistency with pyproject (name and specifier)
 ```
 
 ## Architecture (one asyncio loop)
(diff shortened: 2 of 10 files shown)


skbarber:
## Adversarial review (fresh-context reviewer, three lenses) — report as returned

Checked: the full diff, the pre-PR launcher, `deploy/bootstrap.ps1`, `.gitattributes`, `docs/`, `Planning/`, the root `deploy/` (Linux services box), `GeecsCAGateway/deploy`, `GEECS-MCP/deploy`, `GEECS-DataPortal/deploy`, `scripts/`, every `requirements*.txt` in the repo. Ran the new tests (3 pass) and the exact `stage_wheels.sh` pip command against PyPI: `h5py-3.16.0-cp311-cp311-win_amd64.whl` downloads and its only `Requires-Dist` is `numpy>=1.21.2`, so the h5py-only pin file works end to end as written.

**Redundancy / placement: no surviving findings.** Nothing else in the repo stages wheels or pins fleet externals; the Linux `deploy/` tree is a different fleet (systemd units), and the new files sit beside the launcher they feed. The cmd semantics the diff relies on hold: `%VAR%` is resolved before the parenthesized block, `for %%I in ("<UNC>\..") do set "…=%%~fI\pva-wheels"` normalizes `..` (and a trailing backslash) through `GetFullPathName` on UNC paths, `if exist` on a UNC directory is fine, the balanced parentheses in the in-block `rem` lines are safe, and `pip install --no-index --find-links` accepts a space-containing UNC directory.

### P2 — `DEPLOYMENT.md:97-101`: copying `launch.bat` over a *running* service, then `:restart`, resumes cmd mid-file in the new launcher; the first restart skips the wheel step
cmd does not load a batch file into memory; it re-reads the file by byte offset after each command returns. When `geecs-pva-gateway.exe` exits 86, cmd resumes reading `launch.bat` at the offset it saved after the exec line of the **old** launcher in the **new** file. In this version pair that offset lands inside the block comment, so the resumed script executes a stray block, evaluates `if exist ""` (the `GEECS_PVA_WHEELS` `set` lines sit above the resume point and never ran in this process), skips the pin install, runs the reinstall, and launches. Net effect: the restart that is supposed to deliver the new launcher and h5py delivers the reinstall only; only a *second* `:restart` uses the launcher from the top. The landing offset is an accident of these two file lengths — a future launcher edit can land it inside the pip command line instead.
Fix: stop the service around the copy (`nssm stop` → copy → `nssm start`; `bootstrap.ps1` already does exactly this), or state that the first `:restart` after a live copy is a no-op for the launcher change.

### P3 — `deploy/launch.bat:36` vs `deploy/stage_wheels.sh:16`: the two sides disagree on whether pins carry their dependencies, and one missing transitive wheel blocks every pin
`stage_wheels.sh` downloads `--no-deps`; the launcher installs the same file **without** `--no-deps`, so pip resolves each pin's `Requires-Dist` against the venv under `--no-index`. A second pin whose wheel requires something not installed and not staged → `pip install -r` fails to resolve → because a single `pip install` is all-or-nothing, h5py is not installed on that restart either. Fix: pick one contract and pin it — either the pins file is the closure (`--no-deps` on the launcher too, transitive deps pinned explicitly) or the cache is the closure.

### P3 — `tests/test_deploy_files.py:70-72`: the parse-time-expansion guard passes a launcher that has the bug it documents
The assertion matches the *clearing* line `set "GEECS_PVA_WHEELS="` first; moving the `for … do set` line inside the block (the exact mistake the comment warns about) still passes. Also the pin test checks name membership only: `h5py==3.7.0` passes while violating `pyproject.toml`'s `h5py = ">=3.8"`.
Fix: anchor the ordering check on the assignment that matters, and check the pin against the declared specifier.

### P3 — contract text that did not travel with the behavior change
`pyproject.toml` (h5py "a bootstrap-time dependency"), `file_plugin.py` `available()` docstring, `README.md` ("a re-bootstrap per box"), `DEPLOYMENT.md:114-116` ("reinstalls the four intra-repo packages" — no GEECS-Core, no wheel step) and `:76-77` ("launcher changes need a re-bootstrap"), `CLAUDE.md` deploy/tests trees.

skbarber:
## Dispositions — all fixed in 02576f1c + 0823d64a (verifier confirm/deny to follow)

| Finding | Disposition |
|---|---|
| P2 live copy + `:restart` resumes the new file mid-way | **Fixed.** The runbook's per-box step is now `nssm stop` → copy from the share → `nssm start`, from an elevated console (which can read the share), with the byte-offset reason stated and a **never** on the live-copy pattern. The fleet roll uses exactly this step, so the launcher's wheel step runs from the top on every box. (This also explains why 6.100's earlier live copy worked: its offset happened to land safely.) |
| P3 no-deps on one side only | **Fixed.** The pin file is the closure: `--no-deps` on the launcher's install too; `requirements-fleet.txt` states that a pin's transitive deps must be pinned or already frozen at bootstrap; the test asserts the flag on both sides. |
| P3 test anchors | **Fixed.** The ordering guard anchors on `set "GEECS_PVA_WHEELS=%%~fI` (the assignment), so moving the `for` inside the block fails it; pins are checked against pyproject's specifier with `packaging.specifiers` (`h5py==3.7.0` now fails). |
| P3 contract text | **Fixed.** pyproject comment, `available()` docstring, README, both DEPLOYMENT.md paragraphs (five packages + the wheel step; launcher changes → the stop/copy/start step), CLAUDE.md trees. |

GeecsPvaGateway suite after the fixes: 46 passed.

skbarber:
## Verifier's verdict on 02576f1c + 0823d64a (same reviewer, continued)

Read `git diff 4eee175d..0823d64a`; reran `tests/test_deploy_files.py` and re-applied the mutations against the new guards.

- **P2 live copy + `:restart` — confirmed.** `nssm stop` → copy → `nssm start` from an elevated console, the byte-offset mechanism stated, the live-copy pattern forbidden; a fresh `nssm start` reads the launcher from byte 0, so the resume-mid-file scenario cannot occur.
- **P3 `--no-deps` closure — confirmed.** `--no-index --no-deps` on the launcher, matching `stage_wheels.sh`; the closure rule stated in the pin file; the test asserts both sides. Residual, documented not fixed: a pin whose wheel needs something the venv lacks installs "successfully" and fails at import with something other than `ImportError` (a numpy-ABI `ValueError`), which `available()` would not catch — exactly the re-bootstrap case the pin file names.
- **P3 test anchors — confirmed.** The mutation (the `for … set` line moved inside the block) now fails; `h5py==3.7.0` is rejected against `>=3.8`. Notes: `packaging` reaches the gateway env only via pytest; the `^`→`~=` translation is stricter than Poetry's caret, which can only produce a false failure, never a false pass.
- **P3 contract text — confirmed.** All six places consistent.

Collateral: the bootstrap paragraph rewrite dropped the "a re-bootstrap removes the existing service first … the failure is loud, fix and re-run" note. **Restored in a20a1360.**

PR #825, opened 2026-09-11
GeecsScanLog 0.1.0: the scan logbook's day view (phase 01)

skbarber:
## What

Phase 01 of replacing `LogMaker4GoogleDocs`: a new `GeecsScanLog` package
serving a **read-only day view** over scan folders, mounted by the Data
Portal at `/log`.

Deliberately read-only. The commentary store, editor and screenshot paste
are phase 02 — this phase exists so the shape of the day view could be
judged against real data before anything is built on top of it.

## Why this shape

The old logger conflated two things with completely different properties,
and that is what made it unmaintainable — machine content was injected into
a live human document by string-matching headings.

- **The record** (scan number, parameters, shot counts, status) is derived
  from `ScanInfoScanNNN.ini` on every request and stored nowhere. It cannot
  drift from the data, and an upstream change needs no migration here.
- **The commentary** — the only irreplaceable half — gets its own store,
  in `logbook/` as a sibling of `scans/`, in phase 02.

**A day is a query, not a document.** `/log/day/2026-09-11` lists whatever
`ScanNNN` folders exist at request time. Nothing creates a log, no template
is stamped, no daily job runs — so `createGdoc.py` and its bash polling loop
have no successor, along with their failure modes (the doc nobody made, the
scan appended twice, the entry that landed in yesterday's log past midnight).

**Status is reported, not inferred**: `success` / `failed` (reason surfaced
verbatim) / `incomplete` / `unknown`. `incomplete` deliberately covers a
running scan, an aborted one and development churn alike — those are not
separable from the folder, and guessing between them would be a lie the UI
tells confidently.

## Per-concern LOC

| Concern | Lines | Notes |
|---|---:|---|
| Package skeleton, models, reader | ~560 | `GeecsScanLog/geecs_scan_log/` |
| Template + CSS | ~480 | ported from the design prototype |
| Tests | ~400 | 32, incl. the scan-folder invariant |
| Portal mount | ~40 | `--scan-log`, `log` extra, 3 tests |
| CI + check.sh wiring | ~13 | |
| Docs (root + package CLAUDE.md, CHANGELOGs) | ~180 | |

1. **Campaign grouping.** A 100-scan day has ~5 distinct scan parameters,
   so consecutive scans sharing `(Scan Parameter, ScanStartInfo)` collapse
   into one block above a 20-scan threshold. Derived, not declared. Revert
   by deleting `Campaign` and the `{% if many %}` branch.
2. **Devices in the day strip.** Costs one directory listing per scan.
   Kept because `scan_contents` gets it in the same listing as ScanInfo.
3. **20-scan grouping threshold** — arbitrary, one constant.
4. **Rejected: sourcing the listing from Tiled** (which the portal already
   does). `RunSummary` lacks `Scan Parameter` — what grouping keys on — and
   carries `exit_status: "fail"` where ScanInfo carries *which capture timed
   out*. It would need a per-scan file read anyway, and a facility not yet
   on Tiled has only the files. The file reader is the portable floor.

## Performance

Measured against the live HTU share over VPN (54 ms rtt):

| | before | after |
|---|---:|---:|
| 108-scan day, cold share | 27.3 s | 5.6 s |
| same day, revisited | 27.3 s | 11 ms |
| today (7 scans) | 3.1 s | 0.76 s |

Two independent changes: **fewer round trips** (one `os.scandir` yields both
the ScanInfo path and the device list, replacing a `glob` + `iterdir`; the
day folder's listing supplies each mtime, so no per-scan `stat` — four round
trips became two, measured **1.85× on cold days**, A/B across untouched
August dates), and **concurrency + an mtime-keyed cache** on top. Keying on
mtime is what keeps the cache honest: a scan still being written misses it,
so a running scan is never served stale.

One negative result worth recording: a worker-count comparison at 32 and 48
looked dramatically better but was invalid — the OS page cache was warm from
the previous run. 16 is the figure measured against a genuinely cold share.

## Tests

- `GeecsScanLog`: **32 passed**
- `GEECS-DataPortal`: **212 passed, 32 skipped** (unchanged count + 3 new
  mount tests)
- `./scripts/check.sh`: OK for both
- Pins the scan-folder invariant with a test that monkeypatches
  `Path.mkdir` to explode while reading an absent day

## Hardware verification

**Not applicable** — this PR touches no device, PV, or scan-execution path.
It only reads files the scanner already wrote.

Verified against live data instead: rendered `2026-09-10` (108 scans, 5
campaigns) and `2026-09-11` (7 scans) from the mounted HTU share, and ran
end-to-end through the real portal process with `--scan-log`. Scan 006's
actual `TimeoutError("uc_amp4_ir_input-hdf-capture didn't match True in
10.0s")` renders in the failure strip.

## Facility literals

None added. `create_log_router(experiment)` takes the experiment
explicitly and the package carries no default; the portal refuses to mount
`/log` without `--experiment` rather than falling back to one.

## Deployment

No new service, port, or unit. It rides the portal's existing checkout and
systemd unit, behind the `log` extra and `--scan-log`. Portal 0.21.2 →
0.22.0.

## Base

Targets **`feature/scan-logger`**, the arc's integration branch, not master.


skbarber:
## Adversarial review — fresh-context reviewer

Per `/land`, a reviewer subagent with no access to the author's rationale
reviewed `feature/scan-logger...phase/01-package-and-day-view` (23 files,
+3329) for correctness, redundancy and placement. Its report follows
verbatim. Several findings are backed by measurements against the live
share; **I have independently re-verified findings 1, 2, 3 and 9 before
posting** — see the disposition comment that follows.

---

### 1. (Critical, correctness) On real data the day view degenerates: status and campaigns both collapse

Re-implementing `scan_status` + `DaySummary.campaigns` and running them over four recent days of the actual share:

| Day | scans | `success` | `failed` | `unknown` | `incomplete` | campaigns | largest campaign |
|---|---|---|---|---|---|---|---|
| 26_0908 | 26 | 0 | 0 | 16 | 10 | 8 | 10 × `(None, None)` |
| 26_0909 | 65 | 0 | 0 | 14 | 51 | 11 | **50 × `(None, None)`** |
| 26_0910 | 108 | 5 | 0 | 7 | 96 | 11 | **65 × `(None, None)`** |
| 26_0911 | 7 | 6 | 1 | 0 | 0 | 4 | 3 |

Two independent causes:

- **Most scan folders carry no `ScanInfo` at all.** On 26_0910, 96 of 108 folders contain only `scan.log`. All 96 render grey with every fact `—`.
- **`ScanEndInfo = ""` is a common third state the classifier has no case for.** 37 of 49 `ScanInfo` files have an empty `ScanEndInfo` (written at start, never finalised). `scan_status` falls through to `unknown` → amber `check` chip. `""` is not "a `ScanEndInfo` we do not recognise"; it means "not finalised", i.e. semantically `incomplete`.

Consequence for campaigns is worse than the flat list it replaces: on 26_0910 — the day the CHANGELOG cites as the benchmark — 65 consecutive scans share `(parameter=None, purpose=None)` and collapse into **one** block headed `—`.

The `share` fixture is 2-of-3 well-formed; reality is closer to 1-in-20. No test would fail if the classifier were wrong about the dominant case.

**The design doc's justification is factually wrong.** `GeecsScanLog/CLAUDE.md` says these are "not separable from the folder". They are separable, from `scan.log`, present in every one of those 96 folders, which this repo already parses first-class: `geecs_data_utils/scan_log_loader.py` and `GEECS-LogTriage/geecs_log_triage/parser.py`.

### 2. (Critical, correctness) The mtime cache serves a failed scan's `unknown` status forever

`_read_scan_cached` is keyed on the **scan folder's** `st_mtime`. The docstring claims a scan still being written misses the cache. False for the write that matters: `ScanInfoCallback.on_stop` finalises the outcome by rewriting the file in place (`path.open("w")`). Truncating an existing inode changes no directory entry, so the parent mtime does not move.

1. Folder claimed; `ScanInfoScanNNN.ini` written with `ScanEndInfo = ""`.
2. Someone loads the day. Scan cached as `unknown`, `failure_reason=None`.
3. Scan fails; `on_stop` rewrites in place. Dir mtime **unchanged**.
4. Every subsequent request serves the stale `unknown`, for the life of the process (no TTL).

The failed scan is exactly the case where the s-file that would otherwise bust the cache is absent. Fix: `scan_contents` already `scandir`s the folder — fold the ini's own `st_mtime`/`st_size` into the key.

### 3. (High, correctness) The static mount may be silently dropped on the declared FastAPI floor

`router.mount(...)` puts a `starlette.routing.Mount` in an `APIRouter`. `Mount` subclasses `BaseRoute`, not `Route`. `APIRouter.include_router`'s isinstance chain has no `Mount` branch in 0.110 / 0.115 / 0.118. `pyproject.toml` declares `fastapi = ">=0.110"`. On that floor the mount is dropped without error: `/log/static/scanlog.css` 404s and `url_for('log-static', …)` raises `NoMatchFound` → 500. Both locks pin 0.141.1, which is why nothing catches it.

The precedent the router's own docstring cites avoids this: `scan_analysis/config_editor/app.py` serves assets from a plain `@router.get("/static/{name}")` returning a `FileResponse`, with a **relative** `href` in its template.

### 4. (High, correctness) `started` is whenever anything last touched the folder

Measured against `scan.log`'s own start line on `26_0109`:

```
Scan001  dirmtime=11:15:30  logstart=09:59:44   (+76 min)
Scan002  dirmtime=11:15:32  logstart=10:00:18   (+75 min)
Scan003  dirmtime=11:36:32  logstart=11:09:31   (+27 min)
Scan006  dirmtime=14:27:29  logstart=13:27:14   (+60 min)
```

Offsets are inconsistent, so it isn't "the end time" either. Scan001/002 started 34 s apart but display 2 s apart, because a later bulk pass (the `analysis_status/` creation the task queue does) rewrote the day's dirents. A day's chronology is whatever the last analysis run did. Second-order: every such touch also invalidates the cache for the whole day.

### 5. (Medium, placement) `scan_reader.py` is pure logic under a view package

No HTTP surface — it walks folders, parses `ScanInfo`, classifies. Natural home is `geecs_data_utils`, which already owns `ScanPaths`, `scans_database/`, and `scan_log_loader.py` — whose docstring states the rationale verbatim ("deliberately lives in `geecs_data_utils` … so that any consumer … can read scan logs with the same data model"). Consumers already exist: `GEECS-LogTriage/harvester.py` and `scans_database/builder.py` both need "the scans present on day D".

Also: a whole package (pyproject + 1465-line lock + CI job + CHANGELOG + extra + repo-map entry) for a router only the portal mounts. `geecs_portal.log` would need none of it. Note `GEECS-DataPortal/CLAUDE.md` opens with "**Read-only, except explicit analysis runs and the config editor** (charter amendments: owner rulings …)". Phase 02 is writes — a third write surface is a third charter amendment, unrecorded in this diff.

### 6. (Medium, redundancy) Four pieces already exist

- **`parse_scan_info`** is line-for-line `ScanPaths.load_scan_info` (`scan_paths.py:585`) — same `ConfigParser`, same `optionxform`, same strip.
- **`ScanSummary`'s ini half** is `scans_database.entries.ScanMetadata` (`scan_parameter`, `start`, `end`, `step_size`, `shots_per_step`, `scan_mode`, `scan_description` = this PR's `purpose`, `background`, plus `from_ini_dict`). Only status/`ScanEndInfo` is new — a two-field extension, not a new model.
- **The day walker** is the third: `GEECS-LogTriage/harvester.py:212` and `scans_database/builder.py:149`. `harvester.py` resolves both `Scan042` **and** unpadded `Scan42`; `_SCAN_DIR = r"^Scan(\d{3,})$"` makes an unpadded folder **invisible**.
- **The day page** duplicates the portal's own `/day/{day}`: same prev/next, same filter box, same list, same status chip. One service now serves two day views over two sources that *will* disagree.

### 7. (Medium, process) Contract files did not travel with the behaviour

- **`GEECS-DataPortal/CLAUDE.md`** — untouched; its architecture rules, `__main__.py` flag list, `tests/` list and `Routes:` line are now all wrong.
- **`GEECS-DataPortal/DEPLOYMENT.md`** — has a `## The config editor (--config-editor)` section for the analogous opt-in; its install line is `poetry install --extras analysis`. Following it verbatim leaves `geecs_scan_log` uninstalled.
- **`deploy/bootstrap_host.sh:66`** — `extras_of() { … portal) echo "analysis";; }`. A fresh host never installs the `log` extra.
- **`deploy/site.env.example:88`** — `GEECS_PORTAL_EXTRA_ARGS=--config-editor`; no `--scan-log` counterpart, so no documented way to turn it on.

### 8. (Low) Root-`CLAUDE.md` edit contradicts "Known debt"

Line 23 now says LogMaker is "being replaced by `GeecsScanLog/`", but the Known-debt entry still says the fix is a **refactor** and that `EXPERIMENT_FILE_IDS` moves with it. Those paragraphs no longer name a real event; the config home for `EXPERIMENT_FILE_IDS` is now unowned.

### 9. (Low) Smaller things

- **`CHANGELOG.md`** — the "Fewer round trips per scan…" bullet appears **twice**, verbatim.
- **`_NON_DEVICE_SUFFIXES`** defined and never referenced.
- **`read_scan`** exported in `__all__`, called by nothing, no test.
- **`_as_int` contradicts its docstring**: `Shots per step = inf` → `OverflowError`; `nan` → `ValueError`. Uncaught inside `pool.map` → **the whole day 503s**, contradicting "one bad scan must not take down a whole day's view".
- **Naming collision**: `geecs_scan_log` vs `geecs_data_utils.scan_log_loader` / `GEECS-LogTriage`, which read `scan.log`. Finding #1's fix is *to read `scan.log`*.
- **"Expand all" only expands one level**: `cards()` selects `details.campaign` when any exists, leaving every `details.scan` collapsed. `test_busy_day_starts_collapsed` passes only because its fixture yields exactly one campaign.

---

### What the reviewer checked and found clean

- **Scan-folder invariant** — genuinely upheld; the monkeypatched-`mkdir` test would actually fail if the code were wrong, and the root pinned-tests list was updated.
- **Public-repo hygiene / lock paths** — no `file:///Users/`, `/home/`, or worktree paths in either lock.
- **Facility values** — experiment passed explicitly; portal refuses to mount without it. No lab literals.
- **Versioning** — portal minor bump + CHANGELOG, new package 0.1.0 + CHANGELOG, both per the root rule.
- **CI** — real install+run step, `--extras log` correctly citing the #737 `importorskip` lesson; `check.sh` registered in both the list and the runner case.
- **Template escaping** — Jinja2 autoescape on, so the `ScanEndInfo` traceback text and `data-hay` are safe.
- **Route-name collisions** — `_day_page` / `_day_json` / `log-static` don't clash with the portal's. Relative redirect from `/log/` resolves correctly.

skbarber:
## Disposition

The reviewer re-checked every fix (commit `9b131a7d`) and confirmed all
eight findings fixed — while catching **two regressions the fixes
themselves introduced**, both user-visible on real data. Those are fixed in
`339d8fa2`. Every claim below was independently verified against the live
share or the scanner's own source before acting.

| # | Finding | Disposition |
|---|---|---|
| 1 | `ScanEndInfo = ""` → `unknown`; doctrine wrong | **Fixed** `9b131a7d` — now `incomplete`; zero `unknown` across four sampled days. Doctrine retracted and corrected |
| 1b | *(new)* `incomplete` chip read "no scan info" over a card full of parsed ScanInfo | **Fixed** `339d8fa2` — splits on `has_scan_info`; 2026-05-01 now 21 "not finalised" / 0 "no scan info" |
| 1c | *(new)* `abort` still fell through to `unknown` | **Fixed** `339d8fa2` — own status, reason surfaced |
| 2 | Cache served a failed scan's status forever | **Fixed** `9b131a7d` — keyed on the ScanInfo file's own `(mtime, size)`. Reviewer confirmed the regression test fails on the intended assertion |
| 2b | *(new)* un-stat-able ini cached on path alone, forever | **Fixed** `339d8fa2` — reads uncached |
| 3 | Static `Mount` dropped at the declared FastAPI floor | **Fixed** `9b131a7d` — plain route; reviewer verified traversal-safe. `.path` applied for proxy safety |
| 4 | `started` was folder mtime, up to 76 min off | **Fixed** `9b131a7d` — from `scan.log` |
| 4b | *(new)* no fallback, so a 2025 day showed all `—` | **Fixed** `339d8fa2` — ScanInfo mtime, marked approximate. 2025-06-01: 13 approximate, 0 missing |
| 4c | *(new)* `started` docstring (published JSON schema) now false | **Fixed** `339d8fa2` |
| 6 | `parse_scan_info` duplicated `ScanPaths.load_scan_info`; unpadded folders invisible | **Fixed** `9b131a7d` — one `read_scan_info_file`, `^Scan(\d+)$` |
| 6b | *(new)* no tests for the two new data-utils functions | **Fixed** `339d8fa2` — 9 added. My first log fixture omitted the `scan=ScanNNN -` segment, so it silently never matched; transcribed from a real line |
| 6c | *(new)* root-logger call in library code; stale perf claims | **Fixed** `339d8fa2` — module logger; CHANGELOG now claims only what was measured |
| 7 | Contract files didn't travel | **Fixed** `9b131a7d` — portal CLAUDE.md, DEPLOYMENT.md, `bootstrap_host.sh`, `site.env.example`. Reviewer traced it through the unit template and env loader |
| 8 | Root CLAUDE.md known-debt contradiction | **Fixed** `9b131a7d` — replacement not refactor; `EXPERIMENT_FILE_IDS` declared unowned |
| 9 | Duplicate CHANGELOG bullet, dead constant, `_as_int` 503, expand-all one level | **Fixed** `9b131a7d` |
| 9b | `read_scan` unused | **Withdrawn by the reviewer** — it is the single entry point |

### Deferred, with the owner's agreement

| # | Item | Why deferred |
|---|---|---|
| 5 | Move `scan_reader.py` into `geecs-data-utils` | Owner agreed with the placement argument and directed the *reuse* half be done now (it is). The relocation is a structural call with phase 02 pending |
| 6d | `ScanSummary` vs `scans_database.ScanMetadata` overlap | Same decision |
| 6e | Two day views (`/day/` from Tiled, `/log/day/` from folders) that can disagree | Needs an owner ruling, not an implementation choice |
| 9c | `geecs_scan_log` vs `scan_log_loader` naming collision | Owner call; sharper now that `scan_reader` imports `scan_log_loader` |
| 1d | Reading `scan.log` to split running-vs-aborted | Named as open work in the package CLAUDE.md, with `running` flagged as the status to add |

### Tests

`GeecsScanLog` 41 · `GEECS-DataPortal` 212 · `GEECS-Data-Utils` 334 — all
green via `scripts/check.sh`.

No finding is left undispositioned.

PR #827, opened 2026-09-12
GeecsBluesky 0.82.0: run_action as a queue plan — the action library over the device namespace (#807)

skbarber:
Phase-2 warm-up for the native-Bluesky rebuild (#807): the plan-layer ledger's "action plans not registered as a queue plan (run_action gap)" closed.

## What

- **`run_action(name)`** is registered beside the scan verbs (`plan_names.GEECS_PLAN_NAMES`; `NON_SCAN_PLAN_NAMES = ("mv", "run_action")` names the two queue items that are not scans). It resolves *name* in the experiment's action library (`ConfigsRepoResolver.resolve_action_plan` / `action_plan_registry`) and runs the compiled steps — the existing `plans/action_compiler.compile_action_plan`, unchanged — over the device namespace. No run is opened: nothing is claimed, nothing is written; a `check` mismatch aborts the item with `ActionCheckFailedError`.
- **The namespace is the compiler's `SettableFactory`**: `GeecsNamespace.get_settable` / `get_readable` hand out the child the namespace already built for `(device, variable)` — the Movable for a settable, the served signal otherwise. A `set` step on a read-only variable, or an unknown device/variable, raises `GeecsConfigurationError` (typed, listed by the manager as the item's error). `connect_on_demand` connects the touched children as it does for scans.
- **`registry.bind_plans(profiles, resolver=, settables=)`** replaces `bind_strict_plans` and returns every registered name, so the startup profile and the tests have one call. The hermetic worker (`QS_DEVICE_NAMESPACE=off`) registers a `run_action` that refuses, keeping the manager's plan list identical in every mode (the readiness oneshot asserts it).
- `qs_client`: `submit_plan("run_action", ["Amp4_DUMP_HP"])` works unchanged (the name check reads `GEECS_PLAN_NAMES`); a preset cannot name it (`PRESET_PLAN_NAMES`, same rule as `mv`, message names the verb).
- **Deleted**: `devices/ca/action_signals.py` (`CaActionSignalFactory`) and its test — the session-era CA factory that nothing used since #816; the namespace's children are the signals an action plan touches.

LOC by concern: registry + plan_names ≈ +100/−15; namespace ≈ +30; startup profile ≈ +10/−8; qs_client presets ≈ +15/−6; tests +170 (`tests/test_run_action.py` new, registry/presets/probe edits) / −100 (the deleted test); the deleted factory −190; docs (CLAUDE.md, CHANGELOG) ≈ +30.

Judgment calls, cheap to veto: the `bind_strict_plans` → `bind_plans` rename (the old name became a misnomer); the `NON_SCAN_PLAN_NAMES` constant in the import-light module rather than two literals in two packages.

Follow-up, not in this PR: the Undulator `actions.yaml` on the configs `main` is still the legacy dialect (`convert_action_library` reads it). Per the disposable-configs policy it should be regenerated as an `ActionPlanLibrary` v1 document and the converter deleted — its own small change.

## Tests

`./scripts/check.sh GeecsBluesky` (lint + the suite the way CI runs it): **366 passed, 3 skipped, 1 deselected**. New pins in `tests/test_run_action.py`: the namespace as a `SettableFactory` (Movable child / served signal / read-only refusal / unknown names), a plan's `set` landing on the child's `:SP` and its `check` reading the served readback with no `open_run` in the message stream, a failed check aborting, the three refusals, and registration beside the scan verbs; `test_plan_registry.py` validates a `run_action` item against the manager's own `validate_plan`; the hermetic startup probe pins the profile registering it.

## Hardware verification

**OWED**: one `submit_plan("run_action", [<plan>])` through the production manager on the worker host, with a plan whose steps are safe with the laser off (the library's plans move shutters and PLC channels — the maintainer's pick), expecting the item to complete with the sets visible on the devices' readbacks and no scan number claimed.


skbarber:
# Adversarial review (fresh-context subagent)

Reviewed `gh pr diff 827` (plans/run-action → feature/native-bluesky-plans) against the branch tree (`git show plans/run-action:…`), the queueserver manager in the package venv, and the root/package `CLAUDE.md`s. The PR's new tests were run from an extracted copy of the branch (`tests/test_run_action.py`, `tests/test_plan_registry.py`, `tests/qs_client/test_presets.py`: 24 passed). Findings ranked by severity; verified non-findings at the end.

## Findings

**1. P2 — `user_group_permissions.yaml` `operator` group still excludes `run_action`, and the pin the file claims does not exist.**
`GeecsBluesky/qserver/user_group_permissions.yaml` (untouched by the diff) allows the operator group `:^(count|scan|…|x2x_scan|mv)$` — no `run_action`. The comment above it says "`tests/test_deploy_templates.py` pins this line to that tuple", but `git grep allowed_plans|x2x_scan| plans/run-action -- GeecsBluesky/tests scripts` finds no such pin (only `tests/qs_client/test_readiness.py:45`, which checks a fake verdict against `GEECS_PLAN_NAMES`, not the YAML). Today this is masked: every client (`qs_client`'s `REManagerAPI(...)` at `client.py:593`, the `qserver` CLI, `qserver_ready.py` with `DEFAULT_USER_GROUP = "primary"`) submits as `primary`, which is `:.*`. The file's own note says production tightens `primary` to the operator shape — the day that happens, `submit_plan("run_action", …)` is rejected with "not in the list of allowed plans" while the readiness oneshot (asserting against `primary`) still reports ready. Concrete fix in this PR: add `run_action` to the operator regex and add the test the comment promises (derive the regex from `GEECS_PLAN_NAMES` so the next verb cannot drift either).

**2. P2 — the worker's action library is cached for the process lifetime; edits to `actions.yaml` are invisible to `run_action` until an environment reopen.**
`startup.py` now holds one `_resolver = ConfigsRepoResolver(_experiment)` and `run_action` calls `resolver.resolve_action_plan(name)` / `action_plan_registry()`, both of which go through `_action_library()` (`config_resolver.py:306-317`), which fills `_action_library_cache` on first use and never invalidates it. The other kinds the same resolver serves (presets, trigger profiles, experiment defaults) are re-read from disk on every call via `_load_yaml`; the action library is the only cached one, and this PR is what puts it behind a long-lived process. Scenario: an operator adds or edits `Amp4_DUMP_LP` in the console's Action Library editor (`GEECS-Console/geecs_console/services/action_library_store.py` writes the same `action_library/actions.yaml`), then submits `run_action("Amp4_DUMP_LP")` → refused with "not in the … action library. Known plans: […]" listing the stale set; worse, an *edited* existing plan silently runs its old steps. Nothing in the diff or CHANGELOG states the restart requirement. Fix: drop the cache (read per call like the presets — it is one small YAML per queue item) or key it on the file's mtime.

**3. P3 — native-save camera controls produce a misleading refusal.**
For a triggered device whose settables include both `save` and `localsavingpath`, the namespace hands those to the detector (`NATIVE_SAVE_VARIABLES`, `namespace.py:109`, filtered out of `typed` at ~line 404), so `get_settable("UC_Cam", "save")` raises "device namespace: UC_Cam: no served scalar variable 'save'" (verified with a probe roster). The variable *is* served by the gateway; the message names the wrong cause. Any legacy action step that toggles a camera's saving (the legacy ActionManager allowed any set) now fails with a message that sends the operator to the DB. Either route the step to the detector's own save signal or say "owned by the detector's native saving; not settable from an action plan".

**4. P3 — placement: the `run_action` plan body sits in the registration table.**
`plans/registry.py` is documented (its module docstring and `CLAUDE.md`) as "the registration table"; this PR adds `run_action_plan` — the plan's implementation, the resolver call, the logging — there and imports the compiler into it. The natural home is beside `compile_action_plan` (`plans/action_compiler.py`, or a `plans/action.py`), with `bind_plans` importing it the way it imports `strict_plan`'s inputs. A client-side "preview" (`flatten_action_steps`) verb or a dry-run flag would land in the same file next. Cheap to move now; a judgment call, so flagged not demanded.

**5. P3 — the client half and the runbook do not know the new verb (not a break introduced here).**
No Console/MCP module imports `bind_strict_plans`, `CaActionSignalFactory`, or `geecs_bluesky.devices.ca.action_signals` (repo-wide `git grep` on the branch), so the deletion/rename breaks no consumer. But: `GEECS-Console/geecs_console/app/actions_menu.py:224` passes `submitter.run_action`, and no Submitter on the branch defines it (grep of `geecs_console/services`); `queue_panel.py:209` labels only the funnel's `geecs_run_action_plan`, so a `run_action` item renders as `run_action (plan)`; `qserver/README.md` (lines 60-70, 110-116) documents `mv` as the non-scan queue item with a CLI example and none for `run_action` — the runbook the owed hardware check will be run from. The Console rewire is deferred by policy; the README example and a ledger line are in scope for this PR.

## Verified non-findings (checked, refuted)

- **Manager handling of `name: str`**: `profile_ops.prepare_plan` → `process_parameter_value` leaves a parameter whose annotation carries no `devices`/`plans` untouched (`convert_*` stay `None`, `eval_expressions` False), so an action name equal to a device name is not swapped for the device and nothing is `eval`ed. An unknown action name validates at submission (any `str`) and fails at run time with the resolver's `GeecsConfigurationError` — the same shape as an `mv` typo; `test_queue_items_validate_against_the_bound_plans` covers the submission side.
- **`bps.rd` on namespace children**: probed on a mock namespace — `CaMotor` and `CaSettable` report `hints == {}` and a single read key (the readback), a bare `SignalR` likewise; `rd` returns the value in all three cases. The "settable's child for a check step" path the tests skip does work.
- **`connect_on_demand`**: `is_namespace_object` walks `.parent` to the device carrying `_geecs_namespace_member`, and `set`/`read` messages name the child in `msg.obj`, so an action's children connect on first touch; `ensure_connected` is idempotent.
- **Readiness / deployed `plans_allowed`**: `qserver_ready.py` asserts `GEECS_PLAN_NAMES` against `primary` (`:.*`), so the deployed worker lists `run_action` after this lands and the oneshot passes; the hermetic probe pins the refusing `run_action` under the same name.
- **Redundancy**: the only other `SettableFactory` on the branch is the in-memory one in `tests/test_action_compiler.py`; `GeecsNamespace.get_settable/get_readable` are a thin typed wrapper over the existing `variable()`, not a second lookup; `NON_SCAN_PLAN_NAMES` replaces two literals.
- **Invariants**: no run is opened (pinned by the `open_run not in commands` assertion, which is non-vacuous because `set`/`read` are asserted present), so no scan folder is claimed; no facility literals or account names added; CHANGELOG/version bump/CLAUDE.md travel with the behavior.

skbarber:
## Dispositions (commit 5e275892)

1. **Fixed** — `run_action` added to the operator group's regex, and the pin the comment promised now exists: `tests/test_deploy_templates.py::test_operator_group_allows_exactly_the_registered_plans` parses the YAML and asserts the alternation equals `GEECS_PLAN_NAMES` (and that every name matches the pattern).
2. **Fixed** — `ConfigsRepoResolver._action_library` reads `actions.yaml` on every call (the cache and its field removed; docstring says why); CHANGELOG entry under Changed.
3. **Fixed** — `GeecsNamespace.get_settable` names the cause when the variable is one of the native-saving controls on a `native_save` detector ("owned by the detector's native saving … not settable from an action plan"); pinned by `test_native_save_controls_are_refused_by_name`.
4. **Fixed** — `run_action_plan` moved to `plans/action_compiler.py` beside `compile_action_plan`; `registry.bind_plans` imports it like it imports the strict hooks; the registry's docstring points there.
5. **Fixed (README) / noted (Console)** — `qserver/README.md` gains the `run_action` CLI example and the "Manual moves and action plans" section names the funnel's `geecs_run_action_plan` as gone and the Console's actions menu as waiting for its rewire. The Console/MCP mismatch is the deferred rewire (plan of record §10.5), not fixed here.

`./scripts/check.sh GeecsBluesky` after the fixes: **368 passed, 3 skipped, 1 deselected**.

skbarber:
Re-review of 5e275892 by the same reviewer: all five findings CONFIRMED (the permissions pin was mutation-tested; the cache removal, the native-save refusal, the move to the compiler module, and the README example verified; 52 tests run on a clean extract of the commit). CI green. Ready for the maintainer; the hardware check (one `run_action` item through the production manager with a safe plan) remains OWED as the PR body says.

skbarber:
## Codex review — disposition (commit 6ce3f279)

- **P2, lazy resolution leaves earlier writes applied on a later typo — Fixed.** `run_action` now walks the flattened plan (`flatten_action_steps`: nested `run` steps inlined, cycles and unknown plan names refused) before yielding anything, resolves every `set` / `check` target through the namespace (an unknown device or variable, or a read-only variable in a `set`, fails here) and reads each target once — the message that makes `connect_on_demand` connect it — so a PV the gateway does not serve also fails before the first write. Pinned by `test_a_late_typo_fails_before_the_first_write`: a plan whose third step names `DI.Ch99` fails with no `set` message emitted and both earlier setpoints unchanged. `./scripts/check.sh GeecsBluesky`: 369 passed, 3 skipped, 1 deselected.

skbarber:
## Hardware verification — done 2026-09-11 22:2x

The OWED check: `submit_plan("run_action", ["s1h_exercise"])` through the production RE Manager (worker at the #833 merge, GeecsBluesky 0.82.1), with `s1h_exercise` added to the Undulator action library on the configs repo's `main` (acee1db: set `U_S1H Current` 1.0, wait 3 s, set 0.0) and picked up by the worker with **no restart** (the library is read per item).

- Item `run_action ['s1h_exercise']`: `exit_status completed`, `run_uids []`, 6.5 s.
- `U_S1H Current` readback sampled at 2 Hz through the CA gateway: 0.0001 A → **0.9999 A** held for ~4.5 s → back to 0.0001 A.
- No scan folder claimed (newest folder Scan008 before and after) — no run is opened by the plan.

Side finding during the check, filed as #838: a timed-out "download the lists of existing plans and devices" (during the #834 OOM thrash) had left the manager idle with empty allowed-plan lists, so the first submission was refused; `permissions_reload` / `environment_update` recovered it without a restart.

PR #826, opened 2026-09-12
Planning: #806 acceptance record current (Tiled read, fleet roll, launcher lesson); phase 1 marked complete (#807)

skbarber:
Docs-only warm-up for the native-Bluesky rebuild (#807), first item of the phase-2 docket in `03_clean_room_rebuild.md` §8.

## What

- `07_806_acceptance.md`: the Tiled row now records the verified read (after the server's `readable_storage` + `HDF5_USE_FILE_LOCKING=FALSE` fix); the "Owed" section becomes "Settled after the merge" — the Tiled server fix, the production worker restart on the nine-host plugin list, the #824 fleet roll (nine gateways at 0.7.1) — and carries the launcher lesson from #824's review (never copy `launch.bat` over a running service and `:restart`; `cmd` resumes a batch file by byte offset). A short "Watch period" section says what happens before `Compression=zlib`.
- `06_pva_file_plugin.md` §8: the rollout status line says what is done and what remains (watch period; the later narrowing of `LvNativeFileDataLogic`).
- `03_clean_room_rebuild.md`: the header status and the §2 state table (the worker on the feature branch, #822/#823/#824 merged, #809 closed) and §8's phase-1 paragraph (complete; the small debts carried into phase 2's warm-ups) — per the document's own staleness rule.

No package code changes, so no version bump (Planning/ is not a package). Facility literals: camera-server IPs only (accepted practice); no account names or home paths.

## Hardware verification

None needed — the record describes runs already made (Scan007 of 26_0911, the fleet roll of 2026-09-11).

PR #828, opened 2026-09-12
Planning: 08_gated_batch.md — the phase-2 design (gated batch, non-essential stream, per-frame scalars), argued before code (#807)

skbarber:
The phase-2 design for the native-Bluesky rebuild (#807), written before any code, per the plan of record's rule and the docket in `03_clean_room_rebuild.md` §8.

## What

- **`Planning/native_bluesky/08_gated_batch.md`** (new): the argument, in six parts — what phase 2 delivers; every API claim verified against the installed ophyd-async 0.19.3 / bluesky 1.15.0 source and this repo's plugin, with line numbers (§2); two corrections to `03` (the trigger box is not a flyer, it has no counter; the non-essential stream is a per-plan argument, not RunEngine-level `SupplementalData.flyers`) and one gap it left (where a fly run's scalars come from); the design — `acquisition="gated"` and `non_essential=[…]` on the bound scan verbs, `gated_take_reading` (prepare per step → kickoff → SCAN → complete → OFF → `Rewind` to the quota → collect), the per-plan fly wrapper with the stage/prepare the stock one lacks, per-frame scalars as NDAttributes in the plugin's stack (the stock data logic already describes them), the s-file from stream data, `essential` on `PresetDevice`; a three-PR sequencing with hardware acceptance per PR (`HTU-LaserOFF`'s SCAN = internal source, so gated can be exercised with the laser off); six open questions for the maintainer with recommendations.
- **`03_clean_room_rebuild.md`**, amended in the same PR per its staleness rule: §4.A (the box is Movable + Pausable, not a flyer), §4.B (non-essential per plan; gated as `acquisition="gated"`), §7 (the phase-2 verified facts), §8 item 2 (the three PRs), §10 item 9 (the open questions; old item 9 renumbered 10).

No package code; no version bump (Planning/ is not a package). Facility literals: none added beyond the profile names already in the docs.

Independent of #826 (different regions of `03`); either merges first.

## Hardware verification

None — a design document. The acceptance plan for the code is in §5 of the new file.


skbarber:

Scope: PR #828 at head `ef9c28e1` (`08_gated_batch.md` new, `03_clean_room_rebuild.md` amended), checked against ophyd-async 0.19.3 / bluesky 1.15.0 in `GeecsBluesky/.venv`, the repo's plugin/worker/client code, `05`/`06`, and the configs repo's `origin/main`.

**§2 line-number audit — what holds.** Every cited ophyd-async/bluesky claim checks out: `number_of_events` NonNegativeInt "0 means infinite" (`_detector.py:96`), provider reuse when `collections_per_event` is unchanged (`:464-500`), `collections_written` baselined at prepare (`:511`), `events_to_kickoff` set then `start_acquiring()` for external triggers (`:605-608`), `kickoff`'s `written + requested > ctx.written + number_of_collections` raise (`:685-693`, so `number_of_events=0` does raise on a frame between prepare and kickoff), `complete` = count wait then `wait_for_idle` on `is_last_kickoff` (`:548-549`), `requested=0` returning on the first observed value (`:547`), `observe_signals_value(timeout=…)` per update ("If an update is not produced in this time then raise", `_signal.py:503`), `trigger()` refusing `number_of_events != 1` (`:649`), `make_stream_docs` last_emitted → indices_written with the dropped-frames TODO (`_data_providers.py:139-155`), `ADHDFDataLogic` NumCapture=0 (`_data_logic.py:160-161`), Capture without set completion (`:206-208`), NDAttribute description from the XML (`:104-131`, `:215-229`), `fly_during_wrapper` inserting kickoff/complete/collect only (`preprocessors.py:866-906`), multi-object collect needing a predeclared stream and cutting at the minimum index (`bundlers.py:1076-1130`), `bp.fly` and `collect_while_completing` shapes, the deferred-pause docstring (`run_engine.py:394`). Repo claims hold too: `NumCapture` stored, never read by `_on_frame`; `_rewind` truncates, posts `NumCaptured_RBV=n`, sets `stale_before=now`; `ShotControl.pause()` SCAN → OFF and `resume()` → SCAN (pinned by `tests/test_shot_control_device.py:133-141`); `HTU-LaserOFF` on configs `main` is SCAN=Internal, ARMED/OFF=Single shot, STANDBY=External rising edges. Hygiene: the diff carries no account names, hostnames, home paths or emails.

The findings below survived my attempts to refute them.

### P1

**1. A gated `scan` has nowhere to record the scan variable (or anything read per step).** `geecs_per_step` today reads `[*detectors, *motors, bins]` into the primary event (`plans/strict.py:373-377`). §4.2 replaces the row with `collect(*D, name="primary")` and says `BinCounter` is not needed — but never says where the motors' per-step readback goes. It cannot go into `primary` alongside the collect: `declare_stream(*D, name="primary")` sets `_descriptor_objs["primary"] = D`, and a later `create("primary") / read(motor) / save()` hits `bundlers.py:606-607`: `elif frozenset(d_objs) != objs_read: raise RuntimeError("Mismatched objects read …")`. Declaring the motor in the same stream doesn't help either — `collect` looks up declared names by `frozenset(collect_objects)` (`:1096`), which would then not match. So as written, a gated `scan(U_S1H.current, …)` produces an s-file with no `U_S1H Current` column and `ScanInfo`'s axis has nothing to join to; §6.1 treats this as "scalar-only devices go baseline-only" but the *scanned motor* is not a telemetry question. The design must pick: a second event stream for `[*motors, bins]` per step (and say how the s-file writer joins it to the datum ordinal), or motors as `collect`-able, or something else — and say it in §4.2/§4.5.

**2. §4.4's attribute columns collide across cameras and do not match the strict row's column names.** The stock data logic names each NDAttribute data key **bare** — `data_key=name` from the XML (`_data_logic.py:217`), no detector prefix (`datakey_suffix` only touches the frame dataset; `devices/detector.py:614`). The bundler merges per-object keys with `data_keys.update(dks)` (`bundlers.py:285`), silently overwriting. Consequences: (a) two plugin-backed cameras in one stream — exactly the 2b runbook's "gated scan with two plugin-backed cameras" — already collide today on `acq_timestamp`/`recv_timestamp` (no test covers it: `test_two_cameras_share_one_fire` uses one `hdf_plugins=` fixture); §4.4 would extend the collision to every subscribed scalar (`exposure`, …). (b) The stated goal — "a gated row and a strict row carry the same columns for that device" — is false: strict columns are `<ophyd name>-<var>` (`_column_headers`, `detector.py:623-630`; `ScalarsDataLogic`), the stream columns would be `<var>`. Fix belongs in the design: generate the XML with `<ophyd name>-<var>` attribute names (the existing column convention), which makes the keys unique per device and identical across modes — and then `TIMESTAMPS_DATASET` in `scan_stack.py` and the existing `acq_timestamp` attribute need the same treatment (a breaking rename of a just-shipped file layout; say so).

### P2

**3. "`update` is the whole message" is false; §4.4 needs a gateway subscription change it does not list.** `server.py:207-211` subscribes `[var, "acq_timestamp", "systimestamp"]` and `_parse_subscription` extracts only the *subscribed* names from the push (`tcp_subscriber.py:370-415`, regex anchored on the subscribed list). So `_on_frame`'s `update` is at most three keys. Per-shot scalars beside the frame require the gateway to add the DB `get='yes'` list to that `subscribe([...])` call (cheap, one subscription — the "no second subscription" part survives), but 2a's scope and the "nothing the gateway does not already hold" claim must change.

**4. Placement: the "same rule the namespace uses" lives where the gateway cannot import it.** The subscribed-scalars rule is `GeecsDbScalarPolicy(...).subscribed_by_device()` in `geecs_bluesky/db_runtime.py:60` (used at `namespace.py:240`). GeecsPvaGateway depends on GEECS-Core only. Either the policy moves to `geecs_core.db` (next to `variable_types`, the rule the root `CLAUDE.md` says both gateways and GeecsBluesky share) or the gateway re-derives `get='yes'` — a second copy of the rule whose drift breaks the "same columns" guarantee. Name the move in 2a.

**5. `03` still says the opposite of its own amendments.** At the PR head, `03` line 359 (`save set synchronous flag → … SupplementalData.flyers`), line 366 (`ShotControl: … FlyerController for gated mode`), line 369 (`free run — the contributor job → SupplementalData.flyers`) and line 850 (§11.5 item 3: "A native home for the non-essential stream … `SupplementalData.flyers` (§4.B)") contradict the amended §4.A ("Not a flyer") and §4.B ("per plan"). The mapping table and §11.5 need the same amendment.

**6. `declare_stream` is missing from both pseudo-codes, and the stock `fly_during_wrapper` would put the non-essential datums in `primary`.** §4.2's `collect(*D, name="primary")` asserts a prior `declare_stream(*D, name="primary")` (`bundlers.py:1101-1104`), which must come after the first `prepare` (`describe_collect` raises "Prepare not run", `_detector.py:735`). §4.3 relies on `fly_during_wrapper`, whose collect is `Msg("collect", flyer)` with no name (`preprocessors.py:894`); for a single undeclared object the bundler routes that to stream **`primary`** (`bundlers.py:732`: `message_stream_name or "primary"`), which in a strict run would re-prepare `primary`'s descriptor with the NE's keys. The per-device `<device>_stream` only happens if the bound wrapper emits `declare_stream(ne, name=…)` first (then the bare collect picks the single declared name, `:1105-1107`). Write the declare into both sequences.

**7. Mode inferred from `number_of_events == 1` misclassifies a gated step at the registry's default `shots_per_step=1`.** `strict_plan` defaults `shots_per_step` to 1 (`registry.py:222`, `:267`), so `scan([cam], …, acquisition="gated")` prepares with `number_of_events=1` — the strict signature — and `wait_for_idle` would do the stamp wait against a `_t0` that `baseline()` never set in this mode (`GeecsDetector.trigger` is the only caller of `baseline()`, `detector.py:731`). It "works" only because the undrained queue holds a newer stamp. Make the mode explicit — set fly mode in a `kickoff` override and clear it in `trigger` — rather than reading it off N.

**8. `StackCheckCallback` will warn on every gated or non-essential stack; §4.3's "reports … as it does for primary" is not true today.** `expected` is built from primary *event* rows' `<name>-acq_timestamp` (`callbacks.py:412-418`); with no primary events it is empty and `_check` emits the warning verdict "N frame(s) … but 0 row(s) own a frame" (`callbacks.py:470-475`). 2b must rework the check to take referenced counts from the datums' `indices` (and the stamps from the stack itself) — an unlisted 2b item, and without it the runbook's first gated run logs a false stack failure.

**9. The rewind does not always drop the in-flight frame — the condition is unstated.** Stamp = edge time + drain, drain 0–220 ms per camera (`03` §11.4, `04` M1); `Rewind` drops a later frame only if `stamp < stale_before − 0.1` (`file_plugin.py:STALE_MARGIN_S`, `_on_frame`). So the slipped edge's frame is dropped iff `(t_OFF_effective − t_edge) + (OFF-ack → Rewind) > drain − 0.1`. For a high-drain camera whose edge slipped within ~100 ms of OFF taking effect (a 0.70 s exposure puts OFF right at the next edge: M6's 0.8 s + 155 ms put), the frame is *newer* than the watermark and is written after the rewind. It then either lands between the next step's `prepare` and `kickoff` (→ `kickoff` raises, §2 (b)) or after `kickoff` (→ referenced as frame 1 of the next datum: N+1 frames at the wrong position; `make_stream_docs` is contiguous from `last_emitted`). Not the common case, but it is the exactness claim's one hole; state the condition and make 2b's acceptance include the highest-drain camera, or give `Rewind` a stamp watermark argument instead of `now`.

### P3

**10. Redundant 2b item.** "The s-file writer *skips* stream-only runs with a `scan.log` line" already exists: `SFileCallback.on_stop` logs "no primary events, no scalar files" and returns (`callbacks.py:274-280`). Drop it from 2b or say it is the existing behaviour.

**11. Version label.** 2a calls the Data-Utils change (reading the attribute group in `scan_stack`) a "patch"; it is a new capability → minor per the root `CLAUDE.md` table.

**12. Asymmetric preflight.** §4.3 gives non-essential devices a client-side "must have an `hdf` child" rule but leaves gated *essential* devices to fail at the worker's `kickoff` ("not streamable"). Same rule, both lists — the preflight already walks `item.references` against the device tree (`submit_preflight.py:189-230`), so it is one more membership check.

**13. Split of content.** `08` vs `03` is right (argument in `08`, facts and decisions in `03`); the only misplacement is finding 5 — the unamended table/§11.5 in `03`.

No finding against: the per-frame `exposure_timeout` claim, the `Rewind`-to-`ctx.collections_written + quota` arithmetic (absolute session count, same as `discard_uncollected`'s `last_emitted`), the stock `ADHDFDataLogic` describing NDAttributes from `driver.nd_attributes_file` with `driver=io` (`detector.py:611`, so nothing of ours on the worker for the *describe* step), `essential` on `PresetDevice` in GEECS-Schemas, the per-plan `non_essential` argument over `SupplementalData`, or the trigger-profile facts.

skbarber:
## Dispositions (commit 24bda398)

Finding 2 was **reproduced on mocks before amending**: two `GeecsDetector`s with the plugin's real `NDAttributesFile` XML set on their `hdf` mocks, a strict `count` → `ValueError: Data keys (field names) … collide … {'acq_timestamp', 'recv_timestamp'}`, stop document `exit_status='fail'`, after the scan number is claimed. That is the deployed worker's state for any preset with two plugin-backed cameras; it is being fixed as its own PR ahead of phase 2 (issue filed).

1. **Fixed (design)** — §4.2 gains a second event stream `steps` (the motors' readbacks and `bin_number`, one event per step after the collect); §4.5 joins event *k* of `steps` to datum *k* of `primary`; `BinCounter` stays.
2. **Fixed (design + reproduced)** — §4.4: attribute names are `<ophyd name>-<variable>` via `normalize_component`, the two existing attributes renamed the same way, `scan_stack.TIMESTAMPS_DATASET` becomes a lookup accepting both spellings; the on-disk change is called out; the rename ships first as its own PR (§5 item 0).
3. **Fixed (design)** — §2 corrected (`update` holds only the subscribed names); §4.4/§5 list the widened subscription as a 2a item.
4. **Fixed (design)** — §4.4/§5: `GeecsDbScalarPolicy` moves to `geecs_core.db`; the namespace imports it from there; GEECS-Core minor in 2a.
5. **Fixed** — `03`'s mapping table (three rows) and §12 item 3 amended.
6. **Fixed (design)** — `declare_stream` written into both sequences (after the first prepare; for the non-essential devices before kickoff, naming the per-device stream).
7. **Fixed (design)** — the fly mode is explicit: set by a `kickoff` override, cleared by `trigger`; the `shots_per_step=1` case named as the reason.
8. **Fixed (design)** — the `StackCheckCallback` rework (counts from the datums' `indices`, stamps from the stack) listed in 2b.
9. **Fixed (design)** — the condition stated; the plan waits one period plus the largest drain offset plus a margin after OFF before rewinding (deterministic); the watermark-argument alternative kept in reserve; 2b's acceptance includes the highest-drain camera.
10. **Fixed** — the redundant 2b item replaced by "already skips … 2b leaves that".
11. **Fixed** — Data-Utils minor.
12. **Fixed (design)** — the preflight rule applies to both lists (gated essential devices too).
13. No action (agreed split; 5 fixed).

skbarber:
Re-review of 24bda398 by the same reviewer: all thirteen findings CONFIRMED, no new findings. Its two nits (§4.6 naming the preflight rule for both lists; how a gated `count` batches — one batch, `num` as the quota) are fixed in the follow-up commit. Ready for the maintainer; the six open questions in §6 are the maintainer's to answer before PR 2b.

skbarber:
## Codex review — dispositions (commit 77709ff5)

- **P1, §2 versus §4.3 on the non-essential collect — Fixed as a wording defect; the design stands.** Both statements were true of different cases and §2 dropped the qualifier. bluesky 1.15.0 `bundlers.py:1098-1107`: a nameless `Msg("collect", obj)` (what `fly_during_wrapper` emits) resolves to the one stream declared for exactly that object set ("if one was not provided, but a single stream has been declared, then use that stream"); it lands in `primary` (`:732`) only when nothing is declared for the object. So the `declare_stream(ne, name="<device>_stream")` the bound plan emits before the wrapper's kickoff is what routes the datums, and the stock wrapper needs no replacement. §2 now cites both branches and §4.3 says which one it relies on.
- **P2, `GeecsBluesky/CLAUDE.md` still names `SupplementalData.flyers` and a `FlyerController` — Fixed in #826**, which already carries a GeecsBluesky docs bump (0.81.1) and merges before this PR in the agreed order; its CLAUDE.md now says the box is not a flyer and the non-essential stream is per plan, pointing here. Keeping the one file's edits in one PR avoids a three-way CHANGELOG/pyproject conflict.

PR #830, opened 2026-09-12
GeecsPvaGateway 0.8.0: file-plugin attribute names carry the device — two plugin-backed cameras in one run no longer collide (#829)

### LINKED ISSUE #829: File plugin: NDAttribute data keys are bare, so two plugin-backed cameras in one run collide (`acq_timestamp`, `recv_timestamp`) and the run fails
Found by the adversarial review of #828 and reproduced on mocks (GeecsBluesky at the feature head): two `GeecsDetector`s whose `hdf` plugin serves the real `NDAttributesFile` XML, one strict `count([a, b], 2)` →

```
ValueError: Data keys (field names) from <GeecsDetector …> collide with those from <GeecsDetector …>. The colliding keys are {'acq_timestamp', 'recv_timestamp'}
```

stop document `exit_status='fail'`, after the scan number was claimed (an empty numbered folder is left behind).

**Cause.** The stock `ADHDFDataLogic` names each NDAttribute data key from the XML `name` with no detector prefix (`ophyd_async/epics/adcore/_data_logic.py:217`; `datakey_suffix` only touches the frame dataset), and the plugin's XML (`GeecsPvaGateway/geecs_pva_gateway/file_plugin.py`, `NDATTRIBUTES_XML`) declares `acq_timestamp` and `recv_timestamp` for every camera. The bundler refuses colliding keys across objects.

**Why it matters now.** Since the 2026-09-11 15:00 restart every camera on the nine rolled boxes is plugin-backed, so every preset with two or more cameras hits this on the deployed worker. The #806 acceptance (Scan007) used one camera; `tests/test_strict_plans.py::test_two_cameras_share_one_fire` uses no plugin, so no test caught it.

**Fix (Planning/native_bluesky/08_gated_batch.md §4.4, §5 item 0).** Attribute names carry the device: `<ophyd name>-acq_timestamp`, `<ophyd name>-recv_timestamp` (`geecs_core.pv_naming.normalize_component(device)`, the same rule the worker's `safe_name` uses for its event columns), which is also the strict row's column convention. The on-disk dataset names change with them; `geecs_data_utils.io.scan_stack.TIMESTAMPS_DATASET` becomes a lookup accepting either spelling so the files written between #823 and the fix stay readable. Deploy: merge, pull the share clone, `:restart` on each box (no launcher change). Until then: one plugin-backed camera per preset, or `save_images: false` on all but one.

skbarber:
Fixes #829, found by the adversarial review of #828 and reproduced on mocks: with the plugin's real `NDAttributesFile` XML on two `GeecsDetector`s, a strict `count([a, b], 2)` fails at the descriptor — `ValueError: Data keys (field names) … collide … {'acq_timestamp', 'recv_timestamp'}` — after the scan number is claimed. Since the 2026-09-11 15:00 restart every camera on the nine rolled boxes is plugin-backed, so this is every multi-camera preset on the deployed worker.

## What

- **GeecsPvaGateway 0.8.0** — `file_plugin.attribute_names(device, variable)` / `attributes_xml(device, variable)`: the per-frame attribute datasets and the served XML name `<device>-hdf-<variable>-frame_acq_timestamp` / `-frame_recv_timestamp` (both parts through `geecs_core.pv_naming.normalize_component`). The stock `ADHDFDataLogic` turns attribute names into stream data keys verbatim, so the names must be **unique across the cameras of one run** (the bare names collided) **and disjoint from every event column** — the first cut of this PR used `<device>-acq_timestamp`, which is the camera's own CA stamp column; the review showed the stream key overwriting that column's description and Tiled's normalizer then dropping the column and refusing the stop document (a regression for single-camera runs too). `-hdf-` names the plugin child and the `frame_` suffixes never spell an event column. `NDAttributesFile` is set per instance; the module-level `ATTRIBUTES` / `NDATTRIBUTES_XML` constants are gone. On-disk layout change of a one-day-old format, called out in the CHANGELOG and `06_pva_file_plugin.md`.
- **GEECS-Data-Utils 0.29.0** — `io.scan_stack.timestamps_dataset(f)` resolves either spelling (the current `…-frame_acq_timestamp`, or the bare `acq_timestamp` of stacks written by 0.7.x); `is_stack_file`, `read_stack_timestamps`, `read_shot_for_acq_timestamp` go through it. `TIMESTAMPS_DATASET` stays as the bare-name spelling (the tests of five packages write it; those files remain valid input); `TIMESTAMP_SUFFIX` is the current suffix — deliberately not `-acq_timestamp`, which `tiled_schema.is_acq_timestamp_column` reserves for a device's event column. `CLAUDE.md` layout line updated.
- **GEECS-DataPortal 0.21.4** — the stack cache resolves the dataset the same way.
- **GeecsBluesky** — tests only, no version bump: `_plugin_camera` now serves the real XML shape by default (every plugin-backed strict-plan test ran with no attributes before, which is why #829 went unseen); `test_two_plugin_cameras_in_one_run_keep_distinct_attribute_keys` runs two plugin-backed cameras through a strict count, asserts the CA stamp column is internal (no `external`, CA source) beside the external attribute keys, and drives every document through bluesky's `RunNormalizer` (the front of `TiledWriter`) asserting the stop is emitted and the column survives; `test_bare_attribute_names_collide_on_the_second_camera` pins what #829 looked like. `StackCheckCallback` needs no change (it reads through `read_stack_timestamps`).

LOC by concern: plugin ≈ +50/−20; scan_stack ≈ +35/−4; portal cache 3 lines; tests ≈ +140/−5 across three packages; docs/changelogs ≈ +60.

Facility literals: none. Version-bump note: #826 bumps GeecsPvaGateway to 0.7.2 (docs); whichever of #826 / this merges second needs a one-line CHANGELOG/pyproject rebase.

## Tests

`./scripts/check.sh` (lint + every changed package's suite): GeecsPvaGateway **47 passed**; GEECS-Data-Utils **328 passed, 3 skipped**; GEECS-DataPortal **209 passed, 32 skipped**; GeecsBluesky **368 passed, 3 skipped, 1 deselected**; the unchanged suites green.

## Hardware verification

**OWED** (deploy step, after merge): pull the share clone (`Active Version/GEECS-Plugins`), `:restart` PV on each of the nine boxes (no launcher change — the reinstall picks up 0.8.0), then one strict `count` through the manager on a preset with **two** plugin-backed cameras: expected a `success` stop document, two stacks each with `<device>-hdf-image-frame_acq_timestamp` under `NDAttributes`, the stack check's "match the rows' stamps" line for both, and the run's `<device>-acq_timestamp` columns readable through Tiled. Until then: one plugin-backed camera per preset, or `save_images: false` on all but one.


skbarber:

Baseline: the three named suites pass on the branch head (`test_file_plugin.py` 7 passed, `test_scan_stack.py` 10 passed, `test_strict_plans.py` + `test_callbacks.py` 23 passed). Findings below are what survived reproduction.

### 1. P1 — The new attribute key `<name>-acq_timestamp` collides with the detector's own CA stamp column; Tiled loses the column and never receives the run's stop (a regression for single-camera scans too)

`GeecsDetector` already emits a scalar column named `<name>-acq_timestamp` (`GeecsBluesky/geecs_bluesky/devices/detector.py:559` — `self.acq_timestamp` child; pinned by `tests/test_geecs_detector.py:82`). The plugin now serves an NDAttribute with the identical name, which the stock `ADHDFDataLogic` turns into an `external: "STREAM:"` data key. `StandardDetector.describe` merges readable providers first and streamable providers second through `merge_gathered_dicts`, which is a plain `dict.update` (ophyd-async 0.19.3, `_utils.py`) — so the HDF description silently **overwrites** the CA scalar's. The RunEngine's `read()`/`save()` path does not check for event data on external keys (only `collect()` does, `bundlers.py:976`), so the run reports `success`.

Reproduced on mocks with the PR's own two-camera scenario (descriptor dumped from the run):

```
data_keys["uc_a-acq_timestamp"] = {"source": "file://.../UC_A/UC_A.h5", "shape": [1],
                                   "dtype": "number", "external": "STREAM:", ...}
event["data"] = {"uc_a-meancounts": 0.0, "uc_a-acq_timestamp": 1001.0, ...}   # CA value, LabVIEW epoch on hardware
```

Fed through bluesky's `TiledWriter` document normalizer (`bluesky/callbacks/tiled_writer.py` — pure, no server needed; same class the worker subscribes via `tiled_integration.subscribe_tiled`):

| attribute XML shape | Tiled-side primary columns | stop document reaches Tiled |
|---|---|---|
| old bare names, 1 camera (feature head) | `bin_number, uc_a-acq_timestamp, uc_a-meancounts` | yes |
| **new `<name>-…`, 1 camera** | `bin_number, uc_a-meancounts` | **no — `RuntimeError: Cannot emit StreamDatum for uc_a-acq_timestamp because the corresponding Datum document is missing.`** |
| new `<name>-…`, 2 cameras | `bin_number, uc_a-meancounts, uc_b-meancounts` | no — same error |

Mechanism: the normalizer files `uc_a-acq_timestamp` under `_ext_keys`, strips it from the internal event, then treats the event's `1001.0` as a datum id to resolve; at `stop` it raises. `SafeDocumentCallback` then disables Tiled forwarding for the run, so the catalog holds an un-stopped run with no `<dev>-acq_timestamp` column. Every downstream reader of that column is broken: `assets/tiled_readback.py:249` (`_required_float(event_dict, "<dev>-acq_timestamp")`), `geecs_data_utils.tiled_schema.device_acq_timestamp_column` (the portal's shot join, `geecs_portal/app.py:228`), the Tiled → s-file exporter's `"<Device> acq_timestamp"` header (`_column_headers`, detector.py). The in-process `StackCheckCallback` and s-file writer read raw RunEngine events, where the value is still present — so the PR's proposed hardware check ("stack check's match line") would pass while Tiled is broken, and the #823 Tiled-read acceptance that passed on the feature head with one camera would now fail.

The PR text and `06_pva_file_plugin.md` present "spell the strict row's stamp column" as a feature; it is the defect. The keys need to be unique across cameras **and** disjoint from every detector column — e.g. `<name>-hdf-acq_timestamp` / `<name>-<HDF_PLUGIN_SUFFIX>-acq_timestamp` (the same `child-child` shape the `hdf` child already has). `scan_stack.timestamps_dataset`'s `endswith("-acq_timestamp")` keeps working unchanged.

### 2. P2 — The new worker test cannot fail on the wrong code, and the strict-plan plugin fixtures run with no attributes at all

`test_two_plugin_cameras_in_one_run_keep_distinct_attribute_keys` hand-writes the XML and touches no GeecsPvaGateway code (GeecsBluesky does not depend on it), and the GeecsBluesky diff is tests-only — so it passes identically on the feature head; the "fails on the feature head" claim only holds for the *old* XML, which the test never exercises. It also asserts only key membership and the stream-resource list, which is exactly why it is green with finding 1 present. Two concrete strengthenings: (a) assert `descriptor["data_keys"]["uc_a-acq_timestamp"]` has no `external` and its `source` is the CA PV (`mock+ca://testexp:uc_a:acq_timestamp`), and (b) push the collected documents through `TiledWriter`'s normalizer (the table above) — it is the consumer that actually broke. Separately, `_plugin_camera` (`tests/test_strict_plans.py:322`) never sets `nd_attributes_file`, so every existing plugin-backed strict-plan test (including the missed-frame rewind test) runs with zero attribute datasets; making the helper serve the real XML shape by default is what would have caught both #829 and finding 1.

### 3. P3 — Duplicated suffix rule in the same package

`scan_stack.timestamps_dataset` re-implements the `-acq_timestamp` suffix predicate that `geecs_data_utils/tiled_schema.py` already owns (`_ACQ_TIMESTAMP_SUFFIX = "-acq_timestamp"`, `is_acq_timestamp_column`, lines 57 and 116–129). Checked: `tiled_schema.py`, `native_files.py`, `tiled_export.py` — none resolve HDF dataset paths, so a new resolver is warranted, but the suffix constant/predicate should be shared rather than spelled twice (`TIMESTAMP_SUFFIX` vs `_ACQ_TIMESTAMP_SUFFIX`).

### 4. P3 — Contract docs not travelling with the layout change

`GEECS-Data-Utils/CLAUDE.md:27` still documents the stack layout as `/entry/instrument/NDAttributes/acq_timestamp`. No other contract page names the datasets (checked `GeecsPvaGateway/{CLAUDE,README,DEPLOYMENT}.md`, `GeecsBluesky/EVENT_SCHEMA.md`, `docs/`); `geecs-pva-gateway diff` (`diff.py:99`) reads through `read_stack_timestamps` and is fine. `06_pva_file_plugin.md`'s new paragraph will need rewriting once finding 1 changes the names.

Placement: `attribute_names`/`attributes_xml` belong in the plugin — GEECS-Data-Utils cannot host the rule without a new dependency on `geecs_core.pv_naming`, and the reader's suffix-scan is the right read-side shape. No placement findings.

skbarber:
## Dispositions (commit 725163b4)

1. **Fixed** — the names are now `<device>-hdf-<variable>-frame_acq_timestamp` / `-frame_recv_timestamp`: unique across cameras and never an event column's name (`-hdf-` is the plugin child, the `frame_` suffixes never spell a CA column). The first cut's "spell the strict row's stamp column" was the defect, as you said; the CHANGELOG, `06` and the PR body now say so. `06`'s paragraph and the phase-2 design (#828 §4.4) are amended to the disjointness rule.
2. **Fixed** — `_plugin_camera` serves the real XML shape by default (`_attributes_xml`), so every plugin-backed strict-plan test now runs with attribute datasets; the two-camera test asserts the CA stamp column is internal (no `external`, CA source) beside the external attribute keys and drives every document through `RunNormalizer`, asserting the stop is emitted, the normalized descriptor keeps `uc_a-acq_timestamp` internal and the event carries it; `test_bare_attribute_names_collide_on_the_second_camera` pins the #829 shape with the old bare names.
3. **Resolved by 1** — the stack key no longer ends in `-acq_timestamp` (deliberately: `is_acq_timestamp_column` reserves that suffix for a device's event column, and a stack key matching it would be mistaken for one); `TIMESTAMP_SUFFIX` is a different suffix, so there is no second copy of that predicate. The CHANGELOG says why.
4. **Fixed** — `GEECS-Data-Utils/CLAUDE.md`'s layout line updated; `06` rewritten for the final names.

`./scripts/check.sh` after the fixes: GeecsPvaGateway 47 passed; GEECS-Data-Utils 328 passed, 3 skipped; GEECS-DataPortal 209 passed, 32 skipped; GeecsBluesky 368 passed, 3 skipped, 1 deselected; lint green.

skbarber:
Re-review of 725163b4 by the same reviewer: all four findings CONFIRMED — its RunNormalizer table now shows the CA stamp column intact and the stop document reaching Tiled for one and two cameras; suites green in four packages. The one stale test comment it noted is fixed in the follow-up commit. Ready for the maintainer; deploy and the two-camera hardware check are in the PR body.

PR #831, opened 2026-09-12
GeecsWebTheme 0.1.0: one palette for every web surface

skbarber:
## What

Three web surfaces had grown three unrelated looks — all three mounted in
the same app. Clicking from a scan page to its log entry changed the
product.

| | Before | After |
|---|---|---|
| Portal `/day`, `/run` | dark-only, teal | shared tokens |
| Config editor `/configs` | a copied fork of the portal's 7 tokens | shared tokens |
| Scan logbook `/log` | its own light-first palette | shared tokens |

`GeecsWebTheme` is the shared vocabulary: surfaces, text, rules, accent, the
semantic trio, type stacks. Three palettes, each with a light and a dark
variant, plus a picker in the shared shell so a viewer's choice follows them
across `/day`, `/run`, `/configs` and `/log`.

## The themes

Drawn from the lab rather than invented:

- **`bella`** — the BELLA Center's red on black.
- **`laser`** — 532 nm pump green, with the breadboards' red carrying the
  criticals. Semantically clean: green for what's live, red for what's
  wrong, so the accent and the semantic colours don't fight.
- **`plasma`** — hydrogen Balmer. H-β 486 nm cyan as the accent, H-α 656 nm
  for the agent voice.

Default `laser`, following `prefers-color-scheme`.

## ⚠️ The portal's appearance changes for everyone

Dark-only becomes three palettes defaulting to `laser` following the system.
Flagged here rather than buried. The owner has confirmed the existing
colours were generated rather than chosen, so nothing deliberate is being
overridden — but it is visible to anyone already using it.

## Why its own package

The dependency only runs one way: everything web-facing needs it, it needs
nothing — **not even FastAPI**. It is a stylesheet, a script and
`static_dir()`. Inside the portal it would make the portal the owner of
something two other packages consume, and a fourth surface is coming (the
web scanner). A non-Python surface could link the file directly.

`ScanAnalysis` takes a `theme_url` from its host rather than importing the
theme, so the editor still runs standalone with a fallback palette and the
dependency stays one-way.

## The part worth reviewing: the guardrail

A token layer only works while components use it. One hardcoded colour
silently opts a component out of theming and nobody finds out until they
switch and one thing stays the wrong colour.

`tests/test_no_literal_colours.py` walks every registered surface and fails
on any colour literal outside a small allowlist — each entry carrying its
reason, because the allowlist is where this erodes if it erodes. A second
test checks no palette is missing a token another defines, since a missing
one inherits silently rather than failing loudly.

Writing it earned its keep immediately:

- **Two real leaks in my own code** — `color:#fff` on the accent in two
  places, which fails on a light-mode accent. Now `--on-accent`.
- **One false positive in the test itself** — `#765` in a JavaScript comment
  referencing a PR is a valid three-digit hex to a regex. The detector now
  reads only real CSS (an HTML file's `<style>` blocks, comments stripped),
  because a guardrail that cries wolf gets switched off.

## Per-concern LOC

| Concern | Lines |
|---|---:|
| `GeecsWebTheme` package (css, js, helper) | ~430 |
| Guardrail tests | ~150 |
| Portal ported + theme mount | ~70 |
| Config editor ported + `theme_url` | ~45 |
| Logbook ported (local tokens deleted) | ~-120 |
| Docs, changelogs, CI + check.sh | ~90 |

## Out of scope, deliberately

- **The vendored Plotly bundle** (286 literals). Chart theming is a Plotly
  template object derived from the tokens, not a patched `min.js`.
- **GEECS-Console.** Qt stylesheets are a different mechanism; the same
  token *values* could feed a Qt palette later.

Both confirmed out of scope by the owner.

## Tests

`GeecsWebTheme` 9 · `GEECS-DataPortal` 250 · `GeecsScanLog` 41 ·
`GEECS-Data-Utils` 334 · `ScanAnalysis` 192 — all green via
`./scripts/check.sh`.

Verified end to end through the real app: `/theme/theme.css` and
`/theme/theme.js` serve 200, and both `/day/…` and `/log/day/…` carry the
stylesheet link, the picker, and the before-paint stamp that stops the page
flashing the wrong palette.

## Hardware verification

**Not applicable** — no device, PV, or scan-execution path is touched.

## Base

Targets **`feature/scan-logger`**, the arc's integration branch, not master.


skbarber:
## Adversarial review — fresh-context reviewer

Per `/land`, a reviewer subagent with no access to the author's rationale reviewed `origin/feature/scan-logger...theme/shared-web-theme`. Its report follows verbatim. **I independently re-verified findings 1–6 before posting; all six held** — see the disposition that follows.

---

### 1. HIGH — `--surface-2` is referenced 35 times and defined nowhere the theme is loaded

`theme.css` defines the token as `--surface2` (10 definitions). Every consumer — and the picker's own rules — reference `--surface-2`: `run.html` 17, `scanlog.css` 9, `editor.css` 6, `theme.css` picker 2, `base.html` and `editor.html` 1 each. `background: var(--surface-2)` with an undefined token → `initial` (transparent). Every button and input in the editor, the run page's modal/plot bar, code boxes, chips, table-row hover, logbook `.day:hover`/`.scanrow:hover`, and the picker's *own* active-theme highlight lose their ground. The only place `--surface-2` is defined is the editor's **standalone** fallback — so the standalone editor looks right and the hosted one doesn't, the opposite of the PR's purpose.

`test_every_theme_defines_every_token` cannot see this: it compares theme blocks against each other, never against what surfaces *reference*.

### 2. HIGH — the run page's Plotly figure still carries the old dark-only palette in Python, and the guard cannot see it

`figures.py:39-66` hard-codes `paper_bgcolor "#1a2026"`, `plot_bgcolor "#12161a"`, font `#dde4ea`, grid colours, and `TRACE_COLORS` — all for the retired dark palette. Under a light system the run page is a pale page with a navy-black plot island. The PR's claim "a colour literal added anywhere fails here" is false for the biggest coloured surface in the portal, because the test only walks CSS. Needs: layout colours from live tokens, and `theme.js` emitting an event on change — today `apply()` has no hook.

### 3. MEDIUM-HIGH — hardcoded `/theme` URLs break the documented reverse-proxy deployment

`day.html` hardcodes `/theme/theme.css` and `/theme/theme.js`; `app.py` passes the editor `theme_url="/theme"` as a literal. The portal ships `_ForwardedPrefixMiddleware`, documents `--root-path /portal`, and every one of its own asset URLs goes through `{{ root }}`. Behind a proxy at `/portal/`, `/theme/theme.css` 404s → **every** token undefined including font stacks → the logbook renders as an unstyled document. `base.html` uses `url_for('theme', …)` unwrapped — an absolute scheme+host URL, the only one in the portal templates, exactly the mixed-content case the branch's own comment describes. Also no cache-busting `?v=`.

### 4. MEDIUM — the literal-colour test's allowlist lets a literal through on any line that also uses a token

`_ALLOWED["--"]` passes any line containing `--`, i.e. every line with `var(--…)`. Empirically: adding `background:#ff00ff` to a line using `var(--accent)` still passes 9/9. Minified one-rule-per-line CSS is therefore almost entirely exempt. Also: `pytest.skip` on a missing surface silently drops the guard on rename; the completeness test's regex checks only 6 of the 10 blocks; no coverage of named colours or `hsl()`.

### 5. MEDIUM — contrast: `--muted` fails AA on every palette; the default `laser` light accent fails too

| palette | muted/paper | accent/paper | on-accent/accent |
|---|---|---|---|
| **laser-light (default)** | **3.22** | **4.15** | **4.44** |
| plasma-light | 3.72 | 4.30 | 4.64 |
| bella-light | 3.68 | 6.32 | 6.91 |
| bella-dark | 3.97 | 5.56 | 5.37 |
| plasma-dark | 4.38 | ok | ok |
| laser-dark | 4.87 | ok | ok |

`--muted` is used for 11–12 px text everywhere — under 4.5 on all six. In the default palette, links, the brand, primary-button text and `--ok` badges all miss AA.

### 6. MEDIUM — `base.html` adds Google Fonts CDN links against the portal's own doctrine

`GEECS-DataPortal/CLAUDE.md:64`: "No npm, no CDN assets (control-room machines may lack internet)". `base.html` now adds a render-blocking stylesheet link to `fonts.googleapis.com`. Behind a black-holing firewall that delays first paint until the connect timeout. The token font stacks already carry full fallbacks.

### 7. MEDIUM — `color-scheme` is static; the bare `:root` has no dark variant; bootstrap and `theme.js` can disagree

`color-scheme: light dark` is set once, never per mode: a viewer on a dark OS who picks **Light** gets light tokens with dark native widgets (selects, scrollbars). The inline bootstrap stamps an unvalidated `localStorage` value; `theme.js` validates and falls back — so a stale/foreign value produces exactly the flash the bootstrap exists to prevent.

### 8. MEDIUM (redundancy/placement) — the PR reintroduces the duplication it exists to remove

The editor's standalone fallback is a fourth palette copy. The stated reason (keeping ScanAnalysis from importing the theme) does not hold: GeecsWebTheme has **zero** dependencies, so `ScanAnalysis[editor] → GeecsWebTheme` is one-way and cycle-free; `create_editor_app` can mount `static_dir()` itself. The inline bootstrap is copy-pasted into three templates, each with its own `"laser"` literal — `DEFAULT_THEME` lives in five places. The dark palette is hand-duplicated in the `[data-mode="dark"]` block and the `@media` copy (~90 lines); stamping the *effective* mode from the bootstrap removes the copies.

### 9. LOW

`theme.js` persists `"laser"` on every load for viewers who never chose, so a future default change never reaches them; `check.sh` changed-mode doesn't run the theme guard when only a surface file changes; stale "dark palette" text in the portal CLAUDE.md and `editor.css` header; GeecsScanLog changelog silent on its tokens moving out.

### Checked and not flagged

Lock file clean (no home/worktree paths); no lab literals; scan-folder invariant untouched; `_REPO = parents[2]` resolves from CI and `check.sh`; `url_for('theme', …)` does resolve since the mount precedes the routers.

skbarber:
## Disposition

Three review rounds by the same fresh-context reviewer; every finding independently re-verified before acting. Final verdict from the reviewer on `dc75659d`: **"I have no further findings on this branch."**

| # | Finding | Disposition |
|---|---|---|
| 1 | `--surface2` defined, `--surface-2` referenced 34× — every hover fill and button ground transparent | **Fixed** `a6047979`. New test: every `var(--x)` any surface references must be defined — the check that would have caught it |
| 2 | Plotly figure kept the retired dark palette; chips ≠ traces after round 1 | **Fixed** `101b9e51` — one sentinel palette (`page_figure()` swaps placeholder hex → `$tok:` in the served JSON; plotly's validators reject sentinels in a `go.Figure`) resolved by one client walk; doctrine restated as "identical up to palette". Reviewer verified every colour property is a sentinel and parity with `a6047979` on 12 figure pairs |
| 2b | *(introduced r1)* `base_layout()` dropped legend placement, `showlegend`, `xaxis.automargin`; pin read a dead constant | **Fixed** `63bb1a7d` — restored; `BASE_LAYOUT` deleted; pin reads the *served* layout |
| 3 | Hardcoded `/theme` URLs broke the documented reverse-proxy deployment; absolute `url_for` in the shell | **Fixed** `a6047979` — `{{ root }}` / `request.scope.root_path`, `?v=` cache-bust on all three assets (`dc75659d` lineage). Reviewer verified hosted, standalone and `/portal`-prefixed |
| 4 | Guard excused any literal beside a token; 14/17 probes passed | **Fixed** across `a6047979`→`dc75659d` — definitions exempt only inside `:root` blocks; `var()` fallbacks judged; case-insensitive extended named colours; `%23hex`; `style=`/`.style.x`/`setAttribute("style")`/`setProperty`/SVG `fill`/`stroke` in either quote; `:root .foo` no longer exempt. 32-probe battery pinned |
| 5 | Contrast failed AA on all six palettes | **Fixed** `a6047979` — regenerated, every text pair ≥ 4.5:1; `test_contrast.py` widened `101b9e51` to the pairs surfaces compose (semantic chips on their washes, accent on raised ground) |
| 6 | Google Fonts CDN links against the portal's own doctrine | **Fixed** `a6047979` — removed; token stacks carry local fallbacks |
| 7 | Static `color-scheme`; unvalidated bootstrap could flash | **Fixed** `a6047979` — `theme-boot.js` validates, resolves system, stamps effective mode; `color-scheme` per block |
| 8 | Editor's fourth palette copy; bootstrap copy-pasted ×3; dark palette duplicated | **Fixed** `a6047979` — editor depends on the theme (one-way; it has no deps) and mounts it standalone; one boot script; `@media` copies gone |
| 9 | Persist-on-load; `check.sh` gap; stale text | **Fixed** `a6047979`; ScanAnalysis bumped to a minor `101b9e51` |

### Waived, with reasons (reviewer agreed)
- **No-JS viewers get the bare light `:root` regardless of OS mode** — the boot script owns mode resolution; a `@media` block for the bare root only would be cheap but is not required.
- **`LABELS` in `theme.js` is a third copy of the theme descriptions** — ids fall back; the theme *list* and default are pinned Python↔JS by test.
- **`rgba(0,0,0,1)`, `color-mix()`, allow-marker substrings** — contrived bypasses, not worth the false-positive risk.

### Process note
`63bb1a7d` was pushed with one red probe while its message claimed green; `dc75659d` is the honest follow-up and says so. The commit chain now gates on the check.

### Tests
`GeecsWebTheme` 104 (from 9 at open) · `GEECS-DataPortal` 252 · `GeecsScanLog` 41 · `GEECS-Data-Utils` 334 · `ScanAnalysis` 192 — green via `scripts/check.sh`.

No finding is left undispositioned.

PR #832, opened 2026-09-12
Phase 02: the commentary store — GeecsLogbook 0.2.0 (renamed from GeecsScanLog), GEECS-Schemas 0.21.0, GEECS-DataPortal 0.24.0

skbarber:
Second phase PR of the scan-logger arc, into `feature/scan-logger` (not master). Phase 01 (#825) was the read-only day view; this adds what people write.

## What

**The commentary store.** `geecs_logbook.store.NotesStore` — SQLite (WAL), optimistic locking (`version`; a stale edit is a 409 carrying the current entry). `geecs_logbook.mirror` writes each entry as a front-matter markdown file into the day's `logbook/` folder on the share — a sibling of `scans/`, never inside a scan folder, never creating the day. Store first, mirror second: a save never fails because the share is slow or unmounted; `mirrored_at` stays null and `mirror.sync` (throttled to once a minute, on day views) pays the debt. `render.render_markdown`: markdown-it (commonmark + tables + strikethrough + task lists) → nh3, with `> [!NOTE]`-style callouts and attachment links rewritten at render time only.

**Entry routes** (only when `notes_db` is given): create (201), edit (409 on conflict), keep/un-keep, delete, attachment upload (20 MiB; png/jpeg/gif/webp/pdf; 413/415/503), `GET /api/day/{day}/entries`. The day page grows a composer per scan, per gap, and for the day; entries render with author/chips/stamp/edit/delete; agent drafts carry a Keep button.

**The schema** lives in `geecs_schemas.log_entry` (`LogEntry`, `Attachment`, `AnalysisPayload` / `ProblemPayload` discriminated on `kind`), registered as `log_entry` with a reference example — so GEECS-MCP and a mirror reader can agree on it without importing the logbook. `body_md` is opaque; templates seed it and stop.

**Rules the store enforces** (owner rulings this session, each pinned in `tests/test_store.py`):
- Three anchors: `scan=`, `after=`, or neither — a **day-level** entry (the general logbook's staple). Both refused. No `scan=0`.
- **An agent's entry is born a draft**: `kind != note` with `status=kept` is refused at creation (422); a person keeps it via the status route. A promoted agent entry is a valid stored state, which is why the schema does not carry the rule.
- **`updated_at`** moves on every change; `edited_at` only when text does — so "everything since" sees promotions and uploads.
- **Delete is a tombstone**: `deleted_at` set, listings hide it, writes fail as missing, the sync removes the file. Accidental deletes are a field to clear.
- Additive column migration (`_ADDED_COLUMNS`) for existing database files — nothing more.

**The portal**: `--notes-db PATH` / `create_app(notes_db=)` makes `/log` writable. Default: `logbook.db` under systemd's `$STATE_DIRECTORY`; the unit template now declares `StateDirectory=geecs-data-portal` (same path on every host, so not a site value). Charter amendment in the portal's CLAUDE.md: a write verb for **commentary only**, never the scans tree.

**The rename**: `GeecsScanLog` → `GeecsLogbook` / `geecs_logbook` / `geecs-logbook` (first commit, no behaviour change). The scan logger is the archetype for the month-level general logbook that follows, so the package is named for what it is becoming.

## Per-concern LOC (excluding locks and the regenerated schema reference)

| Concern | Files | ~LOC |
|---|---|---|
| Store + mirror + render (new) | `store.py`, `mirror.py`, `render.py` | +820 |
| Router write path + day page composer/entries | `router.py`, `day.html`, `scanlog.css` | +610 / −140 |
| Schema | `log_entry.py`, `__init__.py`, `docgen.py` | +300 |
| Tests (new: store, mirror; extended: router, portal mount) | 5 files | +740 |
| Portal flag + unit + charter + deployment docs | 8 files | +180 |
| Rename (mechanical) | 12 files | small |
| CHANGELOGs, versions, CLAUDE.md | 8 files | +150 |

- **`updated_at` / tombstones / born-draft**: ruled in this session ("yes"). The two ARIEL-facing endpoints (`?since=` cursors on entries and scans) are **not** in this PR, per the same ruling.
- **`StateDirectory` as the default notes-db home** rather than a `site.env` key: the path is host-invariant, so I read it as not a facility value. Say so if you'd rather it were explicit in `site.env`.
- **`mirror.logbook_root` now maps "share cannot be resolved" to `MirrorUnavailable`** — found writing the portal test on a host with no share config: without it, a save wrote the row and then 500'd. Behaviour change, small, pinned.
- **`GeecsWebTheme` 0.1.1**: only its guard test's surface list moved with the rename; patch-bumped per the docs-only precedent.

## Facility literals

None added. `Undulator` appears only in tests. `/var/lib/geecs-data-portal` is systemd-derived, in a unit comment and docs.

## Tests

- GeecsLogbook 84 (was 41 at phase 01; +43 store/mirror/router write path)
- GEECS-DataPortal 254 (+2 notes-db mount)
- GEECS-Schemas 460 (+7 for `log_entry`: descriptions, registry, example, reference)
- GeecsWebTheme 104 (unchanged)

CI: the `unit-tests` workflow steps follow the rename.

## Not in this PR (next chunks of phase 02)

The editor toolbar, paste-image via the upload endpoint, paste-spreadsheet→table, local-time display; then the month view and typed templates.

## Deployment

Not deployed. When `feature/scan-logger` reaches the host: `poetry install --extras "analysis log"`, re-render the unit (`StateDirectory` is new), restart. `GEECS_PORTAL_EXTRA_ARGS=--config-editor --scan-log` already in `site.env.example`.

## Review


skbarber:

Reviewer's report, verbatim, followed by the disposition. Fix commit: 61fc8b4e. Verification by the same reviewer is in flight; a second comment will carry its verdict.

---

# Review of PR #832 — Phase 02 commentary store

Tests as shipped pass (GeecsLogbook 84, portal mount 5, schema `log_entry` 14). Findings below are ranked; each was probed or refuted before inclusion.

## Findings

### 1. The portal test suite writes into the production share — HIGH
`GEECS-DataPortal/tests/test_scan_log_mount.py:59-88`

`test_writable_with_a_notes_db` calls `create_app(..., default_experiment="Undulator", scan_log=True, notes_db=db)` with no `base_directory`, so `mirror.logbook_root` resolves the share from `~/.config/geecs_python_api/config.ini`. The docstring's premise ("No share is configured in the test environment") holds in CI only. On any dev box with the share mounted, the POST for `2026-09-11` lands a markdown file in the real day folder; the test's own assertion inspects `tmp_path` and passes.

Evidence: the real `26_0911/logbook/` now holds `0308-sbarber-485199.md` (the author's run) and `0313-sbarber-542403.md` (my run of the suite), both body `hello`, author `S. Barber`. **I did not delete them; they need removing by hand.**

Fix: thread `base_directory` from `create_app` into `create_log_router` for tests (or monkeypatch `geecs_logbook.mirror.logbook_root` to raise `MirrorUnavailable`), and assert the entry's `mirrored_at` is null.

### 2. Editing corrupts any body containing `' " < > &` — HIGH
`GeecsLogbook/geecs_logbook/templates/day.html:63` and `:436`

Starlette's `select_autoescape()` HTML-escapes `{{ r.entry.body_md }}`, but a `<script>` element is raw text — the browser does not decode entities there, so `.textContent` returns them literally. Probe: body `don't say "charge < 10 pC & rising"` → the edit textarea receives `don&#39;t say &#34;charge &lt; 10 pC &amp; rising&#34;`; Save persists the entities into the store and the mirror. Apostrophes make this hit most real notes on first edit.

Fix: `<script type="application/json" class="raw">{{ r.entry.body_md | tojson }}</script>` + `JSON.parse` (`tojson` escapes `<` as `<`, so `</script>` cannot break out), or a `data-raw` attribute.

### 3. Mirror sync starves behind permanently-deferred entries — MEDIUM-HIGH
`store.py:196-210`, `mirror.py:294`

`unmirrored(limit=100)` orders by `created_at`; an entry whose day folder never appears (a day-level note on a day with no scans — the PR calls this "the general logbook's staple" — or a mistyped date) is deferred forever and stays at the head. Probe: 101 such entries plus one mirrorable → `sync` returns `(0, 100)` and the mirrorable one is never written. Each page-view sync also re-stats 100 day folders on SMB.

Fix: a `last_attempt_at` column set on deferral, `ORDER BY last_attempt_at NULLS FIRST`; and decide whether `logbook/` on a no-scan day is ever mirrorable — the invariant forbids creating `scans/ScanNNN/`, not the day's `logbook/`, and today's rule strands the general-logbook case.

### 4. UTC stamps in stable filenames and on the page — MEDIUM
`mirror.py:134`, `day.html:49`

The two stray files above prove it: written 20:08/20:13 PDT on 09-11, named `0308-…`/`0313-…`, front matter `day: 2026-09-11` / `created_at: 2026-09-12T03:13`. A directory listing sorts evening entries before morning ones for any site west of UTC after ~17:00, and the page shows `03:13` beside scans stamped local from `scan.log`. Filenames are stable across edits by design, so this cannot be corrected later without renames.

Fix: `entry.created_at.astimezone()` (host TZ, which the unit sets from `site.env`) in `entry_path` and the template; keep ISO UTC in the front matter.

### 5. Same-stem uploads overwrite silently; manifest records both — MEDIUM
`router.py:439-447`, `mirror.py:249-254`

Probe: two `image.png` uploads → one file on disk, two manifest rows both `image.png`, both `id == entry_id`. Every clipboard paste is named `image.png`, so the announced next chunk (paste-image) makes this the common case. `Attachment.id` is documented as "the directory segment the file is stored under" but the router sets it to the entry id, making the field redundant.

Fix: per-attachment id as the directory segment, or dedupe the stem (`image-2.png`).

### 6. `add_attachment` is an unguarded read-modify-write — MEDIUM
`store.py:323-339`

Two concurrent uploads to one entry (two parallel pastes) both read the same manifest; the second UPDATE overwrites the first's → a file on disk with no manifest row, breaking the "orphans are computable" promise. Unlike `update`/`set_status` the WHERE has no `deleted_at IS NULL`.

Fix: one statement — `SET attachments = json_insert(attachments, '$[#]', json(?)) … WHERE entry_id = ? AND deleted_at IS NULL`.

### 7. `_upload` blocks the event loop — MEDIUM
`router.py:414`

It is `async def` but does SQLite and SMB writes inline. A 20 MiB PDF over a slow share stalls every other portal request for the write's duration. Fix: plain `def` with `file.file.read(...)` so FastAPI threadpools it like the other routes.

### 8. Renderer and attachment routes are untested — MEDIUM
No `tests/test_render.py` exists; `test_router.py` has zero references to `/attachments`. The sanitizer is the security boundary the PR itself cites ("an agent writes into the same field a human does"), and the advertised 413/415/503 ladder has no pin. Finding 9 is the kind of regression this would have caught.

### 9. Task lists are advertised but dead — LOW-MEDIUM
`render.py:43,77`

nh3's default allowlist drops `<input>`. Probe: `- [ ] open\n- [x] done` → `<li> open</li><li> done</li>` — checkbox and state vanish; `scanlog.css` styles `.entry-body input[type=checkbox]` for an element that never reaches the page. Fix: extend nh3's tags/attributes for `input` (`type`, `checked`, `disabled`), or drop the plugin and the claim.

### 10. Upload/serve 500 when the share is unresolvable — LOW-MEDIUM
`router.py:448` (`logbook_root` outside the try), `:307`

Probe: `MirrorUnavailable` from `logbook_root` → 500 on upload (docstring promises 503) and 500 on every `` in every entry on a host without share config. Fix: move the call inside the try; map to 503 in `_attachment`.

### 11. The write contract lives in the router, not the schema — LOW (placement/redundancy)
`router.py:80-95`

`EntryCreate` re-declares the `day` pattern, `scan ge=1`, `after ge=0`, and re-types `kind`/`status` as literal copies instead of importing `EntryKind`/`EntryStatus`. The stated reason for `LogEntry` in `geecs_schemas` is MCP validating "when an agent posts one" — but an agent posts an `EntryCreate`, and `LogEntry` requires server-assigned `entry_id/created_at/updated_at`. Fix: move the create shape into `geecs_schemas.log_entry`; the router imports it.

### 12. Third private atomic-write helper — LOW (redundancy/placement)
`mirror.py:198-206` vs `ScanAnalysis/scan_analysis/config_store.py:442` (mkstemp-based) and `ScanAnalysis/scan_analysis/task_queue.py:372-388`. The mirror's uses a fixed `.tmp` name, so the once-a-minute `sync` and a concurrent request `_mirror` on the same entry can clobber each other's temp file (config_store's cannot). Natural home: `geecs_data_utils`, which both depend on. Also `logbook_root` (`mirror.py:99-104`) re-derives tag→daily folder identically to `scan_reader.read_day`.

### 13. Sync can mark a stale mirror as current — LOW
Entry v1 owed; `sync` reads v1; a request edits → v2, writes file, marks mirrored; sync then writes v1 over it and marks mirrored. Millisecond window, self-heals on the next edit. Fix: `mark_mirrored(entry_id, version)` with `WHERE version = ?` plus a re-`get` before writing in `sync`.

### 14. Real colleague's name as an account-style slug — LOW (hygiene)
`agonsalves` / "A. Gonsalves" in `mirror.py:17` and `tests/test_mirror.py:49-54`. `sbarber` has pyproject precedent; the other has none in the repo.

### 15. Doc contradiction + a JS nit — LOW
`test_scan_log_mount.py:3` and `GeecsLogbook/CLAUDE.md` call writes "a second opt-in", but under the rendered unit every host already running `--scan-log` becomes writable at its next restart with no site change (`__main__.py:85-89`). State this in DEPLOYMENT.md's upgrade note. Separately, `day.html:385` renders a pydantic 422 (list `detail`, e.g. body over 200 000 chars) as "[object Object]".

## Checked and found clean
- Sanitizer on the probed cases: raw HTML escaped, `javascript:`/`data:` not linkified, `onerror` never emitted, `html.escape(base)` blocks quote injection into the attachment base, `link_rel` applied.
- Attachment serving traversal: `..`, `%2F`, sibling-prefix `attachmentsX` all 404 (`resolve()` + `is_file()`).
- Scan-folder invariant: write paths guard on the day folder and never traverse `scans/`; pinned by the mkdir-explodes and snapshot tests.
- Optimistic lock on `update`: WHERE repeats the version check; 409 carries the current entry.
- Tombstone semantics, never-mirrored delete, additive migration idempotence.
- Rename completeness: no `geecs_scan_log`/`GeecsScanLog` leftovers outside CHANGELOGs; CI workflow and `check.sh` follow.
- Facility literals: none added; `STATE_DIRECTORY` colon-list parse correct.
- No other markdown/sanitizer or SQLite implementation in the repo to duplicate.

---

## Disposition (61fc8b4e unless noted)

Both HIGH findings were verified independently before fixing: the two 186-byte `hello` files were on the real share and have been removed by hand (nothing else was in that `logbook/`); the entity round-trip reproduces exactly as described.

| # | Disposition |
|---|---|
| 1 | **Fixed.** The portal test monkeypatches `mirror.logbook_root` to raise `MirrorUnavailable` and asserts the entry is left owed in the store. The two stray files are gone. |
| 2 | **Fixed.** Raw body ships as `application/json` via `tojson`; the editor `JSON.parse`s it. |
| 3 | **Fixed** (rotation): `mirror_attempted_at` column (additive migration), set on every deferral by `sync` and by the request-path mirror; `unmirrored` orders never-tried first, then least recently tried. Pinned. The policy half — whether a dayless day-level entry may ever mirror — is **recorded as a deferred decision** in `GeecsLogbook/CLAUDE.md`, to be ruled with the month view; not silently chosen here. |
| 4 | **Fixed.** `entry_path` and the page stamp use `astimezone()` (host local, the clock the day and scans are named by); front matter stays ISO UTC; the stamp carries the ISO instant as its `title`. |
| 5 | **Fixed.** Same-stem uploads are numbered (`image-2.png`); each attachment gets its own id; `Attachment.id`'s doc no longer claims to be the directory segment. Pinned. |
| 6 | **Fixed.** Single-statement `json_insert` append with the `deleted_at IS NULL` guard; `KeyError` on zero rows. Pinned. |
| 7 | **Fixed.** Plain `def`, `file.file.read`. |
| 8 | **Fixed.** `tests/test_render.py` (9: escaping, refused schemes, base escaping, tables/strike/task lists, callouts, plain quotes, link rewrite with and without base, empty) and `TestAttachments` in `test_router.py` (5: upload→serve, dedupe, 415/413/422/404, containment, 503). |
| 9 | **Fixed.** nh3 allowlist extended for `input` with `type/checked/disabled` only. Pinned. |
| 10 | **Fixed.** `logbook_root` inside the try on upload; `_attachment` maps `MirrorUnavailable` → 503. Pinned. |
| 11 | **Partly fixed, remainder waived.** The router now imports `EntryKind`/`EntryStatus` from the schema. Moving the create shape into `geecs_schemas` is deferred to the agent-verbs phase (its first second consumer) and recorded in the deferred-decisions table — not worth a schema-package churn with one consumer. |
| 12 | **Clobber fixed, relocation waived.** Temp files are `mkstemp`-unique. Folding the helper (and `logbook_root`'s tag→folder derivation) into `geecs_data_utils` joins the `ScanPaths`/`ScanData` review already owed at master-merge; recorded in the deferred-decisions table. |
| 13 | **Fixed.** `mark_mirrored(..., version=)` with `WHERE version = ?`; `sync` and the request path pin to the version they wrote. Pinned. |
| 14 | **Fixed.** Fictional `T. Operator`. |
| 15 | **Fixed.** DEPLOYMENT.md upgrade note states that a host on `--scan-log` becomes writable at its next restart; the page joins pydantic's list-shaped 422 detail into a message. |

Counts after the fixes: GeecsLogbook 101, GEECS-DataPortal 254, GEECS-Schemas 460, GeecsWebTheme 104.

skbarber:
## Verification round 1 (same reviewer, on 61fc8b4e) — and the round-2 disposition

Reviewer's verdict, verbatim:

---

# Verification of 61fc8b4e against the #832 review findings

All four suites pass at the claimed counts (GeecsLogbook 101, portal 254, schemas 460, theme 104). Each finding re-probed; verdicts below, then regressions and one new finding surfaced by the probes.

## Per-finding verdict

| # | Claimed | Verdict | Evidence |
|---|---|---|---|
| 1 | fixed | **Confirm** | `test_scan_log_mount.py` monkeypatches `mirror.logbook_root` and asserts `unmirrored() == ["hello"]`. I re-ran the full portal suite with the real share mounted: `26_0911/logbook/` gained nothing. The two stray `.md` files are gone. **Residue:** the empty `26_0911/logbook/` directory (created by the author's 20:08 run) is still on the share — harmless, but it was not made by the scanner and should be removed by hand too. |
| 2 | fixed | **Confirm** | Body `don't say "charge < 10 pC & rising" </script> — ünïcode` → script content is `"don't … </script> …"`; no literal `</script>` in the page; `json.loads` of the block equals the original exactly. |
| 3 | fixed (rotation) | **Confirm** | 101 dayless entries + 1 mirrorable: passes return `(0,100)` then `(1,99)`; the late entry is written. Worst case for an entry deferred at create time behind N older deferred: ~⌈N/100⌉ page-view minutes (250 older → 3 passes). Dayless-day policy is recorded as deferred in `GeecsLogbook/CLAUDE.md`, as stated. |
| 4 | fixed | **Confirm** | File stamp `2026` = local `20:26` (UTC `03:26`); page stamp `20:26` with the ISO instant in `title`. `astimezone()` resolves the offset per instant, so a DST change does not rename files. |
| 5 | fixed | **Partial** | Sequential: `image.png`, `image-2.png`, `image-3.png`, three distinct ids, three files — fixed. **Parallel** (4 same-name uploads in flight): all four return `attachments/…/image.png`, one file on disk, four manifest rows all named `image.png`. The dedupe reads the manifest at request start, so concurrent requests pick the same name. Fix: dedupe on disk with `os.open(O_CREAT\|O_EXCL)` inside `write_attachment` (loop the suffix), or name the file by the attachment id. |
| 6 | fixed | **Confirm** | 8 threads appending → manifest count 8, version 9. Note the floor: `json_insert(…, '$[#]', …)` needs SQLite ≥ 3.31.0 (local 3.45.3; Ubuntu 22.04 ships 3.37). Worth one line in DEPLOYMENT.md. |
| 7 | fixed | **Confirm** | `_upload` is `def`, reads `file.file.read(n)`; 201/413/415/422/503 all behave under the threadpool. |
| 8 | fixed | **Confirm** | `test_render.py` (9) pins raw-HTML escape, `javascript:`/`data:` refusal, `onerror`, base-quote escaping, features, callouts, rewrite-only-with-base. `TestAttachments` (5) pins upload/serve, numbering, limits, containment, 503. These would fail against the pre-fix code for 9 and 10. |
| 9 | fixed | **Confirm** | `<input disabled="disabled" type="checkbox">` / `checked="checked"` survive; raw `<input …>` in source is still escaped by markdown-it before nh3 sees it. |
| 10 | fixed | **Confirm** | Unresolvable share → upload 503, serve 503 (was 500/500). |
| 11 | partial / waived | **Confirm as stated** | `EntryKind`/`EntryStatus` imported; the create-shape move is recorded in CLAUDE.md's deferred table. |
| 12 | clobber fixed / relocation waived | **Confirm, with a regression** (below) | `_replace_with` uses `mkstemp` → unique temp names. Relocation recorded. |
| 13 | fixed | **Deny — narrowed, not fixed** | Same-author interleaving (sync holds v1; a request edits → v2, writes v2, marks v2; sync then writes v1): result **DB version 2, `mirrored_at` set, file body `v1`**. The pinned mark correctly no-ops for sync, but the request's mark already covered the stale file. The version pin only helps when sync's stale write lands *before* the request's write. A real fix needs the write+mark serialised (a process-wide `threading.Lock` around `_mirror` and `sync`'s loop body, with a fresh `store.get` inside it), which the CLAUDE.md/CHANGELOG do not claim. Severity stays Low (ms window, self-heals on next edit) — but the changelog line should not say "fixed". |
| 14 | fixed | **Confirm** | No `agonsalves`/`Gonsalves` left in `GeecsLogbook/`. |
| 15 | fixed | **Confirm** | DEPLOYMENT.md upgrade note present; `api()` joins list-shaped `detail` via `x.msg`. |

## Regressions introduced by the fix

**R1. Mirror files are now mode 0600** — `mirror.py:_replace_with`. `mkstemp` creates the temp at 0600 and `os.replace` carries that mode to the final file; before the fix `write_text`/`write_bytes` honoured the umask (0644). Probe: both the `.md` and the `.png` are `0o600`. On the HTU CIFS mount the mode comes from mount options, so it is moot there; on any local or NFS share the "legible when none of this code runs" copy is readable only by the service user. Fix: `os.fchmod(fd, 0o666 & ~umask)` before the rename (read the umask once at import via `os.umask(0); os.umask(m)`), or `Path.chmod` after. Low.

## New finding surfaced by the probes (pre-existing, not from the fix)

**N1. An edit replaces the entry's author, and that changes the mirror filename — MEDIUM-HIGH.** `store.py:update` sets `author = ?` from the editor's name; `entry_path` uses the author slug. Probe: S. Barber writes "original"; T. Operator fixes a typo via PATCH → the entry's `author` is now `T. Operator` (the page shows T. O.'s initials and name — silent misattribution), and the share holds **both** `2027-sbarber-ecd3ac.md` (stale v1, never removed; nothing is owed) and `2027-toperator-ecd3ac.md`. The docstring's claim "carries the creation time, the author and the id, none of which change" is false. `test_name_is_stable_across_edits` uses the same author for both saves, so it cannot catch it. Fix: keep `author` immutable (drop it from `update`'s SET) and record the editor separately (`edited_by`, or an edit-log later); until then, at minimum stop `update` from touching `author` so both the attribution and the filename are stable.

## Checked and found clean
- Order of `unmirrored()` is never-tried first, then LRU — no head-of-line blocking; `_ADDED_COLUMNS` migration for the new column is idempotent.
- `_delete` still marks unconditionally, which is correct (no writes follow a tombstone).
- `add_attachment` now refuses tombstoned rows (`deleted_at IS NULL`), and the router maps its `KeyError` to 404.
- Traversal containment on the attachment route still 404s for `..`, `%2F`, and sibling-prefix segments.
- No tracked files were modified by me; the share was only listed and read.

---

## Round-2 disposition (2469eb9b)

| Item | Disposition |
|---|---|
| #5 parallel uploads | **Fixed.** `write_attachment` claims the name on disk (`O_CREAT\|O_EXCL`, numbered) and returns the name it got; the router records that. Manifest-based dedupe removed. Pinned with a 4-thread test. |
| #6 SQLite floor | **Fixed.** DEPLOYMENT.md states the ≥ 3.31 floor for `json_insert`. |
| #13 stale sync | **Fixed for real this time.** `mirror.WRITE_LOCK` + `mirror_one()`: re-read under the lock, write, mark that version; `sync`, the request path and delete all go through it. Pinned by `test_sync_never_covers_a_newer_file_with_an_older_one`. CHANGELOG corrected to describe what is actually done. |
| R1 0600 files | **Fixed.** `fchmod(fd, 0o666 & ~umask)`; pinned for both the markdown and an attachment. |
| N1 author overwritten | **Fixed.** `author` is immutable; `edited_by` is a new `LogEntry` field/column (additive migration); `PATCH` takes `editor`; the page shows "edited by X" when it differs; the mirror front matter carries it. Pinned with a different-editor edit: same file, same author. |
| residue | The empty `26_0911/logbook/` directory has been removed from the share by hand. |

Counts: GeecsLogbook 104, GEECS-DataPortal 254, GEECS-Schemas 460, GeecsWebTheme 104. Verification of round 2 by the same reviewer is in flight.

skbarber:
## Verification round 2 (same reviewer, on 2469eb9b) — and the round-3 disposition

Reviewer's verdict, verbatim:

---

# Round-two verification (2469eb9b) — PR #832

Suites: GeecsLogbook 104, portal 254, schemas 460 — all pass. The share's `26_0911/` now lists only `analysis` and `scans` (stray `logbook/` gone). DEPLOYMENT.md carries the SQLite ≥ 3.31 floor; the CHANGELOG now describes #13 as lock + fresh read + pinned mark, not the pin alone.

## Per-item verdict

**#5 parallel same-name uploads — Confirm.** Four threads posting `image.png` with distinct payloads through the router: links `image.png`, `image-2/3/4.png`; four files on disk; each file holds exactly its own thread's bytes; the manifest carries the four claimed names (the router records `link.rsplit("/")[-1]`, not the requested name). `test_parallel_same_name_uploads_all_get_their_own_file` exercises `write_attachment` directly with a 4-thread pool and asserts names on disk, so it would fail against the round-one manifest dedupe.

**#13 stale sync write — Confirm.** Realistic interleaving (sync under `WRITE_LOCK` writing v1; a request thread edits → v2 and calls `mirror_one`, which blocks on the lock; sync finishes and marks v1; the request then re-reads v2, writes, marks): result DB version 2, not owed, **file body v2**. Reverse order (request mirrors v2 first, then a sync pass whose `unmirrored()` snapshot predates the edit): `mirror_one` re-reads and rewrites v2, nothing owed. The delete route also goes through `mirror_one` (file removed, marked). `test_sync_never_covers_a_newer_file_with_an_older_one` pins the snapshot-then-edit half; it does not pin the lock itself, but the fresh read inside `mirror_one` is what closes the hole, and that is what the test exercises.

**R1 0600 mirror files — Confirm.** `.md` and attachment both `0o644` (`0o666 & ~umask`); the `_claim_name` placeholder is also opened with `0o666`, so the mode does not regress through that path either. `test_mirror_files_are_readable_by_others` computes the expectation from `mirror._UMASK`, so it holds on any developer umask.

**N1 edit re-attributes and renames — Confirm.** PATCH by `T. Operator` on S. Barber's entry: `author` stays `S. Barber`, `edited_by = T. Operator`, exactly one file for the entry (`…-sbarber-…md`), front matter carries `edited_by: T. Operator`, the page shows "edited by T. Operator", and the old request shape (`author=`) is a 422 rather than silently accepted. `test_name_is_stable_across_edits` now uses a different editor and asserts author unchanged + same path, so it would fail against the old `update`.

## Regression introduced by round two

**`_claim_name` leaves a served 0-byte placeholder when the real write fails — LOW.** `mirror.py:_claim_name` creates the name with `O_EXCL` and closes it; if `_replace_with` then raises (share hiccup), `write_attachment` maps it to `MirrorUnavailable` → the router returns 503 correctly, but the empty `hic.png` stays on disk. Probe: after a forced `OSError` in `_replace_with`, `hic.png` exists at 0 bytes, `GET /log/attachments/…/hic.png` returns **200 with 0 bytes**, and the user's retry lands as `hic-2.png`. The placeholder is in no manifest, so the "orphans are computable" promise misses it too. Fix: in `write_attachment`, on any exception after `_claim_name` returns, `with contextlib.suppress(OSError): target.unlink()` before re-raising.

## Notes, not findings
- A request's response now waits on `WRITE_LOCK` if the once-a-minute sync is mid-write of another entry on a slow share; the lock is per entry, not per pass, so the wait is one SMB write, not up to 100. Pre-existing design already put the request's own mirror write on the response path.
- `_UMASK = os.umask(0); os.umask(_UMASK)` at import is the standard idiom; the umask-0 window is at module import, before any request thread exists.

No tracked files were modified; the share was only listed.

---

## Round-3 disposition (c7fd9d9e)

| Item | Disposition |
|---|---|
| 0-byte placeholder on a failed write | **Fixed.** The claim and the write are separate steps; a failed write unlinks the claimed name before raising `MirrorUnavailable`. Pinned by `test_failed_attachment_write_leaves_no_placeholder` (forced `OSError` → no file → the retry gets the original name). |
| Lock-wait note | Acknowledged; accepted as-is — one SMB write per entry on the response path is the pre-existing design. |

Counts: GeecsLogbook 105, GEECS-DataPortal 254, GEECS-Schemas 460, GeecsWebTheme 104. Verification of this one fix by the same reviewer is in flight; merge follows its confirm and green CI.

skbarber:
## Verification round 3 (same reviewer, on c7fd9d9e) — final

Reviewer's verdict, verbatim:

---

# Round-three verification (c7fd9d9e) — PR #832

**Regression from round two (0-byte placeholder) — Confirm fixed.**

End-to-end through the router, `_replace_with` forced to raise after `_claim_name` succeeded:

- Failed upload → **503**, `hic.png` **absent** on disk, `GET /log/attachments/…/hic.png` → **404** (was 200 with 0 bytes), manifest unchanged (`[]`).
- Retry → **201** with link `…/hic.png` (the original name, not `hic-2.png`); GET → 200, 7 bytes; disk holds only `hic.png`.

Side paths checked:

- Claim failure (attachment folder read-only) → 503, not 500: the `mkdir`/`_claim_name` block is still inside its own `except OSError → MirrorUnavailable`.
- Write **and** unlink both failing → still 503 (the `suppress(OSError)` does not mask the original error). The placeholder is left in that case, which is the only remaining way one can survive — an unwritable-and-undeletable folder, which is not a realistic share state and not worth more code.

`test_failed_attachment_write_leaves_no_placeholder` exercises the same path directly (forced `OSError`, asserts absence, then asserts the retry gets `hic.png`), so it would fail against the round-two code.

**Regressions introduced: none found.** GeecsLogbook 105 pass. No tracked files were modified; the share was not touched this round.

## Standing state of the review, all rounds

Every finding from the original review is now confirmed fixed or explicitly waived-and-recorded:

| Finding | State |
|---|---|
| 1 test writes to the real share | fixed (monkeypatched `logbook_root`); stray files and the empty `logbook/` directory removed from the share by hand |
| 2 entity-corrupting edit round-trip | fixed (`tojson` + `JSON.parse`) |
| 3 sync starvation | fixed (LRU rotation via `mirror_attempted_at`); dayless-day policy recorded as a deferred decision |
| 4 UTC stamps | fixed (host-local `astimezone()`) |
| 5 same-name uploads | fixed (on-disk `O_EXCL` claim), parallel case included |
| 6 `add_attachment` lost update | fixed (`json_insert`); SQLite ≥ 3.31 floor documented |
| 7 upload blocking the event loop | fixed (plain `def`) |
| 8 untested renderer/attachment routes | fixed (`test_render.py`, `TestAttachments`) |
| 9 dead task lists | fixed (`input` allowlisted) |
| 10 500 on unresolvable share | fixed (503) |
| 11 write shape outside the schema | partial — aliases reused; move recorded as deferred |
| 12 duplicate atomic-write helper | clobber fixed (`mkstemp`); relocation recorded as deferred |
| 13 stale sync write | fixed (`WRITE_LOCK` + fresh read in `mirror_one`), both interleavings probed |
| 14 real name in fixtures | fixed |
| 15 doc contradiction / JS 422 text | fixed |
| N1 edit re-attributes + renames file | fixed (`author` immutable, `edited_by`) |
| R1 0600 mirror files | fixed (`fchmod` to umask) |
| R2 served 0-byte placeholder | fixed (this round) |

---

Review resolved: no undispositioned finding. Merging into `feature/scan-logger` once CI on c7fd9d9e is green, per the owner's standing ruling for phase PRs on this branch.

PR #835, opened 2026-09-12
Phase 03: the foundation for the ops book — GeecsLogbook 0.3.0, GEECS-Schemas 0.22.0, GEECS-DataPortal 0.24.1

skbarber:
Third phase PR of the logbook arc, into `feature/scan-logger` (not master). PR 1 of the path to a functional scans + ops logbook; the editor (PR 2) and the month view + types (PR 3) land on top of this. Owner rulings 2026-09-11 for every item below.

## What

**Two books, one store.** `book: scans | ops` on every entry, chosen by the page the author writes from — never by timestamp (the owner's pushback on a single timeline: the scans book is a curated campaign document, ops is routine operations). An ops entry is day-level only (validated in the schema and the store). The day document renders the scans book; `GET /api/day/{day}/entries?book=` filters either way. The ops page itself is PR 3.

**Tags from the body.** `#laser` in the text *is* the tag; `tags.parse_tags` reads them at every save into an indexed column (narrowly: a letter first, ≤32 chars, not `#1`, not `# Heading`, not inside code spans or fences, not glued to a word). A type button will insert a template whose prefill carries its tag, so a button and a typed tag are the same thing; there is no category column to drift from the text. Rendered as chips.

**History.** `entry_history` snapshots the entry as it was before every edit, keep/un-keep, upload and delete — in the same transaction (`BEGIN IMMEDIATE`; a refused edit leaves no history). `GET /api/entries/{id}/history`, oldest first; works for a deleted entry, which is when it is wanted. Undo is a new edit with an old body.

**One query.** `NotesStore.query(day_from, day_to, book, tag, kind, status, author, include_scan_anchored, limit)` — the month page, its filter chips, a search and a synchroniser's "everything since" share it.

**Attachments are store-first.** `attachments.AttachmentStore` keeps uploads under `attachments/` beside the database (names claimed on disk with `O_EXCL`, contained serving, no placeholder on a failed write); the page serves `/log/attachments/{entry_id}/{filename}` from there; the mirror copies files beside the markdown (size-checked, so the sync costs a stat per file). A screenshot pasted with the share unmounted lands. Everything irreplaceable is now one directory: the state directory.

**The mirror owns its tree.** Entries mirror into `{experiment}/logbook/Y2026/09-Sep/26_0911/{ScanNNN|after-ScanNNN|.}/…` — the data tree's date shape, *outside* the data tree — instead of `logbook/` inside each day folder. Always writable, so a note on a day with no scans has a home and the "deferred forever" state (a recorded deferred decision) is closed; `mirror._assert_own_tree` refuses any path with `scans` in it, and a test monkeypatches `Path.mkdir` to assert every directory the mirror makes is under `logbook/`. The scan-folder invariant is now satisfied by construction *and* pinned in code.

**Router split.** `routes/day`, `routes/entries`, `routes/attachments`, sharing a `Context`; `router.create_log_router` only assembles them. `routes/month` slots in next.

## Per-concern LOC (excluding locks and the regenerated schema reference)

| Concern | ~LOC |
|---|---|
| Store: book/tags columns, history table + snapshots, `query`, migration | +230 |
| Mirror rewrite: own tree, invariant assert, attachment copy, shared `_fs` | +120 / −110 |
| Attachment store + `_fs` helpers (new) | +160 |
| Tag parser (new) | +35 |
| Router split into `routes/` (moved + `Context`) | +330 / −480 |
| Schema: `book`, `tags`, validator, docs example | +45 |
| Tests: tags (11), attachments (4), store (+10), mirror (rewritten), router (+7) | +420 / −140 |
| Docs: logbook CLAUDE.md, root CLAUDE.md, portal charter + DEPLOYMENT, CHANGELOGs | +150 |

- **The mirror tree reverses this morning's "sibling of `scans/`"** — the owner had no strong opinion and took the recommendation; the reasoning is in the mirror's module docstring and the logbook CLAUDE.md.
- **Attachments store-first** likewise reverses "attachments on the share"; the share still gets every file, just not in the save path.
- **`Attachment.id` is a uuid**, not the directory segment (the directory is the entry's id); the schema doc was corrected in #832.
- **The day document does not yet show the ops strip** ("3 ops notes today →") — it would link to a page that does not exist until PR 3. `ops_count` is already in the template context.
- **Git-as-truth was considered and rejected** (attachment load); recorded in CLAUDE.md so it is not re-proposed.

## Facility literals

None added. `Undulator` in tests only.

## Tests

GeecsLogbook 134 (was 105), GEECS-DataPortal 254, GEECS-Schemas 460 (+book/tags coverage via the reference example), GeecsWebTheme 104.

## Deployment

Not deployed; nothing on the host runs this branch. Note for the eventual deploy: the state directory (`/var/lib/geecs-data-portal`) now holds the database **and** `attachments/`; back it up as one unit. No existing entries exist anywhere, so the mirror-tree move migrates nothing.

## Review


skbarber:

Reviewer's report, verbatim, followed by the disposition. Fix commit: 44bff0b7. Verification by the same reviewer is in flight; a second comment will carry its verdict.

---

## Review of PR #835 (`phase/03-foundation` → `feature/scan-logger`)

All three suites pass as claimed (GeecsLogbook 134, GEECS-DataPortal 254, GEECS-Schemas 460). Findings below are ordered by severity; every runtime one was reproduced with a probe against the worktree (probes in the scratchpad; no tracked file touched, no share reached — every store/mirror call used an explicit `base_directory` under a temp dir).

### Findings

**1. Attachment serving escapes the attachment root — `GET /log/attachments/../logbook.db` returns the SQLite database.** — HIGH
`GeecsLogbook/geecs_logbook/attachments.py:56-63` (`AttachmentStore.path`), served by `routes/attachments.py:37-42`.
The containment check is `target.parent != folder`, where `folder = (self.root / entry_id).resolve()`. It proves the file sits directly inside `root/<entry_id>`, but never that `root/<entry_id>` sits inside `root`. With `entry_id=".."` and `filename="logbook.db"`, `folder` resolves to the state directory and the check passes. Reproduced: `blobs.path("..", "logbook.db")` returns the database path; a raw ASGI request for `/log/attachments/../logbook.db` returned **200 with `SQLite format 3`** (httpx's TestClient normalises dot-segments, which is why the route test cannot see it; uvicorn/Starlette do not, so `curl --path-as-is` against the deployed portal would dump the whole logbook — the one irreplaceable thing, per the store's own docstring). `entry_id="."` + any filename at the root of `attachments/` also serves.
The existing test `tests/test_attachments.py::test_serving_is_contained` asserts `blobs.path("..", "secret.txt") is None`, but it wrote `secret.txt` to `blobs.root`, not `blobs.root.parent`, so it passes against the wrong code — the "would the test fail" answer is no.
Fix: check `folder.parent == self.root.resolve()` (or `target.is_relative_to(root)` after resolving both), and move the test's decoy to `root.parent`. The `_static` route in `routes/day.py:47-49` is correct only because its base is a constant; a single `contained(root, *segments)` helper in `_fs.py` would serve both.

**2. `_assert_own_tree` raises `RuntimeError`, which `Context.mirror` does not catch — every save 500s after the row is written.** — MEDIUM
`mirror.py:113-117`, `mirror.py:129-132`, `routes/_common.py:114-127`.
`Context.mirror` catches only `MirrorUnavailable`. `logbook_root` calls `_assert_own_tree`, which raises `RuntimeError` for any path with `scans` among its parts — including a `base_directory`/config `base_path` that merely *contains* a `scans` component (e.g. a share mounted at `/mnt/scans/data`), or an experiment named `scans`. Reproduced: with `base_directory=<tmp>/scans/d`, `POST /log/api/entries` returns **500** although the row is committed — exactly the outcome the store docstring says must never happen ("surfacing it as a 500 after the row was written would tell the writer their entry failed when it did not"). Because the check fires on every request forever, the deployment is dead for writes with a misleading error, while `maybe_sync` (broad `except Exception`) silently swallows the same failure.
Fix: assert on the path *relative to the experiment directory* (`root.relative_to(year_dir.parent).parts`) so only segments this module appends are inspected, and/or have `Context.mirror` catch `Exception`, log, and `mark_deferred` like `maybe_sync` does.

**3. A share that drops after startup makes the mirror write to the local mount point and mark the entry mirrored.** — MEDIUM (host-permission dependent)
`mirror.py:96-118` (`logbook_root`), `mirror.py:243-249` (`write_entry` `mkdir(parents=True)`).
The previous mirror's "day folder must exist" guard was also the only "is the share actually mounted" check. It is gone, and nothing replaced it: `ScanPaths.get_daily_scan_folder` (`GEECS-Data-Utils/.../scan_paths.py:313-343`) builds the path without touching disk, and `GeecsPathsConfig` validates `base_path` only at construction (`geecs_paths_config.py:78-100`), i.e. once per process. Reproduced: `logbook_root(..., base_directory=<empty dir>)` + `write_entry` happily creates `<empty>/Undulator/logbook/Y2026/09-Sep/26_0911/…` and the entry is then `mark_mirrored`. On the portal host, if the CIFS mount drops while the service runs and the mount-point directory is writable by the service account (or autofs re-creates it), the markdown lands on the root filesystem, is marked done, and is hidden when the share remounts — the same "orphaned by a transient" failure class the scan-folder invariant exists for, just for the mirror's own files. `test_writes_a_day_the_scanner_never_made` pins the behaviour deliberately (empty `tmp_path`, nothing exists), so it will not catch this.
Fix: require the **experiment directory** (`{base}/{experiment}`, the year folders' parent, never a scan folder) to exist before any `mkdir`, raising `MirrorUnavailable` otherwise. That keeps the ops-book fix (a dayless day still lands) while restoring a mount sentinel; the test needs to create `tmp_path/Undulator`.

**4. `delete` on a tombstoned entry commits a spurious history snapshot.** — LOW-MEDIUM
`store.py:582-596`.
`_snapshot` runs before the UPDATE, and unlike `set_status`/`add_attachment`, `delete` does not raise on `rowcount == 0`, so the transaction commits. Reproduced: three `store.delete(id)` calls → history `[('delete',1,live), ('delete',2,tombstoned), ('delete',2,tombstoned)]` — two "delete" records that snapshot an already-deleted state at the same version, contradicting "nothing here rewrites history" and the CLAUDE.md rule that history records the state *replaced*. The route pre-checks `store.get`, so this needs two concurrent DELETEs or a direct store caller (the synchroniser the PR names as a future consumer). `test_delete_reports_whether_anything_went` deletes twice but never checks history.
Fix: on `rowcount == 0` raise inside the `_transaction` block (a private sentinel, translated to `False`), so the snapshot rolls back.

**5. `update` after a concurrent delete reports a 409 with a stale "current" instead of 404.** — LOW
`store.py:429-462`.
If the entry is tombstoned between the pre-read and `BEGIN IMMEDIATE`, the UPDATE matches nothing (`deleted_at IS NULL`), `self.get(entry_id)` returns `None`, and `ConflictError(current)` is raised with the *pre-delete* entry — the client is told "saved by someone else (now at version 1)" and offered a body that no longer exists. Reproduced (probe G). Fix: after `rowcount == 0`, re-read with `include_deleted=True`; deleted → `KeyError`.

**6. `query(limit=-1)` is unbounded; the cap does not guard it.** — LOW
`store.py:247-248`. `min(-1, _QUERY_CAP)` is `-1`, and SQLite's `LIMIT -1` means no limit — reproduced (returns all rows). The docstring calls the cap "a guard against an unbounded range". Once the month route or the MCP exposes `limit`, a caller can bypass it. Fix: `max(0, min(limit, _QUERY_CAP))` or validate `limit >= 1`.

**7. Tag parser: in-document anchor links become tags; trailing hyphen kept.** — LOW
`tags.py:22`. `[see](#results)` → tag `results` (the `(` is not in the lookbehind), `?#top` → `top`, `#e-beam-` → `e-beam-`. The first is the realistic one — markdown anchor links are common in longer entries and would surface as filter chips. Consider adding `(` to the lookbehind (or excluding `](#…)`) and forbidding a trailing `-`. Also inconsistent Unicode handling: `#日本` is not a tag but `#eé` is (`[A-Za-z]` first, `\w` after).

**8. Documentation claims the code does not keep.** — LOW (several, cheap)
- `GEECS-Schemas/CHANGELOG.md` 0.22.0: "`Book` exported" — `Book` is not imported or listed in `geecs_schemas/__init__.py` (`EntryKind`/`EntryStatus` are, lines 40-48 and 139-147).
- `GEECS-Schemas/geecs_schemas/log_entry.py:18-19` ("``body_md`` is one markdown string and nothing in this repository parses it") and `:188` ("Opaque markdown; nothing parses it") are now false: `tags.parse_tags` parses it at every save, and the schema's own `tags` field documents that. Same file `:52-53` ("The bytes live on disk beside the entry's markdown, never in the database") and `:191-192`/`:247` ("Files stored beside the markdown") describe the pre-#835 layout; bytes now live in the host attachment store and are only *copied* beside the markdown.
- `GeecsLogbook/geecs_logbook/render.py:532-534`: the `attachment_base` example `/log/attachments/2026-09-11/Scan005` is the removed route shape; the base is now `/log/attachments`.
- `store.py` `unmirrored` docstring (`:296-299`) still motivates the queue rotation with "an entry whose day folder never appears (a note on a day with no scans)" — the case this PR removes.
- PR body / `GeecsLogbook/CLAUDE.md`: "a test monkeypatches `Path.mkdir` to assert every directory the mirror makes is under `logbook/`" — `test_never_creates_anything_in_the_data_tree` exercises `write_entry` only; `mirror_attachments` (the second `mkdir(parents=True)` in the module) runs outside the guard.

### Redundancy

- `_fs.replace_with` is the third temp-file-plus-`os.replace` writer (`ScanAnalysis/scan_analysis/task_queue.py:369-392`, `config_store.py:443-449`). Acknowledged in the CLAUDE.md debt table, so not a new finding — noting only that `task_queue._write_atomic` carries an SMB `DELETE`-permission caveat for `os.replace` on Windows shares that the new helper, which now also targets the share, does not.
- `store.update` hand-rolls `BEGIN IMMEDIATE`/`ROLLBACK`/`COMMIT` (`store.py:437-462`) while `_transaction` exists in the same module; raising `ConflictError` inside a `with _transaction(conn)` block gives the same semantics and removes the duplicate.
- `router.ATTACHMENTS_DIRNAME = "attachments"` and `mirror.ATTACHMENTS_DIR = "attachments"` are two constants for one name; the host layout and the relative link `attachments/<id>/<file>` only coincide because both are the literal. One constant.
- `routes/_common.parse_day` vs `GEECS-DataPortal/geecs_portal/app.py:183 _parse_day` — pre-existing (moved from `router.py`), different semantics (400 vs fall back to today); not introduced here.
- `_entries_json` re-validates `book` by hand (`routes/day.py:105-106`); typing the query param as `Optional[Book]` gets the 422 from FastAPI.

### Placement

- `tags.py`: the `tags` field's meaning is now defined by this regex, and the stated reason `LogEntry` lives in `geecs_schemas` is that GEECS-MCP and a mirror reader "must agree on one definition without depending on each other". A mirror reader reconstructing entries from bodies, or the MCP validating that a posted entry's tags match its body, would need the same parser — it is stdlib-only and would fit next to `LogEntry`. Flagging as code a future consumer moves; not blocking.
- `ops_count` is computed on every day view (`routes/day.py:69`, an extra query) and unused by the template — PR body acknowledges; either drop until PR 3 or render it.

### Checked and found clean

- Migration of a pre-`book`/`tags`/`entry_history` file: `TestMigration` builds a table lacking all six added columns; `ADD COLUMN … NOT NULL DEFAULT` is legal SQLite, indexes are created after the migration, reopening is a no-op (ran).
- History transactionality: `set_status`/`add_attachment` on a tombstone raise inside `_transaction` and roll the snapshot back (probe B2: history count unchanged); a refused edit leaves no history (test passes; the snapshot is inside `BEGIN IMMEDIATE`).
- `query` SQL: `json_each(entries.tags)` correlated EXISTS works; tag filter is lower-cased both sides; `BETWEEN` on `YYYY-MM-DD` strings orders correctly; deleted rows excluded; `include_scan_anchored=False` correct.
- `logbook_root` derivation: four `.parent` hops from `…/scans` land on the experiment directory for every `ScanPaths` layout (including empty experiment); `date.fromisoformat` blocks traversal via `day`; `entry_dir`/`mirror_attachments` never append a `scans` segment; `_assert_own_tree` is applied on `write_entry` and `mirror_attachments`.
- `_fs.claim_name`/`write_claimed`: `O_EXCL` claim, failed write releases the name (test), parallel same-name pastes get distinct files (test); route sanitises the stem and maps the extension from the content type before `save`, and `entry_id` must exist in the store, so `save` cannot be steered to a traversal name via the API.
- Route split versus the old `router.py`: every old route survives with the same name (`_day_page`, `_static`, `_entries_json`, `_attachment` — template `url_for` calls still resolve); `attachment_base`/`api_base` derive from `url_for` so mount prefix/`root_path` still hold; `_attachment` is registered iff the store exists, matching the only caller (`grouped`, which returns early without a store); the only behavioural change beyond the intended ones is the delete-route mirror failure logging at INFO instead of WARNING.
- Day page: renders `book="scans"` only, composer posts `book: "scans"`, `?book=` validated; ops entries excluded from the page and included in the unfiltered API (test).
- Schema: `Book` validator, additive fields with defaults (old rows read unchanged), reference regenerated, portal test updated to expect `attachments/`.
- Scan-folder invariant: no `ScanPaths(read_mode=False)`, no `mkdir` through `scans/`; the portal mount test still asserts nothing but the DB (and `attachments/`) is written on create.
- Public-repo hygiene: no hostnames, account names or home paths added; `Undulator` in tests only.
- Versions/CHANGELOGs bumped for all three packages touched.

---

## Disposition (44bff0b7)

The HIGH was verified independently before fixing (`blobs.path("..", "logbook.db")` returned the database).

| # | Disposition |
|---|---|
| 1 | **Fixed.** `AttachmentStore.path` checks both levels: the entry directory must be directly inside the resolved root, and the file directly inside it. The test's decoys now sit where an escape would land (the database beside the store; a file at the root). |
| 2 | **Fixed, both halves.** `_assert_own_tree` inspects only the trailing `logbook/Y/M/D` segments and raises `MirrorUnavailable`; pinned by a test with a share root at `…/scans/data`. `Context.mirror` also catches anything else after the row landed, logs it and defers — pinned by a route test that makes `mirror_one` raise `RuntimeError` and asserts the save still returns 201 and reads back. |
| 3 | **Fixed.** The experiment directory (the year folders' parent; never a scan folder) must exist before any `mkdir`, else `MirrorUnavailable("share not mounted")`. A dayless day still lands (its test now creates the experiment dir); a missing experiment dir builds nothing (pinned). |
| 4 | **Fixed.** `delete` raises a private sentinel inside the transaction on `rowcount == 0`, so the snapshot rolls back; returns `False`. Pinned (second delete and a missing id leave one history record). |
| 5 | **Fixed.** On `rowcount == 0`, `update` re-reads with `include_deleted=True`: deleted or gone → `KeyError` (404); otherwise `ConflictError` with the current entry. Pinned with a monkeypatched race. |
| 6 | **Fixed.** `max(1, min(limit, cap))`; pinned for `-1` and `10**9`. |
| 7 | **Fixed.** `](#` refused by lookbehind (plain `(#laser)` still a tag), `?` added to the lookbehind, a trailing `-` is not part of a tag, the tail lookahead is Unicode-aware so `#eé` is not `e` and `#日本` is not a tag. Pinned. |
| 8 | **Fixed, all five.** `Book` exported from `geecs_schemas`; the schema's "nothing parses it" and "beside the markdown" wording corrected; the render docstring example; the `unmirrored` docstring; the mkdir-guard test now covers `mirror_attachments` too, so the CLAUDE.md claim is true. |
| Redundancy | **Fixed:** `update` uses `_transaction`; one `ATTACHMENTS_DIR` constant (router imports the mirror's); `?book=` typed as `Optional[Book]`; the SMB `DELETE`-permission caveat noted on `_fs.replace_with`. **Waived:** `parse_day` vs the portal's `_parse_day` — different semantics, pre-existing; the atomic-write helper relocation is already in the deferred table with the owed ScanPaths issue. |
| Placement | **Waived:** `tags.py` stays in the logbook until a second consumer exists (the agent-verbs phase, same as `EntryCreate` — recorded in the deferred table). **Fixed:** the unused `ops_count` query is gone until the month page renders it. |

Counts after the fixes: GeecsLogbook 143, GEECS-DataPortal 254, GEECS-Schemas 460, GeecsWebTheme 104.

skbarber:
## Verification (same reviewer, on 44bff0b7) — final

Reviewer's verdict, verbatim:

---

## Verification of 44bff0b7 against the #835 review

Every probe re-run against the fixed worktree (scratchpad only; explicit `base_directory` under a temp dir throughout; no tracked file touched). Suites: GeecsLogbook 143, GEECS-DataPortal 254, GEECS-Schemas 460 — all green. GeecsWebTheme was not re-run; the commit touches no file in it.

### Per-finding verdict

| # | Finding | Verdict | Evidence |
|---|---|---|---|
| 1 | Attachment `..` traversal serves the DB | **Confirmed fixed** | `blobs.path("..","logbook.db")` → `None`; raw ASGI `GET /log/attachments/../logbook.db` → **404** (was 200 with `SQLite format 3`); `./logbook.db`, `../../state/logbook.db`, filename `.`/`""`, and an entry directory that is a **symlink out of the root** all → `None`/404. The rewritten test puts the decoy at `root.parent/logbook.db`, so it now fails against the old code. |
| 2 | `RuntimeError` from `_assert_own_tree` → 500 after the row landed | **Confirmed fixed** | POST with `base_directory=<tmp>/scans/d` → **201** (was 500); `_assert_own_tree` raises `MirrorUnavailable`; a short path is refused; an experiment literally named `scans` or `logbook` now resolves to `<base>/<exp>/logbook/Y/M/D` — correct, that tree is the mirror's own. The new route test makes `mirror_one` raise `RuntimeError` and asserts 201 + read-back, which exercises the broad catch. |
| 3 | Mirror builds a tree under an unmounted mount point | **Confirmed fixed** | Empty base → `MirrorUnavailable("share not mounted…")`, directory left empty; experiment dir present but no year folder → the ops entry still lands and `Y2026` is **not** created in the data tree; experiment path that is a file → `MirrorUnavailable`, not `RuntimeError`. |
| 4 | Repeated `delete` commits spurious history | **Confirmed fixed** | Three deletes → `[('delete', 1)]` only; `set_status` on a tombstone still leaves history untouched. |
| 5 | Edit after concurrent delete → stale 409 | **Confirmed fixed** | `KeyError`; the rolled-back edit snapshot is absent (history = `[('delete', 1)]`); a genuine conflict still raises `ConflictError` carrying the **latest** entry (v2, `"v2"`). |
| 6 | `limit=-1` unbounded | **Confirmed fixed** | `-1` → 1 row; `10**9` → capped. |
| 7 | Tag parser edges | **Confirmed fixed** | `[see](#results)` → `[]`, `?#top` → `[]`, `#e-beam-` → `[]`, `#日本`/`#eé` → `[]`; `(#laser)`, `#e-beam`, `#a-b`, `#Tag_`, 32-char cap and 33-char rejection all as intended. |
| 8 | Doc claims | **Confirmed fixed** | `Book` imported and in `__all__` (`__init__.py:43,146`); `log_entry.py`, `render.py`, `store.unmirrored` docstrings now match the code; the mkdir-guard test runs `mirror_attachments` under the guard and asserts ≥2 guarded mkdirs. |
| R | Redundancy items | **Confirmed** | `update` runs inside `_transaction`; one `ATTACHMENTS_DIR` (router imports it from `mirror`, no import cycle); `Optional[Book]` query param (the `?book=nope` → 422 test still passes); SMB note on `replace_with`. |
| P | Placement | **Confirmed** | `ops_count` and its extra query gone; `tags.py` stays, recorded. |

### Regressions introduced by the fixes

None functional. Three nits, none blocking:

- **`query(limit=0)` now returns one row** (`max(1, …)`), where before it returned none. Neither behaviour is documented; if a caller ever uses `limit=0` as "count only / nothing", it will get a row. Suggest `limit < 1 → ValueError`, or document that the floor is 1. (`store.py:276`)
- **Tag lookbehind is now ASCII-only**, so `é#tag` yields `tag` where the old `\w` lookbehind refused it (the "not glued to a preceding word character" rule no longer holds for non-ASCII letters). Cosmetic; the trailing lookahead was correctly kept Unicode-aware. (`tags.py:27-29`)
- **`tags.py` docstring says "not ending in `-`"**, which reads as trimming, but the behaviour is that `#e-beam-` is not a tag at all (probe: `[]`, and the new test pins `[]`). One-word wording fix.
- `TestPaths` class docstring ("Nothing here touches the disk") is now slightly untrue — `logbook_root` stats the experiment directory. Trivial.

One caveat, not a regression: `experiment_dir.is_dir()` on a *hung* (not absent) CIFS mount blocks like any other stat; `read_day` already stats the same share on every page view, so the exposure is unchanged.

---

## Disposition of the nits (0ba721f7 + b7c8b37d)

All four taken: `limit < 1` raises `ValueError` (pinned for 0 and −1); the tag lookbehind is `\w`-based again (Unicode-aware on both sides; `é#glued` pinned); the tags docstring says a run ending in `-` is not a tag; the `TestPaths` docstring corrected. The hung-mount caveat is accepted as unchanged exposure.

Candour note: 0ba721f7 was pushed with one red test (the `limit=-1` pin, whose edit had missed a reformatted anchor); b7c8b37d fixed it four minutes later. GeecsLogbook is 144 green on the head.

Review resolved: no undispositioned finding. Merging into `feature/scan-logger` once CI on b7c8b37d is green, per the owner's standing ruling for phase PRs on this branch.

Issue #834, opened 2026-09-12
Data portal reached 5.3 GB RSS and was the host's first OOM casualty — bound the process, not just the pixel cache


skbarber:
## What happened (2026-09-11, ~21:00 PDT, services host)

An unrelated ad-hoc Tiled probe (`read_primary_dataframe()` on an image run, run over ssh) took the host to 147 MB free with 8.4 GB of swap in use. The kernel's OOM killer then chose the **data portal** first, at 21:06:15:

```
Out of memory: Killed process 12546 (python) total-vm:6895627160kB, anon-rss:5273120kB ... task_memcg=/system.slice/geecs-data-portal.service
```

`anon-rss` **5.27 GB** for the portal process. systemd restarted it 14 s later (`Restart=on-failure`), which users saw as `ERR_CONNECTION_REFUSED` and then a portal that took 14 s to redirect while Tiled was starved. The probe was killed and the host recovered; the portal is healthy now.

## Why this is a portal issue even though the trigger was external

`geecs_portal/cache.py` deliberately keeps within-scan data (owner doctrine 2026-08-29: "a diagnostic's full data for one scan is ~100s of MB — trivial against server RAM"). Two observations against that premise:

1. **The byte bound covers only pixels.** `ShotDataCache` is bounded at `_PIXEL_BUDGET_BYTES = 1.5 GB`, but `CachingScanCatalog` is bounded by *count* (`_DETAIL_CACHE_RUNS = 8`), not bytes. The journal shows the portal fetching the same run's camera arrays through Tiled (`array/full/.../uc_amp4_ir_input?expected_shape=5,600,600`, and the 700×700 device) in the minutes before the kill. Whether those land in the detail cache, the pixel cache, or transiently in a render path, the process footprint was **3.5× the only budget that exists**.
2. **"Trivial against server RAM" assumes the portal has the box to itself.** It shares 15.8 GB with Tiled, MySQL, the queueserver, the capture daemon, the CA gateway and the MCP server — and whoever runs a probe. Under any pressure the portal, being the largest resident process, is the one that dies, and it is the one users are looking at.

## Suggested shape (not decided)

- Measure what the process actually holds after browsing one image-heavy run (detail cache entries' event tables — do they carry array columns? — plus pixel cache) so the number is a fact, not an inference from `anon-rss`.
- Byte-bound the detail cache too, or strip array-typed columns from cached run detail; total budget as one number.
[...]

PR #836, opened 2026-09-12
GEECS-Data-Utils 0.29.1: the catalog and the s-file export read the primary stream's scalar table only (#834)

skbarber:
Fixes the portal/host memory runaway of 2026-09-11 (#834): opening a two-camera plugin run (Scan008 of 26_0911) in the data portal took the worker host to 147 MB free twice and the kernel killed the portal each time.

## Cause

`TiledScanCatalog.load_run` (the scalar table every browser uses: the portal's run page, the Console's scan browser) and `tiled_export._fetch_run` (the Tiled → s-file re-export) both did `run["primary"].read().to_dataframe()`. A Bluesky run's `primary` node is a Tiled composite: one `internal` table plus one array node per external data key. `CompositeClient.read()` with no `variables` downloads **every** array part (verified in the installed client: camera stacks and per-frame attributes included), and `xarray.Dataset.to_dataframe()` then takes the outer product of all array dimensions. One 5×600×600 stack survives; two stacks of different shapes (5×600×600 and 5×700×700) multiply into billions of rows. The portal's run page triggers it before any Images tab, and it is independent of the branch the portal runs (master and feature alike) — it needed a plugin run with two cameras to surface.

## What

- `tiled_catalog.read_primary_scalars(primary)`: reads the table part(s) by name through `primary.base` and never touches an array part; `None` for a stream with no table or no rows (the old "no event rows" contract kept). `load_run` and `_fetch_run` use it. Pandas stays a lazy import of the module.
- The test fake is now a composite node (`get_contents` + `base[part]`) whose array parts raise if read; `test_load_run_reads_the_scalar_table_and_never_the_array_parts` pins a two-camera run (scalars only, arrays untouched) and the empty-table case. The module docstring no longer blesses `run["primary"].read()`.

LOC: catalog +45/−10, export +6/−1, tests +75/−25, CHANGELOG +13. No facility literals.

## Tests

`./scripts/check.sh` (lint + every suite): GEECS-Data-Utils **328 passed, 3 skipped** (48 in the two Tiled test files); GEECS-DataPortal **209 passed, 32 skipped**; GEECS-Console **814 passed**; the rest unchanged and green.

## Hardware verification

**Done, 2026-09-11 21:41**: `~/portal-checkout` on this branch, portal restarted, Scan008 (run 72b54ce0, two plugin-backed cameras) opened in the portal: run page, both cameras' shot images and a processed bin image served; the portal process at 415 MB RSS afterwards, host at 13.9 GB free, and **zero** `array/full` requests to Tiled in the portal's journal (the previous two attempts showed the full-array reads and a 14.6 GB process).

Review: per /land, a fresh-context adversarial pass runs in parallel with CI; the maintainer chose to keep it despite the size, because the change sits on every browser's read path.


skbarber:
# Adversarial review (fresh-context subagent)

Branch `fix/catalog-scalar-only-read` → `feature/native-bluesky-plans`, PR #836, HEAD c9eaee52. Tests: 48 passed in the two Tiled files; the new test **fails on the base branch** (ran it against a copy of `origin/feature/native-bluesky-plans`'s `tiled_catalog.py`: `AssertionError: primary.read() downloads every array part`), so it pins the fix.

## Findings

**1. P2 — The same defect survives in `GeecsBluesky/geecs_bluesky/assets/tiled_readback.py:183` (`read_primary_dataframe`), and the PR body's "every browser's read path" claim doesn't cover it.**
`read_primary_dataframe` is verbatim the old code (`run["primary"].read()` → `.to_dataframe().reset_index()`). It is on the public path `load_asset_from_tiled_run` → `load_camera_image_from_tiled` (re-exported from `geecs_bluesky.assets.__init__`, used by `GeecsBluesky/notebooks/tiled_external_asset_readback.ipynb`), and `TILED_SETUP.md:100` names `tiled_readback.py` as a user of the pattern. Opening Scan008 through it reproduces #834 exactly (both stacks downloaded, dimensions outer-joined) — and the only column it needs, `scan_event_index`, lives in the `internal` table. GeecsBluesky depends on GEECS-Data-Utils, so the swap is one line: `read_primary_scalars(run["primary"])` (with `None` → empty frame, as `_fetch_run` does). Its test fake (`GeecsBluesky/tests/test_tiled_asset_readback.py:39`, a `read()` returning a DataFrame) needs the composite-fake shape from this PR. Either in this PR (GeecsBluesky patch bump) or a filed follow-up — but not left silent.

**2. P2 — Three docs in the repo still bless the pattern the PR forbids, and the new docstring points readers at one of them.**
- `GEECS-Data-Utils/CLAUDE.md:284` describes `load_run` as "the repo-blessed `run["primary"].read().to_dataframe().reset_index()`" — the same package this PR versions; it is now false. Same-PR.
- `GeecsBluesky/TILED_SETUP.md:100-103` (post-upgrade verification: "`run["primary"].read()` — the pattern `tiled_export.py` / `tiled_readback.py` use") and `:186` (`df = run["primary"].read()` in the client recipe). The rewritten module docstring says "see `GeecsBluesky/TILED_SETUP.md`" for the scalar-only rule — that page still recommends the opposite.
- `GeecsCAGateway/DEPLOYMENT.md:322-327`: `df = run["primary"].read().to_dataframe().reset_index()   # per-shot scalar table` with the note that "`.to_dataframe()` … is how the repo's own readers get the per-shot table".
The recipes should become `read_primary_scalars(run["primary"])` (or `run["primary"].base["internal"].read()` for the ad-hoc case) with a one-line "never `.read()` a primary with camera parts" warning.

**3. P3 — The multi-table branch (`pd.concat(frames, axis=1)`) aligns on index labels, not row position, and is untested.**
`DataFrameClient.read()` returns whatever index the arrow→pandas conversion yields per partition (dask `from_map` keeps per-partition RangeIndexes, so npartitions>1 gives duplicate labels); `concat(axis=1)` then raises `InvalidIndexError` unless the indexes are identical, and different row counts NaN-pad. Unreachable today — bluesky 1.15.0's TiledWriter creates exactly one `internal` table with one partition (`create_appendable_table` + `append_partition(table, 0)`) — but the branch exists, so make it positional (`reset_index(drop=True)` per frame before concat) or drop it and read the single table. Related nit: `load_run`'s `except KeyError` now also swallows a KeyError raised *inside* the helper (`item["attributes"]["structure_family"]`, `primary.base[part]`) as "no primary stream".

## Checked and refuted (no finding)

- **Client contract**: `CompositeClient.get_contents()`, `.base` (a plain `Container` over the same item), and `DataFrameClient.read(columns=None)` are identical in the installed 0.2.9, in 0.2.14 (fetched `tiled/client/composite.py` at tag v0.2.14), and in the portal venv's 0.2.17. In all three `read(variables=None)` fetches every array part (0.2.14+ is lazy only for `DaskArrayClient`; the default numpy `ArrayClient` is eager), so the diagnosis holds on the deployed versions. The `== "table"` comparison is fine (`StructureFamily` is a str enum; msgpack yields plain strings).
- **Index / columns consumers relied on**: the old frame carried a `dim0` column (table columns get dim `dim0` in `read()`), plus array-dim columns whenever arrays existed. No code in the portal, Console, MCP, or data-utils references `dim0` (grep, venvs excluded). Consumers use `scan_event_index` else `frame.index + 1` (`figures.shot_axis_for_frame`), `frame[column].iloc[shot-1]` (portal `app.py:246`), positional `range(1, n+1)` (`build_legacy_scalar_dataframe`), `tiled_schema.plottable_columns` — all correct on the `RangeIndex` from `reset_index(drop=True)`; `seq_num`/`time` are columns of the internal table anyway. Side effect worth a CHANGELOG sentence: before this PR *any* plugin run with even one camera produced a broadcast frame (5×600×600 = 1.8 M rows for Scan008-shaped single-camera runs), so the portal's row count and shot lookups were wrong there too.
- **Legacy non-composite `primary`**: `get_contents`/`.base` exist only on `CompositeClient` (`Container` has neither, nor `read()`), so such a run raises `AttributeError` — exactly as the old code did; parity, not a regression, and `TILED_SETUP.md:100` records that existing runs read back through the composite client.
- **Dask-backed frames**: handled by the `compute()` branch; `from_uri` in `_connect` uses the default numpy structure clients anyway.
- **Redundancy**: the portal and Console import no `tiled` (only `tiled_schema`/`tiled_drift`); no `column=`/`columns=` Tiled reads exist there; `tiled_drift.py` takes column samples, not runs; the worker's s-file callback (`GeecsBluesky/geecs_bluesky/callbacks.py:284`) builds its frame from its own event rows, no Tiled round trip; `_fetch_run` is the CLI re-export only and is untested (unchanged). The one duplicate is finding 1.
- **Placement**: `tiled_catalog.py` is the right home — foundational package, lazy `tiled` import, both in-package callers, importable by GeecsBluesky for finding 1.

skbarber:
## Dispositions (commit 04b5c6aa)

1. **Fixed in this PR** — `assets.tiled_readback.read_primary_dataframe` now calls `read_primary_scalars(run["primary"])` (empty frame for `None`); GeecsBluesky 0.82.2 with a CHANGELOG entry; its test fake is a composite node whose array part raises if read (the notebook path is covered by the same function).
2. **Fixed** — `GEECS-Data-Utils/CLAUDE.md`, `GeecsBluesky/TILED_SETUP.md` (the post-upgrade check and the client recipe) and `GeecsCAGateway/DEPLOYMENT.md` (its recipe, GeecsCAGateway 0.20.4 docs bump) all say `read_primary_scalars` and carry the one-line warning; the module docstring's pointer to TILED_SETUP.md now points at text that agrees with it.
3. **Fixed** — each table part is `reset_index(drop=True)` before the concat (row position, never labels); `load_run` catches `KeyError` only around the `run["primary"]` lookup, so a KeyError inside the helper propagates. The CHANGELOG gains the sentence about single-camera plugin runs having produced broadcast frames before this fix.

Suites after the fixes: GEECS-Data-Utils Tiled tests 48 passed; GeecsBluesky `test_tiled_asset_readback` + `test_tiled_integration` 21 passed; lint green.

skbarber:
Re-review of 04b5c6aa by the same reviewer: all three findings CONFIRMED, no new findings; the pinning test was verified to fail on the base branch. CI green. Merging into the feature branch (maintainer-delegated tonight); deploy = `~/portal-checkout` back onto `feature/native-bluesky-plans` + portal restart.

PR #837, opened 2026-09-12
Phase 04: the editor — GeecsLogbook 0.4.0

skbarber:
PR 2 of the path to a functional scans + ops logbook, into `feature/scan-logger` (not master). With this, the **scans logger is functionally complete**: writing, editing, screenshots, tables, on the day page, durable and mirrored.

## What

`static/editor.js` — the whole write path in the browser, one implementation for every composer on the page (per scan, per gap, the day, in-place edits; the month page next). Deliberately not WYSIWYG: the body stays the plain markdown the mirror holds.

- **Toolbar**: bold, italic, code, bulleted list, task list, link, table skeleton, callout, image (file picker), preview. Each writes markdown around the selection.
- **Paste or drop a screenshot** into the composer: the file goes to `POST /api/entries/{id}/attachments` and a relative image link lands at the cursor. A brand-new entry has no id until saved, so the first attachment **saves it first** (an "autosave"; the form shows *saved with attachment*) and Save becomes a PATCH; a **Discard** button deletes the stub. Uploads bump the version, so the form re-reads it before its eventual Save.
- **Paste a spreadsheet range**: an HTML table on the clipboard (Excel/Sheets) or tab-separated text with ≥2 lines becomes a markdown table; `|` in cells escaped.
- **Preview**: `POST /api/preview` renders with the page's renderer and the same attachment base, toggled in place of the textarea.
- **⌘/Ctrl+Enter** saves; a 409 shows what happened rather than a generic error.
- **Image grid** (`render._image_grids`): two or more consecutive image paragraphs render as one `figgrid` — the plot table that replaces LogMaker's `gdoc_slot` numbering (no slot assignment, no ceiling at four).

The template no longer carries an inline script; it hands the editor its three facts through `<main id="logbook" data-api data-day data-book>` and loads `editor.js` from the logbook's static route. `editor.js` sets no colours (the theme guard's rule; all styling through tokens in `scanlog.css`).

## Per-concern LOC

| Concern | ~LOC |
|---|---|
| `static/editor.js` (new; includes the ~110 lines moved out of the template) | +330 |
| Template: hooks, inline script removed | +6 / −100 |
| Renderer: image grid | +25 |
| Route: `POST /api/preview` | +20 |
| Styles: toolbar, preview, drop target, grid | +20 |
| Tests: grid (4), preview + page hooks (3) | +60 |
| CHANGELOG, CLAUDE.md "The editor" | +45 |

- **Autosave-on-first-attach.** The alternative — staging uploads under a temporary id until Save — is more machinery for the same outcome. The cost is a stub entry if someone pastes a screenshot and walks away; Discard exists, and the entry is visible on the page, not hidden.
- **The `who` field is still a free-text name** stored in `localStorage`; identity is a later concern (there is no login anywhere in the portal).
- **No JS unit tests** — the repo has no JS test runner, and adding one for one file is out of proportion. `editor.js` is syntax-checked with node in this session and exercised by hand on a local smoke server; the Python tests pin what the page hands the script and the routes it calls.

## Facility literals

None.

## Tests

GeecsLogbook 151 (was 144: +4 grid, +3 preview/page hooks), GeecsWebTheme 104 (the guard still walks the template and stylesheet).

## Hand-checked (local smoke server, temp share + temp db)

Page serves the hooks and the script; `node --check` on `editor.js`. Owner to click through: type → Save; Edit → toolbar → Save; paste a screenshot; paste a spreadsheet range; Preview.

## Review

### DIFF
--- a/GeecsLogbook/CHANGELOG.md
+++ b/GeecsLogbook/CHANGELOG.md
@@ -28,7 +28,35 @@ The editor: what makes people use it.
 
 - The day page's inline write-path script is gone; the page hands the
   editor its facts through `<main id="logbook" data-api data-day
-  data-book>` and loads `editor.js`.
+  data-book data-accept>` and loads `editor.js`.
+
+### Fixed (review of #837)
+
+- An uploaded image link was left selected, so the next keystroke — or
+  the second file of a multi-file drop — replaced it. Attachment links
+  now insert with the caret after them; only the table and callout
+  skeletons stay selected for overtyping.
+- Duplicate entries from a held ⌘↵ (key repeat), a double submit, or two
+  quick pastes on a new composer: one create per form (the pending POST
+  is shared), key-repeat ignored, Save/Discard exclusive with the buttons
+  disabled meanwhile.
+- The version is re-read after uploads even when a later file in the
+  batch failed, and written back to the article for an in-place edit so
+  Cancel-then-Edit does not start stale.
+- Attaching before entering a name no longer creates an entry attributed
+  to "unknown" forever: the name is required first.
+- Paste precedence: an HTML table on the clipboard wins over a bitmap of
+  the same cells (Excel); a file pasted alongside text keeps the text and
+  uploads the file.
+- Tab-indented prose is not turned into a table (consistent column count
+  and a non-empty header cell are required); list/task prefixing at the
+  start of the text and on a triple-clicked line; a failed preview leaves
+  nothing visible; a dropped file of a type the server does not take says
+  so instead of vanishing. Accepted types come from the page
+  (`data-accept`), one list not two. The theme guard walks `editor.js`.
+- Between-scan blocks collapse as well as expand (owner's hand-test):
+  they are `<details>` like the scans, with a count in the header, and
+  Expand all / Collapse all reaches them.
 
 ## [0.3.0] - 2026-09-11
 
--- a/GeecsLogbook/CHANGELOG.md
+++ b/GeecsLogbook/CHANGELOG.md
@@ -4,6 +4,32 @@ All notable changes to `geecs-logbook` are documented here. The format
 follows [Keep a Changelog](https://keepachangelog.com/en/1.1.0/) and this
 project adheres to semantic versioning.
 
+## [0.4.0] - 2026-09-11
+
+The editor: what makes people use it.
+
+### Added
+
+- `static/editor.js`, one implementation for every composer on the page
+  (per scan, per gap, the day, and in-place edits): a toolbar that writes
+  markdown around the selection (bold, italic, code, list, task list,
+  link, table, callout, image, preview); **paste or drop a screenshot**
+  into the composer — the file goes to the upload endpoint and a relative
+  image link lands at the cursor; a brand-new entry is saved first so the
+  upload has somewhere to go ("autosaved"), with Discard to delete the
+  stub; **paste a spreadsheet range** (tab-separated or an HTML table) as a
+  markdown table; a server-rendered Preview; ⌘/Ctrl+Enter saves.
+- `POST /api/preview` — the page's renderer, for the composer.
+- Two or more consecutive images render as one grid (`figgrid`): the
+  plot table, replacing LogMaker's `gdoc_slot` numbering — no slot
+  assignment, no ceiling at four.
+
+### Changed
+
+- The day page's inline write-path script is gone; the page hands the
+  editor its facts through `<main id="logbook" data-api data-day
+  data-book>` and loads `editor.js`.
+
 ## [0.3.0] - 2026-09-11
 
 The foundation for the operations book. Owner rulings 2026-09-11.
--- a/GeecsLogbook/CLAUDE.md
+++ b/GeecsLogbook/CLAUDE.md
@@ -153,7 +153,11 @@ stays the plain text the mirror holds. The two things people actually
 need — paste a screenshot, paste a spreadsheet — are events on the
 textarea: a pasted or dropped file goes to the upload endpoint and its
 relative link lands at the cursor; a tab-separated or HTML-table paste
-becomes a markdown table. A brand-new entry has no id until saved, so
+becomes a markdown table (a text-only paste must have a consistent
+column count and a non-empty first header cell — tab-indented prose is
+not a table, and a cross-tab with a blank corner from a text-only source
+is pasted as text; spreadsheet apps supply the HTML form, which has no
+such rule). A brand-new entry has no id until saved, so
 the first attachment saves it first ("autosaved", with Discard); Save is
 then an edit. The page hands the script its facts through
 `<main id="logbook" data-api data-day data-book>`; nothing is templated
--- a/GeecsLogbook/CLAUDE.md
+++ b/GeecsLogbook/CLAUDE.md
@@ -141,7 +141,23 @@ package. Rules the store enforces, each pinned in `tests/test_store.py`:
   `ALTER TABLE` with a fill expression. Nothing else, deliberately.
 
 `body_md` is opaque: `render.render_markdown` (markdown-it + nh3) is the
-only thing that reads it, and only to draw it.
+only thing that reads it, and only to draw it — plus the tag scan.
+
+## The editor
+
+`static/editor.js` is the whole write path in the browser, one
+implementation for every composer (per scan, per gap, the day, in-place
+edits, and the month page next). It is deliberately **not** a WYSIWYG
+editor: the toolbar writes markdown around the selection and the body
+stays the plain text the mirror holds. The two things people actually
+need — paste a screenshot, paste a spreadsheet — are events on the
+textarea: a pasted or dropped file goes to the upload endpoint and its
+relative link lands at the cursor; a tab-separated or HTML-table paste
+becomes a markdown table. A brand-new entry has no id until saved, so
+the first attachment saves it first ("autosaved", with Discard); Save is
+then an edit. The page hands the script its facts through
+`<main id="logbook" data-api data-day data-book>`; nothing is templated
+into the script. It sets no colours (the theme guard's rule).
 
 ## Status is reported, not inferred
 
--- a/GeecsLogbook/geecs_logbook/render.py
+++ b/GeecsLogbook/geecs_logbook/render.py
@@ -60,6 +60,13 @@ _CALLOUT = re.compile(
 #: Attachment references as the store writes them.
 _ATTACHMENT_REF = re.compile(r'(src|href)="attachments/')
 
+#: Two or more images in a row, each its own paragraph — what a run of
+#: pasted screenshots or published figures renders as. They become one
+#: grid, the plot-table LogMaker's ``gdoc_slot`` numbering used to fake:
+#: no slot assignment, no ceiling at four, no collisions.
+_IMAGE_RUN = re.compile(r"(?:<p>\s*<img\b[^>]*>\s*</p>\s*){2,}")
+_IMAGE = re.compile(r"<img\b[^>]*>")
+
 
 def render_markdown(body_md: str, *, attachment_base: str | None = None) -> str:
     """Render an entry body to safe HTML.
@@ -84,10 +91,23 @@ def render_markdown(body_md: str, *, attachment_base: str | None = None) -> str:
         rendered, tags=_TAGS, attributes=_ATTRIBUTES, link_rel="noopener noreferrer"
     )
     clean = _callouts(clean)
+    clean = _image_grids(clean)
     if attachment_base:
         base = attachment_base.rstrip("/")
         clean = _ATTACHMENT_REF.sub(rf'\1="{html.escape(base)}/', clean)
     return clean
+
+
+def _image_grids(fragment: str) -> str:
+    """Gather consecutive image paragraphs into a ``figgrid``."""
+
+    def swap(match: re.Match[str]) -> str:
+        imgs = "".join(
+            f"<figure>{img}</figure>" for img in _IMAGE.findall(match.group(0))
+        )
+        return f'<div class="figgrid">{imgs}</div>'
+
+    return _IMAGE_RUN.sub(swap, fragment)
 
 
 def _callouts(fragment: str) -> str:
--- a/GeecsLogbook/geecs_logbook/routes/day.py
+++ b/GeecsLogbook/geecs_logbook/routes/day.py
@@ -19,6 +19,7 @@ from geecs_schemas.log_entry import Book, LogEntry
 
 from geecs_logbook.models import DaySummary
 from geecs_logbook.render import render_markdown
+from geecs_logbook.routes.attachments import ATTACHMENT_TYPES
 from geecs_logbook.routes._common import (
     STATIC_DIR,
     Context,
@@ -87,6 +88,7 @@ def register(router: APIRouter, ctx: Context) -> None:
                 "entry_count": sum(len(v) for v in entries.values()),
                 "writable": ctx.writable,
                 "api_base": api_base(request),
+                "accept": ",".join(sorted(ATTACHMENT_TYPES)),
             },
         )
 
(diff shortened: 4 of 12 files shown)


skbarber:

Reviewer's report, verbatim, followed by the disposition. Fix commit: 8caf91f2. Verification by the same reviewer is in flight; a second comment will carry its verdict. The owner also hand-tested the editor on the smoke server ("looks really really good") and found one nit of their own, fixed in the same commit.

---

## Review of PR #837 — Phase 04: the editor (GeecsLogbook 0.4.0)

Diff base `origin/feature/scan-logger...HEAD`; GeecsLogbook 151 passed, GeecsWebTheme 104 passed, `node --check editor.js` clean. Findings below were each reproduced by reading the code as a browser would run it, by a node harness modelling the spec'd `setRangeText`/`setSelectionRange` semantics, or by curl against the smoke server.

### Findings, ranked

**1. HIGH — An uploaded image link is left *selected*, so the next insert or keystroke overwrites it**
`GeecsLogbook/geecs_logbook/static/editor.js:93-98` (`insertBlock`) and `:158-167` (`uploadFiles` loop).
`insertBlock` calls `replaceSelection(ta, lead + text + "\n", lead.length, lead.length + text.length)`, which selects the inserted text. That is right for the table skeleton and callout placeholders, wrong for an attachment link. Two concrete failures (both confirmed by harness):
- Drop or pick two files at once: the loop's second `insertBlock` reads `selectionStart/End` = the first link and replaces it. Value after two uploads: `"note\n\nb\n\n"` — `a.png` is uploaded, attached, version bumped, but its link is gone.
- Paste one screenshot, then type a caption: the first character replaces the whole `image`. Value: `"note\n\nx\n"`.
Fix: give `insertBlock` a "collapse caret after" mode (or call `replaceSelection` without the selection args) for uploads; keep the select-placeholder behaviour only for the toolbar skeletons.

**2. HIGH — No in-flight guard on `save`/`ensureEntry`: duplicate entries from key repeat, double submit, or two quick pastes**
`editor.js:257-273`, `:134-148`, `:302-304`.
`ensureEntry` only short-circuits on `form.dataset.entry`, which is set *after* the POST resolves; nothing disables the submit button or memoises the pending create. Holding ⌘+Enter for a few hundred ms fires auto-repeat `keydown`s (each with `ev.repeat === true`) → N `POST /entries` → N identical entries before `location.reload()` lands. The same race exists between a paste (→`uploadFiles`→`ensureEntry`) and a Save on a brand-new composer, and between two pastes in succession — two stubs, one orphaned with no Discard. For an existing entry the second PATCH gets a 409 and flashes "Someone else saved this first" before the reload.
Fix: `if (ev.repeat) return;` in the keydown handler; store the pending create promise on the form (`form._creating ||= api(...)`) and await it; add a `busy` gate to `save`/`discard`.

**3. MEDIUM — A partially failed batch skips the version refresh → the eventual Save is a misleading 409**
`editor.js:158-170`. The `GET /entries/{id}` that refreshes `dataset.version` sits *after* the loop. Drop two files where the second is over the 20 MiB cap (or an unsupported type slipping past the client filter): file 1 uploads and `store.add_attachment` bumps the version (`store.py:497`), file 2 throws 413, the refresh never runs. Save then PATCHes with the stale `expected_version` and the user is told "Someone else saved this first" although nobody did.
Fix: refresh in `finally` (when `id` is set), or have `POST …/attachments` return the entry's new `version` and set it per file.

**4. MEDIUM — Autosave-on-first-attach records `author = "unknown"` permanently**
`editor.js:137`: `name = who.value.trim() || who() || "unknown"`. A first-time user (empty `localStorage`) who pastes a screenshot before typing a name creates the stub as "unknown". `save()` later insists on a name, but PATCH only sets `editor`; `store.update` deliberately never touches `author` and its docstring notes the author is "part of the mirror file's stable name" (`store.py:415-418`). The entry — and its markdown file on the share — stay attributed to "unknown".
Fix: gate `ensureEntry` the way `save` is gated (focus `.who`, `fail(form, {message: "Enter your name before attaching"})`).

**5. MEDIUM — Edit → paste image → Cancel → Edit again → Save = 409**
`editor.js:169-170` writes the fresh version to `form.dataset.version` only; `:338-339` seeds each new edit form from `art.dataset.version`, which the upload made stale; `:349` Cancel just removes the form without reloading. Second Edit + Save conflicts on the entry's own version.
Fix: after the refresh, also write `art.dataset.version` for `.editing` forms (or reload on Cancel when an upload happened).

**6. MEDIUM (needs a hand check on Excel/Windows) — Paste precedence: any file item wins, dropping the text/HTML alongside it**
`editor.js:239-247`. Files are checked before `text/html`. Excel is widely reported to place a PNG rendering of the copied cells on the clipboard next to `text/html`; where that holds in Chrome, the headline "paste a spreadsheet range" feature uploads a screenshot of the cells instead of producing a table. Independently of Excel, a rich paste carrying an image plus text (Word, a web page) silently discards the text. I could not confirm the Excel clipboard contents from here — the smoke test list in the PR body ("paste a spreadsheet range") is exactly the check; do it from Excel, not only Sheets.
Fix: if `text/html` contains a `<table>`, take the table; only then fall to file items; otherwise let the browser's default paste keep the text and upload the file as well.

**7. LOW — `tsvToRows` converts tab-indented text to a table**
`editor.js:113-117`. Every line must contain a tab, so `"\tfoo\n\tbar"` (two lines copied from inside an indented block in a terminal — no `text/html` there) becomes a two-column table with an empty first column. Harness: a block with one unindented line escapes; a fully indented one does not.
Fix: require a consistent column count ≥ 2 with non-empty first cells, or ignore leading tabs.

**8. LOW — `prefixLines` edge cases**
`editor.js:80-90`. Caret at 0 with a leading newline: `lineStart = 1`, `lineEnd = 0`, `setRangeText(done, 1, 0)` throws `IndexSizeError` (uncaught in the click handler; nothing sticks). A selection ending just after a `\n` (triple-click a line) also prefixes the following line (`"- one\n- two\nthree"`).

**9. LOW — Preview failure leaves an empty, visible pane**
`editor.js:185-190`. The pane is created without `hidden`; on a failed `/preview` call the empty dashed box and the textarea are both visible and `is-on` is unset; the next click on preview *hides* the pane rather than retrying.

**10. LOW — Contract claim not pinned: "editor.js sets no colours (the theme guard's rule)"**
`GeecsWebTheme/tests/test_no_literal_colours.py:37-46` `_SURFACES` lists `scanlog.css` and `day.html` but not the new `static/editor.js`; the CLAUDE.md/PR statement is currently true but unenforced. Add the file to the list (the comment there says a new surface "means adding it here").

**11. LOW (placement/redundancy) — `IMAGE_TYPES` duplicates `ATTACHMENT_TYPES`**
`editor.js:131` vs `routes/attachments.py:26-32`. Two homes for one contract, and the client filter drops non-matching files silently (`if (!list.length) return;` — dropping a `.bmp`/`.svg` produces no message). Either hand the list through `data-accept` on `<main id="logbook">` (the page already hands facts that way) or send everything and surface the server's 415, which already names the accepted types.

**Nits**: `PreviewRequest.entry_id` is a dead field ("unused for now"); `from fastapi import Request` is a second, out-of-order `fastapi` import in `routes/entries.py:22` (passes because ruff `I` is not enabled); after an autosave, Save with an emptied textarea PATCHes an empty body (the `trim` gate at `:268` only covers the create path).

### Checked and found clean

- **`_image_grids` regex** (`render.py:67-68`): probed via `/api/preview` — text between images splits runs; two images on adjacent lines (one `<p>`) are not gridded; a linked image is not gridded; grids form correctly inside `<li>` and `<blockquote>`; `<pre><code>` content is entity-escaped so `<p>` text in a code block cannot match; `alt="a&gt;b"` cannot break `[^>]*`.
- **`POST /api/preview`**: same `render_markdown` + `attachment_base(request)` as the day page (`routes/day.py:40-43`); `<script>` arrives escaped; `{}` → 422; read-only router → 404. `pane.innerHTML = data.html` is therefore server-sanitised output.
- **Hooks under a prefix**: `data-api`/`editor.js` come from `url_for`; `data-after="0"` survives the `!== ""` check and `ge=0`; `/log/static/../pyproject.toml` → 404.
- **Discard**: never built for `.editing` forms (`:290`), `discard()` guarded on `dataset.entry`; no template composer carries a preset `data-entry`, so it cannot reach a real entry.
- **Idempotence**: `mount` (`dataset.mounted`) and `buildToolbar` (`.toolbar` check) guard re-entry; the template's remaining inline script has no leftover submit/keydown/click handlers.
- **`rowsToMarkdown`**: ragged rows padded to max width, `|` escaped, newlines collapsed; single-line TSV → null.
- **`busy`**: only set in `uploadFiles`, always cleared in `finally`.
- **Redundancy search**: `ScanAnalysis/scan_analysis/config_editor/static/editor.js:40` has its own ~10-line `api()` and a `/api/preview` fetch, and `GEECS-DataPortal/geecs_portal/templates/run.html:604,834` inline `fetch`+`detail` handling — different packages, different API roots, and GeecsWebTheme ships only `theme.js`/`theme-boot.js` (no shared JS kit), so extracting a shared helper is not warranted. `geecs.author`/`clipboardData`/`DOMParser`/`setRangeText` appear nowhere else in the repo.
- **Invariants**: no scan-folder creation touched; no facility literals; version bump + CHANGELOG present.

---

## Disposition (8caf91f2)

| # | Disposition |
|---|---|
| 1 | **Fixed.** `insertBlock(ta, text, select)`: attachment links insert with the caret after them; only the table and callout skeletons stay selected. |
| 2 | **Fixed.** One create per form — the pending POST promise lives on the form and every caller awaits it; `ev.repeat` keydowns ignored; `save`/`discard` run through an `exclusive()` gate that disables the Save/Discard buttons meanwhile. |
| 3 | **Fixed.** The version re-read runs in `finally` whenever an entry id exists, so uploads that landed before a failure are accounted for. |
| 4 | **Fixed.** `ensureEntry` requires a name (`authorName()` focuses the field and says so); no more "unknown". |
| 5 | **Fixed.** `noteVersion()` writes the fresh version to the form *and* to the article for an in-place edit, so Cancel-then-Edit starts current. |
| 6 | **Fixed** (precedence): an HTML table on the clipboard wins over a file item of the same cells; a file pasted alongside plain text keeps the browser's text paste and uploads the file. **The Excel-on-Windows hand check is owed** — the owner tested on a Mac; noted for the soak week. |
| 7 | **Fixed.** A range needs a consistent column count ≥ 2 and a non-empty header cell; tab-indented prose is left alone. |
| 8 | **Fixed.** Caret at 0 and a selection ending after `\n` handled; no `IndexSizeError`. |
| 9 | **Fixed.** The pane is created hidden and shown only after a successful render. |
| 10 | **Fixed.** `editor.js` is in the guard's `_SURFACES` (105 theme tests); the ⌘↩ entity that the guard read as `#8984` is a glyph now. |
| 11 | **Fixed.** Accepted types come from the page (`data-accept`, sourced from `ATTACHMENT_TYPES`); a dropped file of another type gets a message naming the accepted ones. |
| Nits | **Fixed.** Dead `entry_id` field removed; import merged; an emptied textarea no longer PATCHes an empty body. |
| Owner's nit | **Fixed.** Between-scan blocks are `<details>` like the scans (header shows the count; "+ note between" opens one; Expand/Collapse all reaches them) — they could be opened but not closed. |

Counts: GeecsLogbook 151, GeecsWebTheme 105.

skbarber:
## Verification (same reviewer, on 8caf91f2) — and the residuals' disposition

Reviewer's verdict, verbatim:

---

## Verification of 8caf91f2 against the #837 review

Environment: worktree at `8caf91f2`; GeecsLogbook 151 passed, GeecsWebTheme 105 passed, `node --check` clean; the smoke server serves `editor.js` byte-identical to the worktree. **Caveat on the smoke server:** its Python process is still the pre-fix module — the template is re-read (the `<details class="between">` and `data-accept` attribute appear) but `routes/day.py` is not reloaded, so the page renders `data-accept=""` and the script falls back to its hard-coded list. Item 11 is therefore verified by the test only, not by curl.

### Per-finding

| # | Finding | Verdict | Evidence |
|---|---|---|---|
| 1 | Uploaded link left selected | **Confirmed fixed** | Harness on the verbatim `insertBlock(ta, text, select)`: two uploads → `"note\n\na\n\nb\n"`, caret collapsed at end; typing after an upload → `"…a\nx"`; table/callout still selected. |
| 2 | No in-flight guard | **Confirmed fixed** | `form._creating` memo cleared in `finally`, `ev.repeat` ignored, `exclusive()` gates `save`/`discard` and disables submit/discard buttons. Residual race noted below (R1). |
| 3 | Partial batch skips version refresh | **Confirmed fixed** | Refresh in `finally` when `id` is set; `id` is only set after `ensureEntry` resolves, so a silent name rejection does no GET. |
| 4 | Autosave author "unknown" | **Confirmed fixed** | `authorName()` required in `ensureEntry`; rejection is `silent`, so exactly one message; `clearFail` runs before the name check, so the message is not wiped. |
| 5 | Cancel-then-Edit starts stale | **Confirmed fixed** | `noteVersion()` writes the enclosing `<article>` when it is the same entry; new composers are siblings of articles, so no cross-write. |
| 6 | Paste precedence | **Confirmed fixed for the ≥2-row case** | HTML table first, then files with default text paste preserved when `text/plain` exists, then TSV. Residual R2 below. |
| 7 | Tab-indented text becomes a table | **Confirmed fixed** | `"\tfoo\n\tbar"` → null, mixed width → null, real 2×3 → rows, CRLF → rows. Trade-off noted (R3). |
| 8 | `prefixLines` edges | **Confirmed fixed** | Caret at 0 with leading `\n` → no throw; triple-clicked line → only that line; other cases correct. |
| 9 | Preview pane shown on failure | **Confirmed fixed** | Pane created hidden, shown only after a successful render. |
| 10 | Theme guard omits `editor.js` | **Confirmed fixed, and the pin is real** | `_offences()` on the real file → `[]`, on a scratch copy with `"#ff0000"` appended → one offence. |
| 11 | `IMAGE_TYPES` duplicates server list | **Confirmed fixed (by test)** | `day.py` hands `sorted(ATTACHMENT_TYPES)` as `data-accept`; test pins the exact string; unsupported files now produce a message. |
| Nits | dead `entry_id`, import order, empty-body edit | **Confirmed** | |
| Owner's nit | collapsible between-scan blocks | **Confirmed, no stale selectors** | `.between > summary` styles, `applyAll` includes `details.between`, "+ note between" sets `open`. |

### Residuals (not regressions of the original review; none blocking)

**R1 — LOW: Save is not gated on an in-flight upload.** Drop a large PDF, then ⌘Enter while it is uploading: `save` PATCHes with the pre-upload version — either a spurious 409 or a reload before the link lands. Fix: have `save` await the upload.

**R2 — LOW, conditional on the Excel-bitmap behaviour: a 1-row Excel copy.** `htmlTableToRows` returns null for fewer than two rows, so the file branch runs and the bitmap of the cells is uploaded. Fix: if `text/html` contains a `<table>` at all, never treat file items as an upload.

**R3 — LOW, design trade-off:** the non-empty-corner rule rejects a text-only cross-tab with a blank corner cell. Mention in the CLAUDE.md "The editor" paragraph if kept.

**R4 — nit:** after an autosave, Save with an emptied textarea silently focuses the textarea instead of hinting that Discard is the way out.

### Also checked, clean
- No tracked-file edits made; probes confined to the scratchpad and `127.0.0.1:8299`.
- `discard` still never reachable from `.editing` forms; `data-after="0"` path unchanged.

---

## Disposition of the residuals (136ccc0e)

All four taken: **R1** — the upload keeps a promise on the form and `exclusive()` awaits it before any Save/Discard; **R2** — a clipboard whose `text/html` contains a `<table>` at all never uploads its file items (the default text paste goes through); **R3** — the rule is documented in CLAUDE.md "The editor"; **R4** — an emptied autosaved entry says "Nothing to save — Discard removes this entry." The smoke-server caveat is understood (the Python process predates `data-accept`); the pinned test covers it.

Review resolved: no undispositioned finding. Merging into `feature/scan-logger` on green CI for 136ccc0e, per the owner's standing ruling for phase PRs on this branch. GeecsLogbook 151, GeecsWebTheme 105.

Issue #840, opened 2026-09-12
Strict mode: deliberate rep-rate throttle (one shot per N seconds)


skbarber:
Sam, 2026-09-12 (raised while answering the phase-2 design questions, `Planning/native_bluesky/08_gated_batch.md` §6): operators sometimes want a strict scan deliberately slower than the laser, e.g. one shot every 5 s, to let a slow device settle or to limit dose.

Today the strict per-shot loop fires as soon as every essential detector is re-armed, so the rep rate is the laser's (M7: 1 Hz through the manager at short exposures). There is no knob to hold it below that.

Proposed shape: one keyword on the bound scan verbs beside `trigger_profile` / `shots_per_step` / `acquisition`, e.g. `shot_period: float | None = None` (seconds between fires; `None` = as fast as the shot allows), implemented as a `bps.sleep` for the remainder of the period at the top of each strict shot. Strict only: in gated mode the box free-runs at the laser rate by design. Rides in the start document like the other GEECS keywords so the s-file / ScanInfo record it.

Not part of phase 2's PRs (2a/2b/2c); small enough to land alongside 2b or after it.

skbarber:
Delivered as **`shot_period`** in **#850** (phase 2b, GeecsBluesky 0.83.0), merged into `feature/native-bluesky-plans` 2026-09-12.

`shot_period` is a keyword on every bound scan verb: seconds between fires, enforced by the strict per-shot hook. Two deliberate constraints:

- It is **strict-only**. Passing it with `acquisition="gated"` is refused at bind time, because a gated batch runs at the box's own rate — throttling it would be a contradiction rather than a setting.
- The wait happens **before the detectors are triggered**. Putting it after was the bug 2b's hardware acceptance (A8) found: the triggered count/stamp wait times the shot out during the pause. Pinned twice in the suite, once for a native camera's stamp wait and once for a plugin camera's count wait, because the plugin-only form passed on the pre-A8 code.

Hardware-accepted as A8 on Scan010 of 26_0912. On the feature branch; reaches master with that arc.

PR #841, opened 2026-09-12
Planning: phase-2 design answers recorded (08 §6 → 03 §10.9)

skbarber:
Docs only. Records Sam's 2026-09-12 answers to the six open questions of `Planning/native_bluesky/08_gated_batch.md` §6 in `03_clean_room_rebuild.md` §10.9 (its staleness rule) and amends 08 to match.

What changed in the design, in one screen:

- **Non-plugin devices in a gated run → a per-shot sampler** (new 08 §4.7), not baseline and not monitors. Most devices are not triggered and publish at ~5 Hz (monitors would flood); baseline would lose every per-shot scalar (ICT, energy meters). The sampler is a small worker-side device (Flyable + EventCollectable, verified against installed source in 2b) clocked by an essential triggered device's `acq_timestamp`; it yields one event per shot into a `shots` stream with the latest value of every non-plugin subscribed signal plus the motors' readbacks and `bin_number`. It also counts, so a gated run needs one essential *triggered* device (camera or scalar), not a camera in particular; a run with none is refused at preflight.
- **The per-step `steps` stream is gone**: the sampler's per-shot row carries the motors and the bin.
- **Pause**: the deferred pause is a real pause between steps; an immediate pause mid-batch repeats the interrupted step on resume (was "fail the step").
- **S-file join**: one row per essential shot; orphan non-essential frames stay in the stack and Tiled, out of the s-file (was "a row of their own").
- Confirmed as recommended: "N frames each" as the gated meaning of essential; the DB subscribed list rides in the stack (PR 2a's assumption); `acquisition="gated"` as a keyword.
- Filed alongside: #840 (strict-mode rep-rate throttle).

Also touched in 03: §4.A (pause → the plan repeats the step), §4.B (gated paragraph names the sampler), §8 item 2.

No code, no version bumps. PR 2a (the plugin's per-frame scalar attributes) follows on its own branch.

### DIFF
--- a/Planning/native_bluesky/03_clean_room_rebuild.md
+++ b/Planning/native_bluesky/03_clean_room_rebuild.md
@@ -207,8 +207,8 @@ Every DB device is one long-lived noun (#808, kept). Two classes:
   abstraction, §11.1), `Pausable` (state-dependent, §10.3: in strict mode
   the RE pausing simply stops the plan firing and the box stays ARMED —
   `pause()` does nothing; in gated mode edges flow on their own, so
-  `pause() → OFF` and `resume()` restores SCAN, and the plan then
-  repeats the interrupted step, §10.9;
+  `pause() → OFF` and `resume()` restores SCAN, and after an *immediate*
+  pause the plan repeats the interrupted step, §10.9;
   the RE calls these on every Pausable it has seen in a message, §7).
   **Not a flyer** (amended 2026-09-11, `08_gated_batch.md` §3): the box
   has no counter, so a `complete` of its own could not know when N shots
@@ -760,8 +760,8 @@ Still open, for Sam:
    not the 1 Hz mode; phase 2's gated batch is.  Recorded here so the
    cadence fix is scoped as "gated batch", not "a faster fire".
 9. ~~Phase-2 design awaiting Sam's answers to `08_gated_batch.md` §6~~
-   **Answered 2026-09-12 (Sam); `08` amended to match (its §4.2, §4.5,
-   §4.7, §5, §6).**  Two amendments already taken as read before:
+   **Answered 2026-09-12 (Sam); `08` amended to match (its §1, §3, §4.2,
+   §4.3, §4.5, §4.6, §4.7, §5, §6).**  Two amendments already taken as read before:
    the box is not a flyer (§4.A) and the non-essential list is per plan
    (§4.B).  The answers:
    - **Q1 — non-plugin devices in a gated run: a per-shot sampler, not
--- a/Planning/native_bluesky/03_clean_room_rebuild.md
+++ b/Planning/native_bluesky/03_clean_room_rebuild.md
@@ -207,8 +207,10 @@ Every DB device is one long-lived noun (#808, kept). Two classes:
   abstraction, §11.1), `Pausable` (state-dependent, §10.3: in strict mode
   the RE pausing simply stops the plan firing and the box stays ARMED —
   `pause()` does nothing; in gated mode edges flow on their own, so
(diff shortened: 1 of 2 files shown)


skbarber:
**Fresh-context review (docs consistency, 08 §4.7 claims checked against the installed bluesky 1.15.0 / ophyd-async 0.19.3 source) — 13 findings, all fixed in 38cb3263.**

1. **P1** 08 §4.3 still said every essential detector of a gated run must be plugin-backed, contradicting the new §4.7 (an ICT can be the clock) and §5's acceptance. → Fixed: the plugin requirement is restricted to the non-essential list and to the essential *cameras* of a gated run; an essential triggered scalar device is legal and rides in the sampler; a run with no essential triggered device is refused.
2. **P1** 08 §4.6's preflight rule had the same contradiction. → Fixed: hdf-child rule for non-essential + essential image devices when gated, plus the shot-clock rule.
3. **P1** 08 §4.2's lead-in defined *D* as "all plugin-backed" and never introduced *S*. → Fixed: *D* = the plugin-backed subset of the essential detectors, *S* = the sampler carrying every essential non-plugin device and every non-plugin device of the run.
4. **P2** The pseudo-code declared/collected `*D` unconditionally while the text allowed an empty *D*. → Marked "D non-empty" on both lines.
5. **P3** §1 "Two things" above a four-item list. → "Three things".
6. **P3** 03 §4.A stated the repeat-the-step unconditionally. → "after an *immediate* pause".
7. **P3** 03 §10.9's list of amended 08 sections was incomplete. → Lists §1, §3, §4.2, §4.3, §4.5, §4.6, §4.7, §5, §6.
8. **P3** `declare_stream` on a collectable needs `collect=True` (bundlers.py:324-345) and `Msg("prepare", obj, value)` takes one positional. → Pseudo-code uses `collect=True` on both declares and `prepare(S, SamplerInfo(quota=…))`.
9. **P3** "the monitor cache the signals already hold" — `SignalR._cache` exists only while staged/subscribed. → Reworded.
10. **P3** §4.7 Columns sentence flipped verb mid-clause. → Split.
11. **P3** §6 Q1 "and no camera in particular" read as a negation. → "— not a camera in particular".
12. **P3** `prepare(quota=shots_per_step)` vs a gated `count`'s `num`. → "the step's quota (`shots_per_step`, or `num` for `count`)".
13. **P3** Status paragraph line wrap. → Re-wrapped.

Verified by the reviewer with no change needed: `complete` on multiple objects is `bps.complete_all`; a single EventCollectable collected with `name=` is supported and requires the prior `declare_stream` the pseudo-code has; no remaining `steps`-stream / "fail the step" / "row of its own" / baseline-only text in 08; 03 §4.B, §8 and §10.9 agree with 08; no numbered-list break in 03 §10.

skbarber:
**Codex review — 2 findings, both fixed in 7521dced.**

1. **P1 — `ShotControl.resume()` restores SCAN before the repeated step is re-prepared.** Correct: the RE calls `resume()` on every Pausable before it hands control back to the plan, so edges would flow into a step whose plugin baseline is stale. → The gated step now **opens with `mv(B, OFF)`** unconditionally (idempotent in the normal path), and the repeat path drains and rewinds to the step baseline only after it (08 §4.2 pseudo-code + prose, the Pause bullet; 03 §4.A says `resume()` restores the standing state and the plan re-establishes OFF itself). Any edge that slips in between the resume and the OFF is rewound with the partial frames.
2. **P2 — the pseudo-code required a non-empty *D* while scalar-only gated runs are legal.** → The pseudo-code branches on `if D:` for the prepare, `declare_stream(primary)`, `kickoff`, the drain/truncate and `collect(primary)`; the sampler's lines are unconditional, and the prose states that an empty *D* declares and collects no `primary`.

Issue #839, opened 2026-09-12
GEECS-Data-Utils: review ScanPaths / ScanData and extract their pure parts into path-free utilities


skbarber:
Owed at the logbook arc's merge to master (`GeecsLogbook/CLAUDE.md` § "Deferred decisions"), filed now so it is not lost.

## Why

During review of the logbook's phase 01 (#825) the reviewer argued that `geecs_logbook.scan_reader` — walking a day's `ScanNNN` folders, parsing `ScanInfoScanNNN.ini`, classifying status — is pure logic that belongs in `geecs_data_utils`, next to `ScanPaths`, `scans_database/` and `scan_log_loader`. The owner agreed with the direction but not the conclusion:

> the data-utils package in this context is quite weak — fairly convoluted and confusing. A simple 'scan info' reader that doesn't care about paths seems like the right thing.

So the shape wanted is a **pure** reader with no path-object ceremony, and moving `scan_reader` under `ScanPaths` today would bind it to the design that needs fixing. The parse itself was shared at the time (`scan_paths.read_scan_info_file`, `scan_log_loader.first_log_timestamp`, Data-Utils 0.27.0); the day-walking and the summary model were duplicated on purpose, **on the condition that this review removes the duplication later**. This issue is that condition.

## Scope

- Review `geecs_data_utils.scan_paths.ScanPaths` / `ScanData`: which parts are path resolution (the `Y{YYYY}/{MM-Mon}/{YY_MMDD}/scans/ScanNNN` layout, the config-driven base path), which are file parsing (ScanInfo INI, s-files, TDMS), which are creation (`read_mode=False`, scanner-side only), and which are convenience wrappers that hide all three.
- Extract the pure parts into path-free utilities: functions that take a `Path` (or file text) and return a model, with no `GeecsPathsConfig` in the call chain. `geecs_logbook/scan_reader.py` is a working sketch: one `os.scandir` per scan, a cache keyed on the ScanInfo file's `(mtime, size)`, status classification from `ScanEndInfo`, a `ScanSummary` model.
- Reconcile the overlapping models: the logbook's `ScanSummary` vs `scans_database.entries.ScanMetadata` (also on the deferred list).
[...]

PR #842, opened 2026-09-12
GeecsLogbook 0.5.0 / Portal 0.25.0: the ops book — month page, seed templates as type buttons, cross-links; portal memory ceiling (#834)

### LINKED ISSUE #834: Data portal reached 5.3 GB RSS and was the host's first OOM casualty — bound the process, not just the pixel cache

skbarber:
Phase 05 of the logbook arc (into `feature/scan-logger`; follows #825, #831, #832, #835, #837). After this PR the **ops book is functionally complete**: the second book gets its page, entries get types, and the two books point at each other.

## What

**GeecsLogbook 0.5.0**

- `GET /log/month/{YYYY-MM}` — the ops book. Every `ops` entry in the month grouped by day, newest day first; one composer with a date picker (defaults to today when the month is current, else the month's edge; "+ note" on a day group points it at that day); prev/next month + a month picker; tag chips as URL filters (`?tag=laser`) — chips count the whole month while the list shows the filtered part. **Reads only the notes store, never the share** (pinned: the test makes the share reader explode). `GET /log/api/month/{YYYY-MM}/entries?book=&tag=` is the JSON peer.
- **Seed templates as type buttons** (`seed_templates.py`): `*.md` files in a directory the host names (`create_log_router(templates_dir=)`). Front-matter `label` / `colour` / `book` / `order`, body = the prefill carrying its `#tag`. A button press inserts the prefill and the entry records the template's name (provenance only); a stored entry shows its template as a chip in the file's tone. `colour` is a **theme token name** from a closed vocabulary (`TONES`, pinned against `scanlog.css` in both directions; unknown → `accent` + warning), so the configs repo can never put a literal colour on a page. Loaded once at start, refreshed in a background thread when stale (60 s); a failed refresh keeps the last set — the share stays off every page's critical path. `examples/logbook_templates/` is the documented starter set (5 files + README).
- Cross-links: the day page's "N ops notes today →" strip (`ops_count` re-added, links into the month page's day heading) and an "Ops book" topbar link; each day group on the month page links to its day document; "Scan log →" in the month topbar.
- Entry + composer markup moved into shared macros (`templates/_entries.html`) so both books draw them identically; `editor.js` takes the day per form (`.when` input / `data-day` / page), reads prefills from one JSON block, sends `template` on create.
- `CLAUDE.md`: the owed ScanPaths review is now "filed as #839" (was "OWED at master-merge"); templates section documents the format and the vocabulary; month page section.

**GEECS-DataPortal 0.25.0**

- The logbook's templates directory is derived, not flagged: `<configs>/logbook_templates/`, the parent of `--processing-configs` (the checkout the portal already reads). Without that tree the composers are plain.
- **Memory ceiling (#834 tail, Sam: yes 2026-09-11):** `MemoryHigh=@PORTAL_MEMORY_HIGH@` / `MemoryMax=@PORTAL_MEMORY_MAX@` in the unit template, rendered by `render_units.sh` from two new **required** site.env keys `GEECS_PORTAL_MEMORY_HIGH` / `GEECS_PORTAL_MEMORY_MAX` (3G / 4G in the example for the 16 GB reference host). `bootstrap_host.sh` requires them too, so a profile missing them fails at render rather than at 03:00. `DEPLOYMENT.md` documents the shape and the `systemctl set-property` shortcut; `site_profile.md` lists the placeholders.

**GeecsWebTheme** (tests only, no version bump): `month.html` + `_entries.html` added to the guarded surfaces; the referenced-token test now accepts a surface-local indirection (`.tone-ok{--tone:var(--ok)}`) as defined *only when its value is itself a token* — a local property holding a literal is still caught by the literal guard.

## Per-concern LOC

| Concern | ± lines |
|---|---|
| GeecsLogbook (month route, seeds, templates, editor, CSS, tests, docs, examples) | 1546 |
| GEECS-DataPortal code + tests + changelog + CLAUDE/DEPLOYMENT | 124 |
| Unit template + site.env.example + render/bootstrap + site_profile.md (memory ceiling) | 37 |
| GeecsWebTheme guard | 8 |

The memory-ceiling ride-along is a separate concern bundled by the owner's prior yes; it is the 37-line block and is cheap to veto.

- **Memory keys are required**, not defaulted — matches the site-profile contract ("every runtime/install-time key is required"). Consequence: the running host's `/etc/geecs/site.env` needs the two lines before its next re-render of *any* unit.
- **Templates load synchronously at construction**, then in the background. If the share is dead at portal start, startup blocks on one directory listing (same exposure the day page already has per request).
- **README.md is skipped silently** in the templates directory (a browsed directory has one); any other unusable stem warns.
- **Last type button pressed wins** as the recorded `template` (provenance only; recorded in the deferred table).
- `?book=scans` on the month JSON returns scan-anchored entries too (it is the raw query); the month *page* is ops-only by design.

## Facility-literal check

No new lab literals outside `site.env.example` (3G/4G are host sizes, labelled as the reference deployment) and docstrings.

## Tests

- GeecsLogbook: **181** (was 151; +30: `test_seed_templates.py` 14, `test_month.py` 16)
- GEECS-DataPortal: **256** (+2 in `test_scan_log_mount.py`)
- GeecsWebTheme: **107** (+1 surface parametrisation)
- root `tests/test_render_units_sh.py` + `test_bootstrap_host_sh.py`: 10
- `scripts/check.sh --base origin/feature/scan-logger`: lint + all three suites green.

## Hardware verification

No device or scan-execution surface. **Smoke-tested** on a local server over a temporary share (never the real one): month page with four seeded ops entries across three days, tag chips, type buttons per book, tone chips on stored entries, the day page's ops strip — headless-Chrome screenshots inspected. **OWED at deploy**: the host's `site.env` gets the two memory keys; the configs repo gets `logbook_templates/` (copy the examples); `systemctl status geecs-data-portal` shows `Memory:` against the cap. Close #834 on merge to master.

### DIFF
--- a/GEECS-DataPortal/CHANGELOG.md
+++ b/GEECS-DataPortal/CHANGELOG.md
@@ -3,6 +3,26 @@
 All notable changes to this package will be documented here.
 Format follows [Keep a Changelog](https://keepachangelog.com/en/1.1.0/).
 
+## [0.25.0] - 2026-09-11
+
+### Added
+
+- The logbook's seed templates (its type buttons) are read from
+  `logbook_templates/` beside the `--processing-configs` tree — the
+  configs checkout the portal already has; no new flag. Without that
+  tree the composers are plain.
+- Unit template: `MemoryHigh=` / `MemoryMax=` on the portal service,
+  rendered from two new **required** site.env keys
+  `GEECS_PORTAL_MEMORY_HIGH` / `GEECS_PORTAL_MEMORY_MAX` (3G / 4G in the
+  example, sized per host), so a runaway portal is throttled and then
+  restarted by systemd instead of being the kernel's first OOM pick on
+  the shared box (#834). Existing hosts add the two keys and re-render.
+
+### Changed
+
+- `/log/month/{YYYY-MM}` (the ops book) is now served under `--scan-log`
+  alongside the day pages (GeecsLogbook 0.5.0).
+
 ## [0.24.1] - 2026-09-11
 
 ### Changed
--- a/GEECS-DataPortal/CLAUDE.md
+++ b/GEECS-DataPortal/CLAUDE.md
@@ -154,7 +154,9 @@ geecs_portal/
   static/        # the vendored Plotly bundle (the ONE committed JS asset)
   __main__.py    # CLI (geecs-data-portal): real TiledScanCatalog + uvicorn;
                  #   --config-editor mounts scan_analysis.config_editor at /configs
-                 #   --scan-log     mounts geecs_logbook at /log (needs --experiment)
+                 #   --scan-log     mounts geecs_logbook at /log (needs --experiment;
+                 #                  type buttons from <configs>/logbook_templates/,
+                 #                  the parent of --processing-configs)
                  #   --notes-db     the logbook's SQLite file; makes /log writable
                  #                  (default: $STATE_DIRECTORY/logbook.db under systemd)
   templates/     # base.html / day.html / run.html (Jinja2; every colour a GeecsWebTheme token)
@@ -166,7 +168,9 @@ tests/
 ```
 
 Routes: `/` (redirect to today) · `/day/{iso}` (run list; `?experiment=`)
-· `/log/day/{iso}` + `/log/api/day/{iso}` (the scan logbook, `--scan-log`)
+· `/log/day/{iso}` + `/log/api/day/{iso}` (the scans book, `--scan-log`)
+· `/log/month/{YYYY-MM}` + `/log/api/month/{YYYY-MM}/entries` (the ops book,
+same flag; store-only, never the share)
 · `/log/api/entries…` (the logbook's entry writes, only with `--notes-db`)
 · `/run/{uid}` (the scan page: rail + Overview/Plot/Images/Analysis tabs
 — Analysis only when runs are possible, see below;
--- a/GEECS-DataPortal/DEPLOYMENT.md
+++ b/GEECS-DataPortal/DEPLOYMENT.md
@@ -106,8 +106,21 @@ host's `site.env` with `deploy/render_units.sh` (or let
 `deploy/bootstrap_host.sh` do the whole host), then install the rendered
 unit and `enable --now` it — see the
 [Site Profile](../docs/platform/site_profile.md). The account, checkout
-root, poetry path, experiment, and timezone all come from `site.env`;
-nothing site-specific is typed into the unit by hand.
+root, poetry path, experiment, timezone, and the memory ceiling all come
+from `site.env`; nothing site-specific is typed into the unit by hand.
+
+**Memory ceiling.** The rendered unit carries `MemoryHigh=` and
+`MemoryMax=` from `GEECS_PORTAL_MEMORY_HIGH` / `GEECS_PORTAL_MEMORY_MAX`
+(both required; 3G / 4G in the example for a 16 GB host shared with
+Tiled, MySQL and the queueserver). Above HIGH systemd throttles and
+reclaims the portal; at MAX it kills it and `Restart=on-failure` brings
+it back — so a runaway portal evicts itself before the kernel picks a
+victim, and the victim is the portal rather than Tiled (#834). To change
+the numbers: edit `site.env`, re-render, `daemon-reload`, restart. For a
+quick change without a re-render, `sudo systemctl set-property
+geecs-data-portal MemoryMax=6G` writes a persistent drop-in;
+`systemctl revert geecs-data-portal` removes it. `systemctl status`
+shows the current `Memory:` line against the cap.
 
 Verify:
 
@@ -226,6 +239,19 @@ Two requirements, or it warn-and-skips rather than serving a broken page:
 Set it on the host through `GEECS_PORTAL_EXTRA_ARGS` in `site.env`
 alongside `--config-editor`; `deploy/bootstrap_host.sh` already installs
 the extra for the portal service.
+
+**Two books.** `/log/day/…` is the scans book; `/log/month/2026-09` is
+the ops book — day-level notes read by month, from the database alone
+(it never touches the share, so it stays fast when the share is slow).
+
+**Type buttons** on every composer are seed templates: `*.md` files in
+`logbook_templates/` at the top of the configs checkout — the parent of
+the `--processing-configs` tree, so the same `GEECS_CONFIGS_ROOT` serves
+both. Copy `GeecsLogbook/examples/logbook_templates/` there to start
+(commit it to the configs repo); each file's header names its label,
+colour (a theme token name), book and order, and its body is the
+prefill. Adding a file adds a button within a minute, no restart. Without
+the directory the composers are plain.
 
 Reading a day of ~100 scans off a VPN-mounted share takes a few seconds
 cold and milliseconds thereafter — per-scan summaries are cached on the
--- a/GEECS-DataPortal/deploy/geecs-data-portal.service
+++ b/GEECS-DataPortal/deploy/geecs-data-portal.service
@@ -54,6 +54,17 @@ ExecStart=@POETRY@ run geecs-data-portal --experiment ${GEECS_EXPERIMENT} --proc
 # here rather than in site.env. Back it up: what people wrote exists in
 # this file and in the markdown mirror on the share, nowhere else.
 StateDirectory=geecs-data-portal
+# Memory ceiling (#834): the portal shares the box with Tiled, MySQL and
+# the queueserver, and as the largest resident process it was the
+# kernel's first OOM pick. MemoryHigh= throttles and reclaims above the
+# soft line; MemoryMax= is the hard cap, at which systemd kills and
+# Restart= brings it back in seconds — the portal, not Tiled, pays for a
+# runaway. Sized per host in site.env (GEECS_PORTAL_MEMORY_HIGH/MAX;
+# systemd cannot expand ${VAR} here, so the render fills them). To change
+# without a re-render: `systemctl set-property geecs-data-portal
+# MemoryMax=4G` writes a persistent drop-in (`systemctl revert` removes).
+MemoryHigh=@PORTAL_MEMORY_HIGH@
+MemoryMax=@PORTAL_MEMORY_MAX@
 Restart=on-failure
 RestartSec=10
 # An analysis run in flight cannot be interrupted (DEPLOYMENT.md): give a
--- a/GEECS-DataPortal/geecs_portal/app.py
+++ b/GEECS-DataPortal/geecs_portal/app.py
@@ -342,6 +342,11 @@ class _DiagInfo:
         )
 
 
+#: The logbook's seed templates, a directory beside the analysis tree in
+#: the configs checkout (``<configs>/logbook_templates/*.md``).
+LOGBOOK_TEMPLATES_DIR = "logbook_templates"
+
+
 def create_app(
     catalog: ScanCatalog,
     *,
@@ -392,6 +397,10 @@ def create_app(
         each entry as markdown into the day's ``logbook/`` folder on the
         share (a sibling of ``scans/``; never inside it). Without it the
         logbook is the read-only day view. Ignored unless ``scan_log``.
+        The logbook's seed templates — its type buttons — are read from
+        ``logbook_templates/`` beside the ``processing_config_dir`` tree
+        (the configs checkout this portal already reads); none when that
+        tree is not configured.
     config_editor : bool, default False
         Mount the analysis config editor (``scan_analysis.config_editor``)
         at ``/configs`` over the same ``processing_config_dir`` tree, with a
@@ -1919,14 +1928,26 @@ def create_app(
             except ImportError as exc:  # the log extra is not installed
                 logger.warning("scan log requested but not installed: %s", exc)
             else:
+                # Type buttons are markdown files in the configs checkout,
+                # a sibling of the analysis tree: one checkout, one flag.
+                templates_dir = (
+                    Path(processing_config_dir).parent / LOGBOOK_TEMPLATES_DIR
+                    if processing_config_dir
+                    else None
+                )
                 app.include_router(
-                    create_log_router(default_experiment, notes_db=notes_db),
+                    create_log_router(
+                        default_experiment,
+                        notes_db=notes_db,
+                        templates_dir=templates_dir,
+                    ),
                     prefix="/log",
                 )
                 logger.info(
-                    "scan log mounted at /log for %s (%s)",
+                    "scan log mounted at /log for %s (%s; templates %s)",
                     default_experiment,
                     f"entries in {notes_db}" if notes_db else "read-only",
+                    templates_dir or "none",
                 )
 
     return app
--- a/GEECS-DataPortal/geecs_portal/app.py
+++ b/GEECS-DataPortal/geecs_portal/app.py
@@ -342,11 +342,6 @@ class _DiagInfo:
         )
 
 
-#: The logbook's seed templates, a directory beside the analysis tree in
-#: the configs checkout (``<configs>/logbook_templates/*.md``).
-LOGBOOK_TEMPLATES_DIR = "logbook_templates"
-
-
 def create_app(
     catalog: ScanCatalog,
     *,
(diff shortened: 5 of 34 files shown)


skbarber:
# Adversarial review of #842

*(Fresh-context reviewer, run on the PR's first commit dd97dae8. Pasted verbatim; the disposition follows in the next comment.)*

Reviewed the full diff (`origin/feature/scan-logger...HEAD`, 34 files) against the store, mirror, entries route, editor script, deploy scripts and theme; ran the four suites (GeecsLogbook 181, GEECS-DataPortal 256, GeecsWebTheme 107, deploy 10 — all green) plus a TestClient probe with hand-built inputs (hostile template files, odd months, slow-share spy). Findings that survived my own refutation, by severity:

## 1. Medium — the month page puts the share on its critical path after all (`routes/month.py:_month_page`, the `ctx.maybe_sync()` call)

The PR, the module docstring, the CHANGELOG and the new CLAUDE.md section all pin "reads only the notes store, never the share … on a day the share is slow, this page is not". But `_month_page` calls `ctx.maybe_sync()`, which runs `mirror.sync` **inline on the request thread**: up to 100 × `mirror_one` → `logbook_root` (`experiment_dir.is_dir()` on the SMB mount) → `write_entry` / `mirror_attachments`, once per `SYNC_INTERVAL_S` (60 s).

Concrete scenario: the share goes slow or hangs → saves succeed in SQLite and the mirror debt grows (that is exactly when unmirrored rows exist) → the first month GET after each 60 s window blocks for the mount's stat/write timeout, which on a hung SMB mount is tens of seconds. Probe: with `mirror.sync` spied to sleep 0.3 s, `GET /log/month/2026-09` made one sync call and took 0.33 s.

`test_never_reads_the_share` would not catch this — it patches only `read_day`, and the fixture share is a local tmp dir, so the sync succeeds silently.

Fix: drop `ctx.maybe_sync()` from `_month_page` (the day page already pays the debt every 60 s and every write calls `ctx.mirror` itself), or hand the sync to a background thread the way `SeedTemplates` does; and extend the test to also monkeypatch `_common.mirror.sync` with the same `AssertionError`. Then the CLAUDE.md sentence "NotesStore.query, one call" becomes true (it is currently two store calls plus the sync).

## 2. Low/Medium — an attachment autosave pins the day before the picker is settled (`editor.js ensureEntry`/`dayOf`; `month.html` `.when` and `+ note`)

The month composer is the first form whose day is mutable after mount. `ensureEntry` reads `dayOf(form)` at the moment of the first paste/drop and POSTs; from then on Save is a PATCH, which cannot change `day`.

Concrete scenario: open `/log/month/2026-09`, paste a screenshot (autosave creates the entry on `compose_day` = today), then change the date to the 3rd (or press "+ note" on the 3rd's group, which just sets `.when`), type, Save → the note is stored on today, and the reload shows it under the wrong day with no message. The day page cannot hit this because its day is fixed.

Fix: once `form.dataset.entry` is set, disable the `.when` input (and make the `+ note` handler refuse or reset on an autosaved form), or have `save` compare `dayOf(form)` to the day the autosave used and fail with "this note is already saved on YYYY-MM-DD".

## 3. Low — a template file may shadow the three reserved provenance names (`seed_templates.load_templates`; `_entries.html template_chip`)

`template_chip` looks up `seeds.labels.get(entry.template)` first, and every hand-typed entry has `template == "blank"`. A file `blank.md` (or `scan_note.md` / `day_intro.md`, the other names the chip deliberately hides) in the configs repo is accepted by `_NAME` and gives every un-buttoned entry in both books a "Blank" chip. Probe confirmed: `blank.md` → `<span class="chip chip-type tone-accent">Blank Button</span>` on a hand-typed note.

Fix: refuse the reserved stems in `load_templates` with a warning, and make that reserved set one constant that the macro reads (it is currently a literal tuple inside `_entries.html`).

## 4. Low — `/log/month/9999-12` and `/log/month/0001-01` are 500s (`_common.parse_month` / `month_step`)

`parse_month` accepts any four-digit year, then `month_step(first, ±1)` builds `date(10000, 1, 1)` / `date(0, 12, 1)` → `ValueError` escapes as a 500 traceback (the JSON peer returns 200 because it never steps). Fix: bound the year in `parse_month` (400 outside 1–9998) or catch in `month_step`.

## 5. Low — `data-hay` triples every entry body on the day page, which never reads it (`_entries.html entry_article`)

The shared macro now emits `data-hay` with the lower-cased full body on every entry, but the day page's filter reads only `details.scan` (`day.html` ~L318). Measured: a 50 kB scans-book note appears three times in the day page (rendered HTML, `.raw` JSON, `data-hay`); bodies are allowed up to 200 kB, and a busy day can have dozens. Fix: emit the attribute only when the caller asks (`hay=true` from `month.html`), or build the haystack in the month page's filter script from `.raw` + the tag chips at filter time — no server-side copy at all.

## 6. Low (placement) — the `logbook_templates` directory name lives in the portal, its documentation in the logbook (`GEECS-DataPortal/geecs_portal/app.py:LOGBOOK_TEMPLATES_DIR`)

`GeecsLogbook/README.md`, `CLAUDE.md`, `examples/logbook_templates/README.md` and `DEPLOYMENT.md` all state the name, but the only code that makes it true is a portal constant. The next host that mounts `create_log_router` (a standalone logbook server, GEECS-MCP's agent verbs) has to re-spell it. Fix: `geecs_logbook.seed_templates.TEMPLATES_DIRNAME`, imported by the portal; the logbook docs then describe something the logbook owns.

## 7. Low (placement/redundancy) — a second front-matter dialect in the same package (`seed_templates.parse_template` vs `mirror.render`)

`mirror.render` writes `---\nkey: value\n---` and its docstring promises "a reader can parse back without a YAML library". This PR adds the package's first such reader, in a different module, with a slightly different grammar (no list values, `color`/`colour` aliasing folded into the parse). The off-site/rebuild reader in the deferred table will want the same function. Fix: a private `parse_front_matter(text) -> (dict[str, str], body)` in one module (beside `mirror` or `_fs`), which `parse_template` calls now.

## Nits (no action needed unless touching the file anyway)

- `month.py:month_entries(first, last, **filters)` — `filters` is never passed; dead parameter.
- `Context.page_seeds` returns a free-form `dict` where the root CLAUDE.md prefers a model (three fixed keys — a small `PageSeeds` model would also give the macros a typed contract).
- `create_app` docstring: the templates paragraph sits under the `notes_db` parameter rather than `processing_config_dir`, which is the input it actually derives from.

## Checked and not flagged

- **Escaping**: a template with label `<b>Evil</b> & co` and body `</script><script>alert(1)</script>` — `tojson` emits `<`, autoescape covers the chip and the button title; nothing raw reaches the page.
- **Tag filter**: `?tag=LASER` matches lower-cased tags; 33 chars → 422; empty → unfiltered; chip hrefs use tag-safe characters only (`tags._TAG`).
- **Name round-trip**: `_NAME` (≤ 64 chars) vs `EntryCreate.template` (`max_length=64`) agree; `Laser-Notes.md` becomes a button and POSTs back as `Laser-Notes`.
- **Month arithmetic**: prev/next across the year boundary, `2020-02-29`, `2026-00`/`2026-13`/`2026-9` → 400, trailing slash → 307.
- **Theme**: all ten `--tone` targets plus `--on-accent`, `--accent-soft`, `--surface-2`, `--slate`, `--r`, `--rule` are defined in `theme.css`; the guard's new local-indirection subtraction removes only `--tone`, so the ten targets are still checked; both new surfaces pass the literal guard.
- **`SeedTemplates`**: lock discipline, single in-flight refresh, failed refresh keeps the last set and retries after the interval; `README.md` skip; the synchronous first load at construction is the declared judgment call.
- **Deploy**: required-keys change is consistent with `site_profile.md` ("keep every key"); the two placeholders are substituted in `render_units.sh` and listed in the doc table; deploy tests read `site.env.example` so they exercise the new keys; the "existing hosts must add two lines before the next re-render, including `--only gateway`" consequence is stated in the PR.
- **Invariants**: no new scan-folder creation (the mirror tree under `{experiment}/logbook/` predates this PR and is `_assert_own_tree`-guarded); no lab literals beyond 3G/4G in the example; no facility default in `create_log_router`.
- **Redundancy searched**: no existing month helpers (`monthrange`/`relativedelta` grep across packages: only this PR); `geecs_portal/cache.py` is a scan-data cache, not a generic stale-while-revalidate helper; `Context(` is constructed only in `router.py`; `_entries.html` genuinely removes the duplicated article/composer markup from `day.html`.

skbarber:
# Disposition of the #842 review

Fix commit: 1cee3955 (findings 1–6 + nits). Verification by the **same reviewer** (pasted verbatim below), then one follow-up it asked for.

| # | Finding | Disposition |
|---|---|---|
| 1 | Month page ran the mirror sync on the request thread | **Fixed** 1cee3955 — `_month_page` no longer calls `ctx.maybe_sync()`; the day page pays the mirror debt. Pin: `test_never_reads_the_share` now patches `Context.maybe_sync` to raise (the reviewer's verification found my first pin, patching `mirror.sync`, was inert because `maybe_sync` swallows it) — e77172b6 + 8842947e (e77172b6 shipped with a mistaken day-page control assertion and was red; 8842947e removes it; suite 184 green). |
| 2 | Attachment autosave pinned the day before the picker settled | **Fixed** 1cee3955 — `.when` is disabled once the entry exists; `+ note` respects the lock; Discard reloads to a fresh picker. Browser-only path, no test (consistent with the rest of editor.js). |
| 3 | `blank.md` etc. would put a chip on every hand-typed entry | **Fixed** 1cee3955 — `RESERVED_NAMES`, refused by the loader with a warning; the macro reads the same constant via `PageSeeds.quiet`. Two tests. |
| 4 | `9999-12` / `0001-01` were 500s | **Fixed** 1cee3955 — `parse_month` bounds the year; both routes covered; test. |
| 5 | `data-hay` tripled every body on the day page | **Fixed** 1cee3955 — `hay=false` default, only the month page asks; test. |
| 6 | `logbook_templates` name lived in the portal | **Fixed** 1cee3955 — `seed_templates.TEMPLATES_DIRNAME`, imported by the portal inside the same ImportError guard. |
| 7 | Second front-matter dialect (reader here, writer in `mirror`) | **Waived** — one reader and one writer with different shapes today; extracting a shared `parse_front_matter` for a single caller is speculative. Recorded in the CLAUDE.md deferred table with its trigger (the mirror reader as second caller) and target. Reviewer: "reasonable". |
| nits | dead `**filters`, dict → model, docstring placement | **Fixed** 1cee3955. |

Tests after the round: GeecsLogbook **184** (was 181), GEECS-DataPortal 256, GeecsWebTheme 107, deploy 10. `scripts/check.sh --base origin/feature/scan-logger` rc=0.

---

# Verification of the #842 fixes

*(same reviewer, on 1cee3955)*

Worktree at `1cee3955` (`origin/phase/05-ops-book`). Suites: GeecsLogbook **184** passed (+3), GEECS-DataPortal **256**, GeecsWebTheme **107**, deploy **10**; `pre-commit run` on the commit's files: ruff / ruff-format / pydocstyle all pass. No `LOGBOOK_TEMPLATES_DIR` leftovers anywhere.

1. **CONFIRMED (fix) / pin is inert** — `GET /log/month/…` makes zero `mirror.sync` calls with `SYNC_INTERVAL_S=0` and mirror debt outstanding; the day page still pays it (1 call). But the new assertion in `test_never_reads_the_share` cannot fail: `Context.maybe_sync` wraps the sync in `except Exception` ("a sync must never take down a view"), so an `AssertionError` raised from the patched `mirror.sync` is logged as "mirror sync failed" and the page still renders 200 — I verified by calling `maybe_sync()` with the raising stub: it returned normally. If someone re-adds `ctx.maybe_sync()` to the month route, this test stays green. Fix: monkeypatch `_common.Context.maybe_sync` itself to raise (nothing swallows that), or assert `"mirror sync failed" not in caplog.text` after the GET.
2. **CONFIRMED** — `ensureEntry` disables `.when` (with title) once the POST returns; `month.html`'s `+ note` only writes `.when.value` when not disabled; Discard → `location.reload()` yields a fresh, enabled picker. No test (browser-only path), consistent with the rest of `editor.js`.
3. **CONFIRMED** — `blank.md` is refused with the warning, no button, and a hand-typed entry shows no chip; `template_chip` reads `seeds.quiet` from `PageSeeds` (default `RESERVED_NAMES`); tests `test_reserved_stems_are_refused` / `test_page_seeds_carry_the_reserved_names` present and passing. (The stem check is case-sensitive, but `Blank.md` would produce name `Blank`, which never matches the stored `blank`, so no collision — fine.)
4. **CONFIRMED** — `0001-01` → 400, `0002-01` → 200, `9998-12` → 200, `9999-12` → 400; both routes go through `parse_month`, so the JSON peer is bounded too.
5. **CONFIRMED** — `entry_article(…, hay=false)` default, `month.html` passes `hay=true`; the month page carries the haystack, the day page no longer does (`test_haystack_only_where_it_is_filtered` passes against the `share` fixture with Scan001).
6. **CONFIRMED** — `seed_templates.TEMPLATES_DIRNAME = "logbook_templates"`; the portal imports it inside the same `try/except ImportError` as `create_log_router`, so a missing `log` extra still degrades to the warning, not an `ImportError`. Portal mount tests pass.
- **Nits: CONFIRMED** — `**filters` gone; `PageSeeds` (pydantic) replaces the dict and `Context.page_seeds` delegates to `SeedTemplates.for_page` (one `current()` read instead of three); the templates paragraph now sits under `processing_config_dir` in the `create_app` docstring.
- **Waiver of 7: reasonable.** One reader and one writer with different shapes today; the CLAUDE.md deferred-table row names the trigger (the mirror reader as second caller) and the target (`parse_front_matter` beside `mirror`), which is the repo's "do not lose, do not re-litigate" pattern. No regression risk from waiting.

**No new regressions found.** The only open item is the inert pin in (1) — the behaviour is fixed, the guard just would not catch its return.

skbarber:
Post-review commit 9991d4dc from the owner's hand-test of the local build: an entry showed its type chip ("Laser") and the tag chip ("#laser") its prefill carries — the same thing twice. A template now records the tags its prefill carries (`SeedTemplate.tags`) and the entry omits those tag chips; a tag the author added, or typed with no type, keeps its chip. Pinned by two tests in `test_month.py` (185 logbook tests). Template-only change plus the model field; no route or store change, so not sent back to the reviewer.

skbarber:
Codex finding (P2, `editor.js` `applyType`/`ensureEntry`): after an attachment autosave a later type-button press showed a type the store did not hold, because `EntryUpdate` carries no `template`. **Fixed** cb753087, the freeze option: once the autosave returns, the type buttons disable (title names the saved type) alongside the date picker, and `applyType` refuses on a form that already has an entry; Discard reloads to a fresh composer. Browser-only path, no test, consistent with the date lock. `node --check` + theme guard (107) green.

PR #843, opened 2026-09-12
GeecsPvaGateway 0.9.0: subscribed scalars as per-frame stack attributes (phase 2 PR 2a)

skbarber:
Phase 2 PR 2a of the native-Bluesky rebuild (#807), per `Planning/native_bluesky/08_gated_batch.md` §4.4 and §5 item 1. Follows the docs PR #841 (Sam's §6 answers; Q3 — the subscribed list rides in the stack — is what this PR builds).

## What

The file plugin writes, beside each frame's two stamps, **one `DOUBLE` per-frame attribute per subscribed (`get='yes'`) scalar of the camera** under `/entry/instrument/NDAttributes/<device>-hdf-<variable>-<scalar>`, declared in `NDAttributesFile` so the stock `ADHDFDataLogic` describes each as a stream column — nothing changes on the worker's describe path. A gated row (phase 2b) therefore carries the same scalar columns for that device as a strict row does today.

- **GeecsPvaGateway 0.9.0**: `CameraSpec.scalar_variables` is built at startup from the subscribed list by `geecs_core.db.variable_types.scalar_attribute_variables` — numeric types only, minus the timestamp ladder the stamps already carry, names without a metadata row skipped, a second name normalizing onto an earlier one's dataset dropped with a warning; `file_plugin.attribute_names/attributes_xml` take the scalar list (descriptions XML-escaped); the values come from the **same TCP push as the frame** — `_CameraWorker.subscription_variables` widens the image variable's one subscription by the list, only where the plugin serves it; a missing key or a non-numeric value is written `NaN` (`scalar_value`); the constructor still refuses a name collision as the backstop; the file's root attributes record the raw scalar names. Tests: the stock-data-logic-over-`pva://` test now runs with two scalars, one omitted on the second shot (keys described as `<f8`, values `[4095, 4000]` / `[12.5, NaN]`, 5 stream resources); the subscription list with and without a plugin; the config roster from a fake DB, and the policy-query-fails path (frames and stamps only, one warning).
- **GEECS-Core 0.6.0**: `db.scalar_policy` — `GeecsDbScalarPolicy` + `ScalarPolicyProvider` **moved** from `geecs_bluesky.db_runtime` (the gateway depends on GEECS-Core alone; one home for the rule both sides build a row from), semantics unchanged, tests moved with it; `db.variable_types.scalar_attribute_variables` + `TIMESTAMP_LADDER`, the per-frame scalar filter beside `image_variables`, with its test.
- **GeecsBluesky 0.82.3** (mostly deletions): the classes and their tests removed from `db_runtime` (no shim, per repo practice); `namespace` imports from the new home; CLAUDE.md pointer.
- **GEECS-Data-Utils 0.30.0**: `io.scan_stack.read_stack_attributes(path)` (numeric members only), `parse_attribute_name(name)` and `stack_scalar_variables(path)` (the manifest), exported from `geecs_data_utils.io` — the read side (2c's s-file writer and the acceptance below read through them). Test added.
- Planning: `06_pva_file_plugin.md` file-layout paragraph; three stale `db_runtime.GeecsDbScalarPolicy` references repointed (01, 08, data_capture/01).

- **Numeric subscribed variables only.** The review showed an enum's wire value is its text label on both gateways (the CA gateway keeps labels verbatim for `enum_index`), so a `choice` attribute would be `NaN` in every row; enums, strings and paths therefore ride only in the strict row, never in the stack. The stack's scalar set is thus the **numeric subset** of the device's row columns (for `UC_Amp4_IR_input` all nine subscribed scalars are numeric, so the two sets are equal, which is what the acceptance below observed).
- **The timestamp ladder is excluded** from the scalar attributes: `acq_timestamp`/`systimestamp` are the frame's `…-frame_acq_timestamp` already (Unix epoch). Including them would duplicate the stamp under a second name in a second epoch.
- **A policy-query failure degrades to frames-and-stamps** with one warning (the policy's contract), rather than failing the gateway's startup — the roster queries just before it succeeded, so this is a narrow window.
- **Filter placement:** `scalar_attribute_variables` and `TIMESTAMP_LADDER` live in `geecs_core.db.variable_types` beside `image_variables` (the reviewer's placement finding): the gateway builds its roster from it and phase 2c's worker-side s-file writer recovers a stack's columns through it; the PVA server's own ladder constant is that one.
- **The stack records its raw scalar names** (`scalar_variables` / `scalar_attributes` root attributes): normalization is one-way and an offline reader has no DB.

## Tests

`./scripts/check.sh` (lint + the changed packages' suites): GEECS-Core **100 passed, 5 deselected**; GEECS-Data-Utils **329 passed, 4 skipped**; GeecsBluesky **370 passed, 3 skipped, 1 deselected**; GeecsPvaGateway **49 passed**.

## Hardware verification

**DONE 2026-09-11 23:37-23:40 PDT, `HTU-LaserOFF`, gateway 0.9.0 on the camera server of `UC_Amp4_IR_input` (share clone on this branch, `:restart` PV; 0.8.0 → 0.9.0 in 14 s, `NDAttributesFile` declares 11 attributes).** One strict `count([UC_Amp4_IR_input], num=5)` through the production manager → **Scan009 of 26_0911**, manager idle after 17 s, Tiled stop `success` with 5 primary events.

- Stack `UC_Amp4_IR_input.h5`: 11 attribute datasets — the two stamps plus `…-hdf-image-{2ndmomw0x, 2ndmomw0x45, 2ndmomw0y, 2ndmomw0y45, centroidx, centroidy, exposure, maxcounts, meancounts}`, each `(5,)` float64.
- **Names**: the nine scalar attribute names equal the s-file's nine columns for the device (normalized): missing `[]`, extra `[]` — equality because every subscribed scalar of this camera is numeric; in general the stack carries the numeric subset (judgment calls above).
- **Values**: joined by `acq_timestamp` (all five frames matched a row within 2 ms), 9 scalars × 5 rows compared, **0 mismatches** — e.g. `maxcounts` `[2164, 1555, 1584, 1825, 1613]` in both. `2ndmomW0x`/`2ndmomW0y` are `NaN` in the stack and blank in the s-file (the device pushes them empty), the documented "no numeric value → NaN" case.
- **Tiled**: the run's primary node has 13 parts, the device's 12 = the image stack + the 11 attribute arrays, each `array (5,)`; ingestion accepted (the #829 failure mode did not recur).
- Read side: the comparison used `read_stack_attributes` / `parse_attribute_name` (this PR's Data-Utils additions).
- **Re-run on the final commit (606de73e, after the review fixes)** 23:51: gateway restarted on it (14 s), **Scan010 of 26_0911**, identical result — 9 names equal, 45 values 0 mismatches, manager idle after 17 s; the new manifest reads back through `stack_scalar_variables`: `{…-maxcounts: MaxCounts, …-meancounts: MeanCounts, …}` (nine raw names).

## Deploy

The share clone (the fleet pin) is **on this branch now** (as #823 was), so a box that restarts picks up 0.9.0; after merge, check the feature branch out there again and pull, then `:restart` the remaining eight boxes (no launcher change; the launcher reinstalls GEECS-Core 0.6.0 from the clone). Worker: pull + `poetry install` + restart at leisure (the import move only; the describe path is stock).

No facility literals added outside tests/docstrings.


skbarber:
**Adversarial review (fresh-context subagent, three lenses; suites re-run by the reviewer) — 7 findings, all fixed in 7948a68b, each confirmed by the same reviewer against the code; two residuals fixed in 606de73e.**

Refuted by the reviewer before reporting (recorded so nobody re-checks): prefix/regex/space hazards in the widened TCP subscription (`_compile_frame_pattern` escapes and sorts longest-first; the pair boundary does not depend on the subscribed set); `_frame_timestamp`/arming unaffected; the stock `ADHDFDataLogic` re-reads `NDAttributesFile` on every `prepare_unbounded` and declares every attribute `<f8` / chunk 16384, matching `_open_file`; the extra stream resources are 1-D parts that `read_primary_scalars` never touches (no #834 regression).

1. **P1 — enums would be all-NaN columns.** An enum's wire value is its text label on both gateways (the CA gateway keeps labels verbatim for `enum_index`), so a `choice` attribute would be `NaN` in every row, forever. → **Fixed:** numeric subscribed variables only (`SCALAR_ATTRIBUTE_VARTYPES = {"numeric"}`); enum/text columns ride only in the strict row, stated in the docstring, CHANGELOGs and the PR body.
2. **P2 — a normalization collision (`Mean Counts` + `mean_counts`) raised in the plugin constructor and would put the whole camera server's service into a restart loop on the next `:restart`.** → **Fixed:** the filter dedupes on the normalized name with a warning (the roster layer that already warns-and-skips); the constructor check stays as a backstop. Pinned in `test_variable_types.py` and `test_config.py`.
3. **P2 placement — the filter belongs beside `image_variables`.** → **Fixed:** `scalar_attribute_variables` + `TIMESTAMP_LADDER` moved to `geecs_core.db.variable_types`; `server._TIMESTAMP_VARS` is that constant (the second literal copy is gone); phase 2c imports the filter from Core.
4. **P2 — the acceptance was stated as a set equality that cannot hold in general** (strings/paths, `systimestamp`, `save`, subscribed settables differ between the namespace's columns and the stack). → **Fixed (wording):** the PR body states the stack carries the numeric subset of the row's columns and explains why equality held for `UC_Amp4_IR_input` (all nine subscribed scalars numeric).
5. **P2 — the 0.9 stack was not self-describing** (normalization is one-way; an offline reader has no DB). → **Fixed:** root attributes `scalar_variables` (raw names) and `scalar_attributes` (dataset names), written at file open; pinned in the stock-data-logic test.
6. **P3 — `read_stack_attributes` not exported and raising on a non-numeric member.** → **Fixed:** numeric-dtype gate; `stack_scalar_variables(path)` reads the manifest; all three exported from `geecs_data_utils.io`; test covers a byte-string member and a pre-0.9 stack.
7. **P3 — stale `db_runtime.GeecsDbScalarPolicy` references** in `01_device_namespace.md`, `data_capture/01_central_pva_capture_scope.md`, `08_gated_batch.md`. → **Fixed.**

Residuals from the confirmation round, both fixed in 606de73e: `stack_scalar_variables` returned a `ValueError` on half a manifest (now `{}` with a warning, like its sibling skips foreign members); 08 §4.4 spelled the ladder by hand (now `TIMESTAMP_LADDER`).

Hardware acceptance (Scan009 of 26_0911) ran on 0c5d4b3f; the plugin's write path is unchanged by the fixes for this camera (no enum in its subscribed list), and the re-run on the final commit is recorded in the PR body once the camera server has restarted on it.

skbarber:
**Codex review — 1 finding, fixed in 0bad0c21.**

**P2 — case-mismatched subscribed names were dropped from the stack.** Confirmed: the namespace lower-matches the subscribed list against the metadata rows (`namespace.py` `by_lower`), so a `get='yes'` row spelled `maxcounts` against a `devicetype_variable` row `MaxCounts` still yields a strict-row column, while `scalar_attribute_variables` looked the name up by exact spelling and silently omitted it, breaking the same-columns contract. → The filter now keys the metadata rows by `name.lower()`, resolves each subscribed name case-insensitively (the ladder exclusion too), and returns the **metadata row's spelling** — the spelling the namespace uses and the TCP subscription needs. Pinned in `GEECS-Core/tests/test_variable_types.py` (`["maxcounts", "EXPOSURE", "Acq_Timestamp"]` → `["MaxCounts", "exposure"]`). Core 100 passed, gateway 49 passed.

Not re-run on hardware: `UC_Amp4_IR_input`'s nine subscribed names match their metadata rows exactly, so the written stack is byte-for-byte what Scan010 verified; the share clone is pulled to this head so the next gateway restart runs it.

PR #844, opened 2026-09-12
GeecsLogbook 0.6.0: navigation polish (calendar, keys, prefetch, today) + the ARIEL change feed; portal run→log link

skbarber:
PR 4 of the logbook arc (`phase/06-nav` → `feature/scan-logger`): navigation polish, plus the ARIEL change feed that was deferred out of #832 and asked for in this PR.

## What

**GeecsLogbook 0.6.0**

1. **Calendar in the rail**, both pages — a `<details class="cal">` that `static/nav.js` fills on open with a month grid. Days are marked when they have notes (soft accent, from the store's per-day counts) and when a day folder exists on the share (a dot). Marks come from a new `GET /log/api/month/{YYYY-MM}/days`, fetched **lazily by the open calendar**: the month page itself still never touches the share (pinned — `test_nav.py` makes the listing explode under the month page). The share side is **one** `scandir` of the month folder (`scan_reader.days_with_folders`, pinned to a single call), never thirty per-day walks. A missing experiment directory comes back as `share: false` — said, not hidden — and the store's marks stand alone.
2. **Faster day switching** — Sam did not specify the mechanism; my reading, flagged for veto: keyboard `←` `→` (day on the day page, month on the month page), `t` today, `c` toggles the calendar (ignored while typing); and a `<link rel="prefetch">` added when the viewer rests on a day/month link for 80 ms, so the click that follows is served warm. Nothing is prefetched on load — every day page is a share read. Collapse state was already kept (localStorage). The rail links remain full reloads; a fetch-and-swap SPA would be a different shape and was not asked for.
3. **Today** — `GET /log/today` and `GET /log/month/today` (this month, anchored at today's day group). The month rail's Today control is always present now (it was only a "Back to this month" from another month).
4. **The change feed** — `GET /log/api/entries?since=<aware ISO 8601>` (`NotesStore.changed_since`): everything whose `updated_at` moved after `since`, oldest change first, **tombstones included** — the one listing that returns them. `until=`, `book=`, `include_deleted=false` narrow it; `limit=` pages with a `next_cursor` that resumes after `(updated_at, rowid)`, so rows sharing an `updated_at` across a boundary cannot lose one. `since` must carry a timezone (422 otherwise: the columns are UTC `isoformat()` strings). Envelope `{entries, next_cursor}`. Relayed from Sam via the OSPREY/ARIEL session as the deferred #832 chunk, now wanted; scans are not in it by their own ruling.

**GEECS-DataPortal 0.26.0** — the run page links its scan's card in the logbook (`/log/day/YYYY-MM-DD#ScanNNN`, in the rail's day-stepper row) when the logbook is mounted; `GET /api/run/{uid}` carries the same URL as `logbook` (null when unlinked). Built from the mount prefix and the resolved day folder — the portal still never imports the logbook. Pinned mounted → present, not mounted → absent, proxy prefix carried.

**GeecsWebTheme 0.1.2** — the literal-colour guard walks `nav.js`.

## Per-concern LOC (insertions)

| Concern | Code | Tests | Docs/meta |
|---|---|---|---|
| Calendar (nav.js, css, marks route, `days_with_folders`, templates) | ~330 | ~150 | — |
| Keyboard + prefetch (inside nav.js; `data-*` on `<main>`) | ~60 | ~25 | — |
| Today names + month-rail Today | ~15 | ~25 | — |
| Change feed (store + route) | ~130 | ~150 | — |
| Portal run → log link | 24 | 29 | 11 |
| Theme guard surface | 1 | — | 7 |
| CHANGELOGs, CLAUDE.md (Navigation + Change feed sections), versions | — | — | ~95 |

- The feed rides in this PR (a second concern) because it was asked for in exactly this PR; it is store-sized and independently tested, and could be split out in five minutes if preferred.
- `include_deleted` defaults to **true** on the feed only. Every other listing hides tombstones; the feed's whole purpose includes them.
- Calendar cell on the month page links to `/log/month/M#day-D` (the book you are in); on the day page to `/log/day/D`.
- A "day folder exists" mark means the `YY_MMDD` folder exists (something wrote there), not that `scans/` inside it is non-empty — that would be a second stat per day.

## Tests

- GeecsLogbook: **207** (was 185; +22: `test_nav.py` calendar marks / single listing / share-down / month page never lists / today names / rail wiring / feed route; `test_store.py` counts + feed ordering, tombstones, exclusive since, shared-timestamp pagination, bad inputs)
- GEECS-DataPortal: **259** (+3)
- GeecsWebTheme: **108** (+1)
- `scripts/check.sh --base origin/feature/scan-logger` rc=0 (captured unpiped).

Smoke-verified on a local temp-share server with headless Chrome: both pages render the open calendar with correct marks (scan days dotted, note days tinted, active day filled, today outlined); the `/days` payload for the smoke month matches the seeded data.

## Hardware verification

None needed — no scan execution or device path is touched. The share-side change is one directory listing.

### DIFF
--- a/GEECS-DataPortal/CHANGELOG.md
+++ b/GEECS-DataPortal/CHANGELOG.md
@@ -12,7 +12,8 @@ Format follows [Keep a Changelog](https://keepachangelog.com/en/1.1.0/).
   when the logbook is mounted, and `GET /api/run/{uid}` carries the same
   URL as `logbook` (null when there is nothing to link). Built from the
   mount prefix and the resolved day folder; the portal still never
-  imports the logbook.
+  imports the logbook. A run from another experiment than the logbook's
+  gets no link — scan numbers restart per experiment (review of #844).
 
 ## [0.25.0] - 2026-09-11
 
--- a/GEECS-DataPortal/CHANGELOG.md
+++ b/GEECS-DataPortal/CHANGELOG.md
@@ -3,6 +3,17 @@
 All notable changes to this package will be documented here.
 Format follows [Keep a Changelog](https://keepachangelog.com/en/1.1.0/).
 
+## [0.26.0] - 2026-09-12
+
+### Added
+
+- The run page links to its scan's card in the scan logbook
+  (`/log/day/YYYY-MM-DD#ScanNNN`, in the rail beside the day steppers)
+  when the logbook is mounted, and `GET /api/run/{uid}` carries the same
+  URL as `logbook` (null when there is nothing to link). Built from the
+  mount prefix and the resolved day folder; the portal still never
+  imports the logbook.
+
 ## [0.25.0] - 2026-09-11
 
 ### Added
--- a/GEECS-DataPortal/CLAUDE.md
+++ b/GEECS-DataPortal/CLAUDE.md
@@ -168,10 +168,15 @@ tests/
 ```
 
 Routes: `/` (redirect to today) · `/day/{iso}` (run list; `?experiment=`)
-· `/log/day/{iso}` + `/log/api/day/{iso}` (the scans book, `--scan-log`)
+· `/log/day/{iso}` + `/log/api/day/{iso}` (the scans book, `--scan-log`;
+`/log/today` names today)
 · `/log/month/{YYYY-MM}` + `/log/api/month/{YYYY-MM}/entries` (the ops book,
-same flag; store-only, never the share)
-· `/log/api/entries…` (the logbook's entry writes, only with `--notes-db`)
+same flag; store-only, never the share; `/log/month/today`) ·
+`/log/api/month/{YYYY-MM}/days` (the rail calendar's marks: note counts
+from the store plus one listing of the month folder)
+· `/log/api/entries…` (the logbook's entry writes, only with `--notes-db`;
+`GET /log/api/entries?since=` is the change feed — tombstones included,
+cursor-paged — for synchronisers such as ARIEL)
 · `/run/{uid}` (the scan page: rail + Overview/Plot/Images/Analysis tabs
 — Analysis only when runs are possible, see below;
 `?tab=&y=&x=&view=&filters=&bincfg=&display=` is the Plot-tab state,
@@ -210,7 +215,9 @@ agent/script surface; served `no-cache`): `/api/day/{iso}?experiment=&filter=`
 (the run table, newest first, the page's filter haystack) ·
 `/api/run/{uid}` (summary + the Overview table verbatim + start/stop
 docs + `devices` + stepper `neighbours`/`day_runs` +
-`processing_options` + `analysis_enabled`) · `/api/run/{uid}/device?device=`
+`processing_options` + `analysis_enabled` + `logbook`, the run's card in
+the scan logbook or null — only when the logbook is mounted and the run
+belongs to its experiment) · `/api/run/{uid}/device?device=`
 (one device's tier/path via the same `device_kind` probe as the page) ·
 `/api/run/jump/{iso}?prefer=` (the day steppers' target as data).
 The page and JSON routes share `_list_day` / `_neighbours` /
--- a/GEECS-DataPortal/geecs_portal/app.py
+++ b/GEECS-DataPortal/geecs_portal/app.py
@@ -54,6 +54,7 @@ from geecs_data_utils.data.row_filters import filter_mask
 from geecs_data_utils.io.images import average_frames
 from geecs_data_utils.scan_frame import PROVENANCE_RUN, scan_frame
 from geecs_data_utils.tiled_catalog import (
+    RunDetail,
     RunSummary,
     ScanCatalog,
     fmt_time_of_day,
@@ -445,18 +446,24 @@ def create_app(
     scan_log_enabled = False  # set when the logbook router mounts (below)
 
     def _logbook_url(
-        request: Request, run_day: Optional[date], scan_number: int
+        request: Request, detail: RunDetail, run_day: Optional[date]
     ) -> str:
         """The run's entry in the scan logbook, or "" when there is none to link.
 
         The logbook's day page anchors each scan card by its folder name
         (``#Scan012``); the portal knows the mount prefix and the day, so
-        the link is built here without importing the logbook — a
-        peer view layer, reached by URL like any other page.
+        the link is built here without importing the logbook — a peer
+        view layer, reached by URL like any other page. The logbook is
+        mounted for the default experiment alone, and scan numbers
+        restart daily per experiment, so a run from another experiment
+        gets no link rather than a wrong one.
         """
-        if not (scan_log_enabled and run_day and scan_number):
+        summary = detail.summary
+        if not (scan_log_enabled and run_day and summary.scan_number):
             return ""
-        return f"{_root(request)}/log/day/{run_day.isoformat()}#Scan{scan_number:03d}"
+        if summary.experiment and summary.experiment != default_experiment:
+            return ""
+        return f"{_root(request)}/log/day/{run_day.isoformat()}#Scan{summary.scan_number:03d}"
 
     def _load_run(uid: str):
         """Load one run, mapping failures to honest HTTP status codes.
@@ -1391,8 +1398,7 @@ def create_app(
             "processing_options": _processing_names(),
             "analysis_enabled": _analysis_enabled_for(folder),
             "config_editor": config_editor_enabled,
-            "logbook": _logbook_url(request, run_day, detail.summary.scan_number or 0)
-            or None,
+            "logbook": _logbook_url(request, detail, run_day) or None,
             "page": f"{_root(request)}/run/{uid}",
             "portal_version": _portal_version(),
         }
@@ -1615,9 +1621,7 @@ def create_app(
                 "next_uid": next_uid,
                 "day_runs": day_runs,
                 "scan_number": detail.summary.scan_number or 0,
-                "logbook_url": _logbook_url(
-                    request, run_day, detail.summary.scan_number or 0
-                ),
+                "logbook_url": _logbook_url(request, detail, run_day),
                 "prev_day": (
                     (run_day - timedelta(days=1)).isoformat() if run_day else ""
                 ),
--- a/GEECS-DataPortal/geecs_portal/app.py
+++ b/GEECS-DataPortal/geecs_portal/app.py
@@ -442,6 +442,21 @@ def create_app(
     # 2026-08-29 — lazy stays the rule ACROSS scans only).
     data_cache = ShotDataCache()
     config_editor_enabled = False  # set when the editor router mounts (below)
+    scan_log_enabled = False  # set when the logbook router mounts (below)
+
+    def _logbook_url(
+        request: Request, run_day: Optional[date], scan_number: int
+    ) -> str:
+        """The run's entry in the scan logbook, or "" when there is none to link.
+
+        The logbook's day page anchors each scan card by its folder name
+        (``#Scan012``); the portal knows the mount prefix and the day, so
+        the link is built here without importing the logbook — a
+        peer view layer, reached by URL like any other page.
+        """
+        if not (scan_log_enabled and run_day and scan_number):
+            return ""
+        return f"{_root(request)}/log/day/{run_day.isoformat()}#Scan{scan_number:03d}"
 
     def _load_run(uid: str):
         """Load one run, mapping failures to honest HTTP status codes.
@@ -1376,6 +1391,8 @@ def create_app(
             "processing_options": _processing_names(),
             "analysis_enabled": _analysis_enabled_for(folder),
             "config_editor": config_editor_enabled,
+            "logbook": _logbook_url(request, run_day, detail.summary.scan_number or 0)
+            or None,
             "page": f"{_root(request)}/run/{uid}",
             "portal_version": _portal_version(),
         }
@@ -1598,6 +1615,9 @@ def create_app(
                 "next_uid": next_uid,
                 "day_runs": day_runs,
                 "scan_number": detail.summary.scan_number or 0,
+                "logbook_url": _logbook_url(
+                    request, run_day, detail.summary.scan_number or 0
+                ),
                 "prev_day": (
                     (run_day - timedelta(days=1)).isoformat() if run_day else ""
                 ),
@@ -1938,6 +1958,7 @@ def create_app(
                     ),
                     prefix="/log",
                 )
+                scan_log_enabled = True
                 logger.info(
                     "scan log mounted at /log for %s (%s; templates %s)",
                     default_experiment,
--- a/GEECS-DataPortal/geecs_portal/templates/run.html
+++ b/GEECS-DataPortal/geecs_portal/templates/run.html
@@ -145,6 +145,9 @@
         <a class="act" href="{{ root }}/day/{{ run_day }}?{{ qs(device="", shot="") }}" title="browse this day's list">{{ run_day }}</a>
         <a class="act" href="{{ root }}/run/jump/{{ next_day }}?prefer={{ scan_number }}&amp;{{ qs(device="", shot="") }}" title="next day (same scan number, else its latest)">day ›</a>
         {% endif %}
+        {% if logbook_url %}
+        <a class="act logbook" href="{{ logbook_url }}" title="this scan's card in the scan logbook">log ↗</a>
+        {% endif %}
       </div>
       <div class="steprow">
         <a class="act {{ "off" if not prev_uid }}" href="{{ root }}/run/{{ prev_uid }}?{{ qs() }}" title="previous scan">‹ scan</a>
--- a/GEECS-DataPortal/pyproject.toml
+++ b/GEECS-DataPortal/pyproject.toml
@@ -1,6 +1,6 @@
 [tool.poetry]
 name = "geecs-data-portal"
-version = "0.25.0"
+version = "0.26.0"
 description = "Scan-browsing web service (read-only except explicit analysis runs): a zero-install browser view over the Tiled catalog and the GEECS data share"
 authors = ["GEECS-BELLA"]
 readme = "README.md"
--- a/GEECS-DataPortal/tests/test_scan_log_mount.py
+++ b/GEECS-DataPortal/tests/test_scan_log_mount.py
@@ -134,3 +134,32 @@ class TestSeedTemplates:
         html = client.get("/log/month/2026-09").text
         assert client.get("/log/month/2026-09").status_code == 200
         assert "typebtn" not in html
+
+
+class TestRunPageLink:
+    """The run page points at its scan's card in the logbook — when there is one."""
+
+    _UID = "uid-002"  # TEST_DAY, Scan 002 in the fake catalog
+
+    def test_present_when_the_logbook_is_mounted(self) -> None:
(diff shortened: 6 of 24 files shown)


skbarber:
# Adversarial review of #844

(Fresh-context reviewer, brief per `.claude/skills/land/SKILL.md`; posted verbatim as the audit record. Disposition follows in the next comment.)

## 1. Arrow-key stepping can navigate away from an unsaved composer

**Severity:** Medium
**Lens:** Correctness
**Where:** `GeecsLogbook/geecs_logbook/static/nav.js` (keydown handler); `editor.js` (no `beforeunload` or dirty-state guard anywhere — the only `keydown` it owns is the textarea's Cmd/Ctrl+Enter)

**Scenario / evidence:** The `typing()` guard only checks the *event target*. A user types a paragraph into a composer `<textarea>`, then clicks the preview toggle (focus lands on the button and the textarea is hidden), or clicks the page background to scroll, then presses `←` expecting to move the caret. Focus is no longer on a text field, so `window.location.href = host.dataset.prev` fires and the draft is gone — no `beforeunload` prompt, no draft persistence (`editor.js` stores only the author name in localStorage). `store.py`'s own module docstring calls unsaved prose "the one thing that cannot be regenerated"; this PR adds the first single-keystroke path that discards it.

**Suggested fix:** Before stepping, bail if any composer is dirty (e.g. `document.querySelector("form textarea")?.value.trim()`), or have `editor.js` register a `beforeunload` handler while a composer holds text. Either is a few lines; the first also protects the `t` key.

## 2. `create()` stamps `updated_at` outside the write lock, so the change feed can permanently miss a row

**Severity:** Medium-Low
**Lens:** Correctness
**Where:** `GeecsLogbook/geecs_logbook/store.py` (`create`: `now = _now()` before the INSERT under a fresh autocommit connection); claim in `routes/entries.py` and `GeecsLogbook/CLAUDE.md` "The change feed" ("misses nothing")

**Scenario / evidence:** Every other writer (`update`, `set_status`, `add_attachment`, `delete`) computes its stamp inside `BEGIN IMMEDIATE`, so stamps are assigned in commit order. `create` computes `now` *before* `_connect()` (sqlite connect + two PRAGMAs). FastAPI runs these sync handlers in a threadpool, so two POSTs interleave: A takes `T1`; B takes `T2 > T1`, inserts, commits; the synchroniser polls and records `T2` as its high-water mark; A inserts with `T1`. The next poll asks `since=T2` — A is below the mark and is never delivered. The window is milliseconds, but the docstring and CLAUDE.md promise "a copy that asks for everything since my last `updated_at` misses nothing", and ARIEL is the named consumer.

**Suggested fix:** Compute `now` inside `with self._connect() as conn, _transaction(conn):` in `create`, matching the other writers. Independently, soften the "misses nothing" wording to "…provided the copy overlaps its `since` by a few seconds".

## 3. The run→log link ignores which experiment the run belongs to

**Severity:** Low-Medium
**Lens:** Correctness
**Where:** `GEECS-DataPortal/geecs_portal/app.py` (`_logbook_url` and its two call sites)

**Scenario / evidence:** The logbook is mounted for `default_experiment` only. The run page and `/api/run/{uid}` accept `?experiment=` and thread `experiment or default_experiment` into `_neighbours`, and `RunSummary.experiment` ("GEECS experiment name from the start doc") is on `detail.summary`. `_logbook_url` checks neither. A run from a second experiment on the same catalog, Scan 012 on 2026-09-11, renders a `log ↗` link and a JSON `logbook` value pointing at `/log/day/2026-09-11#Scan012` — the *default* experiment's Scan012 card, a different scan. Scan numbers restart daily per experiment, which is exactly the trap `_run_day`'s docstring warns about for the day.

**Suggested fix:** Gate on the run's own experiment: `if detail.summary.experiment and detail.summary.experiment != default_experiment: return ""`. One assertion in `TestRunPageLink` with a `FakeCatalog` run whose `experiment` differs pins it.

## 4. The feed cursor is not URL-safe, and a shaped-but-bogus cursor is accepted silently

**Severity:** Low
**Lens:** Correctness
**Where:** `GeecsLogbook/geecs_logbook/store.py` (cursor parse and render); `routes/entries.py`

**Scenario / evidence (probe):** `next_cursor` renders as `2026-09-12T07:31:42.969272+00:00|1`. A client that pastes it into a URL unencoded (curl, a shell script, a browser bar) gets `+` decoded to a space by Starlette's query parser. The stamp then sorts *below* the real one (`" " < "+"`), so `updated_at > ?` re-admits the whole tie group: with `limit=1` page 1 returned `n0`; page 2 via the raw cursor returned `['n0','n1','n2']` — the boundary row repeated (the encoded form correctly returned `['n1','n2']`). The tests pass only because httpx percent-encodes. Separately, `cursor=junk|5` passes the `rpartition`/`isdigit` check and returns `200 {"entries": [], "next_cursor": null}` — a corrupted cursor reads as "caught up". (`since=…+00:00` raw also 422s with pydantic's "unexpected extra characters" rather than the route's own timezone message; `…Z` works.)

**Suggested fix:** Validate the cursor's stamp with `datetime.fromisoformat` (422 on failure), and make the cursor opaque and URL-safe — e.g. `base64.urlsafe_b64encode(f"{stamp}|{rowid}")`. Add a test that sends the cursor through a raw query string.

## 5. `/api/month/{m}/days` lets a share I/O error escape as a 500 where the day page gives 503

**Severity:** Low
**Lens:** Correctness
**Where:** `GeecsLogbook/geecs_logbook/scan_reader.py` (`days_with_folders`: the `is_dir()` check and `ScanPaths.get_daily_scan_folder` with no `base_directory`); `routes/month.py`

**Scenario / evidence (probe):** `Path.is_dir()` swallows only ENOENT/ENOTDIR/EBADF/ELOOP. With the share root unreadable (EACCES; EIO/ETIMEDOUT on a wedged SMB mount behave the same) the experiment-directory check raised `PermissionError` straight out of the route, while the same condition on `/log/day/2026-09-11` returned 503 through `Context.load_day`. The `try/except OSError` covers only the `scandir`. Likewise, with no `base_directory` and no `paths_config`, `get_daily_scan_folder` raises `AttributeError`. The browser degrades to "Marks unavailable." so the user impact is small, but every calendar open logs a traceback and the endpoint's status lies about the cause.

**Suggested fix:** Wrap the whole body of `days_with_folders` in the same `except OSError → None` (or route it through a handler that maps to 503 like `load_day`).

## 6. Hover prefetch fires after 80 ms on a 15-link rail, and prefetched pages carry no freshness headers

**Severity:** Low
**Lens:** Correctness (perf / staleness)
**Where:** `GeecsLogbook/geecs_logbook/static/nav.js` (prefetch); no `Cache-Control` anywhere under `geecs_logbook/`

**Scenario / evidence:** The rail lists fifteen day links vertically. A mouse travelling down that list at ordinary speed dwells well over 80 ms per link, so one sweep to the target can queue several `/log/day/…` prefetches — each a `read_day` share listing plus `maybe_sync` on the server. The 80 ms threshold is a pass-through filter, not the "rests on" the CLAUDE.md text describes. Second, the pages set no `Cache-Control`; Chrome reuses a `<link rel=prefetch>` document for ~5 minutes without revalidation when no freshness is stated, so a day hovered at 10:00 and clicked at 10:04 can show a scan list missing a scan that finished at 10:01 until a reload. Both are trade-offs the design can accept, but neither is stated.

**Suggested fix:** Raise the dwell to ~250–300 ms (a real hover) and cap in-flight prefetches; state the staleness window in the CLAUDE.md "Navigation" section, or send `Cache-Control: no-cache` on the day/month pages.

## 7. Third copy of the share-layout walk and third `YY_MMDD` parser in the repo

**Severity:** Low
**Lens:** Redundancy / Placement
**Where:** `GeecsLogbook/geecs_logbook/scan_reader.py` (`_DAY_DIR`, `month_folder`, `days_with_folders`)

**Scenario / evidence:** `month_folder` derives the month directory as `get_daily_scan_folder(...).parent.parent` and `days_with_folders` derives the experiment directory as a further `.parent.parent` — the same `.parent` chain already written out in `mirror.logbook_root` (which also owns the identical "experiment directory missing means unmounted, not empty" check) and in `read_day`. `_DAY_DIR` + `2000 + yy` is a new parser for a folder name whose one builder is `ScanPaths.get_daily_scan_folder` and whose two existing parsers both use `strptime("%y_%m%d")` (`scan_paths.py`, `scans_database/builder.py`) — a different century rule for the same string. Any future consumer of "which days exist this month" (the portal's day browser, the console's ops panel via `ops_paths.todays_scan_folder`, LiveWatch) would have to lift this out of the logbook.

**Suggested fix:** Put one read-only `day_folder_date(name) -> date | None` and a `month_folder(...)`/`list_day_folders(...)` beside `daily_scan_folder` in `geecs_data_utils.scan_paths`, then have `logbook_root`, `read_day` and `days_with_folders` share it. Not a blocker for this PR — but it is the moment the third copy appeared.

## 8. Portal contract page not updated for the behaviour it describes

**Severity:** Low
**Lens:** Correctness (contract files travel with behaviour)
**Where:** `GEECS-DataPortal/CLAUDE.md` (route index and the `/api/run/{uid}` field list)

**Scenario / evidence:** The diff updates `GeecsLogbook/CLAUDE.md` but the portal's CLAUDE.md, which indexes every mounted `/log/…` route and enumerates the `/api/run/{uid}` payload keys, gains neither the new `logbook` key nor `/log/today`, `/log/month/today`, `/log/api/month/{m}/days`, `/log/api/entries?since=`. The run-page `log ↗` link is likewise undocumented on the portal side.

**Suggested fix:** Add `logbook` to the `/api/run/{uid}` field list and the four routes to the `/log/…` index in the same PR.

---

## Checked and not flagged

- **Change-feed ordering and boundaries:** verified the ISO string comparison is order-preserving across the microsecond/no-microsecond forms (`+` < `.` < digits), `since` exclusive / `until` inclusive, legacy rows (`COALESCE(edited_at, created_at)` fill) share the format, `Book` is a `Literal` so the filter binds cleanly, the `(updated_at, rowid)` cursor covers ties (test pins it), and `mark_mirrored`/`mark_deferred` correctly do *not* move `updated_at`.
- **Tombstone bodies are readable through the feed:** by design (CLAUDE.md "accidental delete is a field to clear"); noted, not a finding.
- **Route ordering:** `/today` vs `/day/{day}`, `/month/today` before `/month/{month}`, `GET /api/entries` before `/api/entries/{id}` — no shadowing; the feed is unregistered without a store (404 pinned).
- **Redirect targets** (`day/…`, `YYYY-MM#day-…`) resolve correctly under the `/log` mount and a proxy prefix; `url_for(...).path` on `data-*` attributes includes `root_path`.
- **nav.js helpers:** `PAGES` regex escaping and `stepMonth` across year boundaries verified under node; `ROOT` derivation from `data-api` works under a prefix; the `c`/`t`/arrow keys collide with nothing in `editor.js` or `theme.js`.
- **Scan-folder invariant:** `get_daily_scan_folder` is pure path construction; `days_with_folders` only `scandir`s; the `Path.mkdir` explode-pin matches the package's existing pattern.
- **Month-page promise:** the existing `test_never_reads_the_share` plus the new `test_month_page_never_calls_it` together pin that `/days` is the calendar's lazy fetch.
- **Public-repo hygiene:** no hostnames, accounts or home paths in the diff; the theme guard now walks `nav.js`, which sets no colours.
- **Redundancy sweep for month/day listings:** `GEECS-Data-Utils/geecs_data_utils/scan_paths.py`, `scans_database/builder.py`, `GEECS-Console/geecs_console/services/ops_paths.py`, `GEECS-DataPortal/geecs_portal/app.py`, `GeecsLogbook/geecs_logbook/mirror.py`, `_fs.py`, `routes/_common.py`. Only the layout-walk/parser overlap in finding 7 survived.
- **Version bumps / changelogs:** logbook 0.6.0, portal 0.26.0, theme 0.1.2 each carry an entry consistent with the diff.

skbarber:
# Disposition of the #844 review

All eight findings dispositioned; fixes in **0c54ebd4**, verified by the same reviewer (report below, verbatim).

| # | Finding | Disposition |
|---|---|---|
| 1 | Arrow keys could navigate away from an unsaved composer | **Fixed**: `nav.js` refuses `←` `→` `t` while any textarea holds text. |
| 2 | `create()` stamped `updated_at` outside the write lock; the feed could skip a row | **Fixed**: the stamp is taken inside `BEGIN IMMEDIATE` like every other writer; the "misses nothing" wording is qualified. |
| 3 | Run→log link ignored the run's experiment | **Fixed**: no link when the run's experiment differs from the logbook's; pinned by a new portal test. |
| 4 | Feed cursor not URL-safe; a shaped-but-bogus cursor read as "caught up" | **Fixed**: opaque unpadded urlsafe-base64 cursor, validated with `fromisoformat` (422 otherwise); tests send it through a raw query string. |
| 5 | `/api/month/{m}/days` let share I/O errors escape as 500 | **Fixed**: EACCES/EIO → `share: false`; anything else → 503 like the day page; both pinned. |
| 6 | 80 ms prefetch dwell on a 15-link rail; staleness unstated | **Fixed**: 250 ms dwell; the trade-off (no freshness headers, minutes of possible reuse) is stated in the CLAUDE.md Navigation section. No `Cache-Control: no-cache` — a revalidated prefetch is a full re-fetch here, which defeats the point. |
| 7 | Third copy of the share-layout walk / `YY_MMDD` parser | **Waived** — this is #839's brief (one reader beside the one builder in `geecs_data_utils.scan_paths`). Recorded on #839 and in the logbook CLAUDE.md deferred table; not lifted here so the logbook keeps depending on `ScanPaths` alone. |
| 8 | Portal contract page not updated | **Fixed**: route index and `/api/run` field list updated. |

Suites after the fix: GeecsLogbook 211, GEECS-DataPortal 260, GeecsWebTheme 108; `check.sh` rc=0.

---

# Verification of the #844 fixes

Checked against `0c54ebd4` on `phase/06-nav` (`git diff 3d034035..0c54ebd4`), by reading the diff, re-running the original probe script against the fixed code, and running the three suites unpiped (GeecsLogbook 211 passed, GEECS-DataPortal 260 passed, GeecsWebTheme 108 passed; rc=0 each).

## 1. Arrow-key stepping vs an unsaved composer — **Confirmed fixed**

**Where:** `GeecsLogbook/geecs_logbook/static/nav.js:165-166, 175`
**Evidence:** `drafting()` scans every `<textarea>` for non-blank text and the step branch bails on `!url || drafting()` before `preventDefault`/navigation, so the `c` toggle still works and the arrows fall through to the browser when a note is pending. Both templates' textareas start empty and are only populated by the user or a seed button, so the guard does not silently disable the keys on a clean page. The original scenario (preview toggle moves focus to a button, `←` discards the paragraph) no longer navigates.
**Residual:** none. A `beforeunload` guard for mouse-driven navigation would be the belt to this brace, but it was not the finding.

## 2. `create()` stamps outside the write lock — **Confirmed fixed**

**Where:** `GeecsLogbook/geecs_logbook/store.py:508-529`; wording at `routes/entries.py:137-146`
**Evidence:** `now = _now()` now sits inside `with self._connect() as conn, _transaction(conn):` (`BEGIN IMMEDIATE`), matching the other four writers, so stamps are assigned in commit order. Probe: 4 threads × 30 concurrent `create()`s → the feed returned all 120 rows in monotone `updated_at` order. The docstring now says why it holds and recommends overlapping `since` as belt-and-braces; the CHANGELOG records the change.

## 3. Run→log link ignores the run's experiment — **Confirmed fixed**

**Where:** `GEECS-DataPortal/geecs_portal/app.py:448-466`; test `tests/test_scan_log_mount.py:167-180`
**Evidence:** `_logbook_url(request, detail, run_day)` returns `""` when `summary.experiment` is set and differs from `default_experiment`; both call sites pass `detail`. The new test replaces the fake run's `experiment` with `"Thomson"` via `dataclasses.replace` and asserts no `/log/day/` in the HTML and `logbook is None` in the JSON — it would fail against `3d034035`. An empty `summary.experiment` (older start docs) still links, which is the right default for a single-experiment catalog. CHANGELOG and the portal CLAUDE.md field list state the gate.

## 4. Cursor not URL-safe / shaped-bogus cursor accepted — **Confirmed fixed**

**Where:** `GeecsLogbook/geecs_logbook/store.py:184-208` (`_encode_cursor`/`_decode_cursor`), 381-384, 404-406; `routes/entries.py:132` (`max_length=128`); tests `test_nav.py:270-287`
**Evidence (probe re-run):** the cursor is now urlsafe base64, unpadded; sent through a raw query string it returned `['n1','n2']` — the boundary row `n0` is no longer re-sent. `cursor=junk|5` → 422; `%%%` → 422 (tested); a base64 payload whose stamp is not `fromisoformat`-parsable → 422. The `max_length` raise to 128 fits the encoded form. `binascii.Error` is a `ValueError` subclass, so the except clause covers it.
**Residual:** the `since=…+00:00` raw-paste 422 message is pydantic's, not the route's — cosmetic, not part of the finding.

## 5. `/api/month/{m}/days` lets share I/O errors escape as 500 — **Confirmed fixed**

**Where:** `GeecsLogbook/geecs_logbook/scan_reader.py:380-404`; `routes/month.py:182-188`; tests `test_nav.py:83-96, 154-165`
**Evidence (probe re-run):** with the share root `chmod 0`, `/days` now returns `200 {"share": false, …}` with one warning line instead of an unhandled `PermissionError`. `month_folder` + the `is_dir` check + `scandir` are inside one `try`; `FileNotFoundError` → empty set, other `OSError` → `None`, and the route maps any non-`OSError` (e.g. the no-config `AttributeError`) to 503 with the day page's wording — the `RuntimeError` monkeypatch test pins that path. The `Path.is_dir` monkeypatch test pins the EACCES → `None` path.

## 6. Prefetch dwell and staleness — **Confirmed fixed as scoped**

**Where:** `nav.js:16-22, 203`; `GeecsLogbook/CLAUDE.md:180-190`
**Evidence:** dwell is 250 ms (a pass over the fifteen rail links no longer queues share reads), and the CLAUDE.md "Navigation" section now states the trade-off in full. The decision not to add `Cache-Control: no-cache` is a defensible call given the stated reasoning; the finding asked for the dwell and the statement, and both are there.

## 7. Layout-walk / `YY_MMDD` parser duplication — **Waiver acknowledged, correctly recorded**

**Where:** `GeecsLogbook/CLAUDE.md:343` (deferred table row)
**Evidence:** the row names all three copies, the two existing `strptime("%y_%m%d")` parsers, the intended home (`geecs_data_utils.scan_paths`: `day_folder_date`, `list_day_folders`) and the tracking issue (#839). Consistent with the earlier `logbook_root` row in the same table. Nothing further to verify in this PR.

## 8. Portal contract page — **Confirmed fixed**

**Where:** `GEECS-DataPortal/CLAUDE.md:170-179, 216-220`
**Evidence:** the route index now lists `/log/today`, `/log/month/today`, `/log/api/month/{YYYY-MM}/days` and `GET /log/api/entries?since=`; the `/api/run/{uid}` field list gains `logbook` with its two conditions — matching the code exactly.

## Not re-flagged

- `days_with_folders` with no `base_directory` on a machine that has a `config.ini` resolves the configured root and reports `share: false` if unmounted — expected; the 503 path is pinned by the monkeypatched test.
- The decoded-cursor path still accepts any parsable stamp with any rowid — correct: that is a valid cursor, not a corrupted one.
- `drafting()` also fires while an *edit* form of an existing entry is open — intended: an open edit is an unsaved change.

No surviving findings against `0c54ebd4`.

PR #846, opened 2026-09-12
GeecsBluesky 0.82.4: ensure-ready re-downloads an empty plan list through environment_update (#838)

### LINKED ISSUE #838: Queueserver: a timed-out 'download the lists of existing plans and devices' leaves the manager idle with EMPTY allowed-plan lists — every submission then fails 'not in the list of allowed plans'
Observed 2026-09-11 on the worker host. At 21:02:12, while the host was thrashing in swap (an unrelated OOM incident, #834), the RE Manager logged:

```
[I] Downloading the lists of existing plans and devices from the worker environment
[E] Failed to download the list of existing plans and devices from the worker process: Timeout occurred while p…
```

From then on `plans_allowed` returned **0 plans for every user group** (primary, operator, admin), `existing_plans_and_devices.yaml` in the startup dir was not rewritten (still the 20:09 copy), and every `submit_plan` failed with `Plan 'run_action' is not in the list of allowed plans` — while `status()` reported `re_state=idle, manager_state=idle, worker_environment_state=idle, worker_exists=True`. A `count` submitted at 20:57 (before the timeout) had run normally.

**Why it matters.** `geecs-qserver-ready` asserts `plans_allowed ⊇ GEECS_PLAN_NAMES` once, after the manager starts (#793: "a running service means ready"). This failure happens *after* readiness and leaves the service looking healthy: the queue is idle, the environment is open, and the only symptom is that every submission is refused with the message that #793's memory note attributes to a closed environment — a misleading trail.

**Recovery without a restart.** `environment_update` (the manager re-downloads the lists from the running worker and regenerates the allowed lists); verified as the fix on the host tonight. `permissions_reload(restore_plans_devices=True)` would reload the stale on-disk copy instead.

**Proposed.** (1) `qs_client.readiness` / the readiness verdict treats an empty `plans_allowed` on an open environment as *not ready* and names `environment_update` as the fix (the Console's readiness banner and the MCP preflight then say the right thing). (2) `geecs-qserver-ensure-ready` gains a re-runnable mode that, when the lists are empty, calls `environment_update` and re-asserts — so an operator's `systemctl start geecs-qserver-ready` (or a periodic timer) heals it. (3) The runbook (`qserver/README.md` Troubleshooting) gets the entry: "allowed plans empty while idle = a timed-out list download; `environment_update`, not a restart".

skbarber:
Closes #838.

## What

`geecs-qserver-ensure-ready` (the `geecs-qserver-ready` oneshot) heals the manager that is **idle, environment open, `plans_allowed` empty**. Observed 2026-09-11 on the worker host: the manager's own download of the lists from the worker timed out while the host thrashed in swap, and from then on every submission was refused "Plan 'X' is not in the list of allowed plans" while `status` reported everything healthy and the readiness unit stayed `active (exited)` from its last successful run.

- `qserver_ready.ensure_ready`: a list still empty or incomplete after the (post-open) settle window is **restored once from the worker's on-disk copy** through `permissions_reload(restore_plans_devices=True)`, and the settle window applies again. The worker writes `existing_plans_and_devices.yaml` from its namespace at every environment open (`--update-existing-plans-devices` default `ENVIRONMENT_OPEN`, `worker.py`), so that file *is* the running environment's list and the manager's `_update_allowed_plans_and_devices(restore_plans_devices=True)` loads it and regenerates the allowed lists unconditionally. Unanswered lists (`plans_unknown`) are never restored — a manager that does not answer is not one to ask. A refused restore (queue locked) is reported as not ready with both the verdict's detail and the manager's reason. The restore call gets a 10 s transport budget (the manager answers it after the disk read and a permissions push to the worker).
- **Correction to #838's own text** (found by the review, verified in the installed 0.0.25 source): `environment_update` is *not* a reliable fix — with an empty script the worker re-downloads only when its regenerated descriptions differ from its stored copy, so on an unchanged namespace it is a no-op (it also needs an idle manager). And `permissions_reload(restore_plans_devices=True)` does *not* load a stale copy: the file the issue saw "not rewritten since 20:09" was the worker's copy from that environment open — exactly the right list. The README says both.
- `qs_client.readiness_verdict`: the shared `plans_empty` sentence (Console banner, MCP preflight) names the timed-out download as a cause and `systemctl restart geecs-qserver-ready` (→ the restore from disk) as the gesture, instead of only "startup profile registered nothing".
- `qserver/README.md` Troubleshooting: the entry (allowed plans empty while idle = a timed-out list download; `qserver permissions reload lists` by hand; why `environment update` is not the fix; a plan set changed without re-opening needs `systemctl restart geecs-qserver`). `CLAUDE.md` and `deploy/DEPLOYMENT.md` prose describe the unit's heal step (the unit template is untouched: deployment files move once at the end of the feature branch).

## Why the shape

The issue proposed three things: (1) the verdict names the state — done; (2) a re-runnable ensure-ready mode that heals it — done as the default path (the unit is already re-run by `systemctl restart geecs-qserver-ready`, no new mode or flag; a periodic timer unit is *not* added: deployment files are touched once at the end of the feature branch, `03` §10.5); (3) the runbook entry — done.

- The restore also covers `plan_missing` (an incomplete list), not only `plans_empty`: the manager's pre-open list is the on-disk copy loaded at start, so a timed-out post-open download leaves a stale-but-non-empty list; same mechanism, same cost.
- One restore per run, never a loop: a second empty read after a restore is a real fault to surface.

## Tests

`./scripts/check.sh GeecsBluesky`: **374 passed, 3 skipped, 1 deselected** (was 369). New pins in `tests/test_qserver_ready.py`: empty list with the environment up → one `permissions_reload(restore_plans_devices=True)` → ready when the list reads back; incomplete list restored then ready; refused restore → not ready with the verdict's detail and the manager's reason; unanswered list never restored; the restore call's transport budget; the post-open and no-open exhaustion counts. `tests/qs_client/test_readiness.py`: the `plans_empty` sentence names #838 and the gesture.

## Hardware verification

**OWED** (worker host, manager idle, no scan, no PV write): (1) `systemctl restart geecs-qserver-ready` on a healthy manager must log `ready: N allowed plans` with **no** "restoring the lists" line in its journal (the heal path is not taken when the first read is complete). (2) The heal path: `qserver permissions reload lists` by hand on the idle manager must log the manager's "Reloading lists of allowed plans and devices" and leave `qserver plans allowed` complete — the deterministic part; reproducing the timed-out download itself on demand is not practical.

### DIFF
--- a/GeecsBluesky/CHANGELOG.md
+++ b/GeecsBluesky/CHANGELOG.md
@@ -5,6 +5,22 @@ All notable changes to `geecs-bluesky` are documented here.
 Format follows [Keep a Changelog](https://keepachangelog.com/en/1.1.0/).
 
 
+## [0.82.4] - 2026-09-12
+
+### Fixed
+
+- `geecs-qserver-ensure-ready` heals the manager that is idle with its
+  environment open and `plans_allowed` **empty** (GEECS-Plugins#838: the
+  manager's own download of the lists from the worker timed out while the
+  host thrashed, and every submission was then refused "not in the list of
+  allowed plans" while `status` looked healthy): a list still empty or
+  incomplete after the settle window is downloaded again once through
+  `environment_update`, with the settle window applied again — so
+  `systemctl restart geecs-qserver-ready` recovers it without a manager
+  restart.  The shared `plans_empty` verdict (Console banner, MCP
+  preflight) names that cause and that gesture; `qserver/README.md`
+  Troubleshooting carries the entry.
+
 ## [0.82.3] - 2026-09-12
 
 ### Changed
--- a/GeecsBluesky/CHANGELOG.md
+++ b/GeecsBluesky/CHANGELOG.md
@@ -14,11 +14,14 @@ Format follows [Keep a Changelog](https://keepachangelog.com/en/1.1.0/).
   manager's own download of the lists from the worker timed out while the
   host thrashed, and every submission was then refused "not in the list of
   allowed plans" while `status` looked healthy): a list still empty or
-  incomplete after the settle window is downloaded again once through
-  `environment_update`, with the settle window applied again — so
-  `systemctl restart geecs-qserver-ready` recovers it without a manager
-  restart.  The shared `plans_empty` verdict (Console banner, MCP
-  preflight) names that cause and that gesture; `qserver/README.md`
+  incomplete after the settle window is restored once from the worker's
+  on-disk copy through `permissions_reload(restore_plans_devices=True)`
+  — the copy the worker writes at every environment open — with the
+  settle window applied again, so `systemctl restart geecs-qserver-ready`
+  recovers it without a manager restart.  (`environment_update`, the fix
+  #838 first named, re-downloads only when the worker's namespace changed
+  and is not used.)  The shared `plans_empty` verdict (Console banner,
+  MCP preflight) names that cause and that gesture; `qserver/README.md`
   Troubleshooting carries the entry.
 
 ## [0.82.3] - 2026-09-12
--- a/GeecsBluesky/CLAUDE.md
+++ b/GeecsBluesky/CLAUDE.md
@@ -207,7 +207,9 @@ namespace, so never import a stray generator into the profile),
 `deploy/` (the manager and `geecs-qserver-ready` units + runbook).  **A
 running service means ready (#793)**: the readiness unit runs
 `geecs-qserver-ensure-ready` after every manager start — wait, open if
-closed, wait for idle, assert `plans_allowed ⊇ GEECS_PLAN_NAMES`.  Read
+closed, wait for idle, assert `plans_allowed ⊇ GEECS_PLAN_NAMES` (and,
+when the list is empty or incomplete with the environment up, restore it
+once from the worker's on-disk copy via `permissions_reload`, #838).  Read
 `qserver/README.md` first; its Troubleshooting section is the empirical
 contract (permissions file, `--keep-re`, manager-restart-after-install,
 failed-items-requeue-at-front, CLI parses Python literals not JSON).
--- a/GeecsBluesky/geecs_bluesky/qs_client/client.py
+++ b/GeecsBluesky/geecs_bluesky/qs_client/client.py
@@ -267,9 +267,12 @@ def readiness_verdict(
             False,
             "plans_empty",
             "The RE Manager lists no allowed plans although its worker environment "
-            "exists — the startup profile registered nothing (import error?) or the "
-            "permissions file allows nothing; see GeecsBluesky/qserver/README.md, "
-            "Troubleshooting.",
+            "exists — its download of the plan list from the worker timed out "
+            "(GEECS-Plugins#838), the startup profile registered nothing (import "
+            "error?), or the permissions file allows nothing. On the worker host "
+            "`systemctl restart geecs-qserver-ready` re-downloads the lists "
+            "(`environment_update`) without restarting the manager; see "
+            "GeecsBluesky/qserver/README.md, Troubleshooting.",
             allowed,
         )
     expected = (
--- a/GeecsBluesky/geecs_bluesky/qs_client/client.py
+++ b/GeecsBluesky/geecs_bluesky/qs_client/client.py
@@ -270,9 +270,10 @@ def readiness_verdict(
             "exists — its download of the plan list from the worker timed out "
             "(GEECS-Plugins#838), the startup profile registered nothing (import "
             "error?), or the permissions file allows nothing. On the worker host "
-            "`systemctl restart geecs-qserver-ready` re-downloads the lists "
-            "(`environment_update`) without restarting the manager; see "
-            "GeecsBluesky/qserver/README.md, Troubleshooting.",
+            "`systemctl restart geecs-qserver-ready` restores the lists from the "
+            "worker's on-disk copy (`permissions_reload`, restore_plans_devices) "
+            "without restarting the manager; see GeecsBluesky/qserver/README.md, "
+            "Troubleshooting.",
             allowed,
         )
     expected = (
--- a/GeecsBluesky/geecs_bluesky/qserver_ready.py
+++ b/GeecsBluesky/geecs_bluesky/qserver_ready.py
@@ -20,7 +20,12 @@ only that check catches it.  The plan list is read through the same
 ``qs_client.readiness_from_reads`` assembly the pre-submit ``worker_ready``
 check runs (one definition of ready); after an open this run requested it
 is re-read for a short settle window, because the manager reports the
-environment up before its own plan-list download has landed.  Exit codes:
+environment up before its own plan-list download has landed.  A list that
+is still empty or incomplete after that is downloaded again once through
+the manager's ``environment_update`` (#838: the manager's own download can
+time out and leave it idle, environment open, knowing no plans — the unit
+re-run, ``systemctl restart geecs-qserver-ready``, heals that without a
+manager restart), and the settle window applies again.  Exit codes:
 0 ready; 1 not ready (the message says exactly what was found); 2 usage.
 
 The address asserted is the manager on **this** host (loopback), or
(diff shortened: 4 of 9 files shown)


skbarber:
# Adversarial review of #846

*(fresh-context reviewer subagent, per `/land`; pasted in full — the review of the first commit `4dcf5511`, then the reviewer's confirmation of the fix commit `40800530`)*

Reviewed the single commit `4dcf5511` on `fix/838-readiness-empty-plan-list` (contains the base tip) against the installed `bluesky-queueserver 0.0.25` source. The 42 tests in `tests/test_qserver_ready.py` + `tests/qs_client/test_readiness.py` pass from the branch. No redundancy found: the only `environment_update` caller in the repo is this one (`grep` over every package, docs, deploy, Planning); `qserver_ready.py` is the right home for a unit-side heal, and the verdict sentence belongs in `qs_client/client.py` where `plans_empty` is defined. No hygiene/facility-literal/scan-folder issues in the diff. The findings below are all correctness.

### 1. Medium — the heal is not the mechanism the PR describes; it is diff-gated in the worker, and the stated reason it is preferred over `permissions_reload` is not supported by the source

`qserver_ready.py:97-104` and the README entry (`qserver/README.md:153-157`) say the update "asks the worker for the lists again and regenerates the allowed lists". In 0.0.25 that is not what happens:

- `environment_update` → `_environment_upload_script(script="", update_lists=True)` (manager.py:2945) → worker `_load_script_into_environment` (worker.py:788). With an empty script the worker regenerates its lists from the same namespace and sets `_existing_plans_and_devices_changed = not compare(...)` against **its own** stored copy (worker.py:823-848).
- The manager re-downloads only when the worker's state poll reports `plans_and_devices_list_updated` (manager.py:756). The manager's list download is not otherwise triggered by the update; there is no unconditional "download again" path.

So on a namespace whose descriptions regenerate identically, an empty-script update produces **no re-download** and the unit's second settle window reads the same empty list. The hand fix on 2026-09-11 therefore worked for a reason the PR does not state — either the worker's pending flag was still set (in which case the manager's own 0.5 s poll would have been retrying anyway, manager.py:227/3834), or the GEECS namespace's descriptions do not compare equal across regenerations (plausible with lazily materialised devices, but unpinned). The PR's own "Hardware verification" section explicitly exercises only the no-op path ("no `environment_update` line"); the heal path is assumed from one observation.

Related: the README's instruction "Do NOT `permissions_reload(restore_plans_devices=True)`: that reloads the stale on-disk copy" (README.md:157-159) is not supported. Only the **worker** writes `existing_plans_and_devices.yaml` (worker.py:782, at environment open via `ENVIRONMENT_OPEN` at :1457 and after a changed script load at :850); the manager never writes it. The issue's evidence — "the file was not rewritten (still the 20:09 copy)" — is exactly the expected state: the 20:09 copy *is* the running worker's namespace as of the open. For the #838 shape (no script changed the namespace after the open), `permissions_reload(restore_plans_devices=True)` (manager.py:2037-2069 → `_update_allowed_plans_and_devices(restore_plans_devices=True)`, :1478-1494) restores the correct lists deterministically, is not diff-gated, and does not require the manager to be idle (it validates only the lock key, `check_queue=True`). The "stale" case is a deploy that changed the plan set *without* re-opening — but then the worker process is equally stale and `environment_update` does not help either.

Suggested disposition: either (a) keep `environment_update` but state the actual gating in the module comment and README, and move the heal path itself into the owed hardware verification; or (b) switch the heal to `permissions_reload(restore_plans_devices=True)` and drop the README's "Do NOT" sentence. I lean (b) — it is the one call whose effect the source guarantees.

### 2. Low — under the #838 condition the update request itself can time out at the transport and the unit reports a false negative

`_zmq_request` uses `timeout=int(POLL_S * 1000)` = 2 s per call (`qserver_ready.py:114-124`). `environment_update` is answered only after the manager has completed `_worker_command_reserve_kernel` and `_worker_command_load_script` round trips to the worker (manager.py:1338-1350, each with `_comm_to_worker_timeout_long` = 10 s). On the thrashing host that produces #838, the manager can accept and start the task yet answer after 2 s; `_call` then raises `NotReady("manager did not answer 'environment_update'")` and the unit exits 1 while the heal proceeds. Worth a longer transport timeout for this one call and a journal line that says whether the update was *accepted*.

### 3. Low — a refused update drops the verdict's own detail from the NOT READY message

`qserver_ready.py:328`: when the manager refuses, `_call` raises with only the manager's reason; the `plan_missing` detail (which plan is missing, what is listed) that the pre-#846 unit reported is now only in the preceding log line. Consider `raise NotReady(f"{verdict.detail}; and the re-download was refused: {exc}")`.

### 4. Low — the unit's behaviour description in the contract files is now incomplete

`GeecsBluesky/CLAUDE.md:208-210`, `qserver/deploy/DEPLOYMENT.md:6-7` and `:159-162`, and the header comment of `qserver/deploy/geecs-qserver-ready.service` still describe the four steps "wait, open if closed, wait for idle, assert". The module docstring was updated; the prose contract files were not. The unit template is under the §10.5 end-of-branch touch rule, so leave it; the two Markdown files are not.

### Not findings (checked and refuted)

- `manager_state == "executing_task"` during the update does not perturb the verdict: `readiness_verdict` inspects only `creating_environment` / `initializing`, so the settle re-reads stay in `_SETTLING_STATES`.
- The tests would fail on the obvious wrong implementations: a loop instead of once, re-download on `plans_unknown` (pinned absent), `plans_empty`-only (the stale-list test would raise).
- `systemctl restart geecs-qserver-ready` does not restart the manager: `PartOf=`/`Requires=` propagate only from `geecs-qserver` to the readiness unit.
- No other package or test quotes the old `plans_empty` sentence; the console/MCP preflight keeps `plans_empty` fail-closed (`submit_preflight.py:176`).

---

## Confirmation of the fix commit `40800530` (same reviewer)

Reviewed the full PR diff at `40800530` against the installed `bluesky-queueserver 0.0.25`; the 43 tests pass from the branch.

- **Finding 1 — CONFIRMED resolved.** `qserver_ready.py:346-348` now calls `permissions_reload` with `{"restore_plans_devices": True}`; the fake asserts exactly those params and `environment_update` is pinned absent. The premises hold in the source: `--update-existing-plans-devices` default `ENVIRONMENT_OPEN` (`start_manager.py:415-419`), `--user-group-permissions-reload` default `ON_STARTUP` (`:438-442`, accepted by `_permissions_reload_handler`), the launcher passes neither flag, manager and worker share one `existing_plans_and_devices_path`, and the worker writes the file inside its open sequence **before** it reports `IDLE` (`worker.py:1457` precedes `:1475`). `qserver permissions reload lists` is a real CLI form. The restore regenerates the allowed lists unconditionally. One property worth knowing, not a defect: `restore_permissions` defaults `True`, so the call also re-reads `user_group_permissions.yaml` from disk — the same file the manager was launched with.
- **Finding 2 — CONFIRMED resolved.** `RESTORE_TIMEOUT_S` (10 s) applies only to `permissions_reload`; the test pins both values and is hermetic. The reply's `msg` is logged. The manager answers this call synchronously after the lists are already regenerated, so the post-restore settle window is belt-and-braces rather than load-bearing — fine.
- **Finding 3 — CONFIRMED resolved.** The refused restore re-raises with `verdict.detail` + `user_group` + the manager's reason; the test pins both halves, which also covers the lock-key refusal path.
- **Finding 4 — CONFIRMED resolved.** `CLAUDE.md` and `deploy/DEPLOYMENT.md` carry the restore step; the unit template untouched per the end-of-branch rule. Cosmetic only: the DEPLOYMENT.md edit leaves a hard break mid-sentence ("A\nseparate unit on\npurpose") — renders fine in Markdown.
- **New findings introduced by the fix: none surviving.** Checked: a missing on-disk file makes the restore return `success: False` → the combined NotReady message, not a crash; the `plan_missing` case after a deploy without re-open surfaces the missing plan with the README's `systemctl restart geecs-qserver` gesture; the worker-side permissions push is a background regeneration and cannot disturb a running plan.

---

## Disposition

- **Finding 1: fixed** in `40800530` (option b — `permissions_reload(restore_plans_devices=True)`; `environment_update` removed from code, README, CHANGELOG; the README's "Do NOT permissions_reload" sentence replaced by the correct account; the PR body and #838's own text corrected). Confirmed by the reviewer.
- **Finding 2: fixed** in `40800530` (`RESTORE_TIMEOUT_S = 10 s` for the one restore call, reply logged, pinned). Confirmed.
- **Finding 3: fixed** in `40800530` (verdict detail + manager reason in the NotReady message, pinned). Confirmed.
- **Finding 4: fixed** in `40800530` (CLAUDE.md + DEPLOYMENT.md prose; the unit template deliberately untouched). Confirmed. The mid-sentence line break in DEPLOYMENT.md is **waived** as cosmetic (renders as one sentence).

PR #848, opened 2026-09-12
GEECS-Schemas 0.22.0 / GeecsBluesky 0.82.4 / GEECS-Console 0.32.0: delete the action-library converter; the corpus is ActionPlanLibrary v1

skbarber:
Follow-up owed from #827 ("the Undulator `actions.yaml` on the configs `main` is still the legacy dialect … should be regenerated as an `ActionPlanLibrary` v1 document and the converter deleted — its own small change"). Per the disposable-configs policy: regenerate once, delete the converter root and stem.

## What

- **Configs repo (commit ready, push owed — see below):** `Undulator/action_library/actions.yaml` (21 plans, `s1h_exercise` included) and `Thomson/action_library/actions.yaml` (2 plans) regenerated as `ActionPlanLibrary` v1 documents — same plans, same steps, same values (`set/wait/get/execute` → `set/wait/check/run`), round-tripped through the converter before its deletion. `assigned_actions.yaml` (the legacy GUI's pinned-button list; nothing in the repo reads it) dropped from both experiments.
- **GEECS-Schemas 0.22.0 (minor, Removed + Added):** `convert.actions` (`convert_action_library`, `convert_assigned_actions`) deleted with its legacy fixtures (`actions.yaml`, `assigned_actions.yaml`); `tests/fixtures/actions/actions_undulator.yaml` is now the regenerated document; the `Amp4_DUMP_HP` golden is unchanged and pins it (`test_action_plan.py::TestCorpusDocument`); the corpus walk validates every deployed library as an `ActionPlanLibrary` and asserts no `assigned_actions.yaml` survives; `generate_golden.py`, README and docstrings updated. **The legacy-dialect guard lives on the schema** (after the review): `ActionPlanLibrary._refuse_legacy_dialect`, a `model_validator(mode="before")` in the `AnalysisDiagnostic._refuse_v1_layout` pattern, refuses an `actions:` document from `model_validate` with the regeneration named — every consumer gets the one message, none carries its own guard.
[...]

PR #850, opened 2026-09-12
GeecsBluesky 0.83.0 / GEECS-Schemas 0.22.0: phase 2b — the gated batch, the per-shot sampler, the non-essential stream (#807)

skbarber:
## Runbook and results (worker host, laser off → `HTU-LaserOFF`)

**Hardware acceptance run 2026-09-12 (daytime, Sam present; laser off, `HTU-LaserOFF`; worker `~/qs-checkout` on this branch's head, environment cycled in place): A1, A2, A3, A4, A6, A7, A8 PASSED; A5 deferred (Sam's call, below); plus a broader set on the devices Sam turned on (second table).** The deploy shape that worked: no reinstall (worker code only), `qserver environment close` / `open` + `geecs-qserver-ensure-ready` after each `git pull` — no `sudo`, no unit restart.

| # | result | scan | evidence |
|---|---|---|---|
| A1 | **PASSED** | Scan003 | 5 frames per stack, stamps 1.0 s apart, one datum `{0, 5}` each, `shots` 5 rows, `shot_clock: UC_Amp4_IR_input`, stop `success`. Third attempt: the first two found the device-dir mkdir (fixed, `1518b1a3`) and the plugin's arm-on-first-frame after the camera-server restart (operator note in the qserver README; #854). |
| A2 | **PASSED** | Scan005 | 3 datums per camera `{0,3}` `{3,6}` `{6,9}`, 9 frames, 9 rows with `u_s1h-current-position` −0.5/0/0.5 ×3 and bins 1/2/3, box OFF between steps. First attempt counted from the previous session's stale `NumCaptured_RBV` (14 frames) → `zero_count` at the first arm (`53479958`, #853). |
| A3 | **PASSED** | Scan015 | `U_BCaveICT` (a Picoscope ICT, `save_images: false` → its `.scalars`) as the only device: no `primary` stream, 6 `shots` rows carrying `u_bcaveict-python_results_cha/chb`, `-timestamp`, `-acq_timestamp` and the `U_S1H` readback, bins 1/1/1/2/2/2, `shot_clock: U_BCaveICT`, stop `success`. (First try used `U_HP_Daq`, an analog control with no trigger and a LabVIEW file writer — refused as a "camera"; the refusal now says "native-saving device(s)", `cb95e0d2`.) |
| A4 | **PASSED** | Scan012 | 5 strict rows, `uc_amp3_ir_input_stream` with one datum `{0, 5}`, stack check "5 frame(s) …, 5 referenced", no close warning. Scan007/Scan011 succeeded with the stream's `complete` timing out at the close (the non-essential arm baselined on the stale count) → the non-essential arm zeroes too, and `complete` / `collect` are separate contingencies (`4f2f89f0`). |
| A5 | **DEFERRED** | — | A gateway dying mid-run needs a human restart of the camera server's service during a scan (the `:restart` PV); Sam judged it unlikely enough to defer. The tolerant close (complete / collect / unstage each skipped and logged, never the item's failure) is pinned on mocks. |
| A6 | **PASSED** | Scan013 | Immediate pause 2.3 s into the batch: "gated step interrupted by a pause after 2/5 shot(s) — retaking the step from its first shot"; box OFF for the 6 s pause (DG645 `Single shot` / 0.5 V); "rewound 3 → 0 frame(s) (step baseline)"; exactly 5 frames per stack, `shots` 5 rows, stop `success`. |
| A7 | **PASSED** | Scan014 | Deferred pause after step 1's OFF (paused 2.3 s after the step's OFF, held 6 s); resume moved to the next step; 9 frames, 3 datums per camera, no retake line, bins 1/2/3, stop `success`. `U_S1H` restored to 0 A. |
| A8 | **PASSED** | Scan010 | 3 strict rows ≥ 4 s apart, `shot_period: 4.0` in the start doc. Two fixes on the way: the strict first arm zeroes a plugin's stale count (`9ef41fbc`), and the throttle sleeps *before* the trigger of each attempt (a sleep after the trigger timed out the shot, `670b21d6`). |

### Broader set (devices Sam turned on, same day)

| run | plan | scan | result |
|---|---|---|---|
| two servers + scalar-only + ICT | `count([UC_Amp4_IR_input, UC_ModeImager, U_ESP302_01, U_BCaveICT.scalars], 5, gated)` | Scan020 | **PASSED**: two stacks of 5 on two camera servers (Amp4's box and ModeImager's), stamps 1 s apart on both; `shots` 5 rows with the ESP302's three untriggered positions, the ICT's scalars and its own stamps (1 s apart, in step with Amp4's), `shot_clock: UC_Amp4_IR_input`. First try (Scan016) failed at ModeImager's arm ("no frame within 8 s"): its plugin needed the same priming as this morning's Amp3/Amp4 — Sam fired shots in internal mode, then it armed with a clean count (#854's boot-order gap is the likely reason it needed it). |
| Aerotech stage, relative | `rel_scan([UC_Amp4_IR_input, U_ESP302_01], U_CompAerotech:Position.Axis1, -20, 20, 5, shots_per_step=2, gated)` (µm, tol 1) | Scan018 | **PASSED**: 5 steps of 2 shots, 10 frames, 10 rows with the axis readback stepping 41322 → 41362 µm, the position restored by the plan (41342.24 → 41342.18, inside the 1 µm tolerance). |
| ModeImager stage, relative | `rel_scan([UC_ModeImager, U_ESP302_01], U_ModeImagerESP:Position.Axis 1, -4, 6, 5, shots_per_step=2, gated)` (mm about −16, tol 0.015) | Scan019 | **PASSED**: 10 frames of 950×1200, readback −20 → −10 in 2.5 mm steps, restored to −16.0. |
| native-saving essential, gated | `count([UC_Amp4_IR_input, U_BCaveICT], 3, gated)` with the ICT's file writer on | — | **REFUSED at preflight, as designed**: "gated acquisition: native-saving device(s) without a file plugin: U_BCaveICT — … Use acquisition='strict' or record its scalars only". |
| native-saving essential, strict | `count([UC_Amp4_IR_input, U_BCaveICT], 3)` | Scan021 | **PASSED**: 3 strict rows with the ICT's scalars, 3 frames matching the rows' stamps, and the ICT's three scope traces written natively (`U_BCaveICT/*.tdms`). |

**Engine bug found and fixed by the broader set (`18010336`): every `rel_*` plan failed at its first move** (Scan017: `unsupported operand type(s) for +: 'SignalR' and 'float'`) — bluesky's relative wrappers fall back to `obj.position`, which on a `CaMotor` is the readback *signal*. `CaSettable` now implements `Locatable` (`locate()` returns the streamed readback as both fields — the gateway's `:SP` is the last put through the gateway, not where the device is), pinned by a `rel_scan` test asserting the five relative puts and the restoring put. The registry tests had only ever validated `rel_*` arguments.

Deploy state after the run: worker on `18010336`; configs `main` at `ddc08af` on the share clone; the nine gateways untouched (2a's 0.9.0). Issues filed from the run: #852 (worker-side liveness gate), #853 (the plugin should post `NumCaptured_RBV=0` at `Capture=1`), #854 (gateway boot-order subscription gap).

The runbook as written before the run (kept for the expected numbers):

| # | run | expect |
|---|---|---|
| A1 | `count([UC_Amp4_IR_input, UC_Amp3_IR_input], 5, acquisition="gated", trigger_profile="HTU-LaserOFF")` | stop `success`; `ScanNNN/<cam>/<cam>.h5` holds **exactly 5 frames each**, stamps 1.0 s apart (±20 ms); one `stream_datum` per camera, `indices {0, 5}`; the `shots` stream has **5 events** with `uc_amp4_ir_input-acq_timestamp` (the clock, `shot_clock: UC_Amp4_IR_input` in the start doc) and `bin_number = 1`; `scan.log` stack check: "5 frame(s) … 5 referenced by the stream's datums" for both; the box reads OFF between kickoff and SCAN, OFF after, STANDBY at the end (`ShotControl.state` in the descriptors' configuration). Wall time ≈ 5 s + 2 drain waits (~1.3 s each). |
| A2 | `scan([UC_Amp4_IR_input, UC_Amp3_IR_input], U_S1H.current, -0.5, 0.5, 3, shots_per_step=3, acquisition="gated", trigger_profile="HTU-LaserOFF")` — **moves U_S1H** | 3 datums per camera (`{0,3}`, `{3,6}`, `{6,9}`), 9 frames per stack, 9 `shots` rows with `u_s1h-current-position` ≈ −0.5/0/0.5 (three each) and `bin_number` 1/2/3; each step's stamps 1 s apart; the box OFF between steps (the descriptor config, and `scan.log`'s put trace). ScanInfo `Shots per step = 3`. |
| A3 | `scan([U_BCaveICT], U_S1H.current, -0.5, 0.5, 2, shots_per_step=3, acquisition="gated", …)` — a triggered scalar as the clock, no camera | no `primary` stream at all; 6 `shots` rows carrying `u_bcaveict-*` scalars with **their own stamps 1 s apart**, `shot_clock: U_BCaveICT`; `scan.log` has no stack check. (Pick any triggered scalar device that publishes `acq_timestamp` with the laser off; an ICT may read zero charge — the stamp is what counts.) |
| A4 | `count([UC_Amp4_IR_input], 5, non_essential=[UC_Amp3_IR_input], trigger_profile="HTU-LaserOFF")` — strict with a non-essential camera | 5 strict rows for Amp4 (frames == rows, as #823); `uc_amp3_ir_input_stream` descriptor + one datum covering every edge Amp3 saw during the run — under `HTU-LaserOFF` ARMED delivers single shots only, so **5 frames, `indices {0, 5}`** (with free-running edges it would be every edge, more than the rows); the stack check on Amp3 reports the count only ("5 frame(s) …, 5 referenced"). Nothing waited on Amp3: run time ≈ the 5 fires. |
| A5 | A4 again with Amp3's gateway restarted mid-run (`UC_Amp3_IR_input:hdf1:` unreachable) | the run completes with its 5 rows; Amp3's stream carries whatever it wrote; no error from the plan. |
| A6 | A1 with an **immediate pause** ~2 s in (`RE.request_pause()` / the console's pause when it is wired, or `qserver re pause immediate`), resume after ≥5 s | `scan.log`: "gated step interrupted by a pause after N/5 shot(s) — retaking the step from its first shot"; the box OFF during the pause (verify at the DG645: no shots); after resume the stack holds **exactly 5 frames** (the partial ones rewound: "rewound k → 0 frame(s) (step baseline)" in the journal), `shots` has 5 rows, stop `success`. |
| A7 | A1 with the manager's **deferred** pause between steps of A2 | the pause lands after a step's OFF; resume continues with the next step; 9 frames total, no retake line. |
| A8 | `count([UC_Amp4_IR_input], 3, shot_period=4.0, trigger_profile="HTU-LaserOFF")` (#840) | 3 strict rows, stamps **≥ 4 s apart**; `shot_period: 4.0` in the start doc. |

Rollback: `git checkout feature/native-bluesky-plans` in `~/qs-checkout`, reinstall, restart. The share clone and the nine gateways are untouched by this PR (2a's 0.9.0 is what it reads).

## What

Phase 2b of the native-Bluesky rebuild (#807, `08_gated_batch.md` §5 item 2 as amended by #841; `03` §10.9): the gated batch, the per-shot sampler, the non-essential stream, the repeat-the-step resume, the preflight rules, `essential` on `PresetDevice`, and #840's `shot_period`. GeecsBluesky 0.82.3 → **0.83.0**, GEECS-Schemas 0.21.0 → **0.22.0**.

- **`plans/gated.py`** — `gated_take_reading` (the §4.2 step: OFF → prepare(D, N) + prepare(S, N) → declare once → kickoff → SCAN → complete → OFF → drain wait → `truncate_to_quota` → collect `primary` (datums) + `shots` (events)), `gated_per_shot` / `gated_per_step`, `non_essential_wrapper` (stage + unbounded prepare + declare + kickoff after `open_run`; complete + collect-alone before `close_run`), `gated_run_bracket` (OFF → STANDBY), `shot_clock` (first plugin camera, else first triggered device — the recommended default; flagged below).
- **`devices/sampler.py`** — `ShotSampler`: Flyable + EventCollectable + Preparable over every non-plugin member; ticks on the clock's `acq_timestamp` advancing past the value read at kickoff (the subscribe echo and re-published stamps are not ticks); one row per tick with every member's latest cached reading and the tick's stamp as the clock column; `complete` after the quota or `GeecsTriggerTimeoutError` when the clock stops; `cancel_step` settles it.
- **`devices/detector.py`** — fly mode (`kickoff` sets, `trigger` clears; `wait_for_idle` a no-op in fly), `complete` translating the count timeout and settling quietly after `mark_abandoned` / `abandon_step`, `truncate_to_quota`, `rewind_to_step_baseline`, `step_baseline`; a fly `prepare` is a **`FlyTriggerInfo`** (`gated_trigger_info`, `UNBOUNDED_TRIGGER_INFO` — the type says fly, not the event count, so a batch of one is one too), takes the streamable logic only and is refused on a camera without a plugin. `ShotControl.pause_count`.
- **`plans/registry.py`** — `acquisition`, `non_essential`, `shot_period` keyword-only on every scan verb (in the start doc with `shot_clock`); strict keeps ARMED → STANDBY, gated brackets OFF → STANDBY; `shot_period` (strict only) in `plans/strict.py`.
- **`qs_client`** — `expand_preset`: `essential: false` → `non_essential` (refused with `save_images: false`), `acquisition` validated; `submit_preflight.acquisition_refusal`: the `hdf`-child rule for non-essential devices and for a gated run's essential cameras, the `acq_timestamp`-child rule for the shot clock.
- **`callbacks.StackCheckCallback`** — rows and datums tracked per stream; a datum-only stream is checked by count.
- **GEECS-Schemas** — `PresetDevice.essential` (default `True`; corpus needs no regeneration); schema reference regenerated.
- **Docs** — `08` status + §5 item 2 (what the build found: the fly-prepare data-logic rule, the `count` quota), `03` §2 row + §10.9, GeecsBluesky `CLAUDE.md` scan path, CHANGELOGs.

## Design defaults taken (flagged for Sam, cheap to veto)

- **The shot clock** with several essential triggered devices: the first plugin-backed camera in the plan's detector order, else the first triggered device (a scalar with a stamp or a camera's `.scalars` view). `shot_clock` rides in the start document so the choice is visible per run.
- **A fly prepare leaves the per-event data logics out.** Not in `08`: the stock `_update_prepare_context` refuses a readable provider with `number_of_collections > 1` ("Multiple collections not supported"), and `LvNativeFileDataLogic.prepare_single` would switch LabVIEW-native saving on for the batch. So in gated mode a plugin camera writes **no PNGs** (the plugin's stack is the record) and its scalars ride in the stack (2a) — the sampler carries every other device's. A camera without a plugin cannot be an essential camera of a gated run (refused by the step itself, worker-side, and by the preflight) or a non-essential device (refused at prepare and at preflight); its `.scalars` view rides in the sampler. A device listed both as a detector and as non-essential is refused by the bound plan — compared by owner, so a camera's `.scalars` view essential with the camera itself non-essential is refused too (its one acquire logic would be in fly mode while the view expects the strict stamp wait). A non-essential device whose complete, collect or unstage fails from the run's close on (a gateway that went away mid-run) is logged and skipped, never the item's failure (the review's R4/R5; A5 in the runbook).
- **A gated `count` takes the whole batch on the hook's first call** (`num` is the quota; the stock plan repeats the hook `num` times and the later calls are no-ops); `num=None` (infinite) is refused, and so is a nonzero `delay` (the repeat loop would sleep it after every no-op iteration — a delay between shots is a strict notion; Codex review).
- **The drain wait** is `TRIGGER_PERIOD_S` (1.0 s, the HTU rep rate) + the largest drain offset of the step's cameras + 0.25 s, waited at the first step's open, after every batch, and before a retake. A module default rather than a site value (the review flagged it against "facility values have one home"): the trigger profile carries no period today; a slower site would see the stack check flag the count mismatch loudly. The knob (a `period` on the trigger profile) is deferred.
- **The interruption check** reads `ShotControl.pause_count` the moment the interrupted wait returns and again after the drain wait (a pause landing in the drain, whose resume restores SCAN, is a retake too); the batch's statuses are marked abandoned **synchronously, before any further message**, so a count timeout landing in the resume-to-settle window is pardoned rather than thrown into the plan (the review's finding 1, pinned by a no-in-flight-edge pause test). The retake costs up to one `shot_timeout` (3 s) for the abandoned `complete` statuses to settle — the alternative (cancelling ophyd-async statuses from the plan) was not taken.
- **`StackCheckCallback`** does not compare a gated stack's stamps with the `shots` rows (2c's join) — count only.

## Not in this PR

2c (the s-file from streams, `08` §4.5): a gated run writes **no s-file** yet (`SFileCallback` skips a run with no primary events, with a log line), and a strict run with a non-essential list writes its s-file from the primary events only. The Console follows the client-expansion change later; the preset editor keeps working (a schema-driven form; `essential` defaults on).

## Version collisions

GEECS-Schemas 0.22.0 is also the bump of #848 (the converter deletion) and GeecsBluesky is bumped by #846 and #848 (both 0.82.4) — all off the same base. Whichever merges later needs a one-line rebase of `pyproject.toml` + the CHANGELOG headings (this PR's GeecsBluesky entry becomes 0.83.0 over 0.82.5, say).

## Tests

`./scripts/check.sh --all` (at `bd33a4f0`; the review fixes `02760e86`, `db812dab` and `3743007d` re-ran `check.sh GeecsBluesky`: **407 passed, 3 skipped, 1 deselected**; the acceptance commits through `cb95e0d2`: **408 passed, 3 skipped, 1 deselected**, `test_gated_plans.py` now 19): **GeecsBluesky 401 passed, 3 skipped, 1 deselected** (was 369); **GEECS-Schemas 419 passed, 7 deselected**; root 82, ImageAnalysis 276, ScanAnalysis 192, Data-Utils 329, CAGateway 137, PvaGateway 49, Console 814, Core 100, Portal 209, MCP 103 — all passed. GEECS-LogTriage did not run locally (its venv lacks `geecs_data_utils`, an environment problem on this machine, not this diff; CI runs it). New: `tests/test_gated_plans.py` (13: gated count with two plugin cameras — 3 frames each, the in-flight fourth trimmed, `shots` rows with the clock stamp and the bin; gated scan — one datum per step, motor readbacks and bins in the rows, rewinds baseline + quota; a triggered scalar as the clock with no camera; no triggered device refused; a stalled camera fails with the box OFF; an **immediate pause mid-batch** — OFF, resume restores SCAN, the step retaken from its baseline, exactly `quota` frames; the bound plans' metadata; bad mode / throttled gated refused; `shot_period` throttling; a non-essential camera streamed for a strict run, one that never delivers not holding the run, a non-essential with a gated run), `tests/test_sampler.py` (6), the detector's fly mode / rewinds / abandon / refusal (4 in `test_hdf_plugin_detector.py`), the datum-only stack check, the preset expansion (3) and preflight rules (4), the registry's new keywords and the manager resolving `non_essential` names. The pause mechanics were verified against the installed bluesky 1.15.0 with a probe before the design was fixed (a non-rewindable region replays nothing on resume; the interrupted wait finishes or raises its status's failure).

## Hardware verification

**Done 2026-09-12 for A1, A2, A3, A4, A6, A7, A8 and the broader set** (the two tables at the top, scan numbers on `26_0912`). **Deferred:** A5 (a human gateway restart mid-scan). The acceptance commits (`1518b1a3`, `53479958`, `9ef41fbc`, `670b21d6`, `4f2f89f0`, `cb95e0d2`) carry the reviewer's confirmation and its disposition (`4f99b55c`) as comments below; `18010336` (`locate`) has its own confirmation comment.


skbarber:
# Adversarial review of #850

*(fresh-context reviewer subagent, per `/land`; pasted in full — the review of `bd33a4f0`, then the reviewer's confirmations of the fix commits `02760e86`, `db812dab`, `3743007d`, `87246263`)*

Scope: `phase/06-gated-batch` (bd33a4f0) against `feature/native-bluesky-plans`, read in a throwaway worktree, API claims checked against the installed ophyd-async 0.19.3 / bluesky 1.15.0 / bluesky-queueserver 0.0.25, and four probe plans run on the PR's own mocks (the mocks reproduced findings 1–4). The PR's named worker suites pass (138 in `test_gated_plans`/`test_sampler`/`test_hdf_plugin_detector`/`test_plan_registry`/`qs_client`/`test_callbacks`).

## 1. HIGH (correctness) — a pause resumed 1.7–3 s later aborts the gated run

`gated.py:228-244`: only `complete_all` is inside the `try/except FailedStatus`. On an immediate pause the camera `complete` statuses and the sampler task keep running with `abandoned=False`; `abandon_step`/`cancel_step` are only reached at line 242, **after** `mv(OFF)` (232) and `sleep(drain)` (236). If a status times out in that gap (T_last + `shot_timeout` landing between the resume and the abandon, i.e. for a resume roughly 1.7–3 s after the last frame with the real 3 s timeout and ~1.35 s drain), `_status_object_completed` (`run_engine.py:2366-2388`) stashes `FailedStatus` and throws it into the plan at the `sleep`/`mv` yield, where nothing catches it. Reproduced on the PR's mocks with no in-flight edge at OFF (the PR's `GatedBox._stop` always lands one 25 ms after every OFF, which resets the count timeout and is why the pause test never sees this), `shot_timeout=0.4`, resume 0.30 s after the pause → `FailedStatus(GeecsTriggerTimeoutError(...))` raised from `gated.py:236 (bps.sleep(drain))`, stop `exit_status: fail`. Real-world: a pause with no edge in flight (a ~90% case at 1 Hz), resumed after a couple of seconds. Fix direction: mark the abandonment synchronously, before yielding any further message, then keep the awaited settle. Add a no-in-flight-edge variant of the pause test.

## 2. MEDIUM (correctness) — a gated step with quota 1 is treated as a strict prepare: LabVIEW-native saving switches on

`detector.py:749-752` classifies fly mode as `number_of_events != 1`. `shots_per_step=1` is the **default** of every gated scan verb, so those steps go through the stock `_update_prepare_context` with every data logic: `LvNativeFileDataLogic.prepare_single` sets `save=on`. Fleet cameras carry both a path provider and the plugin, so this is the deployed shape. Reproduced: a gated scan with `shots_per_step=1` on a dual camera → `save` puts `['off', 'on', 'off']` — PNGs written for the whole scan, unbounded by the quota, contradicting the contract "in gated mode a plugin camera writes no PNGs". Fix direction: make fly-vs-strict explicit at `prepare` rather than inferring it from the event count.

## 3. MEDIUM (contract) — a native camera listed as an essential detector of a gated run is *not* refused by the worker; the docs say it is

`gated.py:176-183` demotes a `native_save` `GeecsDetector` without a plugin to the sampler with a **warning** — its frames are then saved nowhere. But `GeecsBluesky/CLAUDE.md`, the CHANGELOG, `detector.prepare`'s docstring and the PR body all state a worker-side refusal; `prepare` is never called on a member. The only refusal is the client preflight, skipped in the fail-open states and by any submitter that bypasses it. Either raise in `take_reading` or change the three contract texts.

## 4. LOW (correctness) — the same device in `detectors` and `non_essential` fails obscurely

Nothing refuses it. Reproduced: `count([a], 2, non_essential=[a])` → after a full `shot_timeout`, `GeecsTriggerTimeoutError('UC_A: no shot within 3.0s')` — the unbounded prepare/kickoff leaves the camera in fly mode, and the strict trigger path never gets its stamp wait. A one-line intersection check.

## 5. LOW (acceptance runbook) — A4's expected datum cannot be observed as written

With zero frames `make_stream_docs` emits no `stream_resource` and no `stream_datum`; under ARMED + SINGLESHOT the non-essential camera receives the five fired edges, so the observable is one datum `{0, 5}`.

## 6. LOW (placement) — `ACQUISITION_MODES` is defined twice

`qs_client/presets.py` duplicates `plans/registry.py`; `geecs_bluesky.plan_names` is the import-light vocabulary module both already import.

## 7. LOW (redundancy) — two rewind loops and two run brackets

`discard_uncollected` and the new `_rewind_plugins` repeat the provider→`GeecsHdfIO` match; the strict `ARMED → STANDBY` bracket is inline in the registry while the gated one is `gated_run_bracket`.

## 8. LOW (facility literal) — `TRIGGER_PERIOD_S = 1.0` "(HTU: 1 Hz)" in a code default

Root `CLAUDE.md` ("facility values have one home") names the problem; the trigger profile the step already drives is the natural home. A slower site would see `StackCheckCallback` flag the count mismatch, so it fails loudly, not silently.

## Checked and holding

The non-rewindable region (nothing replays; the interrupted `wait` returns and the plan proceeds; a failure stashed before the resume is thrown at the resumed `wait` and caught); deferred pauses land at the stock checkpoints; the sampler's ticks (echo, repeats, placeholder, a fresh queue per kickoff, rows carry exactly the descriptor's keys); the bundler rules (same frozenset for declare/collect, equal datum widths after the trim); the queueserver resolves unannotated `non_essential` names and lists `.hdf`/`.acq_timestamp`/`.save` in the tree; `non_essential_wrapper`'s ordering against the claim preprocessor; no start-document key collisions; the scan-folder invariant untouched; no lab hostnames/accounts.

---

## Round 2 — confirmation of `02760e86` (same reviewer)

| # | Verdict |
|---|---|
| 1 | **Confirmed.** The marks run in the same synchronous stretch as the interrupted wait's return; a status timing out in the next loop iteration is pardoned by `_guarded_complete` / the sampler's cancel branch; the one remaining race (the RunEngine's `asyncio.sleep(0)` letting a status fail before the response reaches the plan) lands at the `wait` yield inside the existing `try`. The new `except FailedStatus` around `mv(OFF)`/`sleep` is a correct safety net. The pinning test is the scenario that failed before. |
| 2 | **Confirmed.** `FlyTriggerInfo` + `isinstance`; the context-reuse check compares the stored `trigger_info`'s class. Quota-1 pinned at the device and through a `shots_per_step=1` gated scan. |
| 3 | **Confirmed**, pinned. (Residual R1 on *where* it fires.) |
| 4 | **Confirmed for identical objects.** (Residual R2: the owner/view overlap.) |
| 5 | **Confirmed** on the PR body. |
| 6 | **Confirmed.** |
| 7 | **Confirmed.** The unconditional rewind put is safe against the gateway (`file_plugin.py:796-815` accepts `n == session.count`, sets `stale_before`); `discard_uncollected` now raises "frames vanished" when `written < last_emitted` instead of blocking 10 s; every step now logs "rewound N → N" even when nothing was trimmed (cosmetic). |
| 8 | Waived; no objection. |

**R1. LOW** — the native-camera refusal fired after the scan folder is claimed (first step's `take_reading`); hoist it to bind time. **R2. LOW/MEDIUM** — `X.scalars` essential + `X` non-essential passed the both-lists guard, and the shared `fly` flag then silently un-gated the view's rows (reproduced: `success` with wrong-shot rows). **R3.** `__all__` listed `run_bracket` twice. **R4.** A5's "no error from the plan" was not guaranteed: `before_close` had no guard for a plugin whose gateway was restarted.

## Round 3 — confirmation of `db812dab`

R1 **confirmed** (bind time, before the claim; predicate `native_save and not plugin_backed`). R2 **confirmed** (by owner both ways; the hazard commented on `GeecsAcquireLogic.fly`). R3 **confirmed**. R4 **confirmed** — the `auto_raise=False` contingency mechanism checked: a failed status does not stay stashed; the `wait` inside the contingency receives it, `skip` runs, nothing reaches `close_run` or the STANDBY move; `close_run` deletes the bundler so `backstop_collect` never re-collects. **R5. LOW/MEDIUM** — the dead flyer's `unstage` (the stock `stage_wrapper`'s finalize) still failed the queue item after a successful run (reproduced).

## Round 4 — confirmation of `3743007d`

R5 **confirmed**: `stage_all` on the way in, per-flyer tolerant unstage in a `finalize_wrapper` with `wait=True` inside the contingency; ordering against `run_bracket`'s finalize unchanged (STANDBY last in every exit path). Two contract sentences were wrong — the RunEngine's teardown does **not** retry a skipped unstage (the object leaves `_staged` before the status resolves) — fixed in `87246263`.

---

## Disposition

- **Finding 1 (HIGH): fixed** in `02760e86` — synchronous `mark_abandoned` / `mark_cancelled` the moment the interrupted wait returns, the OFF/drain messages inside the same `except FailedStatus`; pinned by `test_pause_resumed_inside_the_count_timeout_window_still_retakes`. Confirmed.
- **Finding 2: fixed** in `02760e86` — `FlyTriggerInfo`; a batch of one is a fly prepare. Confirmed.
- **Finding 3: fixed** in `02760e86` (refused by the step) and `db812dab` (R1: refused at bind time, before the claim). Confirmed.
- **Finding 4: fixed** in `02760e86` and `db812dab` (R2: by owner). Confirmed.
- **Finding 5: fixed** in the PR body. Confirmed.
- **Finding 6: fixed** in `02760e86` (`plan_names.ACQUISITION_MODES`). Confirmed.
- **Finding 7: fixed** in `02760e86` (one rewind loop, one `run_bracket`). Confirmed. The "rewound N → N" log line on an untrimmed step is **waived** as cosmetic (it records that the stale watermark was set).
- **Finding 8: waived** — `TRIGGER_PERIOD_S` stays a module default with the knob deferred: the trigger profile schema carries no period today, and the stack check fails loudly on a wrong budget; flagged in the PR body under the design defaults.
- **R3: fixed** in `db812dab`. **R4: fixed** in `db812dab` (per-flyer contingency at the close). **R5: fixed** in `3743007d` + the wording in `87246263`. All confirmed.

skbarber:
## Codex review of #850 — disposition

**[P2] gated `count` passes the stock plan its `delay`, so the `num-1` no-op iterations of the batch hook each sleep `delay` after the batch** — **fixed** in the commit above: a gated count refuses a nonzero `delay` (scalar or per-iteration list) with the same sentence shape as the `shot_period` refusal ("a strict-mode spacing between shots: a gated count is one batch at the box's rate"); a zero delay, the default, passes. Pinned in `test_bound_plan_refuses_bad_mode_and_a_throttled_gated_run` (positional `4.0`, a list `[0.0, 2.0]`, and an explicit `0.0` that runs). Refusing rather than forcing `delay=0`: a delay the operator typed and the plan silently dropped would be a lie in the start document's `plan_args`.

skbarber:
## Broader validation set — devices wanted (2026-09-12)

Sam offered to turn on more devices. What the acceptance still needs, in priority order — reply with device names (and, for a motor, a safe range and the value to restore) and the runs go through the staging manager the same way as today's A1–A8:

1. **A triggered scalar device for A3** (the owed case): something that fires from the DG645 and publishes `acq_timestamp` on every shot with the laser off — an ICT (`U_BCaveICT` / `U_ICT_DUMP`, zero charge is fine; the stamp is what counts), an energy meter, a photodiode. It must have **no LabVIEW file writer** (or `save_images: false` works). Run: `scan([<device>], U_S1H.current, -0.5, 0.5, 2, shots_per_step=3, acquisition="gated")` → no `primary` stream, 6 `shots` rows with the device's scalars and its own stamps 1 s apart, `shot_clock: <device>`.
2. **One or two scalar-only, untriggered devices** for the sampler rows — a vacuum gauge, a magnet readback, a temperature. Expect: their latest value in every `shots` row next to the clock's stamp, at their own cadence.
3. **A second motor** (a stage or another magnet) with a safe range and the restore value — a two-motor gated `grid_scan` / `scan`, bins per step, box OFF between steps.
4. **A plugin camera on a different camera server** than Amp3 / Amp4 — a three-stack gated batch across two gateway hosts, `shot_clock` = the first camera in the list.
5. **A camera without a file plugin** (LabVIEW-native saving) — the gated refusal live ("native-saving device(s) without a file plugin"), then the strict fallback with the mixed detector list (strict rows for it, frames == rows).
6. **A5**: the go to restart Amp3's PVA gateway service mid-run. Scripted: submit the A4 preset, restart the service ~2 s in; expect the run to finish with its 5 rows, Amp3's stream carrying whatever it wrote, no error from the plan.

Not needed but useful if cheap: the same set once under a profile with free-running edges (STANDBY delivering triggers), to see the non-essential stream carry more frames than rows (A4's other branch).

skbarber:
# Reviewer confirmation of the post-review commits of #850 (3743007d..cb95e0d2)

Scope note: `git log 3743007d..cb95e0d2` yields **eight** commits, not six — `87246263` (R5 wording) and `c0ab20a6` (Codex delay refusal) precede the six acceptance commits. All eight were reviewed; the acceptance six are the focus. Paths below are relative to `/Users/samuelbarber/Desktop/Code/Github_repos/GEECS-Plugins-native-bluesky/GeecsBluesky/`.

## Commit-by-commit verdicts

| Commit | What it does | Verdict |
|---|---|---|
| `87246263` | Non-essential docstring/log wording: a skipped unstage is not retried by the RE; the next `stage()` clears a stale `Capture` | CONFIRMED — matches bluesky 1.15 (the object leaves `_staged` before the status resolves) |
| `c0ab20a6` | Gated `count` refuses a nonzero `delay` (`_has_delay`, scalar or iterable, positional or keyword); test + CHANGELOG | CONFIRMED |
| `1518b1a3` | `PluginPathProvider.__call__` mkdirs the device dir with `exist_ok=True`; raises `FileNotFoundError` if the scan folder is absent; test pins no-scan-folder-creation | CONFIRMED — honours the root scan-folder invariant (no `parents=True`); idempotent under ophyd-async context reuse |
| `53479958` | Gated first arm: prepare → `wait_for(zero_count)` → prepare again; `GeecsDetector.zero_count` = `_rewind_plugins(0, ...)` | CONFIRMED (see analysis below) |
| `9ef41fbc` | Strict first arm of a plugin camera does the same, tracked by an `armed` id-set per `geecs_take_reading` closure | CONFIRMED with one LOW (finding 4) |
| `670b21d6` | `shot_period` sleep moved out of `fire()` into `throttle()`, called at the top of every attempt | Code CONFIRMED; **FINDING** — the test does not pin it (finding 1); doc wording (finding 3) |
| `4f2f89f0` | Non-essential `after_open` zeroes and re-prepares; `complete` and `collect` are separate contingencies | Code CONFIRMED (empirically: uc_b datum `{0,2}` survives a failed complete); **FINDING** — tests don't assert the claims (findings 2, 5) |
| `cb95e0d2` | Refusal wording "native-saving device(s)" in plan + preflight; README troubleshooting note (#853/#854) | CONFIRMED; docstring inconsistency (finding 6) |

### Analysis of the questions posed

- **`zero_count` → `_rewind_plugins(0)` and the second prepare.** `_rewind_plugins` reads `written` (the stale PV value N ≥ 0), the `written < target` guard can never fire for target 0, it puts `Rewind=0` and waits for `num_captured == 0`. Plugin side (`GeecsPvaGateway/geecs_pva_gateway/file_plugin.py:796-812`): `_rewind(0)` is valid whenever `0 <= session.count`, posts `NumCaptured_RBV=0` and moves `stale_before = now`. A frame written between the first prepare and the zero (a fresh arming frame — the plugin *does* write the arming frame if its stamp is not stale, `file_plugin.py:606-640`) is dropped by the rewind; in all three call sites the box is quiet (gated: OFF after the drain sleep; strict/non-essential: ARMED), so nothing of the batch is lost. The second `prepare` hits ophyd-async 0.19.3's reuse branch (`_update_prepare_context`: same `collections_per_event` → providers reused, no re-open, no `Capture` toggle) and rebuilds `_PrepareCtx` with `collections_written = await _get_collections_written(...)` = 0 — the re-baseline is correct, so `truncate_to_quota` keeps `0 + quota` and `rewind_to_step_baseline` rewinds to 0 on a first-step retake. The watermark moved by the zero is the same mechanism the `Capture=1` watermark already relies on (same host clock); the first real frame is stamped after the zero in every path.
- **Strict `throttle()` placement.** `last_fire["at"]` is set inside `fire()` (per fire, not per attempt); a retry after a timeout sleeps the remainder since the *failed* fire, which is the right pacing. The sleep is after `prepare` (Capture=1 / `start_acquiring`, a no-op for GEECS) and before `trigger`; the count wait's budget (`observe_signals_value(timeout=exposure_timeout)`) and the stamp wait's budget (`wait_for_idle`, which runs *after* the index wait) both start at `trigger`, so a period longer than the budget no longer times the shot out. Pause-retake path unaffected (gated body is non-rewindable; a strict shot-1 replay re-executes the zero, which is what a retake of shot 1 wants).
- **Non-essential `after_open`.** Every flyer must be plugin-backed (the fly prepare refuses otherwise), so `hasattr(f, "zero_count")` is true for every GeecsDetector; a foreign flyer without it is still prepared, declared and kicked off. Fine.
- **Separate `complete`/`collect` contingencies.** `contingency_wrapper(auto_raise=False)` swallows the `FailedStatus` thrown into the plan at the `wait`; the RE keeps no failure state from it. Verified empirically by running the failure test with extra prints: `uc_b` datum `[{'start': 0, 'stop': 2}]`, no "collect failed" warning, streams `['primary', 'uc_b_stream']`.
- **`test_non_essential_camera_streams_for_the_run`.** Against the pre-fix `gated.py` it fails with datum `{0, 9}` vs `{0, 3}`, so it pins that the zero preceded the fires — but not that it preceded the kickoff baseline (finding 5).

## Findings

**1. MEDIUM — `test_shot_period_throttles_strict_fires` does not pin the A8 fix.** `tests/test_gated_plans.py:548-566`. The docstring says the period "exceeds the camera's shot budget on purpose", but the plugin camera's count wait uses `STRICT_TRIGGER_INFO.exposure_timeout` — the 3.0 s module constant (`geecs_bluesky/devices/detector.py:116-120`), not the fixture's `shot_timeout=0.2` — and the 0.2 s stamp wait only starts after the count wait resolves. Verified: with the pre-A8 `strict.py` (commit `9ef41fbc`) shadowed in, the test **passes** (1 passed in 1.83 s). A regression that moves the sleep back after the triggers ships green. Two variants that do pin it (both fail on pre-A8 with `GeecsTriggerTimeoutError: UC_Cam: no shot within 0.2s`, both pass on HEAD): (a) a native camera, `_camera(RE, box, "UC_Cam", shot_timeout=0.2)` — its stamp wait uses the per-camera budget and starts at the trigger; (b) keep the plugin camera and `monkeypatch.setattr(strict, "STRICT_TRIGGER_INFO", STRICT_TRIGGER_INFO.model_copy(update={"exposure_timeout": 0.2}))`. Suggest adding (a) or (b) (or both, since hardware saw the plugin count wait).

**2. LOW — the separate-contingency claim is not pinned.** `tests/test_gated_plans.py:722` asserts only that the log message changed from "complete/collect failed" to "complete failed". The CHANGELOG (`CHANGELOG.md:87`) claims "a failed complete never costs the datums"; nothing asserts `_datums_by_key(col)["uc_b"] == [{"start": 0, "stop": 2}]` or the absence of a "collect failed" warning. The behaviour is correct (verified above); add those two assertions so a future refactor back to one contingency fails.

**3. LOW — "before the detectors are armed" is inaccurate.** `geecs_bluesky/plans/strict.py:221` (the `throttle` summary line), `CHANGELOG.md:50`, `tests/test_gated_plans.py:554`. In this PR's own vocabulary "arm" = `prepare` ("the run's first arm", the README's "first arm fails with no frame within 8 s"), and the sleep runs *after* the prepare group at `strict.py:306` — before the *trigger*. The `throttle` docstring body says "must precede `trigger`" correctly. Suggest "before the detectors are triggered" in all three places.

**4. LOW — `armed` is keyed to the closure, not the plugin session.** `geecs_bluesky/plans/strict.py:218`. The bound plans build a fresh `geecs_per_shot`/`geecs_per_step` per call, so a queue item is fine; but a hand-assembled `bp.count(..., per_shot=geecs_per_shot(sc))` reused for a second run (as several tests in `test_strict_plans.py` could easily be written) skips the zero on the second run — `unstage` closed the plugin session, the next `stage`/`prepare` opens a fresh one, and the stale-N symptom returns silently. `gated_per_shot` already has this reuse limitation (`state["taken"]`), so it is not new in kind, but the strict hook was reusable before this commit. A more robust key is the detector's own session: set a flag in `zero_count`, clear it in `GeecsDetector.stage()`/`unstage()`, and have the plan test `not d.count_zeroed` instead of `id(d) not in armed`. Not blocking.

**5. LOW — `test_non_essential_camera_streams_for_the_run` does not pin "before the kickoff baselines".** `tests/test_gated_plans.py:571-599`. If the zero ran after `kickoff`, the close's `complete` would time out after 3 s, the contingency would skip it, and `collect` would still emit `{0, 3}` — the test passes either way, only slower. With `caplog`, assert no "non-essential uc_b: complete failed" record (or assert the run's wall time is well under `DEFAULT_SHOT_TIMEOUT`).

**6. LOW — `zero_count` docstring claims a plugin fix that has not landed.** `geecs_bluesky/devices/detector.py:977-979`: "Fixed in the plugin too (GeecsPvaGateway posts 0 at arm); this guard stays for the gateways deployed before that." At `cb95e0d2` the plugin posts `NumCaptured_RBV` only in `_on_frame` and `_rewind` (`file_plugin.py:729, 812`), no local ref contains a post-0-at-arm change, and GEECS-Plugins#853 is OPEN. The CHANGELOG (`:88-89`) and README correctly say the plugin-side fix *is* #853. Reword to "the plugin-side fix is #853; this guard stays for gateways deployed before it lands" so nobody removes the guard on the strength of the docstring.

**7. NIT — vacuous alternative in the strict stale-count test.** `tests/test_strict_plans.py:464-471`: `datums[:3] == [...] or len(col.primary_events()) == 3` — the right-hand side is already implied by `box.fires == 3` with no miss, so the datum-indices check can never fail the test. The test still pins the fix (against the pre-zero `strict.py` it fails hard: bluesky's bundler raises "Received multiple indices in a `stream_datum` document for one event"), so drop the `or` clause and assert the datums outright.

## Test run

```
cd GeecsBluesky && .venv/bin/python -m pytest tests/test_gated_plans.py tests/test_strict_plans.py tests/test_hdf_plugin_detector.py -q -p no:cacheprovider
45 passed in 16.95s   (rc=0)
```

Additional read-only checks performed with PYTHONPATH-shadowed copies of the package (the repo was not modified; temp copies removed): pre-zero `strict.py` fails `test_first_strict_shot_zeroes_a_plugins_stale_count`; pre-zero `gated.py` fails `test_gated_count_two_plugin_cameras` (`[8] != [0, 3]`) and `test_non_essential_camera_streams_for_the_run` (`{0, 9} != {0, 3}`); pre-A8 `strict.py` **passes** `test_shot_period_throttles_strict_fires` (finding 1).

## Overall verdict

The six acceptance commits are correct as code: the arm → zero → re-prepare sequence is sound against ophyd-async 0.19.3's context-reuse rule and the plugin's `Rewind`/watermark semantics, the throttle now sleeps before the trigger so a period longer than the shot budget cannot time the shot out, the non-essential `complete` and `collect` are genuinely independent (the datum survives a failed complete), the path provider honours the scan-folder invariant, and nothing regresses the pause-retake path or the strict/gated refusals. What needs a follow-up commit before the reviewer's sign-off is complete is test fidelity and documentation accuracy, not behaviour: the A8 throttle test passes on the code it was written to catch (finding 1, the one item I would fix before merge), two of the A4 claims are asserted only through log text (findings 2, 5), and the `zero_count` docstring says the plugin is fixed when #853 is still open (finding 6). None of the findings is a defect that would misbehave on hardware.

---

## Disposition (branch head `4f99b55c`)

All seven addressed in `4f99b55c`; `./scripts/check.sh GeecsBluesky` green (408 passed, 3 skipped, 1 deselected); deployed on the worker.

|---|---|---|
| 1 | MEDIUM | **Fixed.** `test_shot_period_throttles_strict_fires` runs both variants: a native camera with its own 0.2 s stamp budget and the plugin camera with `STRICT_TRIGGER_INFO.exposure_timeout` patched to 0.2 s. Either fails on the pre-A8 placement. |
| 2 | LOW | **Fixed.** The close-failure test asserts `uc_b`'s datum `{0, 2}` and that no "collect failed" warning was logged. |
| 3 | LOW | **Fixed.** "before the detectors are triggered" in the `throttle` summary, the CHANGELOG and the test docstring. |
| 4 | LOW | **Fixed, the robust way.** `GeecsDetector.count_zeroed`: set by `zero_count`, cleared by `stage` / `unstage`; the strict hook tests it instead of a closure-keyed id set, so a reused hook zeroes again on its next run (the strict stale-count test pins the clear after unstage). Added to `RESERVED_DEVICE_ATTRIBUTES`. |
| 5 | LOW | **Fixed.** The streaming test asserts the run finishes well under a `shot_timeout` and that no "failed at the run's close" record was logged — a zero after the kickoff baseline would have timed the close out. |
| 6 | LOW | **Fixed.** `zero_count`'s docstring says the plugin-side fix *is* #853 (open) and the guard stays until it lands. |
| 7 | NIT | **Fixed.** The `or` clause dropped; the datums for the image key are asserted outright (the stream_datum documents carry one per external key, so the filter goes through the stream resources). |

skbarber:
# Reviewer confirmation of `18010336` — `CaSettable.locate()`

**Verdict: CONFIRMED.** The implementation is correct against bluesky 1.15.0's `__read_and_stash_a_motor` (`preprocessors.py`: `isinstance(obj, Locatable)` → `Msg("locate")`, RE command registered at `run_engine.py:541`), the readback-as-setpoint choice is sound for `reset_positions_wrapper`, and the test pins both the "not the `:SP`" rule and the restore put. Nothing regresses `_column_headers` or the namespace's reserved-attribute rule. Two LOW notes, neither blocking.

**Process note (my error, corrected):** one of my commands ran `git checkout -q 18010336` on the clone, which detached HEAD from `phase/06-gated-batch` at the same commit. I reattached it immediately (`git checkout phase/06-gated-batch`; now `phase/06-gated-batch @ 18010336`, working tree clean, `git status --short` empty). No file content changed at any point.

## Answers to the posed questions

- **Readback as `setpoint` for `reset_positions_wrapper`.** `reset()` issues `Msg("set", k, v)` for every stashed value. For a `CaMotor` the restore is `set(readback_at_start)`: the GEECS blocking set converges immediately when the stage is already within the DB tolerance, and the `CaMotor` poll resolves on `|readback − setpoint| ≤ tolerance`, so a motor sitting inside tolerance but off its last commanded value is re-put to where it physically is — a no-op move, not a drift. The alternative (`:SP`) is wrong for the reason the docstring gives (a LabVIEW-driven move since the last gateway put), and hardware Scan018/019 confirm the restore. Acceptable.
- **Non-float settables (`datatype=None`/`str`).** `Location` is `Generic[T] TypedDict`, so a string/enum location is well-typed. Behaviourally nothing changes for them: a plain `CaSettable` keeps `_readback_attr="readback"`, so before this commit bluesky's fallback was the `read` branch (no `position` attr), which yields the same readback value `locate` now returns. Only `CaMotor` (`_readback_attr="position"`) hit the `SignalR + float` bug. `CaConfirmSettable` inherits `locate` and reports the *target* variable's own readback (its `_readback_attr` is the default), not `_confirm_readback` — correct, no cross-quantity restore.
- **Does the test pin the restore put?** Yes: `puts == [41322, 41332, 41342, 41352, 41362, 41342]` — the trailing `41342.0` after the scan's last point `41362.0` is `reset()`'s put; and with `:SP` mocked to `0.0` and `loc == {41342, 41342}`, any use of `:SP` would have produced `[-20 … 20]`.
- **`_column_headers` / reserved attributes.** `locate` reads through `_readback_attr_name` and touches no names, so the header map is unaffected. `RESERVED_DEVICE_ATTRIBUTES` is pinned to `dir(GeecsDetector)` (`tests/test_namespace.py:365-371`), and settables bind as *children* of namespace devices (`namespace.py:476, 502-527`), never as the parent, so `CaSettable.locate` cannot be shadowed by a GEECS variable named "locate"; correctly not added to the set.

## Findings

**1. LOW — `CaPseudoMovable` still cannot run a `rel_*` plan.** `geecs_bluesky/devices/ca/pseudo.py:45-102`. It has neither `locate` nor `position`, so bluesky takes the `read` fallback: an absolute pseudo's readback is `NaN` until its first set, so `rel_scan` over it moves to `NaN + offset`. Pre-existing and outside this commit's scope (the CHANGELOG entry names `CaSettable`/`CaMotor` only), but the sentence "the `rel_*` plans … work" reads broader than that. Suggest either a `locate` on the pseudo returning `self.readback` (a relative pseudo's `0.0` is its genuine start) or a one-clause scope note in the CHANGELOG.

**2. NIT — the test drives `locate` off-loop via `RE._loop`.** `tests/test_ca_devices.py:341-345` calls `asyncio.run_coroutine_threadsafe(motor.locate(), RE._loop)` to check the raw `Location`; the `RE(bp.rel_scan(...))` that follows already proves `locate` through the RE's own `locate` command, so the private-attribute call could be replaced by `RE(bps.locate(motor))` — cosmetic.

## Test run

```
cd GeecsBluesky && .venv/bin/python -m pytest tests/test_ca_devices.py tests/test_namespace.py tests/test_plan_registry.py -q -p no:cacheprovider
51 passed in 2.52s   (rc=0)
```

---

## Disposition (branch head `50a8ca54`)

|---|---|---|
| 1 | LOW | **Scope note added** (`50a8ca54`, CHANGELOG): `locate` covers the CA settables and motors; a `CaPseudoMovable` has none yet — an absolute pseudo reads NaN before its first set, so a relative plan over one stays undefined until a follow-up gives the pseudo its own `locate` (a relative pseudo's start is its readback). Not in this PR. |
| 2 | NIT | **Not taken.** The direct `locate()` call checks the raw `Location` (both fields, the `:SP` decoy at 0); the `rel_scan` that follows is the RE-path proof. `RE(bps.locate(...))` would not hand the test the result. |

The worker stays on `18010336` (this commit is the CHANGELOG only).

skbarber:
Merged the feature head into this branch after #846/#847/#848 landed (`0de595f6`): GeecsBluesky **0.83.0** sits above 0.82.5 and 0.82.4 in the CHANGELOG; GEECS-Schemas becomes **0.22.1** (the `PresetDevice.essential` entry) above #848's 0.22.0. `check.sh` green for GEECS-Schemas and GeecsBluesky; CI green.

PR #856, opened 2026-09-12
GeecsWebTheme 0.2.0: the surface kit — one layout vocabulary for every web surface

skbarber:
## What

`theme.css` settled **colour** and nothing else. Everything structural — the
shell, panels, chips, overlays, empty states — is re-authored per surface, and
the two surfaces that exist disagree about all of it. This PR adds `kit.css`
(plus `kit.js` and a reference page) as the layer that settles the rest.

Measured on `origin/master` before this change:

| | Data Portal | GeecsLogbook |
|---|---|---|
| Shell | `.frame`, grid `17rem`, **no breakpoint at all** | `.shell`, grid `228px`, sticky, collapses @860 |
| Panel | `.sect` / `.bincard`, ad hoc | `.card` + `> header` |
| Overlay | `.overlay`+`.modal` hand-rolled, `.cedrawer` | `<details>`, inline `.composer` |
| Status | `queued running done failed no_data dup` | `ok success fail failed warn aborted incomplete unknown quiet` |
| Radius | `3px` / `4px` / `6px` literals | `var(--r)` |

**0** lines of layout are shared between them; **15** status class names cover
about five states; none of the three overlay mechanisms is a real `<dialog>`,
so the portal's modal has no focus trap and does not close on `Esc`.

The forcing function is the third surface: when the console becomes web-based
it would be a third independent set of answers — and it is the surface with the
hardest requirements (live values, write actions, hazards, keyboard), so it
should set them rather than inherit a compromise from two read-mostly surfaces.

## What is in the kit

- **Shell** — topbar / 216px rail / pane, one breakpoint at 900px.
- **Three containers by role** — `.panel` (a thing with its own state),
  `.group` (no edges; controls belonging to the panel around them), `.well`
  (recessed; output the reader did not write). Border, fill, radius and shadow
  each say *separate object*; spending all four everywhere is why dense screens
  go flat.
- **One status vocabulary** — `queued · running · ok · degraded · failed ·
  unknown`, plus `agent` (not a severity — *who wrote this*). Every chip is a
  dot **and** a word, so colour is never the only carrier.
- **Five pane states** — loading, empty, error, **stale**, **denied**. The last
  two exist on neither surface today and a control surface cannot open without
  them: a live number that silently stops updating is worse than no number.
- **The overlay ladder**, as components rather than prose:
  `details.disc` (0) → `.inspector` (1) → `.drawer` (2) → `<dialog>` (3) →
  a route (4, needs no CSS). Rule: take the lowest rung that fits. Decider:
  *can the user lose work by pressing Esc?* → then it is a route.
- **Density as a viewer preference** — `theme-boot.js` stamps `data-density`
  pre-paint beside the palette; `kit.css` redefines `--pad` / `--row-h` /
  `--gap` under `[data-density="compact"]`.

`kit.html` is the kit's reference page: every component in the real theme, at
whichever palette and density is picked. It is a **static file beside the
stylesheets**, so a host already mounting this package serves it at
`<mount>/kit.html` — the portal gets `/theme/kit.html` with no route of its own.

## Scope

**One concern, and no consumer is touched** — the portal, the logbook and the
config editor are byte-identical after this PR. Adoption is a separate PR per
surface, deliberately: the point of landing the kit alone is that the reference
page can be argued with while changing it is still free.

Per-concern LOC: `kit.css` 447 · `kit.html` 350 · `kit.js` 217 ·
tests +93/−8 · `theme.css` +31 · `__init__.py` +70 · `theme-boot.js` +18 ·
CHANGELOG +44.

**Judgment calls, cheap to veto:**
- Rail width **216px** and a single breakpoint at **900px** — neither surface's
  current number (272px / 228px, 860px). Splitting the difference on the rail;
  900 because 860 clipped the topbar controls at common tablet widths.
- **Six status words**, which collapses `dup` and `no_data` into `degraded` and
  `aborted` into `failed`. If either distinction is load-bearing for you, say so
  and it becomes a seventh word rather than a lost one.
- `--r-lg` **as a separate token from `--r`** rather than deriving it. One value
  cannot say both "26px button" and "full-width card".
- **No facility literals added** (root CLAUDE.md § "Facility values have one
  home") — the kit is generic; the example device names in `kit.html`
  (`U_ICT`, `UC_HASO`) are sample content on a static demo page, not defaults
  in code.

## Tests

`GeecsWebTheme`: **114 passed, exit 0**. `./scripts/check.sh` (scoped) exit 0 —
it selects GeecsWebTheme, which is the only package whose files changed.

New guards:
- the literal-colour test walks `kit.css`, `kit.html` and `kit.js`
- `test_kit_defines_no_token_the_theme_does_not` — the kit *overrides* tokens
  and introduces none, keeping `theme.css` the single vocabulary
- `test_kit_reference_page_assets_all_exist` — a renamed asset fails here
  instead of silently blanking the page in a browser
- `test_python_and_boot_script_agree_on_the_density_list` — density pinned
  across Python, the boot script and the CSS, as the theme list already was

`./scripts/check.sh --all` is env-blocked in a fresh worktree (the other 12
packages have no poetry env there); CI covers them, and this diff cannot reach
them — no consumer file changed.

## Hardware verification

This is a stylesheet, a script and a static page.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -19,7 +19,7 @@ tooling. Each subdirectory is an independent Python package with its own
 | `GeecsPvaGateway/` | The PVA peer of GeecsCAGateway: distributed pvAccess server on each Windows camera server, exposing that host's GEECS camera images as NTNDArray PVs (gated subscriptions, latest-wins). Images stay off the central CA gateway by design |
 | `GEECS-MCP/` | The general GEECS MCP server for AI agents (OSPREY) — domains as modules, scans first: read tools (status/history/results/config listings/validation) + control verbs (submit with cap/etiquette/acknowledge-loop, ownership-gated stop, clear_queue) over `geecs_bluesky.qs_client` + the resolver + Tiled. Osprey integrates via `profile.yml` — central HTTP (the multi-machine mode) or stdio; see its `deploy/DEPLOYMENT.md` |
 | `GEECS-DataPortal/` | Scan-browsing web service (FastAPI, port 8200 on the worker host), read-only except explicit ScanAnalysis runs from its Analysis tab: day → scan → metadata/scalar plots/images in any browser, over the same `ScanCatalog` layer as the console's scan browser. Arc spec: `Planning/data_portal/` |
-| `GeecsWebTheme/` | One palette vocabulary for every GEECS web surface: three themes (`bella` red/black, `laser` 532 nm green, `plasma` hydrogen Balmer), each light and dark, plus the picker. No runtime dependencies — a stylesheet, a script and `static_dir()`. The rule it exists to enforce: surfaces style **through tokens, never with a literal colour**, pinned by its own test which walks the portal's and the editor's templates too |
+| `GeecsWebTheme/` | The shared look of every GEECS web surface, in two layers. `theme.css` settles **colour**: three themes (`bella` red/black, `laser` 532 nm green, `plasma` hydrogen Balmer), each light and dark, plus the picker. `kit.css` settles **everything else** — page shell, the three containers by role, one status vocabulary (`queued/running/ok/degraded/failed/unknown` + `agent`), controls, tables, the five pane states (incl. `stale` and `denied`), and the **overlay ladder** (`details` → inspector → drawer → `<dialog>` → route; take the lowest rung that fits). `kit.html` is its reference page, static beside the stylesheets so any host mounting it serves it at `<mount>/kit.html`. No runtime dependencies. Two rules it exists to enforce: surfaces style **through tokens, never with a literal colour**, and every kit rule is scoped to `.kit` so a surface adopts page by page — both pinned by its own tests, which walk the portal's and the editor's templates too. **Read its `CLAUDE.md` before building a new web surface** |
 | `GeecsLogbook/` | The logbook (successor to LogMaker4GoogleDocs): two books in one store — the **scans** book, a day-document view over scan folders, and the **ops** book, routine operations read by month — mounted by the Data Portal at `/log`. A day is a **query**, not a document — no daily job, no template stamping. Derived scan facts are rendered per request and stored nowhere; human commentary lives in the portal's state directory and is mirrored as markdown into `{experiment}/logbook/`, a tree of its own outside the data tree |
 | `LogMaker4GoogleDocs/` | Google Docs/Drive API wrapper for automated experiment logs — being replaced by `GeecsLogbook/` |
 
@@ -184,10 +184,14 @@ GEECS-DataPortal     →  GEECS-Data-Utils (tiled extra — the ScanCatalog
                         GeecsWebTheme (the shared palette; the portal is
                         the host that mounts it at /theme for every
                         surface inside this app)
-GeecsWebTheme        →  (no deps at all — a stylesheet, a script and a
-                        path helper. Everything web-facing depends on it;
-                        it depends on nothing, which is why it is its own
-                        package rather than living inside the portal)
+GeecsWebTheme        →  (no deps at all — stylesheets, two small scripts
+                        and a path helper. Everything web-facing depends
+                        on it; it depends on nothing, which is why it is
+                        its own package rather than living inside the
+                        portal. Carries BOTH the colour tokens and the
+                        layout kit, so a new surface inherits the shell,
+                        the status words and the overlay ladder instead
+                        of inventing a fourth set)
 GeecsLogbook         →  GEECS-Data-Utils (ScanPaths only — it reads scan
                         folders and nothing else; a peer VIEW LAYER of
                         GEECS-DataPortal, which mounts it at /log behind
--- a/GeecsWebTheme/CHANGELOG.md
+++ b/GeecsWebTheme/CHANGELOG.md
@@ -39,14 +39,46 @@ project adheres to semantic versioning.
   from `--r`, the control corner), `--bw`, `--tk`, `--pad`, `--row-h`,
   `--gap`, `--shell-max`, `--scrim`, `--lift`. Declared in the bare
   `:root` only, so every palette shares them until one wants its own.
-- `kit_css()`, `kit_js()`, `kit_html()` path helpers.
+- `kit_css()`, `kit_js()`, `kit_html()` path helpers, and `STATES` /
+  `PANE_STATES` constants so a host renders a vocabulary from a constant
+  rather than hand-typing a `data-state` that silently matches no rule.
+- A package `CLAUDE.md`: the kit's doctrine — the overlay ladder, the
+  container roles, the vocabularies — reachable from where an agent
+  building the next web surface actually looks.
 
 ### Changed
 
 - The literal-colour guard walks `kit.css`, `kit.html` and `kit.js`.
 - New tests: the kit introduces no token `theme.css` does not declare (one
   vocabulary, not two); every asset `kit.html` references exists; the
-  density list agrees across Python, the boot script and the CSS.
+  density list agrees across Python, the boot script and the CSS; both
+  vocabularies are pinned to `kit.css` in both directions; every kit rule
+  is scoped to `.kit`.
+- `README.md` and the root `CLAUDE.md` describe both layers, not just the
+  palette.
+
+### Fixed (from adversarial review, before first release)
+
+- **Every kit rule is scoped to `.kit` on `<body>`.** Ungated component
+  classes collided with both adoption targets, one load-bearingly: the
+  portal's run page is `.pane{display:none}` / `.pane.on{display:block}` —
+  its tab mechanism — which ties on specificity with an ungated `.pane`,
+  leaving which wins to stylesheet order. Every tab pane would have
+  rendered at once. Scoping also lets a surface convert one page at a time
+  instead of every page changing when the `<link>` lands.
+- The drawer honours `autofocus`. A selector list carries no priority, so
+  `querySelector("[autofocus],…")` returned the first match in *tree*
+  order — the header's Close button — while the adjacent `<dialog>` rung
+  honoured `autofocus` natively, so identical markup behaved differently
+  on two rungs.
+- A `data-drawer-open` / `data-dialog-open` naming a missing id warns
+  instead of producing a permanently dead control in silence.
+- `Esc` no longer closes the drawer underneath an open `<dialog>`.
+- The density picker builds into a host containing whitespace (ordinary
+  Jinja formatting previously counted as "already built").
+- The density pin asserted only that the selector existed — an *empty*
+  compact block passed it. It now asserts the block overrides exactly the
+  spacing tokens `theme.css` declares.
 
 ## [0.1.2] - 2026-09-12
 
--- a/GeecsWebTheme/CHANGELOG.md
+++ b/GeecsWebTheme/CHANGELOG.md
@@ -94,6 +94,13 @@ project adheres to semantic versioning.
 - `kit.html`'s dialog no longer carries `class="kit-dialog"`, which stopped
   matching anything when the rules were scoped. A dead class on the page
   people copy from is the wrong thing to copy.
+- **`.picklist`**, extracted from `.rail nav` (Codex review of #856). The
+  selectable-list styling was scoped to where it sat, so the inspector on
+  the reference page rendered three native browser buttons — a component
+  shown on the copy-from page that the kit did not actually style. It is
+  now a named component the rail, the inspector and the console's device
+  list all use, and a new guard fails when `kit.html` shows any class
+  nothing styles.
 
 ## [0.1.2] - 2026-09-12
 
--- a/GeecsWebTheme/CHANGELOG.md
+++ b/GeecsWebTheme/CHANGELOG.md
@@ -79,6 +79,21 @@ project adheres to semantic versioning.
 - The density pin asserted only that the selector existed — an *empty*
   compact block passed it. It now asserts the block overrides exactly the
   spacing tokens `theme.css` declares.
+- The scoping guard itself had three holes, found by re-review and each
+  reproduced before fixing. A regex over selectors consumed the `{` of
+  every `@media` prelude, so the **first rule inside each media block was
+  never examined** — four blocks, four unguarded rules, one of them the
+  mobile shell collapse. It also waved through `.kitchen` (a string
+  prefix, not a class token) and `:root .pane` (a fully global descendant
+  selector). The guard now walks braces instead of matching a regex, skips
+  `@keyframes` stops, and allows exactly two unscoped shapes. Each hole is
+  pinned as a probe case.
+- The vocabulary pins only recognised double-quoted attribute selectors,
+  so `[data-state='aborted']` slipped past — inconsistent with
+  `_INLINE_STYLE` in the same file, which already handled both.
+- `kit.html`'s dialog no longer carries `class="kit-dialog"`, which stopped
+  matching anything when the rules were scoped. A dead class on the page
+  people copy from is the wrong thing to copy.
 
 ## [0.1.2] - 2026-09-12
 
--- a/GeecsWebTheme/CHANGELOG.md
+++ b/GeecsWebTheme/CHANGELOG.md
@@ -4,6 +4,50 @@ All notable changes to `geecs-web-theme` are documented here. The format
 follows [Keep a Changelog](https://keepachangelog.com/en/1.1.0/) and this
 project adheres to semantic versioning.
 
+## [0.2.0] - 2026-09-12
+
+### Added
+
+- **`kit.css` — the layout vocabulary.** `theme.css` settled colour and
+  nothing else, so the portal and the logbook each answered the layout
+  questions separately and disagreed on all of them: two rails (17rem
+  fixed vs 228px sticky), fifteen status class names for about five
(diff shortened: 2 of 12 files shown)


skbarber:
## Adversarial review (fresh-context reviewer, full report)

Posted in full per repo convention. Dispositions follow.

---

### 1. `kit.js` drawer focus ignores `autofocus` — `querySelector` with a comma list is document order, not selector order

`GeecsWebTheme/geecs_web_theme/static/kit.js:132`

```js
var first = el.querySelector("[autofocus],[data-drawer-close],button,a[href],input,select,textarea");
```

The ordering of a selector list conveys no priority: `querySelector` returns the first element in **tree order** matching *any* branch. So the `[autofocus]`-first intent is not what runs.

Concrete: in the shipped `kit.html` drawer (`kit.html:307-333`) the first match in document order is the header's `<button data-drawer-close>Close</button>`. A surface that writes the documented thing — `<textarea autofocus>` in `.drawer > .body` — still gets focus on *Close*, so the keyboard user's first keystroke after opening the drawer is on the dismiss button rather than the field. The adjacent `<dialog>` rung does honour `autofocus` (`showModal()` does it natively, `kit.html:341`), so the two rungs behave differently for identical markup, which is the kind of thing that gets debugged as a browser quirk.

Fix is a preference walk: `el.querySelector("[autofocus]") || el.querySelector("[data-drawer-close],button,a[href],input,select,textarea")`.

### 2. The kit's component classes are ungated, and already mean something else on both adoption targets — `.pane` load-bearingly

The file gates its *element* rules on `.kit` (`.kit table`, `.kit th`, `.kit td`, `.kit a`, `.kit dialog`) but leaves every *component* class global: `.panel .group .well .row .grid .chip .dot .btn .field .seg .state .banner .shell .rail .pane .topbar .brand .drawer .scrim .inspector .bar .tscroll .eyebrow .mono .dim .spacer`.

Measured against the surfaces that are supposed to adopt it:

| Surface | Bare-class collisions today |
|---|---|
| `GEECS-DataPortal/geecs_portal/templates/run.html:53,54,72,25` | `.pane`, `.seg`, `.rail` |
| `GeecsLogbook/geecs_logbook/static/scanlog.css:19,21,27,32,34,47,62` | `.topbar`, `.brand`, `.btn`, `.shell`, `.rail`, `.dot`, `.chip` |
| `GEECS-DataPortal/.../base.html:43,44` | `.dim`, `.mono` (benign — near-identical) |

The sharp one is `.pane`. On the portal's run page it is not decoration, it is the tab mechanism:

```css
.pane { display: none; }
.pane.on { display: block; }
```

The kit's `.pane{display:flex;…}` (`kit.css:134`) is the same specificity (0,0,1,0), so which wins is decided purely by stylesheet order. Today `{% block head %}` renders *after* base.html's `<style>` (`base.html:49-50`), so linking `kit.css` beside `theme.css` in base would leave run.html winning — by luck, not design. Reverse the order (or move run.html's rules into a file linked before the kit, the obvious next refactor) and **all five tab panes render stacked at once**. The logbook's `.dot` is 6px vs the kit's 8px and its `.chip` is a mono/uppercase/3px rectangle vs the kit's sans pill with a `::before` dot — same tie, cosmetic rather than functional.

The consequence beyond the one break: because the component names are global, a surface cannot adopt the kit template-by-template. Adding the `<link>` is a per-surface flag day. Gating the components on `.kit` the way the base rules already are would let `<body class="kit">` convert one page at a time. The PR's own framing — "cheap to change while the reference page can still be argued with" — is the argument for deciding this now.

### 3. The two vocabularies the kit is actually selling are pinned nowhere

Themes and densities each get a three-way pin (`test_python_and_boot_script_agree_on_the_theme_list`, `…_density_list`) plus a Python constant a host can read. The **six status words** and the **five pane states** — the parts the PR says replace fifteen ad-hoc class names — exist only as `[data-state="…"]` attribute selectors in `kit.css`, with no Python constant and no test.

Concrete failure: the logbook renders `chip-success` / `chip-failed` / `chip-incomplete` / `chip-unknown` / `chip-quiet` today (`scanlog.css:120-123`, `templates/day.html:83`). When it adopts the kit, `<span class="chip" data-state="no_data">` or `data-state="incomplete"` matches no rule. It does not disappear or throw — `.chip` still gives a pill with `border-color:transparent`, inherited colour and a `::before` dot in `currentColor`, i.e. a plausible-looking neutral chip. A mistyped or un-migrated state is invisible in review and in the browser.

Same for `.state`: `denied` and `empty` and `loading` have **no** rule at all (only `error` and `stale` get `[data-state]` branches, `kit.css:331-340`), so `data-state="stail"` renders identically to the correct spelling.

The existing arrangement is right there to copy: export `STATES` / `PANE_STATES` from `geecs_web_theme/__init__.py` beside `THEMES` and `DENSITIES` (the docstring already justifies it — "named here so a host can offer them"), and pin them against `kit.css` and `kit.html` with the same test shape.

### 4. Contract files didn't travel with the behaviour

- `GeecsWebTheme/README.md` is untouched and still describes the whole package as "A surface links `theme.css`, adds `theme.js`, and styles through the tokens." No mention of `kit.css`, `kit.js`, `kit.html` or density. The `__init__.py` docstring *was* updated; the one human-facing doc was not.
- Root `CLAUDE.md` still says GeecsWebTheme is "a stylesheet, a script and `static_dir()`" in the package map, and "(no deps at all — a stylesheet, a script and a path helper)" in the dependency graph.
- The package has no `CLAUDE.md`, so the kit's doctrine — the overlay ladder, the container roles, the "lowest rung that fits" rule, the status vocabulary — lives only in CSS comments and a demo page. The next agent building the web console reads `CLAUDE.md`, which currently tells them the package is a palette.

Given that this PR's entire value is that the *third* surface inherits these answers instead of inventing a fourth set, the doctrine not being reachable from `CLAUDE.md` is the failure mode the PR exists to prevent.

### 5. `test_python_and_boot_script_agree_on_the_density_list` passes on an empty compact block

I mutated the compact block to `:root[data-density="compact"]{}` — **still passes**. The test only asserts the selector string is present (`kit.css` line 41). Meanwhile the sibling palette test does check completeness, which is exactly the drift this arrangement invites: add a fourth token to theme.css's density group and forget the compact override, and compact silently keeps the comfortable value with no test firing.

Related placement point: the comfortable values live in `theme.css:71-73` and the compact values in `kit.css:41-45`, so the pair is split across two files with nothing pinning them to the same token set. Asserting the compact block defines the same names the theme's density group declares closes it.

(For the record, the guards I *could* break do bite: a literal in `kit.css`/`kit.js`/`kit.html` (including a plain string and an inline `style=` attr), a new token in `kit.css`, and a renamed asset ref all fail correctly.)

### 6. A typo'd overlay id produces a permanently dead button, silently

`kit.js:180,184` — `document.getElementById(t.getAttribute("data-drawer-open"))` returning `null` makes `showDrawer` return immediately, and `confirmDialog(null)` returns a no-op stub (`kit.js:149`). `<button data-drawer-open="save-set">` against `id="saveset"` does nothing, forever, with no console output. The test file's own comment cites the dead-button finding as prior art in this repo. One `console.warn` in each branch.

### 7. Lower

- `kit.js:65` — `if (host.firstChild) return;` treats whitespace as "already built". `<div data-density-picker>\n</div>` (ordinary Jinja formatting) silently ships no picker. All four current `data-theme-picker` hosts happen to be written without whitespace, so nothing breaks today, but `theme.js:75` uses no such guard, so the two files disagree on the idiom. `host.children.length` or `host.querySelector("button")` is the intended check.
- `kit.js:193-197` — the Esc handler calls `e.preventDefault()` unconditionally whenever a drawer is open. A `<dialog>` opened above a drawer via the public `GeecsKit.confirm()` API would then have its native close suppressed while the drawer underneath closes. Guard on `document.querySelector("dialog[open]")`.
- `kit.html` is auto-served at `<mount>/kit.html` by anything mounting `static_dir()` — both `GEECS-DataPortal/geecs_portal/app.py:440` and `ScanAnalysis/scan_analysis/config_editor/app.py:241` do. Every GEECS site's portal therefore publishes a page full of HTU device names (`U_ICT`, `UC_HASO`, `U_S4H`, `HTU_standard`) as if they were that site's. The PR pre-empts the facility-literals rule and I agree it doesn't violate it (these are sample content on a demo page, not defaults in code), but the auto-mount means it's more visible than "a static file in the repo" implies.

### Redundancy: nothing found

I searched for prior implementations before crediting the kit as new:

- `grep "Escape|showModal|<dialog"` across `GEECS-DataPortal/geecs_portal/templates/`, `GeecsLogbook/geecs_logbook/static|templates/`, `ScanAnalysis/scan_analysis/config_editor/` returns **nothing** — no existing Esc handling, no native `<dialog>`, no focus-return logic anywhere. The PR's claim that the portal's hand-rolled `.overlay`/`.modal` lacks a focus trap and Esc checks out.
- No existing density/compact preference on any surface.
- No surface defines CSS custom properties of its own (`grep '^\s*--[\w-]+:'` over `scanlog.css`, `editor.css`, the portal templates returns nothing), so the nine new `:root` tokens collide with no local token — only the class names collide (finding 2).

### Placement: correct

`GeecsWebTheme` is the right home — it is the one package everything web-facing depends on and which depends on nothing, so the kit cannot create a cycle. `DENSITIES` / `DEFAULT_DENSITY` sit correctly beside `THEMES`. `kit.html` as a static sibling rather than a route is the right call and is what makes it free for both hosts. The only placement issue is the split density scale in finding 5, and the vocabulary constants that belong in `__init__.py` in finding 3.

skbarber:
## Dispositions — all 8 findings **fixed**, none waived

Fix commit: `95045d42`. GeecsWebTheme **117 passed**, `./scripts/check.sh` exit 0.
Every new pin verified by mutation (break the thing → the test fails; restore → green).

| # | Finding | Disposition |
|---|---|---|
| 1 | Drawer ignores `autofocus` (selector list is tree order) | **Fixed** — preference walk: `querySelector("[autofocus]")` first, then the fallback list. Comment records why the one-call form is wrong and that `<dialog>` differs. |
| 2 | Component classes ungated; `.pane` collides load-bearingly | **Fixed** — all 120 selectors scoped to `.kit` on `<body>`. New `test_kit_rules_are_scoped_to_the_kit_class` allows only `:root…` density blocks and `body.kit` through. |
| 3 | Status words and pane states pinned nowhere | **Fixed** — `STATES` and `PANE_STATES` exported beside `THEMES`/`DENSITIES`; two new tests pin statuses to `kit.css` in **both** directions (chip and dot), and reject any `data-state` in `kit.html` outside either vocabulary. |
| 4 | Contract files didn't travel | **Fixed** — `README.md` rewritten for both layers; root `CLAUDE.md` package map and dependency graph updated; new `GeecsWebTheme/CLAUDE.md` carries the doctrine (ladder, container roles, vocabularies, sharp edges). |
| 5 | Density pin passes on an empty compact block | **Fixed** — reproduced, then closed: the block must now define exactly `_DENSITY_TOKENS`, and `theme.css`'s bare `:root` must declare them. |
| 6 | Typo'd overlay id → silent dead control | **Fixed** — `need()` warns naming the id and what was expected. |
| 7a | `host.firstChild` treats whitespace as "already built" | **Fixed** — `host.querySelector("button")`. |
| 7b | Esc closes the drawer under an open `<dialog>` | **Fixed** — returns early when `dialog[open]` exists. |
| 7c | `kit.html` auto-served everywhere with HTU device names | **Fixed** — the page now opens with a banner saying every name and reading on it is sample content, not this installation's data. Kept the realistic names: a reference page whose job is legibility is worse with `device_1`. |

### Two notes the review surfaced that are worth recording

**Finding 2 was the valuable one.** It would not have shown up in this PR at all — nothing here links `kit.css` — and would have landed as "why is the run page showing every tab at once" during PR 2, with stylesheet order as the culprit. Scoping also buys something the PR did not have: a surface can convert one template at a time by adding `class="kit"`, instead of the `<link>` being a per-surface flag day.

**A pre-existing sharp edge, deliberately not widened here.** The literal-colour guard blanks `/* */` comments but not `//` line comments in a *standalone* `.js` file, so a `#NNN`-style reference in a JS comment reads as a three-digit hex and fails the guard. (Inside HTML it is fine — everything outside `<style>` is blanked, and there is a probe pinning exactly that.) I reworded my comment rather than teach the guard about `//`, because naive handling would blank the `//` in every URL. Recorded in the new `GeecsWebTheme/CLAUDE.md` under "Sharp edges".

Sent back to the same reviewer for confirm/deny.

skbarber:
## Re-review round 2 — reviewer confirmed 7/8, then broke the 8th's test

The reviewer verified every fix and found that the **guard I added for finding 2 had three holes**, one of them live on the shipped file. I reproduced all three before fixing, and each is now pinned as a probe case.

### The live hole

`test_kit_rules_are_scoped_to_the_kit_class` matched selectors with a regex whose anchor was consumed by each `@media` prelude, so **the first rule inside every media block was never examined**. Four blocks, four unguarded rules — and the one that matters is `.kit .shell` in the 900 px block, the mobile collapse the whole responsive story rests on.

Reproduced by hand: ungate it in the real file, run the suite, **green**.

```
ungated .shell inside @media  -> exit=0   (test passed; should have failed)
```

The guard now walks braces instead of matching a regex: at-rule preludes are recognised rather than consumed, and `@keyframes` bodies are skipped whole because their `0%` / `from` stops are not selectors.

### The other two

Both were the allow-list being sloppy:

- `.kit` was a **string prefix**, so `.kitchen` — a different class entirely — was exempt.
- `:root` whitelisted the **whole descendant selector**, so `:root .pane{display:flex}` passed while being exactly the global `.pane` collision the test exists to prevent.

Now `.kit` must end on a class-token boundary, and the only unscoped shapes allowed are `:root[data-density="…"]` and `body.kit`.

### Also from the re-review

- The vocabulary pins recognised only **double-quoted** attribute selectors, so `[data-state='aborted']` and `data-state='no_data'` slipped past — inconsistent with `_INLINE_STYLE` in the same file, which already handled both.
- `kit.html`'s dialog still carried `class="kit-dialog"`, dead since the rules were scoped. A dead class on the page people copy from is the wrong thing to copy. Removed.

### Verification

All five evasions now caught, and the suite green with each mutation reverted:

```
ungated .shell inside @media   -> exit=1
.kitchen{display:flex}         -> exit=1
:root .pane{display:flex}      -> exit=1
single-quoted CSS status       -> exit=1
single-quoted HTML state       -> exit=1
restored                       -> exit=0
```

**GeecsWebTheme: 125 passed.** Fix commit `113c82fe`.

### Reviewer's confirmations on the rest

Seven findings confirmed outright. Notably it verified the finding-2 CSS rewrite mechanically — normalising away the `.kit ` prefix, **129 rules before, 129 after, no selector lost or altered** — and independently reproduced the `_blocks()` comment premise behind the finding-5 fix. It also accepted the one thing I chose not to change: removing the `dialog.kit-dialog` alternate means a surface wanting a kit dialog without `.kit` on `<body>` has no hatch. That is a deliberate consequence of the scoping doctrine, not an oversight.

No finding waived. No finding outstanding.

skbarber:
Retargeted from `master` to `feature/web-surface-kit`, the arc's integration branch — the same shape the scan-logger arc used (`phase/*` → `feature/scan-logger` → one promotion PR to `master`).

Reason: master merges are the maintainer's by rule, so a step-per-PR-to-master serialises the whole arc on their availability, and a stacked child PR based on a step branch gets auto-closed unreopenably if that base is deleted on merge. Basing each step on the arc branch avoids both.

The promotion PR from `feature/web-surface-kit` to `master` is the one the maintainer merges, and it carries no fresh adversarial re-review — each constituent PR is reviewed on its own, as this one was (two rounds, both posted above, nothing waived).

skbarber:
## Codex P3 — confirmed and fixed (`354a80d4`)

Verified before changing anything: the only item styling was `.kit .rail nav a, .kit .rail nav button`, which requires `.rail` — and the inspector is not inside one. So the rung-1 selection list on `kit.html` rendered as three native browser buttons, on the page whose entire job is to be copied from.

**Took the second of the two options, and then went one further.** Rather than add `.kit .inspector .list button` rules, the fix removes the cause: the selectable list was styled by *where it sits*, which guarantees the third use needs a third rule. The console's device list is that third caller and it is already visible from here.

Extracted as **`.picklist`**, now used by the rail's navigation and the inspector's list alike. The rail keeps only its responsive override, which is genuinely about the rail. Current item is marked with `aria-current="page"` (navigation) or `aria-pressed="true"` (selection) — the state is the accessibility attribute, never a class, so it cannot be styled-but-unannounced.

### New guard for the general shape

`test_reference_page_demonstrates_only_what_the_kit_provides` — every class `kit.html` uses must be styled by `kit.css` or `theme.css`. A component shown on the copy-from page that nothing styles is worse than one that is simply missing: it gets copied, renders as browser defaults in the surface, and the adopter writes their own CSS, which is the divergence this package exists to stop. Verified by mutation (add an unstyled class to the page → fails; revert → green).

That guard would not have caught *this* instance, since the inspector used bare `<nav><button>` with no class at all — it closes the class-shaped version of the same hole going forward. The instance itself is now covered by `.picklist` being a real component with real rules.

Docs travel with it: `README.md` component list, `GeecsWebTheme/CLAUDE.md` (its own section, with the reason it is a component and not a location), and the CHANGELOG.

**GeecsWebTheme: 126 passed. `./scripts/check.sh` exit 0.**

PR #857, opened 2026-09-13
GeecsLogbook 0.7.0: adopt the surface kit, and fix five url_for calls that silently broke the editor behind TLS

skbarber:
Second step of the surface-kit arc. Targets `feature/web-surface-kit`, which already carries the kit (#856).

## What

The logbook's own copies of the shell, topbar, rail, buttons, selectable lists, card and status chips are **deleted, not overridden** — adopting a kit by overriding it is exactly how two vocabularies survive side by side. `scanlog.css` goes **359 → 311 lines**, and everything removed is something the kit now provides.

What stays is what is genuinely the logbook's: campaigns, the derived-facts block, the entry prose renderer, callouts, the composer, the between-scans block, and the `tone-*` system.

| | before | after |
|---|---|---|
| Rail | `228px`, breakpoint 860 | kit `216px`, breakpoint 900 |
| Scan block | `.card` | `.panel` |
| Rail lists | `.daylist` / `.scanlist`, styled by location | `.picklist` |
| Status | 9 × `chip-*` / `dot-*` | `.chip` / `.dot` + `data-state` |
| Tag | `.chip.chip-tag` | `.tag` |
| Topbar inner | `.wrap` | `.topbar-in` |

## Status is the substantive change

`KIT_STATE` in `models.py` maps each `ScanStatus` onto a kit state:

| ScanStatus | kit state | chip still reads |
|---|---|---|
| `success` | `ok` | success |
| `failed` | `failed` | failed |
| `aborted` | `degraded` | aborted |
| `incomplete` | `degraded` | not finalised / no scan info |
| `unknown` | `unknown` | check |

Two statuses share a colour **on purpose**: an aborted scan and an incomplete one both finished with less than was asked, which is what `degraded` means — and the chip still writes which one it was. Colour carries severity, text carries which, so nothing is lost by the merge.

Pinned both ways, because the failure is silent: an unmapped status renders `data-state=""`, which matches no rule and still produces a plausible neutral chip that survives review *and* the browser.

## The bug this turned up

**Five `url_for(...)` calls had no `.path`, including both `editor.js` script tags.**

Starlette's `url_for` returns an ABSOLUTE url built from the request the app saw. Behind TLS termination that is `http://`, so an absolute URL in a `<script src>` is a mixed-content block: **the editor's script silently never loads and the composer stops working** — with every test green, because the test client is plain HTTP.

Both templates have carried a comment saying to use `.path` *for exactly this reason* since they were written, two lines above one of the offenders. Found because the artifact publisher warned about an off-origin script while I was rendering a preview.

Now pinned by `tests/test_templates.py`, verified by mutation.

## Scope

One concern — kit adoption — plus the `url_for` fix, which is **flagged as a judgment-call addition**: it is a different concern, but it is a live production bug in the exact files this PR rewrites, and leaving it for later meant shipping a deploy that looks new and has a dead composer behind TLS. Easy to veto: it is 5 lines plus a test.

LOC: `scanlog.css` −67/+19 · templates −25/+45 · `models.py` +21 · `router.py` +5 · tests +81 · lock/pyproject +27.

**No facility literals added.** No behaviour change beyond the fix — this is styling and vocabulary.

## Tests

`GeecsLogbook` **217 passed**, `GeecsWebTheme` **126 passed**, `./scripts/check.sh` **exit 0** (it selects both packages).

New: `tests/test_models.py` (every `ScanStatus` mapped, every target a real kit state, no failure ever reading as success) and `tests/test_templates.py` (no bare `url_for` in any template).

Updated: 7 assertions in `test_month.py` / `test_router.py` that named the old class strings.

## Hardware verification

## Verified by rendering, not just by test

`/log/day/2026-09-11` rendered through the real router: `class="kit"` ×1, `kit.css` ×1, `chip data-state=` ×4, `panel scan` ×4, `picklist` ×2, and zero `chip-` or `class="card` remaining. `/log/month/2026-09` returns 200 with the kit wired.

## Known gap

The preview I rendered had **no notes database**, so no entries — scan-anchored or between-scan — appear in it. The entry components are converted but have not been eyeballed with real content. That is the first thing to check on the deploy.


skbarber:
## Adversarial review — three rounds, 10 findings, all fixed, none waived

Full reports below; dispositions in the table. Fix commits `9a4daa7f` and `c3ae598c`.

**Three of these were live breakage that 218 green tests did not see** — because nothing in the suite asserted a container class, a button class, or an active-day marker.

| # | Finding | Disposition |
|---|---|---|
| 1 | `month.html` never converted — `.card` orphaned on the composer and every day group; their own overrides renamed to `.panel.*`, matching nothing | **Fixed.** Both sections → `.panel`. Cause: a `git checkout` used to revert a test mutation discarded the file's uncommitted edits, and the re-application missed `.card` because `month.html` spells it `<section class="card …">` while `day.html` uses `<details class="card scan">`. |
| 2 | `btn-sm` / `btn-primary` orphaned — the Save button in **every** composer | **Fixed.** Renamed in `_entries.html` and `editor.js` (4 sites), including the three-class `btn btn-sm btn-primary` form the first pass missed. `editor.js` had not been touched by the adoption at all. |
| 3 | `incomplete` → `degraded` reverses a litigated decision, and swaps severities with `unknown` | **Fixed — the most important one.** `GeecsLogbook/CLAUDE.md` records that empty `ScanEndInfo` is the most common state on the real share (37 of 49 across four sampled days) and that painting it amber "painted most of a day amber". `incomplete → unknown` (neutral), `unknown → degraded` (amber): both keep the colour they had. The old test asserted only `!= "ok"`, which any wrong answer satisfies. |
| 4 | `.day.is-active` orphaned — current day stops highlighting | **Fixed.** `aria-current="page"`, which also closes a screen-reader gap the class never covered. |
| 5 | `data-state="running"` on "today" — and that pulse has **no** reduced-motion escape | **Fixed both.** Chip is neutral; GeecsWebTheme **0.2.1** adds the escape. It is the kit's only endless animation, so the one that most needed it — and a surface's own reduced-motion rule typically kills `transition` only. |
| 6 | `geecs-web-theme` a main dep nothing imports; root `CLAUDE.md` graph stale | **Fixed.** Moved to dev. Graph now names this edge **and** the pre-existing `GEECS-Schemas` one it had also been missing — it claims to be verified against each `pyproject.toml`. |
| 7 | `.rail section + section` outlived its shell | **Fixed.** Deleted; the kit's `gap` was already doing it. |
| 8 | Retired template name rendered as a `.chip` | **Fixed** → `.tag.tag-retired`. Same rule this PR applied to tags. |
| 9 | Dead `.wrap`, `daylist`, `scanlist` | **Fixed.** `calbody` correctly kept — `nav.js` queries it. |
| 10 | **Regression introduced by fix #8:** `.tag-retired` declared *above* `.tag` | **Fixed.** Both are one class, so source order decided and the base overrode the variant — the retired name rendered accent-coloured, pixel-identical to a real tag beside it. Exactly the confusion #8 existed to remove. The markup assertion passed, because the markup was correct. |

### New guards, each verified by mutation

- `test_the_colour_each_status_had_is_the_colour_it_keeps` — pins the colour per status with the CLAUDE.md rationale in the docstring, not merely "not ok".
- `test_a_single_class_variant_is_declared_after_its_base` — would have caught #10, and generalises (no other offender in the file).
- `test_every_literal_data_state_is_a_kit_state` — three templates write `data-state` literally rather than through `KIT_STATE`; a typo renders an uncoloured chip that looks plausible and passes any markup test.

### Verification method worth recording

The reviewer's approach beat mine: **render the pages, then diff every class *used* against every class *styled*** across `scanlog.css` + `kit.css` + `theme.css`. That is what surfaced 1, 2 and 4 — none of which any assertion covered. Its second pass improved on that again by scanning statically (`class=` in all three quote styles, `classList.*`, `setAttribute`), which reaches the JS-only paths — the editing composer's `innerHTML`, the toolbar, the calendar — that a TestClient render can never hit. It then rendered seven branch-heavy variants anyway (scan-anchored note, interscan note, retired template, agent entry, failure strip, 25-scan campaign day, `?tag=` filter, empty day, empty month): `['calbody']` everywhere, tag balance clean everywhere.

**GeecsLogbook 220 passed · GeecsWebTheme 126 passed · `./scripts/check.sh` exit 0.**

skbarber:
## Codex finding — fixed on this branch (`b2d44eda`)

**It was not fixed before, and the count I'd have quoted would have been wrong.** I made this correction while working on the next phase, so it landed on a downstream branch and never touched #857. Caught only because the maintainer asked me to confirm.

The contradiction Codex described was live here until now: the block comment above `KIT_STATE` said an aborted scan and an incomplete one *"are both finished with less than asked, which is exactly what `degraded` means"* — the mapping as it stood **before** the severity correction — sitting directly above an inline comment reading **"NOT degraded"**. `CHANGELOG.md` repeated the same stale claim.

Both now state what the code does:

- `incomplete` — an empty `ScanEndInfo`, the most common state on the real share — is an **absence of information** and stays neutral, because painting it amber painted most of a day amber.
- `unknown` — something *was* written and we cannot read it — is the case that **earns** the amber.

Verified clean: zero occurrences of either stale phrasing in `models.py` or `CHANGELOG.md`. **GeecsLogbook 220 passed, `./scripts/check.sh` exit 0.**

### Worth recording

This is the **second time in this arc** that prose and code disagreed on the same invariant — the one the new tests exist to protect — and no test can catch it, because a docstring is not executable. Both times the sequence was the same: I wrote the explanation, then changed the decision, and only updated the code. Writing the comment *after* the change lands is the actual fix.

Also a verification failure worth noting: my first check used a grep pattern that spanned a line break, so it returned zero matches on every branch. Had I trusted it I'd have reported this clean. A negative grep result is not evidence until the pattern is proven to match something.

PR #858, opened 2026-09-13
Phase 2c: the s-file of a gated run, from the shots rows joined to the stacks

skbarber:
## Hardware acceptance — PASSED 2026-09-12 (the re-export, no scans taken)

*Re-run in full on `0a49604b` after the review fixes changed the join
arithmetic: eleven scans, all OK; 411 of 411 values still match their stack
frame; Scan021 still identical to the s-file on the share. See the acceptance
comment for the new output and the review comment for the disposition of all 18
findings.*

The worker host's `~/qs-checkout` was put on this branch (`git fetch` + checkout,
no reinstall, no sudo) and phase 2b's accepted scans of **26_0912** were
re-exported from Tiled into a scratch tree. **The data share was not written**:
every s-file below went to `/tmp/accept_2c/`, so a real s-file was neither
overwritten nor created. Backfilling the gated runs' s-files on the share is
Sam's call.

| scan | shape | rows | s-file |
|---|---|---|---|
| Scan012 | strict, `UC_Amp3_IR_input` non-essential | 5 from `primary` | 5 rows; both stacks' stamps exact; **the 13 columns already on the share are identical and the non-essential camera's 9 columns are new** |
| Scan013 | gated count, a pause retake | 5 from `shots` | 5 rows, both cameras joined, 0 orphan rows (the retake left none) |
| Scan014 | gated scan over `U_S1H`, 3 shots × 3 steps | 9 from `shots` | 9 rows with `Bin #` 1/2/3 and `U_S1H Current` |
| Scan015 | `U_BCaveICT` as the clock, no camera | 6 from `shots` | 6 rows, the ICT's two channels + `U_S1H Current`, no stack to join |
| Scan018 | `rel_scan` over `U_CompAerotech` | 10 from `shots` | 10 rows, the ESP302's three axes in the rows |
| Scan019 | `rel_scan` over `U_ModeImagerESP` | 10 from `shots` | 10 rows, `UC_ModeImager` joined |
| Scan020 | gated, two cameras on two servers + ICT + ESP302 | 5 from `shots` | 27 columns: both stacks, the ICT and the ESP302 |
| Scan021 | strict with a native-saving ICT | 3 from `primary` | **identical to the s-file on the share** — 17 shared columns, 0 differing, 0 new |

Every stamp the s-file carries for a camera lands on a frame in that camera's
stack to **0.000 ms**, and no row is on no frame. Scans 013, 014, 015, 018, 019
and 020 had **no s-file at all** before this change.

**Values, not just stamps.** The stamp check cannot catch a positional
misalignment of the clock camera's own scalars (its stamp column comes from the
row, not the stack). So each camera's scalars were also looked up independently
— row stamp → frame index → the stack's value for that frame, through the
stack's own `scalar_variables` manifest — and compared with the s-file:
**411 (column, row) values across five runs, 0 mismatched.**

The worker is now live on this branch: environment cycled
(`qserver environment close` / `open`), `geecs-qserver-ensure-ready` reports
**20 allowed plans, all expected present**, and the running worker imports
`geecs_data_utils.shot_join` from the checkout.

**The stack check's new gated verdict** was driven with the real `shots` rows
from Tiled and the real stacks on disk for Scan013, Scan014, Scan018 and
Scan020 — every camera reports *"N frame(s) in X.h5 match the shots rows'
stamps"*, including the two-server Scan020 and the pause-retake Scan013. One
detail worth keeping: Scan014's window came out at **±0.499 s**, not ±0.500 —
its closest row gap was 0.999 s, so the cap in `join_window` took over from the
nominal period. The rule is doing real work on real data.

Not covered, because it needs a new scan rather than a re-export: that verdict
appearing in a live `scan.log`, and the threaded write path on the worker (the
re-export runs the same join synchronously). Both are pinned by the unit suite.

---

## What this is

Phase 2c of the native-Bluesky rebuild (#807,
`Planning/native_bluesky/08_gated_batch.md` §4.5, §5 item 3) — the last piece
of phase 2. **Before this a gated run wrote no s-file at all**: the callback
found no `primary` events and logged the skip.

A gated run's `primary` is datum-only (the plugin-backed cameras' frames) and
its per-shot record is the sampler's `shots` stream, so the writer needed a
second row source and a join. A non-essential camera has the same shape in
either mode.

## The join (`geecs_data_utils.shot_join`)

One home for the rule, shared by the worker's live callback and the offline
Tiled re-export so the two cannot drift. Three things the build settled that
the design left loose:

- **The window is per row.** Half the shot period, narrowed to half the
  distance to that row's closest neighbour, and half-open so a frame exactly on
  the boundary two rows share belongs to one of them. No frame can be claimed
  by two rows, a faster rep rate narrows the windows by itself, and one
  anomalous pair of row stamps narrows only those two rows rather than the whole
  run. Each row then takes the **nearest** frame in its window, not the first
  in the file.
- **Both sides are corrected from one drain-offsets map** before the comparison
  (a shot's cross-device stamps differ by a per-device constant, `03`
  §11.3/§11.4). One map, because correcting only one side does not lose columns
  — it shifts the s-file by a row, which is wrong data. They are all `0.0` in
  the field until the §4.F sync calibration exists, and the code does not assume
  otherwise.
- **A stack attribute becomes the event key a *strict* row would use**:
  `…-frame_acq_timestamp` → `<device>-acq_timestamp`, a subscribed scalar →
  `<device>-<scalar>`. So one `geecs_scalar_headers` map renames both shapes
  of run, and a column the row already carries is never overwritten — the
  clock camera's own stamp column stays the sampler's tick, which *is* the
  shot id it counted on.

A frame with no row inside its window is an **orphan**: it stays in the stack
and in Tiled and is left out of the s-file (Sam, 2026-09-12, `08` §6 Q4). One
s-file row per essential shot, always.

## The worker side

- `SFileCallback` takes its rows from `primary` when it has events and from
  `shots` otherwise, and joins every datum-only stream's per-frame columns
  onto them. Such a run's s-file is written on a daemon thread that waits for
  the plugin's `finalized` attribute — the stop document precedes `unstage`,
  when the plugin closes the file. A stack that never finalizes costs its own
  columns and a warning, never the s-file. **A run with no datum-only stream
  is still written synchronously from its `primary` events: nothing about a
  plain strict run changed.**
- `StackCheckCallback` now compares a *gated* stack's stamps with the `shots`
  rows as well as counting its frames. The batch trims every essential stack
  to the quota and the sampler ticks once per shot, so a frame the trim
  missed is a defect the count alone hides — its own datum covers it. A
  non-essential stream keeps the count check alone; an orphan there is
  normal, not a defect.
- The two stream-reading callbacks share one piece of document bookkeeping
  (`_StreamCallback`) and one finalize wait, and document dispatch unpacks
  `event_page` — which is how a `collect`ed stream's rows arrive.

## The offline re-export

`write_scalar_files_from_tiled(uid)` now reproduces a gated run's s-file too:
`read_run_rows` falls through to `shots`, and `read_frame_columns` reads a
table-less stream's 1-D attribute arrays **by name**. A frame stack, the only
multi-dimensional part, is never downloaded (the #834/#836 lesson). That is
what makes the acceptance above a check of the live path rather than a second
implementation of it.

## Tests

`./scripts/check.sh GeecsBluesky GEECS-Data-Utils` green — **419 + 350
passed** (exit 0, read unpiped), lint green.

- the join's own unit suite (window, offset correction, orphans, duplicates,
  non-finite stamps, unsorted rows, the clock-column rule, the attribute→key
  mapping);
- a synthetic **gated scan driven through the real plans** — two plugin
  cameras, a gauge and a motor, with the stacks written as the plugin writes
  them (per-frame attributes, `finalized`, renamed into place): the `shots`
  rows carry the gauge, the motor readback and the bin, the cameras' scalars
  and the second camera's own stamps come from the stacks, the in-flight
  frame after OFF is dropped;
- a gated run whose stack never finalizes: the s-file still lands, without
  that camera's columns, with the warning;
- a strict run with a non-essential camera: the rows stay the `primary`
  events and the streamed camera joins by stamp, its between-steps frame
  orphaned;
- the gated stack check, both verdicts;
- the Tiled re-export over a gated run's node shape (fake composite nodes,
  as this package's Tiled tests do — the frame-stack part raises if read).

## Notes

- The lock files are not refreshed; they already lag the path deps' versions
  on this branch (2a and 2b did the same) and the worker installs them
  editable.
- `SHOTS_STREAM` moves its definition to `geecs_data_utils.shot_join` and is
  re-exported from `geecs_bluesky.plans.gated`: the stream name is a
  *document* contract the plan, the callback and the re-export all share.

Docs: `08` §4.5 + §5 item 3 and its status block, `03` §2's phase-2 row,
both packages' `CLAUDE.md`, `EVENT_SCHEMA.md` (a new section for the `shots`
stream), both `CHANGELOG.md`.


skbarber:
## The re-exported s-file of a gated scan (Scan014 of 26_0912, six of its 22 columns)

```
Bin #  scan  UC_Amp4_IR_input acq_timestamp  UC_Amp3_IR_input acq_timestamp  U_S1H Current  Shotnumber
1      14    3872082729.591                  3872082729.627                  -0.49966       1
1      14    3872082730.590                  3872082730.627                  -0.49966       2
1      14    3872082731.590                  3872082731.627                  -0.49966       3
2      14    3872082742.806                  3872082742.842                   0.00008       4
2      14    3872082743.806                  3872082743.843                   0.00008       5
2      14    3872082744.805                  3872082744.842                   0.00008       6
3      14    3872082749.213                  3872082749.250                   0.50008       7
3      14    3872082750.213                  3872082750.250                   0.50008       8
3      14    3872082751.214                  3872082751.251                   0.50008       9
```

Worth pointing at: the two cameras' stamps for the same shot differ by a steady
**36 ms** — `03_clean_room_rebuild.md` §11.4's per-device drain constant, measured
here for free. Both cameras' `drain_offset` signals still read `0.0` (the §4.F
sync calibration that would set them does not exist yet), so the join matched
them on the window alone, which at 1 Hz is ±500 ms. That is the case the window
rule exists for, and it is also the reason the offsets are wired through to the
join now rather than later: once a sync button sets them, the same code gets
tighter rather than needing a change.

The shot spacing inside a step is 1.000 s, the gaps between steps are the moves,
and `Bin #` / `Shotnumber` / `U_S1H Current` are what the legacy scanner would
have written. Scan014 had **no s-file at all** before this change.

skbarber:
## One defect the acceptance found, fixed in `e76aadf2`

The re-export reported `offset=0.0` for every camera, which I took at face value
at first — every `drain_offset` signal really does read `0.0` until the sync
calibration (`03` §11.7, §4.F) sets them. Probing the catalog directly showed it
was not reading them at all: a Tiled stream node carries the descriptor's
`configuration` block **at the top of its own metadata**, not under a
`descriptors` list, so `_descriptor_drain_offsets` found nothing and defaulted.

Harmless today, and exactly the drift this PR exists to prevent: once the offsets
are real, the worker's live join would correct for them and the offline
re-export silently would not. `_descriptor_drain_offsets` now reads both shapes,
the test fake models the one Tiled really serves, and the nested shape is pinned
beside it.

Verified against the lab catalog on Scan020:

```
offsets from the primary node: {'uc_modeimager': 0.0, 'uc_amp4_ir_input': 0.0}
frame sources: [('uc_amp4_ir_input', 0.0, 5), ('uc_modeimager', 0.0, 5)]
```

read now rather than defaulted.

## Acceptance re-run on the pushed commit

The worker's checkout was reset to `e76aadf2` and the whole table re-run from it:
**Scan012, 013, 014, 015, 018, 019, 020, 021 all OK**, and the value-alignment
probe still reports 411 of 411 matching with no share column changed. The worker
environment was cycled again; `geecs-qserver-ensure-ready` reports 20 allowed
plans, all expected present.

`./scripts/check.sh GeecsBluesky GEECS-Data-Utils` is green on the new commit
(419 + 351 passed, exit 0).

skbarber:
## The gated stack-check verdict, over the real scans

A re-export cannot reach `StackCheckCallback`'s new comparison, so it was driven
directly with the real `shots` rows from Tiled and the real stacks on disk:

```
=== Scan013 ===   rows=5 from 'shots', clock='uc_amp4_ir_input-acq_timestamp', window=±0.500 s
  [ OK ] uc_amp3_ir_input: 5 frame(s) in UC_Amp3_IR_input.h5 match the shots rows' stamps
  [ OK ] uc_amp4_ir_input: 5 frame(s) in UC_Amp4_IR_input.h5 match the shots rows' stamps
=== Scan014 ===   rows=9 from 'shots', window=±0.499 s
  [ OK ] uc_amp3_ir_input: 9 frame(s) in UC_Amp3_IR_input.h5 match the shots rows' stamps
  [ OK ] uc_amp4_ir_input: 9 frame(s) in UC_Amp4_IR_input.h5 match the shots rows' stamps
=== Scan018 ===   rows=10 from 'shots', window=±0.500 s
  [ OK ] uc_amp4_ir_input: 10 frame(s) in UC_Amp4_IR_input.h5 match the shots rows' stamps
=== Scan020 ===   rows=5 from 'shots', window=±0.500 s
  [ OK ] uc_amp4_ir_input: 5 frame(s) in UC_Amp4_IR_input.h5 match the shots rows' stamps
  [ OK ] uc_modeimager:    5 frame(s) in UC_ModeImager.h5 match the shots rows' stamps
```

Scan013 is the pause-retake run, so its stacks had partial frames rewound out of
them — the verdict confirms nothing survived. Scan020 is the two-camera,
two-server run.

Note Scan014's **±0.499 s**: its closest row gap measured 0.999 s, so
`join_window`'s cap took over from the nominal 1 Hz period. The window rule is
not decoration.

What still needs a live scan rather than a re-export: that verdict landing in a
real `scan.log`, and the threaded write path on the worker. Both are pinned by
the unit suite; neither is reachable without taking a shot.

skbarber:
## Acceptance widened to the rest of the day, including the two failed runs

Beyond the eight in the PR body, the re-export was run over the remaining 2b
scans of 26_0912:

| scan | shape | result |
|---|---|---|
| Scan003 | gated count (A1) | OK — 5 rows, both stacks exact |
| Scan005 | gated scan, 3 shots × 3 steps (A2) | OK — 9 rows |
| Scan010 | strict count (A8) | OK — 3 rows |
| Scan016 | gated count, **exit=fail** (the ModeImager priming failure) | no s-file |
| Scan017 | gated `rel_scan`, **exit=fail** (the `rel_*` engine bug) | no s-file |

The two failures deserve a word, because "no s-file" could look like a
regression. Both failed on their *first* step, before the step's `collect`, so
the sampler's rows were never emitted: those runs have **zero** per-shot rows in
any stream, and the callback's skip line is the truthful outcome. An aborted
gated run that had finished earlier steps keeps those steps' rows and does get an
s-file — the batch collects per step, and `08`'s rule that an aborted scan's rows
are data still holds. Worth knowing when triaging a failed gated scan: no s-file
means no step completed, not that the writer gave up.

Eleven of the day's runs are now verified. `gh pr checks 858` is green on
`e76aadf2` (exit 0, unpiped).

skbarber:
## What the bug cost, on the share

An inventory of `26_0912/analysis/` makes the split exact: **every strict run of
the day has an s-file and every gated run has none.**

```
s-file present : Scan007 008 009 010 011 012 021   (all strict)
NO s-file      : Scan001 002 003 004 005 006 013 014 015 016 017 018 019 020
```

Thirteen of twenty-one. Two of those (Scan016, Scan017) failed before any step
completed and have no per-shot rows to write from, so eleven are backfillable
from Tiled with this branch:

```bash
cd ~/qs-checkout/GeecsBluesky
~/.local/bin/poetry run python -c "
from geecs_data_utils import write_scalar_files_from_tiled
write_scalar_files_from_tiled('<run uid>')"
```

I have **not** run that — it writes into the data share, and every s-file in the
acceptance went to `/tmp/accept_2c/` instead. Whether to backfill the day is
yours; the re-export is proven on all eleven.

skbarber:
## The Data-Utils consumers also pass

`geecs_data_utils` is the foundational layer, so the three packages that depend
on it were run too, not just the two the diff touches:

```
ScanAnalysis      192 passed
ImageAnalysis     276 passed
GEECS-DataPortal  209 passed
== OK — ScanAnalysis ImageAnalysis GEECS-DataPortal   (exit 0)
```

Nothing outside `GeecsBluesky` and `GEECS-Data-Utils` calls
`write_scalar_files` or `build_legacy_scalar_dataframe` (checked), and both
grew only keyword arguments with defaults, so the existing call shape is
untouched. `read_primary_scalars` is unchanged — it is just now also handed a
`shots` node, which its docstring says is fine.

skbarber:
# Adversarial review of #858 — report and disposition

A fresh-context reviewer read the diff against `08` §4.2/§4.4/§4.5/§4.7/§5/§6,
`03` §11, `plans/gated.py`, `devices/sampler.py`, `EVENT_SCHEMA.md`,
`io/scan_stack.py`, the installed `bluesky/bundlers.py` and the `event_model`
StreamResource/StreamDatum schemas, ran both suites, and proved findings with
four throwaway probes. **18 findings, 2 HIGH.** Both HIGH were real. Everything
at MEDIUM or above is fixed; nothing was waived.

## Disposition

|---|---|---|---|
| 1 | HIGH | The offline re-export never supplied `clock_drain_offset`, so it corrected only the frame side of the join and did **not** reproduce the worker's s-file | **Fixed.** One `drain_offsets` map now covers both sides, in both paths; `read_drain_offsets(run)` collects it for the whole run offline. A test proves the half-corrected case shifts the s-file by one row — wrong data, not missing data |
| 2 | HIGH | `read_frame_columns` and `read_run_rows` used different tests for "is this an event stream": a gated `primary` with an **empty** table part lost every camera column, silently | **Fixed.** The test is now "not the row stream", which is exactly what `read_run_rows` decided. Pinned by a test with an empty `internal` part |
| 3 | MED | The s-file thread was unjoinable in production; three un-finalizing stacks delayed it by 3× the timeout | **Fixed.** `subscribe_scan_outputs` returns `ScanOutputs` with a `join()`, kept on the RunEngine as `geecs_scan_outputs`; the finalize wait is one window for all stacks |
| 4 | MED | `join_window`'s min-gap cap was global: one anomalous pair of row stamps collapsed the window run-wide, below the real drain spread | **Fixed.** Windows are now **per row** (`row_windows`). Pinned: one 0.3 s pair narrows those two rows and leaves the rest at half the period |
| 5 | MED | A column whose spelling no header claims was dropped at DEBUG; `stack_scalar_variables` went unused | **Fixed.** A drift warning names the unclaimed columns, in GEECS's own vocabulary via the stack's `scalar_variables` manifest, which the live path now reads |
| 6 | MED | The join read the whole stack file, not the frames the datums referenced; `_Stack.width` was collected and unused | **Fixed.** `FrameColumns.truncated(width)`; the strict non-essential test now pins that an unreferenced frame stays out |
| 7 | MED | Unequal-length attribute arrays became silent `NaN`s | **Fixed.** A ragged column is dropped with a warning — a short dataset is a structural defect, not missing data |
| 8 | MED | `_squash` is a second encoding of the naming contract, needed only because the md stored the device name, not the column | **Fixed at the source.** The plan now records `shot_clock_column`; `_squash` stays as the fallback for the 2b-era runs (Scan012–021), which is the reason it cannot simply go |
| 9 | MED | `contested` was keep-first-in-file-order, so the s-file could carry the *worse* match | **Fixed.** Row-centric nearest-wins. The reviewer was right that keep-first was carried across a semantic boundary: it is the rule for identical millisecond keys, not for a tolerance join |
| 10 | MED | The `stream_resource` descriptor lookup was dead code, and the tests fed a field the RunEngine never emits | **Fixed.** Attribution comes from the first datum; `descriptor` removed from the stream_resource docs in the tests |
| 11 | MED | Row identity silently changed from append-order to seq_num-keyed, and bluesky reuses seq_nums after a rewind | **Fixed.** Rows are kept in arrival order with the sequence number alongside: the s-file gets the rows the run emitted (a partial row is data) and the stack check still maps datums onto them |
| 12 | MED | Buffer and thread growth: every stream's rows buffered, stopless runs leaked, `_threads` never pruned | **Fixed.** Only `primary`/`shots` events are buffered, threads are pruned on each spawn, and a stopless run is evicted after a few more runs with a warning |
| 13 | LOW | The clock column was resolved twice, independently | **Fixed** by #1's single map — the join resolves it once and looks its offset up itself |
| 14 | LOW | Uncommented bare `Any` where the concrete type is the point | **Fixed.** `Sequence[FrameColumns]`, `Sequence[float]`, `np.ndarray` under `TYPE_CHECKING` |
| 15 | LOW | `await_finalized` / `DEFAULT_FINALIZE_TIMEOUT_S` absent from `__all__` | **Fixed**, with `ROW_STREAMS` and `ScanOutputs` |
| 16 | LOW | Dangling `#:` comment in `gated.py` documenting a constant no longer defined there | **Fixed** — moved to the import it describes |
| 17 | LOW | A string-typed NDAttribute would raise out of the offline path | **Fixed.** `frame_columns_from_attributes` skips a non-numeric attribute with a warning, so both paths behave alike |
| 18 | LOW | A 7 s real-time sleep added to the suite | **Fixed.** The shared fixture keeps a safe 6 s window (the other tests write their stacks *after* the run and the thread must still be polling) and only the never-finalizes test shortens its own to 0.4 s — that test is now 1.2 s |

## One defect of my own, found while fixing #4

Per-row windows introduced a hazard the global window did not have: a frame
exactly midway between two rows sits inside **both** closed windows and could be
assigned to each. The windows are now half-open, `[stamp - w, stamp + w)`, so
such a frame belongs to the later row alone — and that is now an assertion, not
a comment.

## On the circular tests

The reviewer was right that three tests could not detect the errors they looked
like they covered, the re-export one most of all: it rewrote the fixture's row
stamps so they agreed with the stacks, *constructing* the agreement it asserted,
and so could never have caught #1. The replacement is the test the reviewer asked
for as gap 7 — a gated run driven through the real plans, its s-file written live
from the documents and the stacks, then the same run assembled as a Tiled node
and re-exported, asserting the two tables identical. It runs on a **5 Hz** box so
every window is 0.1 s, tighter than the 0.2 s between the two cameras' drain
offsets, and each camera's stack carries *its own* stamps rather than the rows'.
I checked it bites: drop the offsets from either side and it fails.

Also added, from the reviewer's gap list: a non-zero offset through
`write_scalar_files` (gap 1), the empty-table gated `primary` (gap 2), the
per-row window with offsets that matter (gap 3), a gated run **with** a
non-essential camera so both stream shapes join at once (gap 4), a ragged
attribute array (gap 5), a contested frame where the later one is nearer
(gap 6), and the live-vs-offline identity (gap 7).

Not done: gap 8, a byte-for-byte comparison of a strict s-file against a
pre-change reference. The hardware acceptance covers it on real data instead —
Scan021's re-export is identical to the s-file already on the share, 17 shared
columns, 0 differing.

## What the reviewer checked and found sound

Worth recording because it is the part that did not change: the strict path's
synchronous write and byte-identical column projection; the preserved stack-check
verdict strings; the epoch direction and the absence of a double application;
`searchsorted` boundaries, NaN, empty and unsorted inputs; that `_squash` never
disagrees with `normalize_component` (0 of 20 000 fuzzed names); that the
`detectors` fallback cannot pick a non-triggered device; that stream attribution
is correct in every reachable case; that descriptor `configuration` really does
carry `<name>-drain_offset` for a `collect`-declared stream; that `shots`
sequence numbers are unique across steps; that `event_page` unpacking is needed
and correct; the scan-folder invariant; no facility literals; docstrings,
`ruff`, and no signature breakage for out-of-package consumers. It also
confirmed `shot_join` is in the right package and duplicates none of
`timestamp_key` / `stack_frame_index_map` / `frame_index_for_timestamp` — those
implement an exact same-device join that explicitly refuses neighbours and
cannot express a cross-device tolerance match.

skbarber:
## Acceptance re-run on `0a49604b`, after the review fixes

The join arithmetic changed (per-row half-open windows, nearest-wins, one
offsets map), so the whole acceptance was re-run from the worker's checkout at
the new commit — not just the tests.

**Eleven scans of 26_0912, all OK**: Scan003, 005, 010, 012, 013, 014, 015, 018,
019, 020, 021. Every s-file camera stamp still lands on a frame in that camera's
stack to 0.000 ms, and no row is on no frame.

**Values**: 411 of 411 (column, row) values still match their stack frame,
checked independently through each stack's own `scalar_variables` manifest.

**The strict path is still untouched on real data**: Scan021's re-export remains
identical to the s-file already on the share — 17 shared columns, 0 differing, 0
new.

**The gated stack-check verdict**, driven over the real rows and stacks, now
reports the per-row window range it measured:

```
=== Scan013 ===  windows=±0.500-0.500 s, drain offsets={'uc_amp4_ir_input': 0.0, 'uc_amp3_ir_input': 0.0}
  [ OK ] uc_amp3_ir_input: 5 frame(s) in UC_Amp3_IR_input.h5 match the shots rows' stamps
  [ OK ] uc_amp4_ir_input: 5 frame(s) in UC_Amp4_IR_input.h5 match the shots rows' stamps
=== Scan014 ===  windows=±0.499-0.500 s
  [ OK ] uc_amp3_ir_input: 9 frame(s) in UC_Amp3_IR_input.h5 match the shots rows' stamps
  [ OK ] uc_amp4_ir_input: 9 frame(s) in UC_Amp4_IR_input.h5 match the shots rows' stamps
=== Scan018 ===  windows=±0.500-0.500 s
  [ OK ] uc_amp4_ir_input: 10 frame(s) in UC_Amp4_IR_input.h5 match the shots rows' stamps
=== Scan020 ===  windows=±0.500-0.500 s
  [ OK ] uc_amp4_ir_input: 5 frame(s) in UC_Amp4_IR_input.h5 match the shots rows' stamps
  [ OK ] uc_modeimager:    5 frame(s) in UC_ModeImager.h5 match the shots rows' stamps
```

Scan014 is the one where the per-row cap does visible work: most of its rows keep
half the period and the pair with the 0.999 s gap between them narrows to 0.499.
Under the old global minimum, that one pair set the window for all nine rows.

The offsets are now **read** rather than defaulted, and they are all 0.0 — the
sync calibration that would set them does not exist yet (`03` §11.7, §4.F). What
the acceptance therefore proves about the offsets is the plumbing, in both
paths; the arithmetic itself is proven by the unit tests, where a 0.2 s
difference is load-bearing.

`./scripts/check.sh GeecsBluesky GEECS-Data-Utils` green on the new commit: 421 +
364 passed, exit 0.

skbarber:
# Codex review — disposition (fixed in `56573799`)

## [P2] One frame could be given to two rows — **confirmed and fixed**

Reproduced exactly as described:

```
rows    [1001.0, 1001.0002]   frame [1001.0001]
windows [0.5, 0.5]
join    ShotJoin(frame_for_shot=(0, 0), orphans=(), contested=())
```

Both rows own frame 0. The diagnosis is right and it is precisely the hole my own
`_SAME_SHOT_S` skip opened: `row_windows` deliberately refuses to let two row
stamps a fraction of a millisecond apart narrow each other, because that pair is
one shot's publish race rather than two shots — so both keep a half-period window
and both reach the single frame between them. The half-open window I added in the
previous round only separates rows whose windows *touch* at a shared boundary; it
does nothing when the windows overlap almost entirely. My invariant test covered
the exactly-midway case and walked straight past this one.

Taking the suggestion to resolve ownership globally. `join_frames_to_shots` now
ranks every (row, frame) pair a window allows by distance and lets the closest
pair take both, then the next, and so on:

```python
candidates.sort()                      # (delta, row, frame) — deterministic ties
for _delta, row, frame in candidates:
    if row in owner or frame in claimed:
        continue
    owner[row] = frame
    claimed.add(frame)
```

One frame per row and one row per frame, whatever the windows look like. Since the
invariant no longer rests on boundary arithmetic, the half-open window is **gone**
— the windows are closed again and ties break by row then frame order, which is
one rule instead of two interacting ones. Nearest-wins is preserved and is now
global rather than per row.

Your case is a regression test, and the invariant itself is fuzzed over 400 random
row/frame configurations, asserting that every frame ends up owned, orphaned or
contested exactly once:

```python
owned = [i for i in join.frame_for_shot if i is not None]
assert len(owned) == len(set(owned))
assert set(owned) | set(join.orphans) | set(join.contested) == set(range(len(frames)))
assert len(owned) + len(join.orphans) + len(join.contested) == len(frames)
```

Result for your inputs is now `frame_for_shot=(0, None)` — one row gets the frame,
the other reads `NaN`, which is the honest answer when one frame exists.

## Open question — does Tiled clip the 1-D attribute parts to the datum ranges?

**I could not answer it from the lab's data, so I removed the dependence on the
answer.**

First, the measurement. Across every run of 26_0912 I compared, per camera, the
frames on disk, the shape Tiled reports for the stack part, and the length of an
attribute part:

```
scan     stream                   device              disk  tiled   attr
Scan012  uc_amp3_ir_input_stream  uc_amp3_ir_input       5      5      5
Scan013  primary                  uc_amp4_ir_input       5      5      5
Scan014  primary                  uc_amp3_ir_input       9      9      9
Scan020  primary                  uc_modeimager          5      5      5
…
```

All three agree everywhere, including Scan012, the one strict run with a
non-essential camera. So **no run of that day has more frames on disk than its
datums referenced**, and the day cannot distinguish "Tiled clips" from "there was
nothing to clip". I am not willing to assert a server guarantee I have not
observed.

What is solid is the other half: Tiled builds an external array's shape from the
`StreamDatum` `indices`, not from the file. So the stack part's **shape** is the
referenced count. `read_frame_columns` now reads that shape out of the node's
structure metadata — free, no download, and the frames array itself is still never
fetched — and truncates the device's columns to it:

```python
referenced = _referenced_frames(contents, device)   # shape[0] of the stack part
if 0 <= referenced < len(columns):
    columns = columns.truncated(referenced)
```

which is exactly what the live callback does with `stack.width`. The two paths now
use the same frames whether or not the server clips, so your divergence scenario —
a non-essential camera's frames written between `collect` and `unstage` reaching
the offline s-file only — cannot happen either way. A part that reports no shape is
left untruncated, on the grounds that a full join beats a silent cut.

Pinned by a test where Tiled reports 3 frames while the attribute arrays hand back
5, asserting the two extra values stay out.

## Verification after both changes

- `./scripts/check.sh GeecsBluesky GEECS-Data-Utils` green: **421 + 367 passed**, exit 0.
- The hardware acceptance re-run from the worker's checkout at `56573799`: **eleven
  scans of 26_0912 all OK**, 411 of 411 values still matching their stack frame,
  and Scan021's re-export still identical to the s-file already on the share (17
  shared columns, 0 differing). The gated stack-check verdict still reports
  "match the shots rows' stamps" for every camera.

Thanks — the P2 was a genuine correctness bug in a function whose docstring
claimed the opposite, and the open question was the right thing to be suspicious
about even though the data could not settle it.

PR #859, opened 2026-09-13
docs: fleet map — the portal hosts the logbook at /log

skbarber:
Two-line docs change to `docs/platform/fleet_map.md`: the portal's diagram node and its fleet-table row now say the logbook (GeecsLogbook) is served at `/log` on the same unit and port, where its state lives (`/var/lib/geecs-data-portal`, mirrored to `<experiment>/logbook/` on the share), that the memory ceiling comes from `site.env`, and add `GET /log/` → 307 as a health check. Records the deploy of #849 + #851 on the worker host (2026-09-12): checkout on master, unit re-rendered with `StateDirectory` and `MemoryHigh/Max`, verified over the network.

The runbook's rule: the fleet map is updated in the same PR when a deployment changes. The deploy itself was a host action, so this is its PR.

Review: docs-only, two lines, no cross-references added; adversarial review skipped and this sentence is the record. Tests: none affected. Hardware verification: none.

Known follow-up (not this PR): the logbook is slated to become its own service and port after the first round of real use, at which point this row splits.


skbarber:
# Disposition of the Codex review of #859

| # | Finding | Disposition |
|---|---|---|
| 1 | Runbook column references "§ The scan logbook" but `GEECS-DataPortal/DEPLOYMENT.md` has no such section | **Refuted.** On this branch (master + this PR) the section exists: `GEECS-DataPortal/DEPLOYMENT.md:190` — `## The scan logbook (\`--scan-log\`)`, added by #849 (merged 2026-09-12). A checkout that predates #849, e.g. `feature/native-bluesky-plans`, lacks it, which is the likely source of the read. No change. |
| 2 | Snapshot note still says 2026-09-04 while the portal row says "since 2026-09-12" | **Fixed** in the follow-up commit: the note now records a 2026-09-12 amendment scoped to the portal row (logbook at `/log`, memory ceiling; #849, #851) and states that no other row was re-observed that day, so the 09-04 snapshot claim stays honest for the rest of the table. |

PR #860, opened 2026-09-13
GeecsLogbook 0.8.0: the log reports the scans, it does not group them

skbarber:
Third step of the surface-kit arc. Targets `feature/web-surface-kit`.

Net **−19 lines** — this mostly removes a wrong idea.

## Two changes, one principle

The scan log had started **interpreting**. It grouped consecutive scans sharing a parameter and a purpose into a "campaign", and it wrapped notes falling between scans in a container announcing which gap they sat in. Neither is the log's job: it is note-taking about what happened, and this package already holds itself to exactly that rule in **"Status is reported, not inferred"** — *report what the files say and do not guess*.

## `Campaign` is deleted outright

Two things were wrong with it.

**It inferred.** Nothing in `ScanInfo` says those scans belong together; the code decided it from two fields matching.

**It borrowed a word that means something else here.** A campaign in this lab is *weeks* of work. A day page cannot contain one — it is a slice of one — so the rail reading **"Campaigns · 15"** told an operator it held fifteen multi-week efforts. That is the part that was actively misleading, and it rendered on screen.

**And it did not pay for itself.** Measured across four real days on the share:

| day | scans | groups | single-scan groups | longest |
|---|---|---|---|---|
| 26_0908 | 26 | 8 | 4 | 10 |
| 26_0909 | 65 | 11 | 9 | 50 |
| 26_0910 | 108 | 11 | 6 | 65 |
| **26_0912** | **21** | **15** | **11** | **3** |

A sweep day collapses 108 scans into 11. An acceptance run of 21 produced **15 groups, eleven of them wrapping a single scan** — because the threshold asked *"is the day long?"* when the question was *"does grouping help?"*.

Gone: the model, the property, seven CSS classes, the grouped template branch, and the filter and jump logic that had to reach two nesting levels. `grep campaign` across the package returns nothing.

What survives is the honest half: **a long day opens collapsed** — a fact about volume, not a claim about meaning. It keeps its own test (`TestLongDay`).

## A note between two scans is just a note

It wore a `<details class="between">` announcing "Between scans" and its range. That was ceremony, and it made identical content look like a different *kind* of thing from the same note in the ops book. It now renders as an ordinary entry at document level with its own timestamp — usually out of step with the scans either side, which is the point.

This also fixes a placement bug: under the grouping, a note anchored "after Scan005" was emitted **inside** Scan006's group.

## Seeded example notes

`scripts/seed_demo_notes.py` writes worked examples to a **separate** database. The store is authoritative — what people wrote exists nowhere else — so invented content must never be seeded into it. But a day with no notes shows none of what the page is for, which is how an unstyled composer shipped unnoticed in #857. The entries say in their own body that they are seeded, and the script refuses to write to a file that already exists.

## Scope

Two concerns, bundled deliberately because the second is only visible once the first lands — you cannot judge "notes in time order" while a grouping is reordering them.

Per-concern LOC: `Campaign` removal −91 `models.py` / −40 CSS / −70 `day.html` · between-block −20 · seed script +145 · tests −61/+56 · changelog.

**No facility literals added.** The seed script names real devices, but as example *content* in a dev-only script, not as defaults in code.

## Tests

**214 passed**, `./scripts/check.sh` exit 0.

Verified by rendering the real `2026-09-12` (21 scans) with seeded notes: **zero** campaign wrappers, **zero** "Between scans", 21 top-level scan blocks, notes interleaved by timestamp. Orphan check (every class used vs every class styled) clean but for `.calbody`, a `nav.js` hook orphaned before this arc.

## Hardware verification

### DIFF
--- a/GeecsLogbook/CHANGELOG.md
+++ b/GeecsLogbook/CHANGELOG.md
@@ -1,3 +1,9 @@
+# Changelog
+
+All notable changes to `geecs-logbook` are documented here. The format
+follows [Keep a Changelog](https://keepachangelog.com/en/1.1.0/) and this
+project adheres to semantic versioning.
+
 ## [0.8.0] - 2026-09-12
 
 ### Removed
@@ -51,12 +57,6 @@
   described the mapping as it was *before* the severity correction —
   claiming `aborted` and `incomplete` share `degraded`. Prose is not
   covered by a test suite; both now say what the code does.
-
-# Changelog
-
-All notable changes to `geecs-logbook` are documented here. The format
-follows [Keep a Changelog](https://keepachangelog.com/en/1.1.0/) and this
-project adheres to semantic versioning.
 
 ## [0.7.0] - 2026-09-12
 
--- a/GeecsLogbook/CHANGELOG.md
+++ b/GeecsLogbook/CHANGELOG.md
@@ -49,14 +49,36 @@ project adheres to semantic versioning.
   else), so invented content must never be seeded into it; but a day with
   no notes shows none of what the page is for, which is how an unstyled
   composer shipped unnoticed. Entries say in their own body that they are
-  seeded, and the script refuses to write to a file that already exists.
+  seeded, and the script refuses to write to a file that already exists —
+  or, on import, to a database that already holds entries for the day.
+
+  One entry is anchored to the day rather than a scan, because it is the
+  only shape that renders whether or not the share is reachable: an anchor
+  naming a scan the day folder does not contain is stored and counted in
+  "Notes N" but never drawn, so seeding on a checkout with no share
+  mounted otherwise produced exactly the empty page the script exists to
+  prevent.
 
 ### Fixed
 
-- The `KIT_STATE` comment and the 0.7.0 changelog entry both still
-  described the mapping as it was *before* the severity correction —
-  claiming `aborted` and `incomplete` share `degraded`. Prose is not
-  covered by a test suite; both now say what the code does.
+- **A dangling selector left `.insert[hidden]` visible.** Removing
+  `.between[hidden]` from `.insert[hidden],.between[hidden]{display:none}`
+  took the declaration block with it, so `.insert[hidden],` merged into the
+  *next* rule and inherited `display:flex`. Clicking "+ note after …" hid
+  nothing: the row stayed on screen above the composer it had just opened,
+  mis-spaced, accumulating until a save reloaded the page. There is no
+  global `[hidden]{display:none}` to fall back on.
+- **The filter left notes between scans floating unlabelled.** Deleting the
+  wrapper removed the only thing that said which gap a note sat in, and the
+  `#q` handler hid scans only — so filtering to one scan left a note from
+  a different gap sitting directly above it, reading as commentary on it.
+  Notes and insert rows now carry the bracketing scan labels as a
+  haystack and hide in the same pass.
+- `editor.js` no longer sets `.open` on the reveal target, which stopped
+  being a `<details>` in this release.
+- Dead after the deletion: `.scanrow .count` (the grouped rail row was its
+  only emitter), and the package `CLAUDE.md`'s "campaign shaping" and
+  "curated campaign record", which named the concept this release removes.
 
 ## [0.7.0] - 2026-09-12
 
--- a/GeecsLogbook/CHANGELOG.md
+++ b/GeecsLogbook/CHANGELOG.md
@@ -1,3 +1,57 @@
+## [0.8.0] - 2026-09-12
+
+### Removed
+
+- **The `Campaign` concept, entirely.** It grouped consecutive scans that
+  shared a `Scan Parameter` and a `ScanStartInfo`, and rendered them in a
+  second collapsible above the scans themselves. Two things were wrong with
+  it. It **inferred** — nothing in `ScanInfo` says those scans belong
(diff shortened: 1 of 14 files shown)


skbarber:
## Re-review round 2 — seven confirmed, five new, **two of them caused by my own fixes**

Fix commit `d253b46b`. **216 passed, `./scripts/check.sh` exit 0.** Every fix verified by mutation.

### The two I broke while fixing something else

**The filter and the editor were writing the same property.** `editor.js` uses `hidden` on an insert row to mean *"this one has been used"*. The filter I added last round wrote `hidden` too — so typing one character into the filter box and deleting it un-hid every used row, putting **"+ note after Scan006"** back on screen above the composer it had just opened. That is the same visible symptom as the dangling-selector bug fixed in the previous round, reached by a completely different path. Filtering now has its own `data-filtered` channel; two meanings no longer share an attribute.

**The haystack I gave those notes hid them from the most natural filter.** It held only the two bracketing labels, while a scan's holds label + parameter + purpose + devices. So typing `u_s1h` — the string the rail prints in *every row* — left all 25 scans on screen and hid every note and every insert affordance. Since a note's body is in no haystack, nothing brought it back: the day read as if nobody had written anything. Notes now borrow the haystacks of the scans that bracket them. Measured on a 25-scan day: **19 rows match where 0 did.**

### The grouping test missed two shapes

A `<section>` wrapper rather than a `<details>` — the obvious next attempt, since the stated complaint was a second *collapsible* — and a `<details>` whose `class` is not its first attribute. Both slipped a class-attribute regex.

It now **parses** the page and asserts every scan block's ancestor chain below `<main>` is empty, which is what "not grouped" actually means. Worth admitting: my first attempt at that assertion was itself wrong — it flagged consecutive sibling blocks, i.e. the normal flat layout. Both reintroduction shapes now fail the test; verified.

### The seed guard did not guard

It asked only whether `DAY` had entries, so a store holding a human's notes for **every other day** passed. It ignored tombstones. And it consulted the store — which meant `NotesStore` had already opened and schema-migrated the real file before the check ran.

It now reads the database directly, read-only, **before opening anything**; refuses any store holding any row including tombstones; and refuses a path named `logbook.db` outright, because a fresh deployment's store does not exist yet and an existence check alone lets that worst case through. Verified: a busy store on another day and a `logbook.db` path are both refused.

Also: the usage caveat had landed *inside* the docstring's literal block, splitting the two commands it sits between.

### On `.betweens`

The reviewer tried to break it as "a wrapper by another name" and could not — it carries exactly one rule (`[hidden]{display:none}`), no border, background, radius, padding or generated content, and `<main>` has no gap, so the rendered page is identical but for two tags. Recording that because it was the thing I most expected to have got wrong.

### What this round says about the work

Two of five new findings were regressions introduced by the previous round's fixes, both in the same small area, and both invisible to a green suite because they are *interaction* bugs — two pieces of code writing one property, and a haystack that is correct in isolation but wrong beside its neighbour's. Neither would surface without someone using the page. That is an argument for the deploy, not for more tests.

skbarber:
## Coordination — confirmed, with one number the other arc needs

Verified both points.

**Base.** Merge-base with `master` is `c2e3693d`, the #851 merge — so yes, this carries the deployed logbook, the memory ceiling and the OOM backport. `master` is now `7377872b`, three commits ahead, and all three are #859's fleet-map docs. Nothing in this arc touches `docs/`, so it merges clean. I will merge `master` forward into `feature/web-surface-kit` once this lands, so the eventual promotion PR is a fast-forward of content rather than a reconciliation.

**The day template — this is the part worth planning around.** My footprint there is larger than "both arcs touch this file" suggests:

```
GeecsLogbook/geecs_logbook/templates/day.html | 177 +++++++-----------
1 file changed, 82 insertions(+), 95 deletions(-)
```

Roughly half a ~350-line file, and it is structural rather than additive:

- the **entire grouped branch is deleted** — the `{% if many %}` campaign loop, `campbody`, and the `summary.campaigns` iteration
- `between_block` is rewritten (the `<details class="between">` wrapper gone, a `.betweens` div and a haystack in its place)
- a new `scan_hay(s)` macro, now the single definition of what a scan matches on, used by both `scan_block` and `between_block`
- the inline `<script>`'s filter, collapse-all and rail-jump logic all rewritten — the filter moved off the `hidden` property onto `data-filtered`, and the two-level traversal is gone

**Silver lining for the other arc:** if their change touches anything inside the grouped branch, that conflict evaporates — the code is deleted, not moved.

**Suggested order: this lands first, the split-service change rebases onto it.** Not seniority — cost. Re-applying a restructure on top of their edit is harder than rebasing their edit onto a restructured file, and this one has been through two adversarial review rounds already (12 findings, two of them regressions from my own fixes). Happy to invert if their change is the more structural of the two; I have not seen it, and it is not among the open PRs, so I am guessing at its size.

**On the redeploy.** Agreed on one restart rather than two — but the trigger is the **arc → `master` promotion**, not this merge: this lands on `feature/web-surface-kit`, which the host does not track. Two things to flag for whoever does it:

- the host is on **0.6.0** and this arc will present **0.8.0** — two releases at once, so the changelog to read spans both
- 0.7.0 is the kit adoption, which changes how every page *looks*; 0.8.0 changes what the day page *contains*. If the soak is meant to catch regressions, those are two different kinds of thing to look at

The owner has said the box can go to the arc branch before promotion (no users right now), so an earlier soak is available if the split-service work wants it.

Issue #855, opened 2026-09-12
GeecsBluesky: CaPseudoMovable needs locate() — rel_* plans over a pseudo scan variable are undefined


skbarber:
## What

`CaSettable` / `CaMotor` gained `locate()` (bluesky's `Locatable`) in #850 (`18010336`) after the 2b acceptance found every `rel_*` plan failing at its first move: bluesky 1.15's relative wrappers (`relative_set_wrapper`, `reset_positions_wrapper`) fall back to `obj.position`, which on a `CaMotor` is the readback *signal*, so `initial + rel_pos` raised `unsupported operand type(s) for +: 'SignalR' and 'float'` (Scan017 of 26_0912).

`CaPseudoMovable` (`geecs_bluesky/devices/ca/pseudo.py`) has neither `locate` nor `position`, so bluesky takes its last fallback — `read()` and the first hinted key. An **absolute** pseudo's readback is NaN before its first set, so a `rel_scan` over one moves to `NaN + offset`; a **relative** pseudo's readback starts at `0.0`, which happens to be its genuine start. Neither case is tested or hardware-verified.

## Asked

- `CaPseudoMovable.locate()` returning its own readback as both fields (the same rule as `CaSettable`: where the device is, not the gateway's `:SP`), and a defined answer for the absolute pseudo before its first set (refuse the relative plan with a clear message, or locate from the component motors' readbacks through the inverse — the inverse is the deferred item of the pseudo arc).
- A test on mocks: `rel_scan` over a relative pseudo moves about 0 and restores; over an absolute one is refused or located.
- A hardware run of a pseudo that combines two motors/magnets under the native engine — the 2b acceptance (2026-09-12) exercised only direct settables (`U_S1H`, `U_CompAerotech`, `U_ModeImagerESP`); no pseudo has been driven through the native path yet.

Found by the reviewer of #850's `locate` commit.

skbarber:
## Hardware note and the relative-pseudo question (2026-09-12)

[...]

PR #861, opened 2026-09-13
Phase 3: the shot-offset calibration and its preflight (#807 §4.F)

skbarber:
Phase 3 of the native-Bluesky rebuild (#807, `Planning/native_bluesky/03_clean_room_rebuild.md` §4.F) — the last backend piece before the merge gate's web front end.

## Hardware verification — PASSED 2026-09-12

HTU, `HTU-LaserOFF`, 10 shots, over UC_Amp3_IR_input + UC_Amp4_IR_input + UC_ModeImager + U_BCaveICT. **The first real shot-offset calibration:**

```
shot offsets over 10 shot(s), reference u_bcaveict:
  u_bcaveict            +0.0 ms  (scatter   0.8 ms)   <- reference
  uc_amp4_ir_input     +58.9 ms  (scatter   0.8 ms)
  uc_amp3_ir_input     +96.0 ms  (scatter   1.5 ms)
  uc_modeimager       +160.3 ms  (scatter   1.5 ms)
```

**Reproducibility.** Two independent 10-shot runs agreed to **under 1 ms** on every device (−0.8, −0.7, +0.0 ms). Amp3 − Amp4 came out **37.0 / 37.1 ms** against the **36 ms** phase 2c measured independently from a scan's s-file, and the **+37 ms** read off the stalled stamps before any shot was fired. Three independent routes to the same number.

**The delivery path, proven end to end.** `check_shot_sync` before and after the worker loaded the measurement:

| | verdict |
|---|---|
| before (`drain_offset` all 0.0) | `shot sync FAILED: corrected stamps span 161.0 ms against a 50 ms tolerance` |
| after one environment cycle | `shot sync OK: corrected stamps span 0.7 ms against a 50 ms tolerance` |

161 ms → 0.7 ms is the whole chain working: file → resolver → namespace → `drain_offset` → descriptors. The "never been calibrated" warning fired correctly on the first, and the refusal stopped the queue exactly as designed. Scan022 (strict `count`, 5 shots) then carried the measured offsets in its run documents — `read_drain_offsets(run)` returns them — and its s-file rows reproduce the calibration to the millisecond.

**The document written** (uncommitted in the configs repo, mode 644, for review):

```yaml
schema_version: 1
reference: u_bcaveict
devices:
  u_bcaveict:       {offset_s: 0.0,      scatter_s: 0.00250, shots: 10, geecs_device: U_BCaveICT}
  uc_amp3_ir_input: {offset_s: 0.09600,  scatter_s: 0.00150, shots: 10, geecs_device: UC_Amp3_IR_input}
  uc_amp4_ir_input: {offset_s: 0.05890,  scatter_s: 0.00075, shots: 10, geecs_device: UC_Amp4_IR_input}
  uc_modeimager:    {offset_s: 0.16030,  scatter_s: 0.00150, shots: 10, geecs_device: UC_ModeImager}
measured_at: '2026-09-12T22:48:49-07:00'
trigger_profile: HTU-LaserOFF
```

**Two findings from the measurement itself.**

The set spans **160 ms, not the ~100 ms §11.4 estimated**, and the reason is physical: **drain offset tracks frame size.** UC_ModeImager is a large chip with no ROI; Amp3/Amp4 are significantly ROI'd (Sam, 2026-09-12). That is §11.4's "4 MB versus 0.5 MB camera" measured directly, and it carries an operational consequence now in the doc: **re-ROI'ing a camera invalidates its calibration** — re-run after an ROI change, not only after a server rebuild.

The ~1 ms scatter here is **not** evidence about the ~10 ms host dither: with the laser off and the box self-triggering there is no laser jitter in the loop, so this is near the measurement floor. The dither during real running is the larger figure, which is why the averaging exists.

**Gated join in use — done 2026-09-13:** gated Scan023/Scan024 joined cleanly using the stored offsets, including at 5 Hz. (Original note:) Strict rows carry every device's own stamp in the same event, so they need no join at all; the join is for a gated run's stacks and for non-essential devices. Gated mode needs external edges, which come from the laser, and the laser was off. Everything up to the join's *input* is verified.

## Why this matters more than it reads

Every `drain_offset` in the field reads `0.0`. Phase 2c threaded the offsets through both sides of the s-file join and both of its review rounds spent most of their effort on that plumbing; this puts real numbers into it.

Two devices stamp the *same* shot at different times — the stamp is the trigger's arrival plus that device's own frame-drain latency (§11.3/§11.4). At 1 Hz the join windows are ±0.5 s and swallow the 36–100 ms spread, so nothing is broken today. They narrow with the rep rate: at 5 Hz they are ±0.1 s, the same order as the spread itself, where an uncalibrated offset costs rows. **This is what makes faster running safe**, which is why the gated batch exists.

## What landed

Two registered queue plans (`geecs_bluesky/plans/calibration.py`), neither opening a run — no scan number, no s-file, like `run_action`:

**`measure_shot_offsets`** drives the box OFF, waits the set quiet, then fires single shots through the existing `fire_and_await_shot` and reads every device's stamp. Three things the build settled that §4.F's sketch did not say:

- **Several shots, not one.** Each host's clock dithers around its own average by up to ~10 ms while the domain holds the averages on a common target (Sam, 2026-09-13). One shot would measure a 36 ms difference to ±10 ms. The quiet wait dominates the cost and is paid once, so the default is 10 shots — ten seconds of beam time for ~3 ms of residual noise — and each device's **peak-to-peak scatter** is recorded beside its mean, so a host with a timekeeping problem shows up instead of hiding inside an average.
- **Each shot is anchored on the mean of its own stamps** before averaging, so the laser's phase drifting between shots cancels and only the device-to-device differences survive. The mean rather than the minimum: the minimum of noisy values is biased and drags the earliest device's full dither into everyone else's figure. Only *complete* shots contribute — a shot one device missed would shift that shot's anchor and bias every other device — so an incomplete shot is discarded and retaken.
- **It reports by default and stores only with `write=True`.** The configs repo is a git working tree on the share, committing is a human act, and a re-run must not silently replace a good calibration with a worse one. The scatter column is exactly the thing you want to look at before committing.

**`check_shot_sync`** is Sam's validation shortcut (§11.7) and **costs no shot at all**: with the box OFF and the set quiet, every device still holds the stamp of the same last real shot, so correcting those stalled stamps by the stored offsets says whether the calibration still holds. It raises when out of tolerance, so a queue that puts it ahead of its scans stops before taking data against a stale calibration.

**Both are queue items and never scan steps.** The cost floor is the longest device timeout in the set, every run: a device's timeout event carries an *unchanged* stamp and the gateway's change suppression drops it (§11.2, measured in M1), so nothing announces quiescence — the only way to know the set is quiet is to watch the stamps not move. Being queue items also means neither can drive the trigger box while a scan is using it.

### The delivery path

`ShotOffsets` (GEECS-Schemas) → `ConfigsRepoResolver.resolve_shot_offsets` → `GeecsNamespace(drain_offsets=…)` seeds each detector's `drain_offset` **at construction** → descriptors → the join. The startup profile now builds the resolver *before* the namespace to pass them, which means a re-measurement reaches the worker at its **next environment open**, not mid-session.

Failure modes made loud rather than silent, because the thing they cause downstream is invisible:
- an **invalid** `shot_offsets.yaml` raises in the resolver (never a quiet fall back to zeros — a calibration that silently reverted would misjoin rows with nothing in the log to say why); the worker startup catches and warns so it still comes up.
- an offset naming **no detector** in the namespace warns as a stale calibration — the device it meant to correct would otherwise sit at `0.0`.
- `write_shot_offsets` refuses to create a missing experiment folder, and writes atomically (same-directory temp + `os.replace`).

## Proving the tests bite

This arc has been bitten twice by tests that did not test their invariant — one built its fixture from the numbers it then asserted. So every guarded production line here was **mutated and the guarding test confirmed red** before this landed:

| mutation applied to the production line | test |
|---|---|
| drop the per-shot anchor (use raw stamps) | `test_common_drift_between_shots_does_not_change_the_offsets` |
| anchor on the latest device instead of the earliest | `test_the_reference_is_the_earliest_device` |
| let a tie follow hash order instead of the name | `test_two_devices_stamping_together_resolve_the_tie_by_name` |
| drop the explicit `0.0` for the reference | `test_two_devices_stamping_together_resolve_the_tie_by_name` |
| skip the quiet confirmation window | `test_measure_refuses_a_set_that_never_goes_quiet` |
| average incomplete shots in instead of retaking | `test_measure_retakes_an_incomplete_shot` |
| compare raw stamps without correcting by the stored offsets | `test_sync_passes_when_the_stored_offsets_describe_the_hardware` |
| ignore the seeded `drain_offset` at construction | `test_drain_offset_reaches_the_descriptor` |
| write the measurement even without `write=True` | `test_measure_writes_nothing_without_write_true` |
| drop the namespace offset lookup | `test_measured_drain_offsets_reach_the_detectors` |
| create the experiment folder instead of refusing | `test_write_shot_offsets_refuses_a_missing_experiment_folder` |
| fall back to zeros on an invalid document | `test_an_invalid_shot_offsets_document_is_loud` |

All twelve went red. The fixtures choose the latencies and assert the measurement recovers them — never the reverse.

One thing the tests revealed and the code now documents: the scatter column is a **relative** quantity. The per-shot anchor is the mean of the set, so one device's wobble is shared with the anchor and the reported figure understates the true dither by `(N-1)/N`. That is the right quantity for judging a calibration, because the join is relative too; it is not an absolute measure of one host's clock.

## Checks

`./scripts/check.sh GeecsBluesky GEECS-Schemas` → exit 0. GeecsBluesky 489 passed, 3 skipped, 1 deselected; GEECS-Schemas 424 passed (head `73de1a1f`, after three review rounds plus a Codex review — see the comments; each round's fixes were confirmed by the next).

**Note:** CI does not execute the calibration tests — the unit-tests job installs GeecsBluesky without the `ca` extra, so the mock-CA layer skips (#866). The numbers above are the local run with the extras installed.

`GEECS_PLAN_NAMES` grows to **22**; the operator permissions regex and the readiness check follow.

## Not in this PR: the pseudo arc

#855 (`CaPseudoMovable.locate()`) was scoped to ride with phase 3. Surveying it first changed the answer, and **#855 as filed is the wrong fix** — the survey is recorded in `03` §8 ("The pseudo arc") so it is not redone:

- **ophyd-async already has the mechanism**: `Transform` + `DerivedSignalFactory` (0.19.3, installed). A `Transform` declares `derived_to_raw` (our `forward`) *and* `raw_to_derived` — the **inverse** the pseudo arc has been deferring since #600 — and yields a plain `SignalRW` that reads, locates and moves through stock machinery.
- #855's hard case is "an absolute pseudo's readback is NaN before its first set, so `rel_scan` moves to NaN + offset". A derived signal computes the readback from the component motors' live readbacks, so it is always defined and `locate()` is free. Hand-rolling `locate()` would build what the migration then deletes.
- **`mode: relative` may be a workaround for the missing inverse.** 11 of 13 corpus entries are relative, carried by hand-rolled baseline capture and a custom restore plan. With a real readback, "relative" is just `rel_scan` over the pseudo — stock bluesky, stock restore.
- **The corpus is nearly all affine**: 24 of 26 target formulas are `a*x + b`, so one `AffineTransform` fed by the catalog covers them with a free inverse; the other two are `±sqrt(…)` (`R56_at_100MeV`), invertible in a line.
- **Open physics question, Sam's to answer**: reading a pseudo back is over-determined (`ALine_e_beam_angle_offset_x` drives S3H ×1 and S4H ×−2 — two raw values, one derived number), so `raw_to_derived` needs a rule for which raw defines it and what to do when the components have drifted apart.

Sam confirmed composite variables are scanned regularly on HTU, so restoring them is a merge-gate item and gets its own PR with its own hardware acceptance.

## Versions

GEECS-Schemas 0.22.1 → **0.23.0**; GeecsBluesky 0.84.0 → **0.85.0**.

### DIFF
--- a/GEECS-Schemas/CHANGELOG.md
+++ b/GEECS-Schemas/CHANGELOG.md
@@ -5,6 +5,34 @@ All notable changes to GEECS-Schemas are documented here.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+
+## [0.23.0] - 2026-09-13
+
+Phase 3 of the native-Bluesky rebuild (GEECS-Plugins#807,
+`Planning/native_bluesky/03_clean_room_rebuild.md` §4.F): the vocabulary for
+the measured shot offsets.
+
+### Added
+
+- **`geecs_schemas.shot_offsets`** — `ShotOffsets`, the per-experiment
+  document of each device's measured edge-to-stamp latency, and its
+  `DeviceOffset` entries (`offset_s`, the peak-to-peak `scatter_s`, the
+  `shots` that contributed, the GEECS device name).  Written by the
+  `measure_shot_offsets` calibration plan, read at worker startup into each
+  detector's `drain_offset` config signal, and ultimately what the s-file
+  join corrects stamps by (`geecs_data_utils.shot_join`).  Registered in
+  `SCHEMA_REGISTRY` under `shot_offsets`.
+
+  Two properties the model enforces, because a document violating either
+  would anchor readers differently from the measurement that wrote it: the
+  `reference` device must be among the measured devices, and its own
+  `offset_s` must be `0.0`.  Only *differences* between offsets matter to
+  the join, so the reference exists purely to anchor the set.
+
+  No legacy dialect and no converter: the legacy scanner stored no
+  calibration at all — the sync ritual was run by hand and the numbers
+  lived in the operator's head.
+
 ## [0.22.1] - 2026-09-12
 
 ### Added
--- a/GEECS-Schemas/geecs_schemas/__init__.py
+++ b/GEECS-Schemas/geecs_schemas/__init__.py
@@ -56,6 +56,7 @@ from geecs_schemas.scan_request import (
     ScanRequestMode,
     SubmissionRecord,
 )
+from geecs_schemas.shot_offsets import DeviceOffset, ShotOffsets
 from geecs_schemas.scan_variables import (
     CompositeMode,
     PseudoComponent,
@@ -128,6 +129,9 @@ __all__ = [
     "Line1DConfig",
     "RendererOptions",
     "ScanRuntime",
+    # shot_offsets
+    "ShotOffsets",
+    "DeviceOffset",
     "SCHEMA_REGISTRY",
 ]
 
@@ -142,6 +146,7 @@ SCHEMA_REGISTRY: dict[str, type[VersionedSchemaModel]] = {
     "action_plan_library": ActionPlanLibrary,
     "experiment_defaults": ExperimentDefaults,
     "derived_channels": DerivedChannels,
+    "shot_offsets": ShotOffsets,
     "analysis_diagnostic": AnalysisDiagnostic,
     "analysis_group": AnalysisGroup,
 }
--- a/GEECS-Schemas/geecs_schemas/docgen.py
+++ b/GEECS-Schemas/geecs_schemas/docgen.py
@@ -189,6 +189,28 @@ derived_channels:
     precision: 6
     description: "Convectron pressure from U_VacuumGauge analog input 0"
 """,
+    "shot_offsets": """\
+schema_version: 1
+# Written by the measure_shot_offsets calibration plan, not by hand.
+reference: uc_amp3_ir_input   # stamped first; its own offset is 0.0 by definition
+devices:
+  uc_amp3_ir_input:
+    offset_s: 0.0             # seconds after the reference that this device stamps
+    scatter_s: 0.004          # peak-to-peak over the shots (host clock dither)
+    shots: 10                 # complete shots that contributed to the mean
+    geecs_device: UC_Amp3_IR_input
+  uc_amp4_ir_input:
+    offset_s: 0.036
+    scatter_s: 0.009
+    shots: 10
+    geecs_device: UC_Amp4_IR_input
+measured_at: "2026-09-13T18:22:04-07:00"
+trigger_profile: HTU-LaserOFF
+description: "after the Amp4 server rebuild"
+# Only DIFFERENCES matter: the join subtracts each device's offset from its
+# stamp before matching frames to rows, so adding a constant to every entry
+# changes nothing. A device absent here keeps 0.0.
+""",
     "analysis_diagnostic": """\
 schema_version: 2
 name: UC_TopView                 # the device folder under scans/ScanNNN/
--- a/GEECS-Schemas/geecs_schemas/shot_offsets.py
+++ b/GEECS-Schemas/geecs_schemas/shot_offsets.py
@@ -1,0 +1,183 @@
+"""ShotOffsets — each device's edge-to-stamp latency, measured once.
+
+A GEECS device's ``acq_timestamp`` is the shot's identity in everything but
+name, but two devices stamp the *same* shot at different times: the stamp is
+the trigger's arrival plus however long that device took to drain the frame,
+a per-device constant (``03_clean_room_rebuild.md`` §11.3/§11.4).  The
+constant differs by tens of milliseconds across a camera set — 36 ms between
+two amplifier cameras, near 100 ms across a set mixing a 4 MB camera with a
+0.5 MB one.
+
+Joining frames to shot rows therefore corrects each side by its own offset
+before matching (``geecs_data_utils.shot_join``).  Until this document
+exists every offset reads ``0.0``, which is harmless only while the join
+windows are wide: at 1 Hz they are ±0.5 s and swallow the spread, but they
+narrow with the rep rate, and at 5 Hz they are ±0.1 s — the same order as
+the spread itself, where an uncalibrated offset costs rows.  Measuring these
+numbers is what makes faster running safe.
+
+This document is what the ``measure_shot_offsets`` plan writes and what the
+worker reads at startup to populate each detector's ``drain_offset`` config
+signal.  It is per experiment, and it is written by a *measurement*, not by
+hand — edit it only to correct a mistake.
+
+Only differences matter
+-----------------------
+The join subtracts each device's offset from its stamp before matching, so
+adding the same constant to every offset changes nothing.  The offsets are
+therefore anchored: :attr:`ShotOffsets.reference` names the device measured
+as stamping first, whose offset is ``0.0`` by construction, and every other
+device's :attr:`DeviceOffset.offset_s` says how long *after* that device it
+stamps the same shot.  A later re-measurement may pick a different
+reference without any of the numbers meaning something different.
+
+Why the scatter is recorded
+---------------------------
+A device's offset is not perfectly steady: each machine's clock dithers
+around its average by up to ~10 ms (a consequence of the host's own
+timekeeping — higher-end boxes hold ~1 ms), while the domain keeps the
+averages on a common target.  One shot therefore estimates the offset only
+to that precision, so the plan averages several shots and records the
+observed scatter alongside the mean.  The scatter is evidence in its own
+right: a device whose scatter is much larger than its peers' has a
+timekeeping problem that an average would hide.
+
+Developer notes
+---------------
+There is no legacy YAML dialect behind this model — the legacy scanner had
+no stored calibration at all (the sync ritual was run by hand and the
+numbers lived in the operator's head), so there is no converter for it.
+"""
+
+from __future__ import annotations
+
+import math
+from typing import Optional
+
+from pydantic import Field, model_validator
+
+from geecs_schemas._base import SchemaModel, VersionedSchemaModel
+
+
+class DeviceOffset(SchemaModel):
+    """One device's measured edge-to-stamp latency, relative to the reference."""
+
+    offset_s: float = Field(
+        description=(
+            "Seconds after the reference device that this device stamps the "
+            "same shot — the mean over the shots that contributed. The "
+            "reference device's own value is 0.0. Subtracted from this "
+            "device's acq_timestamp before frames are matched to shot rows."
+        )
+    )
+    scatter_s: float = Field(
+        0.0,
+        ge=0.0,
+        description=(
+            "Peak-to-peak spread of this device's per-shot offset across the "
+            "measurement, seconds. Expect up to ~10 ms from ordinary host "
+            "clock dither; markedly more than the other devices in the set "
+            "means this machine's timekeeping is worth looking at."
+        ),
+    )
+    shots: int = Field(
+        1,
+        ge=1,
+        description=(
+            "How many shots contributed to the mean. Fewer than the "
+            "measurement requested means this device missed shots."
+        ),
+    )
+    geecs_device: str = Field(
+        "",
+        description=(
+            "GEECS device name this offset was measured for, e.g. "
+            "'UC_Amp3_IR_input'. Informational: the mapping key is the ophyd "
+            "object name the runtime uses."
+        ),
+    )
+
+    @model_validator(mode="after")
+    def _finite(self) -> "DeviceOffset":
+        if not math.isfinite(self.offset_s):
+            raise ValueError(f"offset_s must be a finite number, got {self.offset_s}")
+        if not math.isfinite(self.scatter_s):
+            raise ValueError(f"scatter_s must be a finite number, got {self.scatter_s}")
+        return self
+
+
+class ShotOffsets(VersionedSchemaModel):
+    """The experiment's measured per-device drain offsets.
+
+    Written by the ``measure_shot_offsets`` calibration plan and read at
+    worker startup to populate every detector's ``drain_offset`` config
+    signal, which rides in each run's descriptors and is what the s-file
+    join corrects by.
+    """
+
+    reference: str = Field(
+        description=(
+            "The ophyd object name of the device the offsets are measured "
+            "against — the one that stamped first. Its own offset_s is 0.0. "
+            "Only differences matter, so which device this is carries no "
+            "meaning beyond anchoring the numbers."
+        )
+    )
+    devices: dict[str, DeviceOffset] = Field(
+        default_factory=dict,
+        description=(
+            "Ophyd object name (e.g. 'uc_amp3_ir_input') → that device's "
+            "measured offset. A device absent from this mapping keeps the "
+            "0.0 default, which is correct only if it really stamps with the "
+            "reference."
+        ),
+    )
+    measured_at: Optional[str] = Field(
+        None,
+        description=(
+            "ISO-8601 timestamp of the measurement, with offset. "
+            "Informational, but the thing to look at when a join goes wrong: "
+            "a calibration older than the last camera or server change is "
+            "suspect."
+        ),
+    )
+    trigger_profile: Optional[str] = Field(
+        None,
+        description=(
+            "Trigger profile the measurement fired through. Recorded because "
+            "a profile that drives a different trigger box would measure "
+            "different latencies."
+        ),
+    )
+    description: str = Field(
+        "",
+        description="Optional note about this measurement.",
+    )
+
+    @model_validator(mode="after")
+    def _reference_is_consistent(self) -> "ShotOffsets":
+        """The reference must be listed, and must be the zero.
+
+        A document whose reference names a device it does not list, or whose
+        reference carries a non-zero offset, is not self-consistent: readers
+        would anchor the set differently from the measurement that wrote it.
+        """
+        entry = self.devices.get(self.reference)
+        if entry is None:
+            listed = ", ".join(sorted(self.devices)) or "none"
+            raise ValueError(
+                f"reference {self.reference!r} is not among the measured "
+                f"devices ({listed})"
+            )
+        if entry.offset_s != 0.0:
+            raise ValueError(
+                f"reference {self.reference!r} must have offset_s 0.0 "
+                f"(it is the device the others are measured against), "
+                f"got {entry.offset_s}"
+            )
+        return self
+
+    def offset_for(self, object_name: str) -> float:
+        """The device's offset in seconds; ``0.0`` when it was not measured."""
+        entry = self.devices.get(object_name)
+        return entry.offset_s if entry is not None else 0.0
--- a/GEECS-Schemas/pyproject.toml
+++ b/GEECS-Schemas/pyproject.toml
@@ -1,6 +1,6 @@
 [tool.poetry]
 name = "geecs-schemas"
-version = "0.22.1"
+version = "0.23.0"
 description = "Versioned Pydantic schemas for GEECS scanner configs (scan requests, save sets, scan variables, trigger profiles, action plans) plus legacy-YAML converters."
 authors = ["Sam Barber <sbarber@lbl.gov>"]
 readme = "README.md"
--- a/GeecsBluesky/CHANGELOG.md
+++ b/GeecsBluesky/CHANGELOG.md
@@ -5,6 +5,85 @@ All notable changes to `geecs-bluesky` are documented here.
 Format follows [Keep a Changelog](https://keepachangelog.com/en/1.1.0/).
 
 
+
+## [0.85.0] - 2026-09-13
+
+Phase 3 of the native-Bluesky rebuild (GEECS-Plugins#807,
+`Planning/native_bluesky/03_clean_room_rebuild.md` §4.F): the shot-offset
+calibration and its preflight.  **Every `drain_offset` in the field read
+`0.0` before this** — phase 2c threaded the offsets through both sides of
+the s-file join; this puts real numbers into them.
+
+At 1 Hz the join windows are ±0.5 s and swallow the ~36–100 ms device
+spread, so nothing was broken.  They narrow with the rep rate: at 5 Hz they
+are ±0.1 s, the same order as the spread, where an uncalibrated offset costs
+rows.  This is what makes faster running safe.
+
+### Added
+
+- **`measure_shot_offsets`** (`geecs_bluesky.plans.calibration`), a
+  registered queue plan: drives the trigger box OFF, waits the device set
+  quiet, then fires single shots through `fire_and_await_shot` and reads
+  every device's `acq_timestamp`.  The spread across devices is the
+  calibration.
+
+  Several shots are averaged (default 10) because each host's clock dithers
+  around its own average by up to ~10 ms while the domain holds the averages
+  on a common target (Sam, 2026-09-13) — one shot would measure a 36 ms
+  difference to ±10 ms.  The quiet wait, which dominates the cost, is paid
+  once; the shots after it cost about a second each.  Each device's
+  peak-to-peak scatter is recorded beside its mean, so a host with a
+  timekeeping problem is visible instead of hidden inside an average.
+
+  Shots are anchored on the *mean of their own stamps* before averaging, so
+  the laser's phase drifting between shots cancels and only the
+  device-to-device differences — the quantity the join uses — survive.
+  Only complete shots contribute: a shot one device missed would shift that
+  shot's anchor and bias every other device's offset, so it is discarded and
+  retaken.
+
+  Measures and reports by default; stores only with `write=True`, so a
+  re-run cannot silently replace a good calibration with a worse one.  No
+  run is opened — no scan number, no s-file — and the box is left in
+  STANDBY however the plan ends.
+
+- **`check_shot_sync`**, the preflight (§11.7): with the box OFF and the set
+  quiet, every device still holds the stamp of the same last real shot, so
+  correcting those stalled stamps by the stored offsets says whether the
+  calibration still holds — **at no shot cost at all**.  Raises when the set
+  is out of tolerance, so a queue that puts it ahead of its scans stops
+  before taking data against a stale calibration.
+
+  A queue item, deliberately, and never a step inside a scan: it costs at
+  least the longest device timeout every time it runs (§11.2 — a device's
+  timeout event carries an unchanged stamp, which the gateway's change
+  suppression drops, so nothing announces quiescence), and running it from
+  the queue also means it cannot drive the trigger box while a scan is using
+  it.
+
+- **`ConfigsRepoResolver.resolve_shot_offsets` / `write_shot_offsets`** over
+  the experiment's `shot_offsets.yaml`.  Absent reads as "never measured"
+  (every offset stays `0.0`); an *invalid* document raises rather than
+  falling back to zeros, because a calibration that silently reverted would
+  misjoin rows with nothing in the log to say why.  The write is atomic
+  (same-directory temporary plus `os.replace`), so a reader never sees a
+  half-written document and a failure leaves the previous calibration
+  intact, and it refuses to create a missing experiment folder.
+
+- **`GeecsNamespace(drain_offsets=...)`** seeds each detector's
+  `drain_offset` at construction, and `GeecsDetector(drain_offset=...)`
+  takes it.  The startup profile now builds the resolver *before* the
+  namespace to pass them; a re-measurement therefore reaches the worker at
+  its next environment open, not mid-session.  An offset naming no detector
+  in the namespace is warned about loudly as a stale calibration — the
+  device it meant to correct would otherwise be left at `0.0` silently.
+
+### Changed
+
+- `GEECS_PLAN_NAMES` grows to **22** plans; the operator permissions regex
+  and `NON_SCAN_PLAN_NAMES` gain both calibration plans.  The readiness
+  check (`geecs-qserver-ensure-ready`) asserts all 22.
+
 ## [0.84.0] - 2026-09-12
 
 Phase 2c of the native-Bluesky rebuild (GEECS-Plugins#807,
--- a/GeecsBluesky/geecs_bluesky/config_resolver.py
+++ b/GeecsBluesky/geecs_bluesky/config_resolver.py
@@ -19,6 +19,8 @@ The client seam expands a preset into a stock plan queue item
 from __future__ import annotations
 
 import logging
+import os
+import tempfile
 from pathlib import Path
 from typing import Protocol, runtime_checkable
 
@@ -33,6 +35,7 @@ from geecs_schemas import (
     Preset,
     ScanVariables,
     ScanVariableSpec,
+    ShotOffsets,
     TriggerProfile,
 )
 from geecs_schemas.convert import convert_shot_control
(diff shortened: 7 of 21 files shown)


skbarber:
## Adversarial review — fresh context, 19 findings

A fresh-context reviewer was given the diff, `§11` of the plan of record, and the instruction to assume there are bugs. It reproduced its two HIGH findings with probe scripts. **Both would have written a corrupt calibration to the share**, which is why the hardware run was held rather than taken on `991a0d4a`.

All fixes are in `89cabb3b`. Disposition table first, then the review verbatim.

### Disposition

| # | Severity | Finding | Disposition |
|---|---|---|---|
| 1 | HIGH | `GeecsDetectorScalars.trigger` never clears `fly`, so after any gated run the stamp wait is a no-op and the plan measures whole trigger periods | **Fixed** — the view now leaves fly mode exactly as the parent's `trigger` does. Fixed at the view, so it is right for every caller, not just calibration |
| 2 | HIGH | The quiet confirmation window (0.5 s) is shorter than the trigger period, so a running box is caught ~half the time | **Fixed** — window sized from the trigger period (`QUIET_CONFIRM_PERIODS = 1.5`); plus a second free signal, the whole set advancing across a wait that already exceeds the device timeout |
| 3 | MED-HIGH | Nothing bounds the measured numbers before they are written | **Fixed** — `MAX_PLAUSIBLE_OFFSET_S` / `MAX_PLAUSIBLE_SCATTER_S` refuse the *write*; the table is still reported, and `max_offset` is the deliberate escape hatch |
| 4 | MED-HIGH | `check_shot_sync` stops the queue on a healthy machine (a device holding the previous shot; range statistic grows with set size) | **Fixed, partly my shape** — whole trigger periods are folded out and reported rather than failed, and per-device median deviation names the culprit. I kept the **verdict on the pairwise spread**: see the note below |
| 5 | MED | No test pins that `check_shot_sync` reads the stored offsets at all | **Fixed** — the passing fixture's raw spread is now 120 ms against a 50 ms tolerance, so only correcting makes it pass |
| 6 | MED | The atomic write silently changes the file's mode to 0600 | **Fixed** — destination mode preserved (0644 for a new file), plus `fsync` before the rename |
| 7 | MED | A failed write leaks a temp file; the test only covers the success path | **Fixed** — whole temp lifetime in `try/finally`; new test forces the *write* to fail and asserts no leftover and the previous calibration intact |
| 8 | MED | `_read_stamps` does not do what its docstring says — `bps.rd` uses the monitor cache, which awaits its first update | **Fixed** — a genuine uncached get through `bps.wait_for`. The reviewer is right that the stated reason was inverted, and the blocking worst case is real under OFF |
| 9 | MED | The quiet-guard test runs the pacer 20× the real rep rate, so it cannot detect finding 2 | **Fixed** — pacer at 1.0 s, phase parametrized over four values |
| 10 | MED-LOW | `DeviceOffset.shots` cannot mean what its description says | **Fixed** — description rewritten to what it is (complete shots averaged); the retake counts stay in the log |
| 11 | LOW-MED | The connect-touch comment claims an ordering that `run_bracket` does not give | **Fixed** — comment corrected; the box is already OFF by then and the finalizer returns it |
| 12 | LOW-MED | No test asserts the box ends STANDBY on the exception paths | **Fixed** — asserted on the raising path; and the docstring/CHANGELOG no longer overstate `halt`, which skips finalizers by bluesky contract |
| 13 | LOW | `geecs_device` assertion is tautological in the fixture | **Fixed** — the fixture's GEECS name and ophyd name now differ |
| 14 | LOW | `pytest.raises(Exception)`; "could not check" reported as failure; `_RETAKE_BUDGET` wording; the `ConfigResolver` protocol; no ARMED/SINGLESHOT preflight; `shots=1` accepted silently; the plans' return values are unreachable from a queue client; `%.1f` renders a test's 0.01 s as "0.0 s" | **Fixed** except the last two. `SyncVerdict.comparable` separates "could not check" from "failed"; write capability and profile writes are checked before a measurement is spent; fewer than three shots warns; `resolve_shot_offsets` joins the protocol; `%.3g` for the times. **Not fixed:** a queue client cannot retrieve a plan's return value — that is a queueserver property, and the operator-facing report is the worker log until the web front end exists. Noted, not papered over |

**Where I kept my shape (finding 4, the median half).** The reviewer proposed judging each device's deviation from the set median instead of the range. I took that for *naming* the culprit but not for the verdict, because the join matches a frame to a row by **two corrected stamps** — pairwise disagreement is what costs rows. With two devices the median sits midway, so an 84 ms disagreement reads as ±42 ms and passes a 50 ms budget while the join sees all 84 ms. I confirmed that by making the change and watching `test_sync_fails_when_the_stored_offsets_are_stale` go green when it should not have. The reviewer's underlying concern is real and is now documented instead: a spread is a range statistic and a set of a few dozen devices may need a wider `tolerance_s`.

### Mutation results

**21 mutations, every one turns its guarding test red** — including one for each fix above:

```
test went RED (good)   drop the per-shot anchor (use raw stamps)
test went RED (good)   anchor on the latest device instead of the earliest
test went RED (good)   let a tie follow hash order instead of the name
test went RED (good)   drop the explicit 0.0 for the reference
test went RED (good)   skip the quiet confirmation window
test went RED (good)   average incomplete shots in instead of retaking
test went RED (good)   compare raw stamps without correcting by the stored offsets
test went RED (good)   ignore the seeded drain_offset at construction
test went RED (good)   write the measurement even without write=True
test went RED (good)   drop the namespace offset lookup
test went RED (good)   leave fly set in the scalars view (HIGH 1)
test went RED (good)   size the quiet window at half a period instead of 1.5 (HIGH 2)
test went RED (good)   drop the whole-set-advanced signal
test went RED (good)   store a physically impossible measurement anyway
test went RED (good)   skip the ARMED/SINGLESHOT profile preflight
test went RED (good)   stop folding whole trigger periods out of the sync verdict
test went RED (good)   fold ANY deviation away, not just whole periods
test went RED (good)   treat 'could not check' as a failure again
test went RED (good)   let os.replace hand the file the temp's 0600
test went RED (good)   guard only the replace, not the write (leak the temp file)
test went RED (good)   check write capability only for None, not for the method
test went RED (good)   create the experiment folder instead of refusing
test went RED (good)   fall back to zeros on an invalid calibration document
```

Two earlier probes were themselves no-ops — a `min`→`max` over already-sorted names, and an `and`→`or` that the existing test could not distinguish. Both were replaced with probes that do discriminate, and the second exposed a genuine test gap (no coverage of a resolver that satisfies the protocol *without* `write_shot_offsets`), now covered.

`./scripts/check.sh GeecsBluesky GEECS-Schemas` → exit 0. GeecsBluesky 470 passed, GEECS-Schemas 424 passed.

### Things the reviewer examined and found correct

Recorded so a later round does not re-derive them: the reduction arithmetic (the per-shot mean anchor removes the shot instant exactly and preserves every pairwise difference; mean-over-minimum is right because the min of noisy values is biased and injects the earliest device's dither into everyone else's); float precision (epoch stamps at 1.79e9 resolve to ~2.4e-7 s, three orders below the millisecond target); the sign convention against `shot_join.py:325`; `separate_devices` with the view, and the dedupe by `id(owner)`; that `.scalars` really does bypass `prepare` so no plugin is armed and no file claimed; `run_bracket` returning STANDBY on exceptions and `RE.stop()`; `connect_on_demand` reaching the view through `.parent`; and the plan-name plumbing at 22.

---

<details>
<summary>The review, verbatim</summary>

### 1. HIGH — after any gated run, the calibration's stamp wait is a no-op, and it silently measures whole trigger periods

`calibration.py:623` (via `_stamp_views` → `owner.scalars`, `detector.py:541-545`)

`GeecsAcquireLogic.fly` is set in exactly one place (`GeecsDetector.kickoff`) and cleared in exactly one place (`GeecsDetector.trigger`). Nothing else — not `complete`, not `unstage`, not the gated plan's teardown. `GeecsDetectorScalars.trigger` (the one this plan uses, deliberately, to avoid arming plugins) baselines and waits but **does not clear `fly`**. And `wait_for_idle` returns immediately when `fly` is true.

Failure scenario, entirely ordinary: run a gated scan, then queue `measure_shot_offsets` on those cameras in the same environment session. Every `fire_and_await_shot` returns instantly with `missed == []`, so a device that did not receive the shot at all is recorded as a complete shot with its stale stamp — no retake, no warning; and `_read_stamps` runs before the frames drain, so different devices can be recorded from different shots.

Reproduced, two cameras, one never delivering, `fly=True` on it:

```
reference: amp4 shots: 3
offsets  : {'amp3': 1.9640000000000555, 'amp4': 0.0}
scatter  : {'amp3': 1.0, 'amp4': 1.0}
```

A +1.964 s "drain latency" — against a documented physical range of 36–100 ms — passes `offsets_from_shots`, passes both schema validators, and `write=True` stores it.

### 2. HIGH — the quiet check is a coin flip at the real rep rate

`_settle_quiet` proves quiescence by reading the stamps, sleeping `QUIET_CONFIRM_S = 0.5 s`, reading again, and failing if anything moved. At the machine's 1 Hz, a 0.5 s window contains an edge with probability 0.5. The docstring says the window "only has to catch a box that never went OFF" — it catches that exactly half the time.

Reproduced with the pacer at `interval=1.0` and the phase swept over 12 values: `caught 6/12, NOT caught 6/12`. At 0.5 Hz it would catch 25 %.

Separately: `_settle_quiet` already computes `drained` and only logs it — if **every** device moved across a 2.25 s wait that is near-certainly a running box, and that is a free second signal.

### 3. MEDIUM-HIGH — nothing bounds the measured numbers before they are written

`offsets_from_shots` will return any finite number; the models validate only finiteness, `scatter_s >= 0` and reference-consistency. So a measurement corrupted by finding 1 or 2 is written without complaint. The evidence needed is already computed and thrown away: a scatter of 1.0 s, and an offset two orders past the documented range.

### 4. MEDIUM-HIGH — `check_shot_sync` will stop the queue on a healthy machine

Two ways it misfires. **(a)** A device that legitimately missed the last edge before OFF: §11.1 says STANDBY passes edges, so the set acquires right up to the moment this plan drives OFF; a long-exposure camera whose exposure was cut holds shot *k−1* while the fast ones hold *k* → spread ≈ one trigger period → `shot sync FAILED` on a perfectly calibrated machine. **(b)** `max − min` is a *range* statistic and grows with the number of devices; with ~10 ms dither plus 5–10 ms NTP, a 40-device set routinely ranges past 50 ms. The 50 ms budget was reasoned against a pair.

### 5. MEDIUM — no test pins that `check_shot_sync` reads the stored offsets at all

All three plan-level sync tests use raw spreads that already fall on the correct side of the tolerance without any correction. Mutating the plan so it never reads `drain_offset`: `3 passed, 27 deselected`. The unit-level test bites only via its second assertion, and its docstring's claim is false as written.

### 6. MEDIUM — the atomic write silently changes the file's permissions to 0600

`NamedTemporaryFile` creates 0600; `os.replace` preserves the temp inode's mode. Reproduced: `previous mode: 0o664 / new mode: 0o600`. The configs repo is a shared git checkout on the data share — after one `write=True` the file is readable only by the worker's service account.

### 7. MEDIUM — a failed write leaks a temp file, and the test that claims otherwise only covers the success path

Only the `os.replace` failure is cleaned up. Reproduced: `temp files left behind: ['.shot_offsets-arzw1o_0.tmp']`. The test's glob assertion runs only after a successful write — it would pass identically against a plain `write_text`.

### 8. MEDIUM — `_read_stamps` does not do what its docstring says, and the stated reason is inverted

`bps.rd` → `SignalR.read(cached=None)`, which uses the cache **whenever one exists** — and `attach()` creates it at connect. So this is a monitor-cache read, not a get. `_SignalCache.get_reading()` awaits `self._valid`, so a signal whose monitor has never delivered **blocks** until the signal timeout rather than returning the PV's held value.

### 9. MEDIUM — the quiet test cannot detect finding 2

The pacer runs 20× the real rep rate. The test passes with `QUIET_CONFIRM_S` shrunk to 0.06, and passes today with the guard broken for the actual hardware.

### 10. MEDIUM-LOW — `DeviceOffset.shots` cannot mean what its description says

Every device gets `measurement.shots`, and the plan raises unless that equals the request. So the field is always exactly the request and can never signal a missed shot.

### 11. LOW-MEDIUM — the connect-touch comment is wrong about ordering

`run_bracket` drives the box to OFF as its first message, before `inner()` runs. A missing device fails *after* the box has left STANDBY.

### 12. LOW-MEDIUM — no test asserts the box ends STANDBY on the exception paths

Also worth stating rather than fixing: `finalize_wrapper` does not run under `RE.halt()`, so "the box is left in STANDBY however the plan ends" overstates it.

### 13. LOW — `geecs_device` assertion is tautological in the fixture

The fixture builds `GeecsDetector("amp4", …, name="amp4")`, so passing the ophyd name instead of the GEECS name keeps the test green.

### 14. LOW — assorted

`pytest.raises(Exception)` would be satisfied by a typo; "could not check" is reported as "failed"; `_RETAKE_BUDGET` is documented as extra shots but used as a total; the `ConfigResolver` protocol was not extended, so a protocol-satisfying resolver without the write method fails with `AttributeError` after the measurement is spent; no preflight that the profile defines ARMED and SINGLESHOT; `shots=1` produces `scatter_s: 0.0` for every device; both plans return a value a queueserver client cannot retrieve; `%.1f` renders `quiet_time=0.01` as "0.0 s".

</details>

skbarber:
## The `fly` fix is not only a calibration fix — it closes silent data corruption in a shipped path

Following up review finding 1. The reviewer found it through the calibration plan, but the bug is **older than this PR and affects the `save_images: false` scan path**, so it is worth stating separately from the calibration story.

**The mechanism.** `GeecsAcquireLogic.fly` is one flag shared by a detector and its `.scalars` view. `GeecsDetector.kickoff` sets it; only `GeecsDetector.trigger` cleared it. Nothing else — not `complete`, not `unstage`, not the gated plan's teardown. And `wait_for_idle` returns immediately while it is set.

A preset with `save_images: false` expands to `X.scalars` and runs a stock **strict** scan. So:

> gated run on a camera → `fly = True`, never cleared
> → strict scan over `X.scalars` on that camera, same worker environment session
> → every shot's stamp wait returns instantly

**What that produces.** Probe on mocks, camera producing **no frames at all**, pre-fix:

```
fly left set by a preceding gated run; the box produces NO frames.
  strict count outcome : COMPLETED          <- no error raised
  rows recorded        : 2
    stamp 1000.0  meancounts 42.0           <- stale
    stamp 1000.0  meancounts 42.0           <- the same stale values again
  camera missed_shot   : False              <- nothing flagged
  fires                : 2
```

Two rows of real-looking per-shot data that are the same pre-scan monitor value twice. No exception, no `NaN`, no `missed_shot`, nothing in the log. The refire gate cannot fire because nothing reports a miss.

Post-fix, the same probe:

```
  strict count outcome : GeecsTriggerTimeoutError
  rows recorded        : 3
    stamp nan  meancounts nan
    stamp nan  meancounts nan
    stamp nan  meancounts nan
  camera missed_shot   : True
```

— the partial rows kept with empty columns and the step failing loudly, which is the designed behaviour (`06_pva_file_plugin.md` §2.1).

**No data on the share is affected.** I checked every s-file of `26_0912` (the day the 2b/2c acceptances ran, and the only day with gated runs on this branch) for the tell — consecutive identical `acq_timestamp` values in a row stream:

```
s7.txt : 6 rows, 1 stamp col -> no consecutive repeats
s8.txt : 2 rows, 1 stamp col -> no consecutive repeats
s9.txt : 1 rows, 1 stamp col -> no consecutive repeats
s10.txt: 3 rows, 1 stamp col -> no consecutive repeats
s11.txt: 5 rows, 1 stamp col -> no consecutive repeats
s12.txt: 5 rows, 1 stamp col -> no consecutive repeats
s21.txt: 3 rows, 2 stamp cols -> no consecutive repeats
```

Every stamp strictly advances. The acceptances listed the detectors themselves rather than their `.scalars` views in the sequences that followed a gated run, so the window never opened on real data.

**Exposure window:** the branch only, from #850 (when `kickoff`/`fly` arrived) to this PR. Master is unaffected — it has no gated mode.

The fix is at the view rather than in the calibration plan precisely because of this: `GeecsDetectorScalars.trigger` now leaves fly mode exactly as the parent's `trigger` does, so it is right for every caller. `test_a_camera_left_in_fly_mode_still_waits_for_its_stamp` pins it, and the mutation that removes the line turns it red.

skbarber:
## Second adversarial review — the delta, and confirmation of round 1

The first review covered `991a0d4a`. Four commits then landed **unreviewed**, and round 1's fixes were never confirmed by the reviewer (I had self-verified them with mutation probes, which is not the same thing). A fresh-context reviewer was given both gaps. **The HIGH finding was introduced by round 1's own fix.**

All fixes in `de032b9d`.

### Part A — round 1's claimed fixes, verified

| # | Claimed fix | Verdict |
|---|---|---|
| 1 | `GeecsDetectorScalars.trigger` clears `fly` (HIGH) | **CONFIRMED** — mutation red |
| 2 | Quiet window sized from the trigger period + whole-set signal (HIGH) | **CONFIRMED** — reverting to 0.5 turns 2 of 4 phase cases red, exactly the coin flip |
| 3 | Write-time plausibility bound | **Confirmed but incomplete** → finding 3 |
| 4 | Whole trigger periods folded out | **NOT CONFIRMED — defective** → finding 1 |
| 5 | `SyncVerdict.comparable` | **CONFIRMED** |
| 6 | Uncached stamp get | **Code confirmed, rationale rejected** → finding 5 |
| 7 | Mode + fsync + temp cleanup | **Confirmed**, but the chmod is itself a new hazard → finding 6 |
| 8 | Pre-flight checks before a measurement is spent | **CONFIRMED** |
| 9 | Rejecting the reviewer's median verdict | **Reasoning correct, execution half-done** → findings 1 and 4 |

### The HIGH: the sync check stopped the queue on its own routine case

`check_shot_sync` folded whole periods against the set's **median**. A median is not an instant any device reported — for an even-sized set it sits *between* the groups, so a device exactly one period out lands half a period from it, `round(±0.5)` is `0` (banker's rounding), nothing folds:

```
ONE period apart    synced=False  shots_out={}                     spread=1.000  -> RAISES
TWO periods apart   synced=True   shots_out={'fast': 1, 'slow': -1}  spread=0.000  -> passes
```

So the plan **stopped the queue** on the case its own docstring calls routine (a slow camera holding the previous shot, which STANDBY makes normal), while a two-period gap produced a *fabricated* "one ahead, one behind" attribution and waved through a device that had genuinely stopped being triggered. The earlier test covered only the odd-sized three-device case, which is why it survived.

Now anchored on the **latest corrected stamp** — an instant some device actually reported — so everything else is a whole number of periods behind it or genuinely mis-calibrated. Pinned for one, two and three periods, for an even-sized 2-2 split, and for the half-period case that must still fail. The failure message names the two devices at the ends of the disagreement instead of pretending to know which is wrong.

### Disposition of the rest

|---|---|---|---|
| 2 | MED | `trigger_rate_hz` derived from `trigger_period`, which this plan does **not** fire at — it fires single shots with a stamp wait between | **Fixed.** Declared by the caller (`measured_at_rate_hz`) or left unset. A wrong value in a field whose own description tells readers to trust it is worse than none |
| 3 | MED | The plausibility bound cannot catch a whole-period error above ~3 Hz (at 5 Hz a period is 0.2 s, under the 0.3 s cap) | **Fixed** — also bounded by half the trigger period |
| 4 | MED | Three passages still described the per-device-median verdict that was deliberately rejected | **Fixed** |
| 5 | MED | `_read_stamps`' rationale is contradicted by the plan's own first message (`bps.read(view)` goes through the same cache) | **Fixed** — docstring corrected; the uncached get is the right read, just not for the stated reason |
| 6 | MED | `os.chmod` can EPERM on the CIFS share, and it sat inside the `try` whose `finally` unlinks the temp — every `write=True` would fail and throw away the ten shots | **Fixed** — warns and carries on. A cosmetic permissions problem must not destroy a measurement |
| 7 | MED | The string-annotation test is a proxy with a real false negative: removing `detectors` from the mapping keeps it green while losing device-name conversion | **Fixed** — `resolve_annotations` now raises on an unmapped parameter (map to `None` to drop one deliberately). It immediately caught `measured_at_rate_hz` missing |
| 8 | LOW | One unreadable or never-acquired device permanently disabled the "whole set advanced" backstop | **Fixed** — denominator counts only devices usable in both reads |
| 9 | LOW | The confirmation window's guarantee rests on a defaulted `trigger_period` | **Documented, not enforced** — deriving the period from observed cadence is a bigger change than it earns |
| 10 | LOW | The refusal advises raising `max_offset`, which cannot fix a scatter refusal | **Fixed** — branched |
| 11 | LOW | `plan_report_sink` entered per statement, so everything between blocks was still discarded — including the resolver's own "written to `<path>`" | **Fixed** — one sink per plan, package-scoped |
| 12 | LOW | A second `__signature__`-rewriting mechanism in `registry.py` | **Not fixed.** `registry.py` builds a signature from scratch with added parameters — a different job; merging them would be forced. Noted |
| 13 | LOW | `trigger_rate_hz` landed with no CHANGELOG entry | **Fixed** |
| 14 | LOW | Prose asserting a 36–100 ms physical range, superseded by this branch's own 0–160 ms hardware run | **Fixed** |

### Mutations

Four new guards, each confirmed red:

```
RED (good)   anchor the period folding on the median again (the HIGH)
RED (good)   let chmod failure abort the write and waste the shots
RED (good)   derive the stored rate from trigger_period again
RED (good)   silently drop an unmapped parameter's annotation
```

`./scripts/check.sh GeecsBluesky GEECS-Schemas` → exit 0, **481 + 424 passed**.

### Confirmed safe by the reviewer

Worth recording so a third round doesn't re-derive it: `RunEngine._run` closes every generator on the plan stack even on `halt()`, so `plan_report_sink`'s `finally` always runs and no global logging state survives a stop/abort/halt; enter and exit both happen on the RE loop thread; `StreamHandler.close()` does not close `sys.stdout`; the worker's own stdout handler is scoped to the `bluesky_queueserver` logger, so there is no duplication; and queueserver's `return_value` plumbing is the `function_execute` path only — a plan's return value genuinely is unreachable by a client, so the sink is the right shape rather than a workaround.

skbarber:
## Third review — confirmation of round 2's fix commit `de032b9d`

Round 2's 14 fixes landed in one commit that nobody but its author had confirmed, and both earlier rounds had found a defect inside the previous round's own fix. A fresh-context reviewer was given the round-2 disposition table and `git diff d38c985f de032b9d`, told to assume the same, and asked to probe the HIGH's arithmetic directly. **It happened again: round 2's finding-3 fix refuses the real calibration at 5 Hz.** All fixes are in `55f442cc`.

### Part A — round 2's claimed fixes, verified

| # | Sev | Claimed fix | Verdict | Evidence |
|---|---|---|---|---|
| 1 | HIGH | Fold anchored on the latest corrected stamp | **CONFIRMED** | `anchor = max(corrected.values())`; probe below passes 1/2/3 behind, 2-2, 3-1, 1-3, half-period fails, sub-half fails, 5 Hz and epoch magnitudes; median mutation → 4 tests red |
| 2 | MED | `trigger_rate_hz` declared, not derived | **CONFIRMED** (but validated late → N3) | mutation red |
| 3 | MED | Cap also bounded by half the trigger period | **Present — wrong for real hardware, and untested** | removing it left all 46 tests green; and it refuses the HTU set at 5 Hz → **N1** |
| 4 | MED | Median-verdict prose removed | **PARTIAL** | still at `calibration.py:341`, `:353`, `:397` ("the median names the culprit"), `:501-503`; `CHANGELOG.md:176` |
| 5 | MED | `_read_stamps` rationale corrected | **CONFIRMED** | docstring-only |
| 6 | MED | chmod failure warns, does not abort | **CONFIRMED** | mutation red |
| 7 | MED | `resolve_annotations` raises on an unmapped parameter | **CONFIRMED** | mutation red; `action_compiler.py:318` maps `run_action` fully, nothing legitimate trips it |
| 8 | LOW | Backstop denominator counts only usable devices | **PARTIAL — the comment claims what the code does not do** | excludes only `None`; a camera holding `0.0` in both reads sits in `comparable`, never in `drained`, and disables the backstop. Reproduced with the RE: three cameras, one at 0.0, two advancing across the wait → no raise, 2 shots fired against a "running" box. Reverting the denominator left the suite green — untested either way |
| 9 | LOW | Documented, not enforced | as stated | — |
| 10 | LOW | Refusal advice branched | **NOT CONFIRMED** | the advice sentence was unconditional; a scatter-only refusal still said "raise max_offset" |
| 11 | LOW | One sink per plan, package-scoped | **CONFIRMED** | — |
| 12 | LOW | Not fixed, noted | as stated | — |
| 13 | LOW | CHANGELOG entry for the rate field | **PARTIAL** | Schemas had it; GeecsBluesky's 0.85.0 entry mentioned none of round 2's changes |
| 14 | LOW | 36–100 ms prose superseded | **PARTIAL** | still in `test_calibration.py:611`, `CHANGELOG.md:17` and `:164`, `03_clean_room_rebuild.md:455` |

### The finding that mattered — N1, MED/HIGH: the half-period bound refuses the real calibration at 5 Hz, with no escape

`bound = min(max_offset, trigger_period / 2)`. The hardware table in this PR has UC_ModeImager at **+160.3 ms**. An operator calibrating for 5 Hz running passes `trigger_period=0.2` (the docstring: "the machine's trigger period") → bound 0.1 s → the probe: `offset 0.16 @5Hz, default cap: REFUSED … "this is a bad measurement, not unusual hardware"`. Raising `max_offset` does nothing against the `min`, so the only escape is lying about `trigger_period`, which also mis-sizes the confirmation window.

The premise "a frame drain does not take a whole trigger period" is false at exactly the rate the feature exists for: 160 ms > 100 ms. At 5 Hz a whole-period error and a genuine slow drain are the same magnitude, and **no magnitude bound can tell them apart**. The defence against a whole-period error is upstream — the stamp wait that actually waits (round 1's `fly` fix) and the completeness rule (a shot any device missed is retaken) — not the cap. **Fixed:** the cap is `max_offset` alone; `trigger_period` no longer reaches `_refuse_implausible`. Pinned by `test_a_real_drain_longer_than_half_a_period_is_stored` (0.16 s at `trigger_period=0.2`, default cap, written).

### Disposition of the rest

|---|---|---|---|
| N2 | MED | The escape-hatch test passed for the wrong reason: with period 1.0 the effective bound was 0.5 and the fixture's offset exactly 0.5, so it passed only because `0.5 > 0.5` is False (`>=` mutation turned it red) | **Resolved by N1's fix** — with the period bound gone the test exercises what its comment says: 0.5 refused under the 0.3 default, accepted under `max_offset=1.0` |
| N3 | LOW/MED | `measured_at_rate_hz` validated by the document's `gt=0` only *after* the shots were spent (probe: `0.0` → 3 fires, then `ValidationError`) | **Fixed** — preflighted with the other arguments; `test_a_bad_rate_is_refused_before_any_shot` asserts no fire and no box write |
| 8 | LOW | Backstop denominator hole (above) | **Fixed** — `comparable` requires a usable stamp (read, non-zero) in both reads and `drained` is computed over `comparable`; `test_a_never_acquired_device_does_not_disable_the_backstop` reproduces the reviewer's scenario and expects the refusal |
| 10 | LOW | Advice unconditional | **Fixed** — branched per refusal kind; `test_a_scatter_refusal_does_not_advise_raising_max_offset` |
| N4 | LOW | `'behind' if k < 0 else 'ahead'` — with `anchor = max(...)`, `'ahead'` is unreachable | **Fixed** — dead branch removed |
| N5 | LOW | When the *latest* device is the mis-calibrated one, the good devices are the ones reported "behind" (acknowledged in a comment, but the message read as fact) | **Fixed** — one sentence added to the failure message saying so |
| 4, 13, 14 | LOW | Stale prose | **Fixed** — median wording gone from docstrings, comments and CHANGELOG; round-2/3 block added to the GeecsBluesky CHANGELOG; 36–100 ms replaced by the measured 0–160 ms in the test, CHANGELOG and `03` |
| N6 | cosmetic | `test_a_half_period_offset_is_still_a_failure` pinned `shots_out == {}` for exactly −0.5, which is banker's rounding, not the invariant | **Fixed** — asserts the failure and the named ends only |
| N6 | cosmetic | two long docstring/comment lines (E501 not selected) | **Not fixed** |

### Probe — `sync_verdict_from_stamps` (period 1 s, tolerance 50 ms unless noted)

```
1 / 2 / 3 periods behind         synced=True  shots_out={'slow': -1/-2/-3}   spread=0.000
2-2 split exact / +30 ms         synced=True  shots_out={'c': -1, 'd': -1}   spread=0.000 / 0.030
3-1 split, 1-3 split             synced=True                                 spread=0.000
exactly -0.5 / ±1e-9             synced=False                                spread=0.500
-1.5 / -2.5                      synced=False shots_out={'slow': -2}         spread=0.500
60/100/250/400/490 ms out        synced=False shots_out={}                   spread=as given
latest +0.4 (others fine)        synced=False shots_out={}                   spread=0.400
latest +0.6 (others fine)        synced=False shots_out={'a':-1,'b':-1}      spread=0.400
latest +1.0                      synced=True  shots_out={'a':-1,'b':-1}      spread=0.000
fresh calib 120 ms               synced=True ; stale (36 stored / 120 real) synced=False spread=0.084
5 Hz: 1 and 3 behind             synced=True ; 5 Hz 0.1 (half) out           synced=False
epoch 1.7577e9 magnitudes        all as above
```

The arithmetic of the round-2 HIGH fix is right in every case constructed.

### Mutations

Round 2's four guards re-confirmed by the reviewer (median anchor → 4 red; chmod abort → red; derived rate → red; dropped annotation → red). Round 3's four new guards, each confirmed red before this was pushed:

```
RED (good)   half-period bound reinstated (0.1 s)     -> test_a_real_drain_longer_than_half_a_period_is_stored
RED (good)   0.0 counted as a comparable stamp         -> test_a_never_acquired_device_does_not_disable_the_backstop
RED (good)   rate preflight dropped                    -> test_a_bad_rate_is_refused_before_any_shot
RED (good)   scatter advice says "raise max_offset"    -> test_a_scatter_refusal_does_not_advise_raising_max_offset
```

### A CI gap worth knowing

**CI is green on every head of this PR but has executed no calibration test.** The unit-tests job installs GeecsBluesky without the `ca` extra, so the mock-CA layer skips (`248 passed, 25 skipped`) and every test in `test_calibration.py` sits behind `importorskip("aioca")`. The only evidence for this module is local (`check.sh` with the extras installed: `./scripts/check.sh GeecsBluesky GEECS-Schemas` → exit 0, 485 passed + 424 passed). Filed as #866.

skbarber:
## Codex review — one finding, fixed in `73de1a1f`

Sam ran a Codex review alongside round 3. Its finding, verbatim:

> **P2: Calibration plans don't preflight OFF/STANDBY before using `run_bracket`.** In `calibration.py:929`, `measure_shot_offsets` checks only ARMED and SINGLESHOT, but the plan also relies on OFF at entry and STANDBY in the finalizer. `check_shot_sync` does not preflight either state. Since trigger profiles can omit states, a profile missing STANDBY can spend the quiet wait / shots and then fail during finalization, leaving the trigger box in the prior calibration state instead of the documented STANDBY. Add upfront checks for OFF and STANDBY in both calibration plans, before any device reads or waits, and pin with missing-state tests.

| Sev | Finding | Disposition |
|---|---|---|
| P2 | OFF/STANDBY not preflighted; a profile missing STANDBY fails in the finalizer after the shots, box left in the calibration state | **Fixed.** `measure_shot_offsets` refuses a profile missing any of OFF, STANDBY, ARMED, SINGLESHOT before any wait or shot; `check_shot_sync` refuses one missing OFF or STANDBY before the quiet wait. The refusal message now says why every driven state is required. Four parametrized tests (`test_measure_refuses_a_profile_that_cannot_bracket[OFF|STANDBY]`, `test_check_shot_sync_refuses_a_profile_that_cannot_bracket[OFF|STANDBY]`) assert the refusal happens before the box is written at all (`box.puts == []`) |

The mechanism is exactly as Codex describes: `run_bracket` (`gated.py:518`) does `mv(shot_control, OFF)` first and `mv(shot_control, STANDBY)` in a `finalize_wrapper`, and `ShotControl._drive` raises "defines no writes" for an undefined state — so a missing STANDBY raised only inside the finalizer. A missing OFF failed early anyway (first message of the bracket), but is now refused with the same message as the rest rather than `_drive`'s generic one.

Mutations, each confirmed red before pushing:

```
RED (good)   OFF/STANDBY dropped from the measure preflight   -> 2 failed (both parametrizations)
RED (good)   OFF/STANDBY preflight dropped from check_shot_sync -> 2 failed
```

`./scripts/check.sh GeecsBluesky GEECS-Schemas` → exit 0, **489 passed** + 424 passed (local, with the `ca` extra — CI still runs none of these, #866).

PR #864, opened 2026-09-13
Fix: native-path 1D scans classify as NOSCAN (motors vs motor)

skbarber:
Every 1D scan taken on the native path shows as a **`NOSCAN`** in the data portal. Sam hit this looking for the 26_0912 motor scans and found none of them.

## What is wrong

`tiled_schema.scan_mode`, `scan_variable_columns` and `is_stepped_scan` all classify from the start document's **singular `motor`** key — what the retired GEECS funnel wrote. The stock `bluesky.plans` verbs the native-Bluesky scanner registers write **`motors`**, plural, a list.

So for every native-path scan those three readers see a motorless run.

Measured on a real run — 26_0912 Scan018, a `rel_scan` over `U_CompAerotech Position.Axis1`:

```
   plan_name       = 'rel_scan'
   motors          = ['u_compaerotech-position_axis1']
   num_points      = 5
   -> scan_mode() = 'NOSCAN'
```

while the run's own `ScanInfoScan018.ini`, written by the same scan, correctly says:

```
Scan Parameter = "U_CompAerotech Position.Axis1"
ScanMode = "standard"
Start = -20.0 / End = 20.0 / Step size = 10.0
```

which is exactly what makes it confusing to look at.

## Why it is more than a wrong chip

**Four** readers (the first draft of this PR found three; the review found the fourth):

| reader | consequence for every native 1D scan |
|---|---|
| `scan_mode` | the portal's mode chip says NOSCAN |
| `scan_variable_columns` | returns `[]` — **the scanned axis is not offered as an X axis at all** |
| `is_stepped_scan` | `False` — per-step statistics are disabled |
| `tiled_catalog.metadata_rows` | the Overview table's **"Scan variable" row vanishes** — on the same page the chip is on. Rendered by both the portal run page and the console scan browser |

Plus the same bug class in a different key: `metadata_rows` read `acquisition_mode` where the native scanner writes `acquisition`, so the **Mode row lost its strict/gated suffix** on every native run.

All 8 of 26_0912's 1D scans are affected (Scan004/005/006 and 014/015 over `U_S1H Current`, 017/019 over `U_ModeImagerESP`, 018 over `U_CompAerotech`).

## The fix

One `scan_motors()` helper reading `motors` with `motor` as the fallback; the readers share it. Runs from either backend classify identically, so re-exports and old funnel runs are unaffected.

**`plan_pattern` decides GRID, not the motor count.** The first draft classified `len(motors) > 1` as a grid — safe only because the funnel wrote a list *solely* for grids. Stock `scan` / `rel_scan` / `list_scan` correlate N motors along **one** trajectory (`inner_product`, `inner_list_product`) and are 1D however many motors they name; only `grid_scan`'s `outer_product` is a grid. So the first draft would have turned a multi-motor `scan` from wrongly-`NOSCAN` into wrongly-`GRID`. Caught in review; `scan_mode` now reads `plan_pattern` where it exists and keeps the count-based reading for funnel documents, which have none — the same discriminator GeecsBluesky's ScanInfo writer already uses.

## Tests

New cases use the **real start-doc shapes from both backends**, taken from actual runs rather than invented. Verified they bite: reverting `scan_motors` to the singular key turns the five native-path assertions red

```
FAILED test_a_native_1d_scan_is_not_a_noscan          - assert 'NOSCAN' == '1D'
FAILED test_a_native_grid_is_a_grid                   - assert 'NOSCAN' == 'GRID'
FAILED test_is_stepped_scan_sees_a_native_scan        - assert False is True
FAILED test_the_scanned_axis_column_is_found...       - assert [] == ['u_compaerot...sition_axis1']
FAILED test_scan_motors_normalises_both_shapes        - assert [] == ['u_compaerot...sition_axis1']
```

while the two funnel-compatibility cases stay green — which is the point: the fix must not trade one backend for the other.

The review reproduced that reversion experiment independently rather than taking my word for it, and confirmed no test was weaker than claimed. It also named four reachable shapes that were uncovered — a tuple (the real in-memory shape for `grid_scan`; only JSON makes it a list), both keys present, `motors: None` with a populated `motor`, and a non-iterable — all now pinned, along with the Overview rows for both backends, which had no test at all.

`./scripts/check.sh GEECS-Data-Utils` → exit 0, **388 passed**, 4 skipped.

## Note

This is **not** why those scans have no s-file — that is the separate, known issue that a gated run could not write one until #858 landed, and it is backfillable with `write_scalar_files_from_tiled(uid)`. The two together are why the day's 1D work is invisible in the portal.

**Known residue, deliberately not in this PR:** `metadata_rows`' "Scan axes" and "Grid" rows read the funnel-only `scan_axes` / `num_grid_points`, so they are dead on native runs. They need remapping to the stock keys (`shape`, `plan_pattern`) rather than a rename, which is a different change. The Console has its own copy of the `acquisition_mode` read and is left alone — it is being deleted.

GEECS-Data-Utils 0.31.0 → **0.31.1**.


skbarber:
## Adversarial review — fresh context, 7 findings

This PR was opened without a review (my miss — it came out of mid-conversation debugging rather than a deliberate landing). A fresh-context reviewer was given the diff and the instruction to assume a small diff is not automatically a correct one. It was right to: **two of the findings are in the fix itself**, including a regression I introduced.

All fixes in `c52cb3e2`. Disposition first, then the review.

### Disposition

| # | Severity | Finding | Disposition |
|---|---|---|---|
| 1 | HIGH | The fix missed a **fourth** reader of the singular key — `tiled_catalog.metadata_rows`, the Overview table on both the portal run page and the console scan browser. On Scan018 the "Scan variable" row silently vanished while the chip beside it was fixed | **Fixed.** My commit message claimed three call sites; there were four. Neither that row nor the Mode row had any test — both now pinned for both backends |
| 2 | MEDIUM | **My fix caused a regression**: classifying GRID on `len(motors) > 1` turned a multi-motor `scan` from wrongly-`NOSCAN` into wrongly-`GRID` | **Fixed** by reading `plan_pattern` — stock `scan`/`rel_scan`/`list_scan` correlate N motors along one trajectory (`inner_product`), only `grid_scan`'s `outer_product` is a grid. Funnel documents have no `plan_pattern` and keep the count-based reading. This is the discriminator GeecsBluesky's own ScanInfo writer already uses |
| 3 | MEDIUM | `motors: []` shadows a populated `motor` — the fallback tested `is None`, not falsiness | **Fixed** (`or`). Latent today, but the helper's whole job is to be the one tolerant reader |
| 4 | MEDIUM | The `grid_shape`/`scan_axes` branch is funnel-only vocabulary, inert on the native path it was preserved for | **Fixed** — folded into the `plan_pattern` handling |
| 5 | LOW | Tests bite exactly as claimed, but four reachable shapes uncovered | **Fixed** — tuple (the real in-memory shape for `grid_scan`), both keys present, `motors: None` + `motor`, non-iterable |
| 6 | LOW | Same bug class in three more keys the PR didn't sweep | **Partly fixed.** `acquisition_mode` → `acquisition` fixed in the same function (the Mode row lost its strict/gated suffix on every native run). `scan_axes`/`num_grid_points` and the Console's copy left alone — the Console is being deleted, and those rows want their own change |
| 7 | LOW | Changelog dated 2026-09-13; today is 2026-09-12 | **Fixed** |

The reviewer **independently reproduced** my "five red on reversion, two green" claim by patching scratch copies, rather than taking my word — and confirmed no test in the new class was weaker than claimed.

### Mutation results

Five new mutations, one per fix above, each turning its guarding test red:

```
RED (good)   metadata_rows back to the singular motor key
RED (good)   Mode row back to acquisition_mode only
RED (good)   count motors instead of reading plan_pattern (the GRID regression)
RED (good)   `is None` fallback: an empty motors shadows a real motor
RED (good)   the whole fix reverted: singular key only
```

`./scripts/check.sh GEECS-Data-Utils` → exit 0, **388 passed**, 4 skipped.

### Examined and found correct

Recorded so a later round doesn't re-derive it: patch is the right version level (no API removed, the visible change is a *correction*); backward compatibility is clean and all five pre-existing `TestScanClassification` cases stay green; `tiled_schema.py` is the right home, and its own docstring plus `EVENT_SCHEMA.md` already named `motors` as the stock key — so reading `motor` there violated the file's own cited contract; correctly not exported, matching its siblings; no pre-existing helper duplicates it (`callbacks.py` has a plural-only read but lives in GeecsBluesky, which Data-Utils must not import — wrong direction in the dependency graph); no template or JS reader of either key anywhere in the repo.

### Follow-up worth filing separately

`tiled_catalog`'s "Scan axes" and "Grid" rows read the funnel-only `scan_axes` / `num_grid_points`, so they are dead on native runs. Not fixed here — they need their own mapping from the stock keys (`shape`, `plan_pattern`) rather than a rename, which is a different change from this one.

skbarber:
## Confirmation review of `c52cb3e2` — fresh context

The review above was dispositioned by the fix's author and self-verified with mutation probes; nobody else had confirmed the fix commit. A fresh-context reviewer was given the disposition table and `git diff f46a5853 c52cb3e2` with the instruction to assume the fix commit may itself be wrong, as the first commit's fix was.

**Verdict: the fix commit is correct and introduces no regression.** The one change it produced is the comment fix in `8e4876d9`.

### The 7 findings

| # | Claimed fix | Verdict | Evidence |
|---|---|---|---|
| 1 | `metadata_rows` fourth reader, both rows pinned | **CONFIRMED** | `tiled_catalog.py:655-658` reads `scan_motors(start)`; mutation back to the singular key turns 2 tests red |
| 2 | GRID from `plan_pattern`, count fallback for funnel docs | **PARTIAL** | Code correct (`outer_*` → GRID; `inner_*`/`linspace`/`logspace` → 1D; absent → count rule); mutation red. But the corroboration was false: GeecsBluesky's ScanInfo writer (`callbacks.py:513-518`) writes `ScanMode = "standard"` for any motors and reads `plan_pattern` only for Start/End/Step — it has no grid concept. The comment at `tiled_schema.py:544` said otherwise → fixed in `8e4876d9` |
| 3 | `or` fallback so `motors: []` doesn't shadow `motor` | **CONFIRMED** | `tiled_schema.py:478`; `is None` mutation red |
| 4 | `grid_shape`/`scan_axes` branch "folded into the plan_pattern handling" | **PARTIAL** | Not folded — it survives verbatim after the pattern early-return, with a funnel-vocabulary comment. Functionally right (still needed by `test_grid_from_grid_shape`, unreachable for native docs); the wording overstated the change |
| 5 | Tuple / both keys / `None`+`motor` / non-iterable covered | **CONFIRMED** | `TestScanMotorsShapes`, all four present; full reversion turns 12 tests red |
| 6 | `acquisition` Mode row; Console copy left alone | **CONFIRMED** | Base `plans/registry.py:286` writes `md["acquisition"]`; mutation red. Console's `browser_window.py:856` still reads `acquisition_mode` only — accepted, the Console is being deleted |
| 7 | Changelog date | **CONFIRMED** | `## [0.31.1] - 2026-09-12` |

### Mutations — all five re-run on a scratch copy of `c52cb3e2`

```
RED   metadata_rows singular motor        2 failed, 87 passed
RED   Mode row acquisition_mode only      1 failed, 88 passed
RED   count motors, ignore plan_pattern   2 failed, 87 passed
RED   is-None fallback                    1 failed, 88 passed
RED   whole fix reverted (singular only)  12 failed, 77 passed
      UNMUTATED                           89 passed, rc 0
```

### Edge-case probe (23 start-doc shapes)

`count` with or without `plan_pattern` → NOSCAN; `plan_pattern` with zero/missing motors → NOSCAN; funnel `motor` str → 1D, funnel `motor` list → GRID (the funnel wrote a list only for its outer-product branch, `origin/master:scan_request_runner.py:1177-1183`); `scan_axes`+`grid_shape` → GRID with Scan axes / Grid rows intact; `motors` tuple → GRID for `outer_*`; `motors: []`+`motor` → 1D; `num_points` missing → 1D; unknown pattern → 1D. The five pre-existing `TestScanClassification` cases stay green.

### Callers checked

Portal `_default_x` (`app.py:265-267`), Console `browser_window.py:873/923/929/977` (its `test_grid_metadata_rows_from_scan_axes` still holds — `scan_axes` wins the `if`), MCP `read_tools.py:210`. No template or JS reads either key. `is_stepped_scan` flipping True on native runs only unlocks the default-X pick and the per-step combo.

### New findings

| Sev | Where | Finding |
|---|---|---|
| LOW | `tiled_schema.py:544` | False corroboration comment — **fixed** in `8e4876d9` |
| LOW, pre-existing | the 6 spiral verbs in `plan_names.py` | Stock `spiral*`/`rel_spiral*` write **no `motors` key** (only `plan_args.x_motor`/`y_motor`, `extents`, `plan_pattern`), so they still classify NOSCAN with no X column and no Scan variable row. The ScanInfo writer has the same blind spot (`scan_parameter` → `"Shotnumber"`), so the ini and the portal agree. Belongs with the "Scan axes / Grid rows dead on native runs" follow-up — same shape of fix (read `plan_args.x_motor`/`y_motor` or `hints.dimensions`) |
| INFO | `tiled_catalog.py:655,658` | `scan_motors(start)` evaluated twice; cosmetic |

### Checks

`./scripts/check.sh GEECS-Data-Utils` from a detached scratch worktree at `c52cb3e2` → **exit 0**, 392 passed (the 4 skips in the earlier count are the tiled/pandas extras, present in this env).

Issue #866, opened 2026-09-13
CI never runs GeecsBluesky's mock-CA tests (no `ca` extra in unit-tests.yml)


skbarber:
## CI executes none of GeecsBluesky's mock-CA tests

`.github/workflows/unit-tests.yml` installs GeecsBluesky with `--extras qserver --extras qs-client` only (line 76). Every test that imports `aioca` / `p4p` — the whole mock-CA layer, including all of `tests/test_calibration.py` (50 tests as of #861) — sits behind `pytest.importorskip("aioca")` and **skips** in CI: the job reports `248 passed, 25 skipped` on a head where the local run with the `ca` extra installed reports `485 passed`.

Found during the third review round of #861, where CI was green on every head while the calibration plans' tests had never run anywhere but a developer machine.

## Why it matters

The green check on a GeecsBluesky PR currently says nothing about the detector, shot-control, or calibration plan code paths. Two of #861's review rounds found HIGH defects in exactly that code; CI would not have caught a regression in any of them.

## Options

1. Add `--extras ca` to the GeecsBluesky install step. Cost: `aioca` and `p4p` wheels on the Ubuntu runner (p4p ships manylinux wheels; aioca depends on `epicscorelibs`, also wheeled) — needs a trial run to confirm install time and that the mock-CA tests are hermetic on a runner with no network.
2. If (1) is too slow for every PR, run the `ca` layer as a separate job gated on paths under `GeecsBluesky/`.

Either way the skip count should become an assertion: a job that expects the CA layer should fail if it skipped.

PR #845, opened 2026-09-12
CI: console-windows job off by default, toggled by a repository variable

skbarber:
Keeps the `console-windows` job (the GEECS-Console suite plus the qs_client tests on `windows-latest`) in `.github/workflows/unit-tests.yml` but gates it: `if: ${{ vars.CI_CONSOLE_WINDOWS == 'true' }}`. The console is not being changed and is slated for retirement, and the job added ~5 minutes to every CI check (owner request, 2026-09-12). The first cut deleted the job; the owner reconsidered and asked for a toggle instead — this is that.

**Turning it on/off, no commit:** repository Settings → Secrets and variables → Actions → Variables → `CI_CONSOLE_WINDOWS` = `true`. Unset or anything else = off. While off the job reports *skipped* on every run (no branch-protection rule names it — master is unprotected — so nothing blocks).

**What is off while it is off, stated plainly (review finding 1):** that job is the console suite's *only* CI leg — the Ubuntu job has never run GEECS-Console. So with the variable unset the console suite is local-only: `scripts/check.sh` still runs it when console files change. The qs_client suite keeps its Ubuntu leg (`--extras qs-client`).

Docs: `GEECS-Console/CLAUDE.md` records the gate and how to flip it; console CHANGELOG 0.30.1 (docs-only patch bump per convention); `CONTRIBUTING.md` § Tests rewritten to match the workflow as it actually is (it still described a greenfield-branch Windows job and mis-listed which suites are in CI).

Tests: none affected (workflow + docs). pre-commit (check-yaml) passed at commit; the workflow parses with both jobs present and the `if:` on `console-windows`. The CI run on the final commit is the live proof that the job skips.

Adversarial review: 2 findings, both fixed and verified by the same reviewer (comments below); the rework from delete to toggle post-dates the verification and touches the same three doc surfaces plus the workflow, so the reviewer's checks (YAML validity, CONTRIBUTING list vs workflow) were re-done by hand.

Hardware verification: none — CI only.

PR #867, opened 2026-09-13
Merge master into feature/native-bluesky-plans (logbook, web kit, portal 0.26, #817/#818/#820)

skbarber:
## What

Merges `master` @ d6f74211 (95 commits since the merge base 39e86924) into `feature/native-bluesky-plans` @ 29f5c44e, so the web console can be built on a branch that has both the rebuilt backend and the web surface kit. One true merge commit (two parents), opened as a PR so CI runs on the resolved tree before the worker host picks anything up.

What comes over from master: the logbook (GeecsLogbook, `/log` on the portal unit), the web surface kit (GeecsWebTheme), the portal at 0.26.0, the #817/#818/#820 GeecsBluesky fixes, the CI changes (#845 console-windows gate, #866 note), docs.

## How the 24 conflicts were resolved

**Mechanical (13 files).** Version + CHANGELOG pairs for GEECS-Console, GEECS-Data-Utils, GEECS-DataPortal, GEECS-Schemas, GeecsBluesky and the portal `poetry.lock`.

| Package | master | branch | merged |
|---|---|---|---|
| GEECS-Console | 0.30.1 | 0.32.0 | **0.32.1** |
| GEECS-Data-Utils | 0.27.1 | 0.31.1 | **0.32.0** (minor: new public API on the branch line) |
| GEECS-DataPortal | 0.26.0 | 0.21.5 | **0.26.1** |
| GEECS-Schemas | 0.22.0 | 0.23.0 | **0.24.0** (minor: new public API on the branch line) |
| GeecsBluesky | 0.78.0 | 0.85.0 | **0.85.1** |

Both changelog lines are kept, interleaved in version order. Where the two lines reused the *same* version number for different releases (Data-Utils 0.27.0, Schemas 0.21.0/0.22.0, Bluesky 0.77.0/0.78.0) master's block is kept and marked *(master line, parallel release)*. The portal takes master's 0.26.0 code; its lock was relocked from the merged tree (the diff is the four path-dep versions) and the `scrub-lock-paths` hook normalized the URL — nothing hand-edited.

**Delete vs modify (7 files) — deletions win, each master fix checked against the rebuilt code by reading:**

- `plans/single_shot.py`, `plans/step_scan.py`, `tests/test_single_shot_plan.py`, `tests/test_pause_semantics.py` (#817/#818): the branch's `fire_and_await_shot` already refires only on a `GeecsTriggerTimeoutError` cause and re-raises a refused fire untouched (pinned by `test_a_failed_fire_is_not_a_dropped_frame`). What it lacked was #817's **ERROR line naming the cause** — the `FailedStatus` propagates with the status repr as its text, so without that line the scan log never names the refused PV. Carried as `_log_non_frame_failure` in `plans/strict.py`, wrapping both the fire (`bps.mv` raises right there, before any wait — found by the new test) and the wait loop; `is not None` + `%s`, never `or` + `repr`, because a failed `aioca.CANothing` is falsy and its repr is the bare error code. The gated path's batch-failure message had the same `{cause!r}` loss; now `type: str`. The #861 fly-flag history was read first; it is unrelated (a scalars view left in fly mode) and untouched.
- `session.py`, `tests/test_session.py` (#820): the namespace already binds a `CaMotor` with the DB's per-variable `tolerance` when positive and a plain `CaSettable` otherwise (`GeecsNamespace._movable`), so the per-axis-tolerance half of #820 is native to the branch. The ULP-slack half auto-merged into `motor.py` and `confirm.py`. Not carried: the span-relative "suspect tolerance" warning (a diagnostic; noted below as a follow-up candidate). `motor.py`'s auto-merged docstring named the deleted `GeecsSession`; reworded to the namespace.
- `tests/test_ca_devices.py` (content conflict): master's two motor on-boundary tests had auto-merged; the confirm on-boundary test sat in the conflict block and is carried. The rest of master's block (disconnect/save-control-only tests) covers code the branch deleted — dropped.

**Real content (4 files + 2 docs):**

- `tiled_catalog.py`, `tiled_export.py`, `tests/test_tiled_catalog.py`: the branch's versions. Verified by diffing master's files against the result: master carries nothing the branch lacks (its singular `motor`/`acquisition_mode` readers are the branch's fallbacks). **#864 intact**: the plural `motors` key with `motor` fallback, GRID from `plan_pattern`, and the `TestOverviewRowsReadBothBackends` suite all survive.
- `geecs_schemas/__init__.py`: union — master's logbook exports (`LogEntry`, `Book`, …) plus the branch's `ShotOffsets`/`Preset` exports — minus master's `save_set` import (the branch deleted that module in 0.21.0).
- `docs/platform/fleet_map.md`: master's logbook amendments to the portal row/node, but **not** the capture-daemon row — the branch deleted that service in #806. Service count stays "four". `site_profile.md`: union (the `[pva]` keys and the portal memory-ceiling keys).

## Per-concern LOC beyond the merge itself

Hand edits on top of the mechanical resolution (everything else is one side or the other):

| Concern | Files | ~LOC |
|---|---|---|
| #817 counterpart: ERROR line at the fire + wait loop, gated message | `plans/strict.py`, `plans/gated.py` | +30 |
| Pinning test (falsy, repr-vs-str stand-in) | `tests/test_strict_plans.py` | +30 |
| #820 counterpart: confirm on-boundary test, motor docstring | `tests/test_ca_devices.py`, `devices/ca/motor.py` | +14 |
| Merge blocks in five CHANGELOGs | `*/CHANGELOG.md` | +45 |

Judgment calls, cheap to veto: carrying the ERROR line rather than filing it (the branch otherwise logs nothing when a fire is refused); the `(master line, parallel release)` changelog marker.

## Console

Per Sam (this session): the console cannot run a scan against the branch's worker (it submits `geecs_scan_request_plan`, which the branch's worker no longer registers) and its tests are local-only everywhere since #845. Its conflicts here were version + changelog only and were resolved mechanically; nothing was done to make it work. On the branch it is dead code pending the web console.

## Tests

Run locally, the way CI runs each suite (worktree envs have no Poetry envs of their own; each run below used the CI invocation with the merged tree first on `sys.path`, verified by `__file__`):

| Suite | Env | Result |
|---|---|---|
| GEECS-Schemas | root env, `-m "not integration and not gui"` | 436 passed, 7 deselected |
| GEECS-Data-Utils | root env, same markers | 401 passed |
| GeecsBluesky | ca-extra env (#866: CI runs no mock-CA tests), CI markers | 492 passed, 3 skipped, 1 deselected |
| GEECS-DataPortal | fresh env, `--extras analysis --extras log` | 260 passed |
| GeecsLogbook | fresh env | 240 passed, 1 skipped |

The new test was proven to bite three ways before landing: `or` instead of `is not None` (logs the status repr), `repr` instead of `str` (logs `ECA_BADTYPE`), and no log at the fire — each fails; restored, passes. Not run: GEECS-Console (see above), GeecsWebTheme / GEECS-Core / gateways / MCP (no conflicts, auto-merged; CI covers them).

Lint: pre-commit on every hand-resolved file passes; the merge commit went through `scripts/commit.sh`.

## Hardware verification

OWED (the carried #817 line): against the live gateway, a refused `SINGLESHOT` put (e.g. a `:SP` put to a nonexistent PV with a camera armed read-only, as #818 did on 2026-09-10) should show **one** fire, **no** refire, no event row, and one `geecs_bluesky.plans.strict` ERROR line naming `CANothing` with the PV and the CA message. Nothing else in this PR changes scan-execution behavior.

## Facility literals

None added outside docstrings/tests.

## Follow-ups (not in this PR)

- #820's span-relative "DB tolerance looks like a units mismatch" warning has no home on the branch (`GeecsNamespace._movable` reads only `tolerance`); file if wanted.
- #868: the move-side half of #817 (a refused motor `:SP` put leaves no PV name in the scan log) — found by the review, filed.
- After merge: move the worker host's checkout from `phase/08-shot-offset-calibration` to the branch head once, reinstall, restart the queueserver (Sam's action).
- Out of scope, each its own session: deleting GEECS-Console, the web console, the portal's image route for gated HDF5 stacks.

## Review

Bulk integration merge whose constituent PRs were each reviewed; the adversarial review here is scoped to the conflict resolutions and the carried counterparts (the hand-edited files above), posted as a comment.


skbarber:
# Adversarial review of #867 (f52da4be)

**Scope.** The conflict resolutions and hand-carried counterparts only: `plans/strict.py`, `plans/gated.py`, `tests/test_strict_plans.py`, `devices/ca/motor.py` + `confirm.py` + `tests/test_ca_devices.py`, the three Data-Utils tiled files, `geecs_schemas/__init__.py`, `docs/platform/fleet_map.md` + `site_profile.md`, the five CHANGELOG/pyproject pairs, the portal lock, and the six delete-vs-modify files (read master's versions and diffed master's changes since the base 39e86924 against the rebuilt code). Not reviewed: the 95 master commits themselves. Read-only; no files modified.

## Findings

### 1. Low — The fire-path handler mislabels a trigger-timeout status that the RunEngine throws into `fire()`

`GeecsBluesky/geecs_bluesky/plans/strict.py:154-159` catches *every* `FailedStatus` around `yield from fire()` and calls `_log_non_frame_failure`, whose text asserts "shot failed, but not from a missing frame". The wait loop right below (`strict.py:168-170`) guards on `isinstance(exc.__cause__, GeecsTriggerTimeoutError)` first; the fire path does not.

Concrete scenario: the detectors are triggered before the fire and their frame-wait budget starts then. `ShotControl` puts with `put_timeout=10.0` (`devices/shot_control.py:139`); a camera's shot timeout is sub-second to a few seconds. A gateway stall on the SINGLESHOT `:SP` put longer than the shortest camera timeout fails that camera's status first. `RunEngine._status_object_completed` (bluesky `run_engine.py:2378-2386`) stores it in `self._exception`, and the `_run` loop throws it at the plan's current suspension point — the `wait` yield inside `bps.mv` inside `fire()`. The new handler then logs `... not from a missing frame — another shot cannot help ...: GeecsTriggerTimeoutError: UC_Cam: no frame within 0.5 s` — a self-contradictory line pointing the triage at the wrong fault class. Behavior is unchanged from the branch (the status propagated untouched before too — this was never treated as a missed frame), so this is diagnostics only. Fix is the same one-line guard the wait loop uses, or a second message for that case.

### 2. Low — #817's move-side half has no counterpart and is not dispositioned

Master's `step_scan.py` change (`git diff 39e86924 d6f74211 -- GeecsBluesky/geecs_bluesky/plans/step_scan.py`) was the `is not None` fix on the FAILED MOVE line, pinned by `_FalsyMoveError` in master's `tests/test_pause_semantics.py:596-611`. The PR body lists `step_scan.py` under the #817/#818 bullet but discusses only the fire. On the branch, motors move via stock `bps.mv` in the stock plans, and nothing in `geecs_bluesky` handles a move's `FailedStatus`: `grep -rn "FailedStatus\|__cause__" GeecsBluesky/geecs_bluesky` hits only `plans/strict.py` and `plans/gated.py`. What does reach the scan folder is the stop document's `exit_status: reason` (`callbacks.py:548-550`), i.e. the `FailedStatus` repr — exactly the "status repr instead of PV and CA message" gap #817 closed on master for both handlers. The RE's own `log.exception("Run aborted")` (`run_engine.py:1753`) carries the chain, but it fires after the plan has unwound and the stop doc has closed the scan log (`callbacks.py:739`), so it lands in the journal, not the scan log.

Scenario: a refused motor `:SP` put (the #779-style wrong-catalog case, or a nonexistent PV) on the branch — the scan log's only record is `fail: <status repr>`; the PV is not named anywhere in the scan folder. The PR's own justification for carrying the fire line ("the branch otherwise logs nothing when a fire is refused") applies verbatim to a refused move. Not a merge-breaking gap, but the body's "each master fix checked" claim needs this dispositioned (follow-up filed, or an explicit "not carried because …" like the tolerance warning got).

### 3. Low — `TOLERANCE_SPAN_FRACTION` is orphaned and `DEFAULT_TOLERANCE`'s comment now contradicts itself

`GeecsBluesky/geecs_bluesky/devices/ca/motor.py:40-45` keeps `TOLERANCE_SPAN_FRACTION = 0.01` with a comment saying a larger DB tolerance "is logged as suspect". Its only consumer was `GeecsSession._warn_if_tolerance_implausible` (master's `session.py`), deleted by the merge; nothing on HEAD imports it (`grep -rn TOLERANCE_SPAN_FRACTION GeecsBluesky` → the definition only). The PR body correctly says the warning was not carried, but the constant that describes a behavior no code performs survived, and the repo's stated norm is to delete the old path in the same change rather than leave shims. Same file, lines 35-38: the reworded comment says `DEFAULT_TOLERANCE` is "used only when the GEECS DB records no usable tolerance", then in the next sentence that a non-positive DB tolerance "binds a plain setpoint" — the second sentence is the true one (`namespace.py:573-582` builds `CaSettable` in that case), so the constant is now only a bare-constructor default reached by tests. Delete the constant and fix the sentence, or carry the warning into `GeecsNamespace._movable` (which has the DB row in hand).

### 4. Low — Version choices: two packages take a *patch* for what the convention calls a *minor*

Root `CLAUDE.md`'s table: patch = "bug fix, no behaviour or API change"; minor = "new feature … backwards-compatible". Relative to the branch line these releases replace, GEECS-Schemas 0.23.1 gains a new schema kind (`log_entry` module, eight new public exports, a new `SCHEMA_REGISTRY` key — `geecs_schemas/__init__.py:42-51, 145-153, 171`), and GEECS-Data-Utils 0.31.2 gains two new public functions (`scan_paths.read_scan_info_file`, `scan_log_loader.first_log_timestamp`, per master's 0.27.0 block). Both are API additions on the branch line, i.e. 0.24.0 / 0.32.0 by the table. GeecsBluesky 0.85.1 (fixes only), Console 0.32.1 (CI gate only) and DataPortal 0.26.1 (master's code as-is) are consistent with patch. The convention does not speak to merge commits, so "patch on the higher line" is a defensible house rule — but it is a new rule, and the merge blocks do not say so; either state it in the changelog block or bump the two.

### 5. Low (redundancy) — The `is not None`/`str` idiom now lives twice, with the same explanatory comment twice

`strict.py:86-104` (`_log_non_frame_failure`) and `gated.py:316-323` each select the cause with `x if x is not None else exc` and render it as `type(cause).__name__: cause`, each with a comment re-deriving the `CANothing` falsiness argument. `gated.py` already imports from `strict.py` (`gated.py:99`, `BinCounter`). A `_name_failure_cause(exc) -> str` in `strict.py` returning the `"Type: text"` string would serve the log line and the gated message, and the falsiness reasoning would have one home. Cosmetic today; it is the kind of pair that drifts (the fire-path guard in finding 1 is a sibling of this).

## Checked and refuted

- **#818 refire gating** — `fire_and_await_shot` refires only via `missed` (devices' own `missed_shot`), re-raises any non-timeout `FailedStatus`, and the fire's failure exits before the wait loop; `test_a_failed_fire_is_not_a_dropped_frame` pins `fires == 1`.
- **#820 per-axis DB tolerance** — native in `GeecsNamespace._movable` (`namespace.py:569-582`): positive DB `tolerance` → `CaMotor(tolerance=…)`; the ULP slack auto-merged into both `motor.py:141-142` and `confirm.py:198-199`; master's three on-boundary tests are all present in `tests/test_ca_devices.py` (two auto-merged, one carried).
- **Dropped test_ca_devices tests** — master's additions since the base were exactly the three on-boundary tests, all present; nothing else of master's was in the conflict block to drop.
- **The new test bites** — `len(errors) == 1` fails with no fire-path log; the `_RefusedPut: TestExp:sc2:SP: …` substring fails under `or` (selects the `FailedStatus`) and under `repr` (`_RefusedPut(ECA_BADTYPE)`); `__bool__` returning False makes the stand-in actually falsy; the `__cause__` is attached by the RE (`raise FailedStatus(ret) from exc`, `run_engine.py:2383`).
- **Tiled files** — HEAD equals the branch version; master's post-base changes (`read_primary_scalars`, the `_fetch_run` rewrite) exist on HEAD (`tiled_catalog.py:466`, `tiled_export.py:335`); a def-level `comm` of master's two modules and its test file against HEAD shows nothing missing; `tiled_schema.scan_motors` reads `motors` with `motor` fallback (`tiled_schema.py:478`) and GRID is decided from `plan_pattern` (`tiled_schema.py:546-548`).
- **Schemas `__init__` vs modules** — every imported module exists (`log_entry.py`, `preset.py`, `shot_offsets.py`); `save_set` is neither a module nor imported; ruff has no `I` rule selected, so the non-alphabetical `preset`/`log_entry` import order is not a lint failure.
- **Docs consistency** — fleet_map says "four" services and its table has four repo-managed rows (gateway, qserver, MCP, portal); no capture node, edge, or row survives; the data-flow link target `docs/sites/data_flow/index.html` exists; site_profile's named keys/placeholders all exist in `deploy/site.env.example:81-82,108-109`, `deploy/render_units.sh:47,84-85`, and `GEECS-DataPortal/deploy/geecs-data-portal.service:66-67`.
- **Portal lock** — the diff vs master is exactly the four path-dep versions (data-utils 0.31.2, logbook 0.9.0, schemas 0.23.1, web-theme 0.2.2); the memory-noted "stale path-dep versions" debt is closed by this relock.
- **Portal changelog claim** — the branch's only portal code delta since the base is `tests/test_resources.py` (#829 attribute rename), and that is the only portal code difference between HEAD and master; the merge block's description matches.
- **Changelog interleaving** — version order holds in all five; the *(master line, parallel release)* marker is applied exactly where the same number was reused (Schemas 0.21.0/0.22.0, Data-Utils 0.27.0, Bluesky 0.77.0/0.78.0).
- **Facility literals / hygiene** — the hand edits add none; test strings use `TestExp`; docs keep only `192.168.6.x`.
- **Scan-folder invariant** — nothing in the hand-edited files touches path creation.
- **`CaConfirmSettable` has no runtime consumer on the branch** — true (exported only), but the module was kept by the branch before this merge, so carrying master's fix and test into it is the right call, not a placement issue for this PR.

skbarber:
## Disposition of the review findings (fix commit 346e599f)

| # | Finding | Disposition |
|---|---|---|
| 1 | Fire-path handler labels a detector timeout thrown into the `mv` as "not a missing frame" | **Fixed** in 346e599f — the fire handler guards on `GeecsTriggerTimeoutError` like the wait loop and re-raises unlabelled |
| 2 | #817's move-side half (FAILED MOVE cause by `str`) has no counterpart on the branch | **Not carried — filed as #868.** Stock `bps.mv` in stock plans leaves no plan-side seam; the issue names the two candidate homes (`CaSettable.set` device-side log, or the stop-doc callback rendering the chain) |
| 3 | `TOLERANCE_SPAN_FRACTION` orphaned; `DEFAULT_TOLERANCE` comment contradicts itself | **Fixed** — constant deleted, comment states the one true thing |
| 4 | Schemas / Data-Utils took a patch for new public API on the branch line | **Fixed** — GEECS-Schemas 0.24.0, GEECS-Data-Utils 0.32.0, each merge block says why; portal relocked |
| 5 | `is not None` / `str` idiom duplicated between strict.py and gated.py | **Fixed** — `failure_cause_text` in strict.py is the one renderer; gated imports it |

Tests after the fix: GeecsBluesky affected files 107 passed (strict, gated, ca_devices, calibration, lazy import); Schemas + Data-Utils 837 passed, 7 deselected; portal 260 passed.

### Confirmation review by the same reviewer (fresh read of 346e599f)

## Confirmation review of #867 fix 346e599f

Read `git diff f52da4be 346e599f` (9 files) and `gh issue view 868`; re-derived each disposition from the code, not the commit message. Read-only.

**1. Fire-path label — CONFIRMED FIXED.** `plans/strict.py:162-171`: the fire handler now guards `if not isinstance(exc.__cause__, GeecsTriggerTimeoutError)` before `_log_non_frame_failure` and re-raises either way. The stalled-put scenario (detector timeout thrown into the `mv`'s wait yield) now propagates unlabelled, which is the pre-merge behavior; a refused put still gets the line. The wait loop's guard (`strict.py:180`) and the fire's are now the same predicate.

**2. Move-side #817 half — ACCEPTED AS FILED (#868).** The issue describes the gap correctly: both master handlers named, the branch's stock-`bps.mv` reason for having no seam, the stop-doc `reason` as the only scan-folder record, and the RE's `Run aborted` log landing after the scan log closes. Its one non-obvious claim — that the reason reads `<AsyncStatus ..., done>` — I checked against ophyd-async `core/_status.py:97-102`: `if e := self.exception()` is falsy for a failed `CANothing`, so a refused CA put's status does render as `done`. Accurate, and a good detail for whoever picks it up. The candidate homes (`CaSettable.set` logging at the device, or the stop-doc callback rendering the chain) are the right two seams.

**3. Orphaned constant + comment — CONFIRMED FIXED.** `TOLERANCE_SPAN_FRACTION` deleted; `git grep` at 346e599f finds no reference outside CHANGELOG. The rewritten `DEFAULT_TOLERANCE` comment (`motor.py:35-38`) now states one consistent thing — bare-constructor default only; the namespace passes the DB value or binds `CaSettable` — matching `namespace.py:573-582`.

**4. Version choices — CONFIRMED FIXED.** GEECS-Schemas 0.24.0 and GEECS-Data-Utils 0.32.0 in `pyproject.toml`, the merge blocks retitled with a "Minor, not patch" line naming the new API each carries, and `GEECS-DataPortal/poetry.lock` updated to both (0.32.0 / 0.24.0 — the only two lock lines that changed). The Data-Utils 0.32.0 bump does not require a DataPortal version bump: the portal's code did not change, only its lock.

**5. Duplicated idiom — CONFIRMED FIXED.** `failure_cause_text(exc)` at `strict.py:86-98` is the one place the `is not None` / `str` selection lives; `_log_non_frame_failure` renders through it (`strict.py:106-110`) and `gated.py:319` imports it beside `BinCounter`. `gated.py` still checks the `GeecsTriggerTimeoutError` cause first and raises it bare (`gated.py:313-315`), so the helper only sees non-timeout failures there — correct.

**New findings introduced by the fix: none.** Checked: the test's message assertion (`"_RefusedPut: TestExp:sc2:SP: GEECS refused the set" in errors[0]`, `"ECA_BADTYPE" not in`) still matches the new `%s` rendering — the log text shape is unchanged; the name `failure_cause_text` is public without an underscore, which is right for a cross-module import; the GeecsBluesky changelog block accurately records the three code changes and points at #868; `motor.py` still imports `sys` for `ULP_SLACK`, so the deletion left no unused import.

PR #869, opened 2026-09-13
Planning: the web scanner arc brief (10_web_scanner.md)

skbarber:
Docs only. Adds `Planning/native_bluesky/10_web_scanner.md` — the brief for the arc that replaces GEECS-Console with a web surface on the kit — and points `03` §11 item 8 at it.

**What it records**

- The 2026-09-13 survey: the branch-current `qs_client` verbs and the 22-name plan surface (gated vs strict is a kwarg), the `Preset` as the one submission object (`ScanRequest` orphaned), and the gaps a web front end must build — no HTTP near the queue, a pickled document stream, a blocking client, no optimization plan, no pseudo axes.
- Sam's rulings: strict is the default acquisition; free-run == gated; optimization paused, not dropped; clear-all is enough; operator naming gets one centralized mechanism if adopted; GEECS-MCP is inspiration, not canon; three surfaces run as separate processes behind one front door.
- The design: one `GeecsScanner` package (port 8300, the MCP's dependency cone), a service layer over `qs_client` + a FastAPI app/router, the HTTP API table, the SSE bridge for the document stream, the page exactly as the reviewed mock, and the kit additions it forces (GeecsWebTheme 0.3.0: `.live`, `.meter`, field validation, `dialog.ack`, `paused`, styled `denied`).
- Six PRs with acceptance each, ending with the deletion of GEECS-Console — the closing PR that lifts the branch's merge gate.
- §5: the doors kept open for pseudo positioners and optimization (plan-name-generic API, `kind` on the variable catalog, `planned_total` from the start doc).

**Hardware verification:** not applicable — planning document.

No package versions change; no facility literals added.

PR #870, opened 2026-09-13
GeecsWebTheme 0.3.0: live controls — .live with age, .meter, .field validation, dialog.ack, paused, a styled denied

skbarber:
PR 1 of the web scanner arc (`Planning/native_bluesky/10_web_scanner.md`, #869). Targets **master** per Sam's ruling: GeecsWebTheme has no Bluesky dependency, the additions are additive, and the portal and logbook can use them now. One more master → `feature/native-bluesky-plans` sync carries them to the arc branch before the page PR.

## What

The kit was built for three surfaces and shipped against two read-mostly ones. Building the scanner's mock (artifact reviewed 2026-09-13) found what a surface that **writes and watches** needs and neither existing surface ever forced. Each addition is written in the kit's own idiom — tokens only, `.kit`-scoped, the `data-state` vocabulary — so nothing here is a third dialect:

| addition | what it is | why |
|---|---|---|
| `.live` (+ `.grid.tight`) | a reading with label, value + unit, and **age**; `data-age="stale"` turns it warn and says "stale" | the pane-level `stale` doctrine at value granularity; a number that silently stops updating is worse than none |
| `.meter` | determinate progress: track, fill, two-ended label; `data-state` colours the fill at a terminal state | `.bar` is indeterminate; a scan knows exactly how far along it is |
| `.field` validation | `data-invalid="true"` on the field shows its `.err` slot and hides the hint; `.req` marks required; disabled inputs styled | focus was the only state a field had; a scan form has cross-field rules |
| `.chip.lg` | the one state a room watches, at a size it can read | |
| `.tscroll.sticky` | a capped, scrolling table that keeps its header | a live queue |
| `dialog.ack` | rung 3 widened **once**: a list of tickable preflight questions under one decision | the plain rung forbids forms; the acknowledgement is N questions, one decision |
| **`paused`** joins `STATES` | chip + dot, warn wash, no pulse | a run holding between steps is neither running nor degraded; the console rendered it as one or the other for want of a word |
| `denied` styled | dashed edge on the recessed ground, on `.state` and `.banner` | named-but-neutral until now; ownership refusal in the scanner is its first real use |

`kit.html` demonstrates every one (a **Live controls** section, a validation specimen, the sticky table, the `denied` banner, the acknowledgement dialog behind "Submit scan…"). Rendered copy to look at: the reference-page artifact published 2026-09-13.

## Scope

One concern — the kit additions the scanner needs — in eight small pieces. Per piece: `.live` 12 lines CSS · `.meter` 11 · `.field` validation 8 · `.chip.lg` 2 · `.tscroll.sticky` 2 · `dialog.ack` 10 · `paused` 2 + 1 (`STATES`) · `denied` 4; `kit.html` +116 specimens; docs (CHANGELOG, CLAUDE.md, README) +56.

**Judgment calls, cheap to veto:** `dialog.ack`'s `label:has(input:checked)` — a `:has()` selector, the first in the kit (baseline in every current browser; degrades to an unhighlighted ticked row where absent). `.grid.tight` as a modifier rather than a new container. `denied` takes an *edge* rather than a colour, on the reasoning in the CSS comment.

**No facility literals added.** The reference page's device names and the operator "tony" are example content in the same register as the U_ICT/UC_HASO already there.

## Tests

- **GeecsWebTheme: 135 passed** (127 + 8). New: `test_reference_page_demonstrates_the_live_controls`, parametrized over the eight additions — a rule without a specimen is the quiet failure the existing reverse pin (nothing on the page without a rule) does not catch. **Verified to bite:** renaming the meter's class on the page fails exactly that case (1 failed, 8 passed) and passes restored.
- Existing pins exercised by the change: status vocabulary both directions (`STATES` ↔ `.chip`/`.dot` rules, now including `paused`), pane states (`denied` is in `PANE_STATES`), `.kit` scoping over every new rule, reference page uses only styled classes, no literal colours, `[hidden]` still ahead.
- **GeecsLogbook: 240 passed, 1 skipped** against the new `STATES` (its `kit_state` mapping and literal-`data-state` guard).
- `./scripts/check.sh GeecsWebTheme` exit 0. `--all` was not run locally: this is a fresh worktree without the other packages' envs; CI is the full run.

## Hardware verification

**Not applicable** — stylesheet, a Python constant, the reference page.


skbarber:
## Adversarial review (fresh-context subagent) — and disposition

**Reviewer's verdict on e8ae5a76:** approve after fixes — the CSS is all `.kit`-scoped and token-only (135/135 tests pass; `paused` is pinned both ways by `test_status_vocabulary_is_pinned_to_the_kit`; the logbook's `KIT_STATE` ⊆ `STATES` guards tolerate the new word; the portal does not consume `STATES`), but the validation component contradicts the kit's own accessibility rule, one specimen hides information, and the docs carry a broken cross-reference and two stale claims.

### Findings → disposition (all fixed in dd244e2e, each CONFIRMED by the same reviewer)

1. **MEDIUM — `.field` validation styled on `data-invalid`, announced on `aria-invalid`.** The kit's own picklist rule is "the state is the accessibility attribute, never a class". A surface setting one and forgetting the other gets a red field a screen reader never hears. **Fixed:** rules key on `input|select|textarea[aria-invalid="true"]`; `.err` shows via `~ .err`; `data-invalid` is gone everywhere.
2. **MEDIUM — hiding `.hint` on error deletes the unit** (`psi`) the operator needs to fix the value. **Fixed:** the hint stays; the specimen's error names the unit.
3. **MEDIUM — CHANGELOG linked `Planning/native_bluesky/10_web_scanner.md`, which no master ancestor has.** **Fixed:** cites #869.
4. **LOW — the `paused` rationale said the console rendered it as running/degraded; the Qt console had its own amber pill.** **Fixed** in kit.css, CHANGELOG, CLAUDE.md: the kit lacked the word.
5. **LOW — two contradictions:** the pane-state test docstring still called `denied` neutral; `.meter` was documented as colouring "at a terminal state" yet styles `paused`. **Fixed:** docstring updated; "colours the fill by state — paused while it holds, ok/degraded/failed once it ends".
6. **LOW — `data-age="fresh"` was undefined vocabulary on the reference page.** **Fixed:** fresh readings carry no attribute; `stale` documented as the one value; the pane-state test now asserts `data-age` values ⊆ {stale}.
7. **LOW — under `border-collapse:collapse` the sticky `th` loses its rule (painted on the table, scrolls away).** **Fixed:** inset `box-shadow` on `.tscroll.sticky th`, cause commented.
8. **LOW — "Try it" promised checkbox gating `kit.js` does not ship.** **Fixed:** caption says the surface wires Submit; the dialog footnote says the page ships no wiring.
9. **LOW — a colleague's first name in a committed demo string.** **Fixed:** "another operator".

**Checked and not flagged by the reviewer:** `:has()` degrades cleanly; `position:sticky` inside `overflow:auto` works; the parametrized specimen test does fail when a specimen is removed; the reverse test covers every new class; no existing sticky/meter/age component in the portal or logbook to duplicate; the kit is the right home.

**Waived:** none. GeecsWebTheme 135 passed at dd244e2e; CI re-running on the new head.

PR #873, opened 2026-09-13
GeecsWebTheme 0.4.0: geecs_web_theme.web — the FastAPI glue every surface copied, written once

skbarber:
## What

`geecs_web_theme.web`, behind a new `web` extra (fastapi + jinja2): the one copy of what every FastAPI surface builds around the theme.

- `ForwardedPrefixMiddleware` + `clean_prefix` — the portal's `X-Forwarded-Prefix` → `root_path` middleware, unchanged in behaviour (path re-prefixed so a mount named like a route head still routes; trailing-slash redirects keep the prefix; malformed headers ignored).
- `root_of(request)`, `mount_theme(app)` (named `theme` for `url_for`), `make_templates(dir, globals=, filters=, context_processors=)` — a `Jinja2Templates` with `root` in every context.

`geecs_web_theme.testing` (stdlib only): the three template guards the logbook and the scanner each hand-wrote — `bare_url_for_calls`, `unknown_data_states`, `inline_scripts` + `javascript_syntax_error` (`node --check`), plus `node_available`.

## Why

The portal, the logbook and the scanner each carried these verbatim by intent and drifted anyway: `root` is computed in Python in the portal, in Jinja in the logbook, in a context processor in the scanner. Arc PR 4 and the logbook split would each have copied them again. Comes out of the web-foundation assessment (2026-09-13); the same assessment's other items (portal onto the kit, guard pruning on tinycss2) are separate PRs.

## Scope

One concern. **No consumer changes** — each switches to the shared copy and deletes its own in its own PR (portal and logbook on master; the scanner on the branch in arc PR 4). Judgment calls, cheap to veto:

- The static-only import stays dependency-free; fastapi/jinja2 are optional under the `web` extra. The dev group installs them so CI's `poetry install --with dev` runs the new tests.
- `make_templates` takes `globals`/`filters`/`context_processors` so the logbook's `kit_state` global and `initials` filter can move over without a second environment.
- The guard helpers return findings and never import pytest; a suite chooses skip-vs-fail for a missing `node`.
- Root `CLAUDE.md` dependency-graph line for GeecsWebTheme amended ("no intra-repo deps"; the extra named).

Version note for the sync: `feature/native-bluesky-plans` carries GeecsWebTheme 0.3.1 (the scanner added to `_SURFACES`); at the next master → branch sync that entry folds under this 0.4.0 or becomes 0.4.1.

LOC: web.py ~200, testing.py ~170, tests ~190, docs/changelog ~80, lock ~480.

Facility-literal check: none added.

## Tests

GeecsWebTheme: **158 passed** (127 before + 31 new: 22 in `test_web.py` incl. the portal's re-homed prefix cases, 9 in `test_testing.py`), `check.sh` exit 0.

## Hardware verification

Not applicable — no runtime surface changes in any deployed service; consumers switch in later PRs.


skbarber:
## Adversarial review of #873 (fresh-context reviewer, commit 39928713)

Verified before writing: `poetry check --lock` rc 0 and no `file://` home/worktree URLs in the lock; 158 tests pass; `pre-commit run` (ruff, ruff-format, pydocstyle) passes on every changed file; a scratchpad mutant of `ForwardedPrefixMiddleware` without the path re-prefix returns 404 on `/run/1` under `X-Forwarded-Prefix: /run` and drops the prefix from the trailing-slash redirect, so `test_mount_named_like_a_route_head_still_routes` and `test_trailing_slash_redirect_keeps_the_prefix` do fail on the wrong code. The middleware, `clean_prefix`, and `root_of` are byte-for-byte the portal's (`GEECS-DataPortal/geecs_portal/app.py:114-177`) and the scanner's (`GeecsScanner/geecs_scanner/web/app.py:25-60` on `feature/web-scanner`); the guard regexes are supersets of the logbook's and the scanner's (`test_page.py` on `feat/scanner-page`) — the helpers additionally accept single-quoted attributes. No behaviour drift found. Placement in the theme package is consistent with the web-surface-kit doctrine (every web surface already depends on it, nothing in the repo depends on it), and shipping `testing.py` inside the runtime package follows the `numpy.testing` precedent — considered and not flagged.

Four findings survive, all low-to-medium and all documentation/coverage rather than runtime.

### 1. Medium — `geecs_web_theme/__init__.py:17-31` still tells a new surface to copy the glue the PR forbids copying

The package docstring (unchanged by this diff) says "This package ships stylesheets and small scripts and nothing else. It has no runtime dependencies on purpose", shows the hand-rolled `app.mount("/theme", StaticFiles(directory=str(static_dir())), name="theme")` idiom, and ends "Behind a reverse proxy build those from the request's `root_path`; the portal does this with its `{{ root }}` idiom." That is exactly the copy `GeecsWebTheme/CLAUDE.md` now bans ("A new surface imports these; it does not copy them"), and it is the first thing `help(geecs_web_theme)` / an IDE hover shows. Scenario: the next surface's author reads the module docstring, hand-mounts, hand-computes `root` — the fourth drift this PR exists to end.

Fix: rewrite that docstring block to point at `geecs_web_theme.web.mount_theme` / `make_templates` / `ForwardedPrefixMiddleware` under the `web` extra, keeping the static-only idiom only as the "no-FastAPI host" alternative.

### 2. Low — the consumer inventory names three surfaces; there is a fourth copy

`CHANGELOG.md`, `GeecsWebTheme/CLAUDE.md` and the PR body all say "the portal, the logbook and the scanner switch … and delete theirs". `ScanAnalysis/scan_analysis/config_editor/app.py` is also a FastAPI surface on the theme: `create_editor_app` hand-mounts the theme at line 241 and builds a module-level `_TEMPLATES = Jinja2Templates(...)` at line 51 with no `root` processor. Its `editor` extra already pulls `fastapi`, `jinja2` and `geecs-web-theme`, so it can take `extras = ["web"]` on the path dep at no dependency cost.

### 3. Low — `inline_scripts` documents its output as "ready for `javascript_syntax_error`", but the test's own pinned output is a SyntaxError

`test_inline_scripts_skip_src_json_and_jinja_comments` asserts the blanked script is `'var a = "jinja"; "jinja"b()"jinja"'`. Run through `node --check` that is `SyntaxError: Unexpected identifier 'b'`. Blanking a block tag to the expression `"jinja"` only yields valid JS when the tag sits on its own line. Inherited from the logbook, but this module is now the advertised one copy.

Fix: blank `{% … %}` to `;` instead, and change the fixture so the pinned output is something the downstream helper accepts.

### 4. Low — `bare_url_for_calls` has an unpinned false negative and a false positive

`bare_url_for_calls("{{ url_for('a', x=f(g(1))) }}")` returns `[]` (two levels of nesting defeat the pattern), and `bare_url_for_calls("{# {{ url_for('a') }} #}")` reports a call inside a Jinja comment. Both inherited from the logbook's regex; neither is covered by `test_testing.py`.

Fix: blank Jinja comments before matching and document/pin the nesting limit.

### Nit

Root `CLAUDE.md:195` — one over-long line inside the dependency-graph code fence.

Duplicate search covered: `GEECS-DataPortal/geecs_portal/app.py`, `__main__.py`, `tests/test_app.py`, `tests/test_scan_log_mount.py`; `GeecsLogbook/geecs_logbook/router.py`, `routes/_common.py`, `templates/day.html`, `templates/month.html`, `tests/test_templates.py`; `ScanAnalysis/scan_analysis/config_editor/app.py`; `GeecsScanner/geecs_scanner/web/app.py`, `web/pages.py`, `tests/test_api.py`, `tests/test_page.py`; `GeecsWebTheme/geecs_web_theme/__init__.py`, `tests/test_no_literal_colours.py`; repo-wide grep for `Jinja2Templates(`, `context_processors`, `x-forwarded-prefix`, `url_for(`, `data-state`, `--check`, `static_dir()`; CI workflow, `scripts/check.sh`, and all four `pyproject.toml` consumers of `geecs-web-theme`.

---

## Disposition (author) — all fixed in ba32108c, awaiting the reviewer's confirmation

1. **Fixed** — `__init__.py` docstring rewritten to point at `geecs_web_theme.web`; the static-only idiom kept only as the non-FastAPI alternative.
2. **Fixed** — the config editor named as the fourth consumer in CHANGELOG and CLAUDE.md.
3. **Fixed** — `{% … %}` blanks to `;`; the fixture now asserts the blanked script parses under `node --check`; docstring names the remaining double-quoted-string limit.
4. **Fixed** — Jinja comments blanked first; nesting extended to two levels with the two-level case pinned; the three-level miss pinned as the documented limit.
Nit **fixed**.

GeecsWebTheme: 159 passed.

skbarber:
## Confirmation review of ba32108c (same reviewer, fresh pass over the commit)

Verified against the commit itself: `git show ba32108c` read in full; GeecsWebTheme suite 159 passed; `pre-commit run` passes on every touched file.

- **Finding 1** (package docstring) — **confirmed fixed.** The hand-rolled mount idiom is gone; the static-only case is kept as the explicit non-FastAPI alternative. Residue: the docstring counted three drifted surfaces where the other docs now count four.
- **Finding 2** (fourth consumer) — **confirmed fixed.** CHANGELOG and package CLAUDE.md name the analysis config editor.
- **Finding 3** (blanked fixture was a SyntaxError) — **confirmed fixed for the reported scenario**, re-run through `node --check` by the reviewer. Probe table: inline `{% if %}f(){% endif %}` OK; block tags on own lines OK; if/else/endif OK; `f({{ x }}, {{ y }})` OK; single-quoted `'{{ x }}'` OK; double-quoted `"{{ x }}"` FAIL (documented); a block tag inside an object literal / array / assignment RHS FAIL — not introduced by this commit (the old blanking failed the same inputs), but the docstring's "still parse" claim needed scoping to statement position.
- **Finding 4** (url_for guard) — **confirmed fixed as specified.** Comment-wrapped call → not reported; a real call after a comment → reported; two-level nesting → reported; three-level → the documented, pinned limit; no false positive on `.path` forms at any depth.
- **Nit** — confirmed fixed.

Summary: all four findings and the nit fixed, nothing regresses; two one-clause documentation residues, neither blocking.

### Disposition (author)

Both residues **fixed** in 8de24eab: the package docstring names four consumers; `inline_scripts` scopes its parse claim to block tags in statement position and names the expression-position exception. 159 passed. No findings remain open.

skbarber:
## Codex review finding — fixed in 31864f20

**P2, `testing.py` `unknown_data_states`** — the regex matched only values starting with a lowercase letter or underscore, so `data-state="FAILED"`, `"ok "` and `""` (each an uncoloured chip in the browser) were never reported.

**Fixed:** any quoted value is captured; a Jinja-rendered value (`{{ … }}` / `{% … %}`) is skipped as not-a-literal; everything else is judged exactly as written, case, whitespace and emptiness included. Pinned by `test_unknown_data_states_judges_malformed_values_and_skips_jinja`. GeecsWebTheme 160 passed.

PR #875, opened 2026-09-13
GeecsWebTheme 0.5.0: CSS guards on tinycss2, pruned to the checks that earn their keep

skbarber:
## What

The theme's CSS guards (`tests/test_no_literal_colours.py`) now read stylesheets through **tinycss2**, a real CSS parser, and HTML through the standard library's parser, via new helpers in `geecs_web_theme.testing`: `css_rules`, `rule_selectors`, `colour_literals`, `html_style_sources`, `js_colour_literals`, `referenced_tokens`, `defined_tokens`, `styled_classes`, `attribute_selector_values`, `classes_used`. The guard file goes from 703 to 401 lines.

**Kept** (each with its reason in the file's docstring): no literal colours in any registered surface; every referenced token defined and every palette complete; the kit introduces no token; the four vocabularies agree across Python, `theme-boot.js` and the CSS; every kit rule scoped to `.kit`; the reference page shows only what the kit styles.

**Deleted on purpose:** the `[hidden]` ordering check (a one-time bug, now a comment beside the rule), the per-component specimen list for the reference page (a maintenance list, not an invariant), and the probe cases that pinned holes in the old regexes (moot with a parser).

## Why

Six regex CSS scanners in this package and the logbook each had a silent hole found across three review rounds — a consumed `@media` anchor, a filter testing the empty string, a match starting inside a Jinja comment, a first-declaration lookup where the cascade reads the last — and every hole passed green. That churn is most of what made the web-surface-kit arc feel slow. A parser-based check that is wrong fails loudly. Owner ruling: "ask a real parser". Item 1 of the web-foundation follow-ups (2026-09-13).

## Proof the guards bite

Each kept guard was mutation-tested against the real files (mutate, run, restore): a literal in `kit.css`; a literal in a `style=` attribute of `run.html`; a literal in an inline-script string of `run.html`; an unknown status styled on `.chip`; an unscoped kit rule; an unstyled class on `kit.html`; an undefined token referenced from `scanlog.css`; a palette missing `--ok`; a token introduced by the kit; an emptied compact-density block. Ten mutations, ten single failures, all real surfaces green unmutated.

## Scope

One concern: the guards. Judgment calls, cheap to veto:

- tinycss2 lands as an **optional `testing` extra** (plus the theme's dev group), so the HTML guards in `geecs_web_theme.testing` stay standard-library only; a consumer's test suite takes the extra only if it uses a CSS helper. The scanner's "every class on the page is styled" test (arc PR 4) becomes `classes_used(template) - styled_classes(kit, theme, own_css)`.
- The allowlist is by **selector** with a stated reason (`.themepick .sw`, `img.plot`, `.ce-preview img`) rather than by line substring.
- `js_colour_literals` judges string literals only (a colour built by concatenation is not seen); documented in CLAUDE.md.
- The `.kit` scoping check is kept: the invariant is load-bearing (the portal's `.pane` tab mechanism) and is six lines over the parser.

No consumer changes. No facility literals.

## Tests

GeecsWebTheme: **162 passed** (160 before; the guard file has fewer, larger tests; three direct helper probes added), `check.sh` exit 0.

## Hardware verification

Not applicable — test infrastructure only.


skbarber:
## Adversarial review of #875 (fresh-context reviewer, commit c07f9753)

Suite baseline: 162 passed. Every mutation below was applied to a real file, run, and reverted. Old-guard comparisons ran master's `test_no_literal_colours.py` against the same mutation.

### 1. Medium — `var(--x)` inside script strings is no longer checked by the referenced-token guard (coverage regression, proven on a real surface)
The test passes script bodies and `.js`/`.py` files to `referenced_tokens(..., css=False)`, whose regex only knows `getPropertyValue(` and `$tok:`. The old guard's raw-text regex included `var\(\s*(--[\w-]+)` and so saw JS-built CSS. Scenario: `run.html` builds `note.style.cssText = "... background:var(--surface-2); ..."`. Changing only that to `var(--surfce-2)`: new guard **1 passed**; old guard **failed** with `run.html: --surfce-2`. Same gap for `kit.js`, `editor.js`, `nav.js`, `figures.py`.

### 2. Medium — `css_rules` skips an at-rule whose direct children are only at-rules; docstring says the opposite
`@media (x){@supports (y){.pane{color:#ff00ff}}}` — the `@media` body contains no qualified rule directly, so the whole subtree is dropped from every helper at once. Appended to `kit.css`: colour, scoping and token tests **3 passed**. Master's brace-walker would have flagged both.

### 3. Low-Medium — native CSS nesting is dropped silently
`parse_declaration_list` returns a nested rule as an error token, which the declaration filter discards. `.kit .probe{color:var(--ink); &:hover{color:#ff00ff}}` appended to `kit.css` → **2 passed**. tinycss2 1.5.1 ships `parse_blocks_contents`.

### 4. Low — unquoted `url()` with a quote inside becomes a `bad-url` error and is dropped; conversely any `#` in an unquoted URL is flagged
`url(data:image/svg+xml,%3Csvg fill='%23ff00ff'/%3E)` → **1 passed** (master caught it). `url(icons.svg#frag)` → reported as a colour, a false positive.

### 5. Low — the allowlist is still a substring match, not "by selector"
`img.plot` waves through `.kit img.plotwrap{background:#fff}` and, for `img.plot, .other{color:#fff}`, the whole comma list.

### 6. Low (redundancy) — two checks in the guard file bypass the helpers the file introduces
The `kit.html` `data-state` sweep reimplements `unknown_data_states` inline; the `.tone-` local-indirection check is a raw-text CSS regex in the file whose docstring says "not regular expressions".

### 7. Nits
CHANGELOG says "703 → 401 lines"; the file is 432. `_GROUND_FREE` bakes an ink-black `17,24,33` into the shared helper that no palette uses any more.

Searched for duplicates: the logbook's and portal's test suites, the config editor's tests, the theme's own modules and tests, a repo-wide grep for `tinycss2`. (`GeecsLogbook/tests/test_seed_templates.py:174` carries a pre-existing `.tone-` regex CSS scanner in a consumer — not added here.)

---

## Disposition (author) — all fixed in 67353688, awaiting the reviewer's confirmation

1. **Fixed** — `var\(\s*(--[\w-]+)` is a third alternative in the script-side reference pattern; pinned by `test_referenced_tokens_sees_css_a_script_builds`; re-proven: the `surfce-2` typo in `run.html` now fails the guard.
2. **Fixed** — `css_rules` descends when the body holds any qualified rule **or** any at-rule with content; pinned in the probe table and `test_nested_rules_keep_their_parent_scope`; re-proven on `kit.css` (2 failures: colour + scoping).
3. **Fixed** — `parse_blocks_contents`; nested rules flattened as `parent child` so the scoping check reads the parent's scope; re-proven on `kit.css`.
4. **Fixed** — a `bad-url` error token is reported ("unparseable url(); quote it"); `url()`/string values are judged on `(%23|#)` followed by 3–8 hex, so `icons.svg#frag` is clean; both pinned; re-proven on `kit.css`.
5. **Fixed** — per comma-part, exact or prefix at a class boundary, and every part must be covered; pinned by `test_allowlist_matches_whole_selector_parts`; re-proven (`img.plotwrap` fails).
6. **Fixed** — `unknown_data_states` for the kit page; new `token_indirections` helper over `css_rules` replaces the regex.
7. **Fixed** — count corrected; ground-free exemption is plain black only (the palettes' `--shadow` definitions are `:root` and exempt on that ground).

**Waived, out of scope:** the logbook's `.tone-` regex in `test_seed_templates.py` predates this PR; it moves to `token_indirections`/`styled_classes` when the logbook switches to the shared module (a master PR already on the list).

GeecsWebTheme: 172 passed, `check.sh` exit 0.

skbarber:
## Confirmation review of 67353688 (same reviewer, fresh pass)

Suite 172 passed. All seven findings **confirmed fixed**, each re-proven on the real file: the `surfce-2` typo in `run.html`'s `cssText` fails the token guard; nested `@media{@supports{.pane{#ff00ff}}}` in `kit.css` fails twice (literal + unscoped); a nested `&:hover` literal is reported as `.kit .probe &:hover: color: #ff00ff`; an unquoted data URL is reported as `bad-url`; `url(icons.svg#frag)` is clean; `img.plotwrap` under the `img.plot` allowlist fails; `@font-face` / `@page` yield nothing; nested selectors report parent-first.

Introduced by the fix, none blocking:

- **a.** Declarations directly inside a nested at-rule (`.kit .a{ @media (x){ color:#ff00ff } }`) were dropped — valid CSS nesting, same category as finding 2 one level down.
- **b.** Only the first comma-part of a nested selector inherited the parent (`&:hover, &:focus` → `['.kit .a &:hover', '&:focus']`); a parent comma list cross-multiplied wrongly.
- **c.** `_allowed_part` treated `-` as a class boundary, so `img.plot-x` passed under `img.plot`; the docstring said "class boundary" and `-` is a class-name character.
- **d.** The `var(` text alternative also ran over CSS text, so a token named in a CSS comment became a "referenced but never defined" failure — contradicting the file's "comments are structure" premise.
- **e.** (nit, unchanged from the old regex) `url(sprite.svg#add)` — a hex-looking fragment — is still reported.

---

## Disposition (author) — a–d fixed in 5f4afd06, awaiting the reviewer's confirmation

- **a. Fixed** — a nested at-rule's body is walked with `parse_blocks_contents`; bare declarations attach to the parent selector. Pinned: `.kit .a{ @media (min-width:1px){ color:#ff00ff } }` → `.kit .a: color: #ff00ff`.
- **b. Fixed** — `compose()` cross-multiplies every parent part with every own part, parent first. Pinned: `.a, .b{ &:hover, &:focus{} }` → six selectors.
- **c. Fixed** — an entry extends only into a descendant, pseudo-class, child or attribute (`" :.>["`), never into the class name; the three swatch selectors are listed explicitly with their reason. Pinned: `img.plot-x` fails, `img.plot:hover` passes.
- **d. Fixed** — the `var(` scan applies only when `css=False`. Pinned: `/* was var(--old) */ a{color:var(--ink)}` → `{--ink}`.
- **e. Waived** — same behaviour as the old guard; a hex-looking SVG fragment id is rare and the failure is loud and named. Revisit if one appears.

GeecsWebTheme: 173 passed, `check.sh` exit 0.

skbarber:
## Second confirmation pass on 5f4afd06 (same reviewer)

Suite 173 passed. Items **a–d all confirmed fixed**, each probed directly: nested at-rule declarations attach to the parent (also two at-rules deep and under a comma-list parent; a nested `@keyframes` still leaks no stop); comma lists cross-multiply parent-first and `.kit .x{ &:hover, &:focus{} }` appended to `kit.css` stays clean under the scoping check; `img.plot-x` is reported and `img.plot:hover` is not, with the three swatch selectors now explicit; a token named in a CSS comment is not a reference. Nothing introduced that changes a verdict; one cosmetic note — a parent with both a nested at-rule and nested rules appears twice in `css_rules()` output (every consumer de-duplicates or reports per finding), noted so it is not mistaken for a bug later. Item e waived as agreed. No further findings.

---

## Codex review findings — fixed in 5039d4e3

- **P2 — `html_style_sources` skipped a `style`/`fill`/`stroke` attribute wholesale if it carried any Jinja**, so `style="color:#ff00ff; width:{{ w }}px"` passed. **Fixed:** the Jinja fragments are blanked and the residue judged; pinned (mixed `style`, and a Jinja `fill` beside a literal `stroke`); re-proven on `run.html` — a mixed attribute now fails the guard.
- **P2 — `js_colour_literals` scanned only quoted strings; template literals were invisible** although the guarded pages build HTML with backticks. **Fixed:** backtick strings are scanned (a nested template literal inside `${…}` ends the outer match early — documented, accepted); pinned both ways (`` `color:#ff00ff` `` reported; a backtick HTML builder with `${s}` interpolations clean); re-proven on `run.html`.
- **P3 — `rule_selectors` split on every comma**, cutting `.kit :is(.panel, .well)` in two, and `styled_classes` saw only `kit`. **Fixed:** `selector_parts()` splits on top-level commas through the tokenizer and every internal `split(",")` uses it; `styled_classes` descends into `:is()`/`:not()`/blocks; pinned; re-proven — an unscoped `:is(.pane, .well)` rule in `kit.css` fails the scoping check, and a class styled only through `:is()` is not a false miss on the reference page.

GeecsWebTheme: 178 passed, `check.sh` exit 0. Awaiting the reviewer's confirmation of this round.

skbarber:
## Third confirmation pass on 5039d4e3 (same reviewer)

Suite 178 passed. All three Codex fixes **confirmed**: a mixed-Jinja `style=` value is reported and a wholly rendered one clean (a `{% if %}`-wrapped literal inside `style=` is reported too; `color:#{{ hex }}` is clean, correctly — the colour comes from the server); template literals are scanned, including multi-line, with a `${…}` HTML builder clean; `.kit :is(.panel, .well)` is one selector with both classes styled, nesting composes with `:is()` on both sides, and an unscoped `:is(.pane,.well)` in `kit.css` fails the scoping check.

Template-literal early termination characterized: the one true blind spot is a colour written inside a **nested** template literal within `${}`; a colour after the nest in the outer literal, and quoted colours anywhere, are still reported. No surface does that today; matches the comment on `_JS_STRING`.

Introduced (Low): `styled_classes` descended into `:not()`, so a class named only there counted as styled — the reference-page guard would accept a class nothing paints. Nit, pre-existing: `attribute_selector_values` does not descend into functions (`kit.css` writes those rules flat; the vocabulary pin is two-way, so a rewrite through `:is()` would fail loudly).

---

## Disposition (author)

- **`:not()`/`:has()` descent — fixed in 5b97e0e2**: the class walk skips those two functions; pinned (`.kit :is(.panel, .well):not(.dim)` → `{kit, panel, well}`; `.kit .a:has(.thumb)` → `{kit, a}`).
- **`attribute_selector_values` inside functions — waived**: pre-existing shape, no rule in the kit is written that way, and the two-way vocabulary pin fails loudly if one ever is.
- **Nested-template-literal blind spot — waived, documented** on `_JS_STRING`; no surface writes a colour there.

GeecsWebTheme: 178 passed, `check.sh` exit 0.

skbarber:
## Fourth confirmation pass on 5b97e0e2 (same reviewer)

Suite 178 passed; the reference-page guard passes on the real `kit.html` + `kit.css` + `theme.css`. **Confirmed fixed:** `.kit .a:not(.dim)` → `{a, kit}`; `.kit :is(.a, .b):where(.c)` → `{a, b, c, kit}`; `:has()` skipped; a class named in `:not()` but styled by its own rule elsewhere still counts; nested `:not(:is(…))` / `:is(…, :not(…))` and case-insensitive `:NOT()` behave; attribute blocks still descended. Nothing introduced.

Reviewer's closing line: across the four rounds every finding raised has been confirmed fixed against the real files, and no new defect survived probing. No further findings.

**No open findings on this PR.**

PR #877, opened 2026-09-14
GeecsLogbook 0.10.0 + GEECS-DataPortal 0.27.0: the logbook is its own service

skbarber:
## What

The logbook leaves the Data Portal's process and becomes its own service: `geecs-logbook` (console script, port **8400**) serves one experiment behind its own unit template (`GeecsLogbook/deploy/geecs-logbook.service`, `StateDirectory=geecs-logbook`) with a runbook (`GeecsLogbook/deploy/DEPLOYMENT.md`) that includes the one-time move of the entries out of `/var/lib/geecs-data-portal`.

The portal's `/log` mount is **deleted** in the same change (no parallel path): `--scan-log`, `--notes-db`, the `log` extra, the router mount, its `StateDirectory` and `tests/test_scan_log_mount.py` are gone. The run page's link to a scan's logbook card is now built from **`--logbook-url`**: an absolute URL verbatim, a path (`/log`, the front door's future route) same-origin with the portal's own proxy prefix, empty = no link.

## Why

The logbook was a router inside the process the fleet kills on purpose: the portal carries `MemoryHigh/MemoryMax` (#834) because it is the box's largest resident process and *should* be the first to go, plus a 600 s stop timeout for analysis runs. A write path holding irreplaceable text (what people wrote, in SQLite + uploads) does not belong in that process — the web-scanner brief's own argument (separate processes behind one front door). Item 2 of the web-foundation follow-ups; independent of the scanner arc, and nicer to have landed before arc PR 5's Caddy routes `/log` to a port.

## Per-concern breakdown

| Concern | Files | +/- |
|---|---|---|
| Logbook app factory + CLI + routes at root + templates on `{{ root }}` + tests converted | 16 | +470 / -381 |
| Logbook unit template + runbook + guide/README/changelog | 6 | +287 / -26 |
| Portal: mount removed, `--logbook-url` link, new link test, unit loses `StateDirectory` | 6 | +131 / -283 |
| Portal runbook/guide/changelog | 3 | +56 / -102 |
| Fleet tooling: `render_units.sh`, `bootstrap_host.sh` (logbook service on the portal's clone, own poetry env), `site.env.example`, `fleet_status.sh` (:8400 `/health` probe, role tables), CI portal install drops the `log` extra | 6 | +45 / -19 |
| Docs: fleet map row + diagram, site-profile ports/clone table, root `CLAUDE.md` repo map + dependency graph | 3 | +31 / -24 |
| Relocks (logbook: uvicorn + web-theme `web` extra at runtime; portal: geecs-logbook and its deps drop out) | 2 | +52 / -108 |

## Design decisions (judgment calls — cheap to veto)

- **Port 8400**, not 8250: keeps the hundred-step convention (8000 Tiled, 8100 MCP, 8200 portal, 8300 scanner reserved by the web-scanner brief).
- **No `MemoryHigh/Max` on the logbook unit**, deliberately (comment in the template): it is a small process and the write path; the portal keeps its ceiling.
- **`--logbook-url` is opt-in with no default.** The brief suggested a same-origin `/log` default; that would ship a link that 404s on every host until Caddy exists. No flag = no link, exactly like `--scan-log` off did.
- **No redirect from the portal's `/log` to the new service.** Old bookmarks to `:8200/log/...` 404 until Caddy's front door restores `/log`. Adding a forwarding route to the portal seemed like the parallel path the remove-old-implementations rule forbids; say so if a 307 is wanted for the interim.
- **Shared clone**: the logbook runs from `portal-checkout` with its own poetry env inside `GeecsLogbook/` (the brief's default). A pull there is a deploy of both; the runbook and fleet map say to restart both. A `logbook-checkout` is a one-line change in `bootstrap_host.sh` if the co-versioning turns out to bite.
- `create_log_router` is gone rather than kept beside `create_app`; every test now goes through the app factory.

## Facility-literal check

The only new lab literal is `http://192.168.6.14:8400` in `deploy/site.env.example` (the allowed place, as the reference profile's `--logbook-url`). The runbook uses `<service user>` / `<host>` placeholders.

## Tests

Run from inside each package the way CI does:

| Suite | Result |
|---|---|
| `GeecsLogbook/tests` | 240 passed, 1 skipped (pre-existing: a template with no inline script) |
| `GEECS-DataPortal/tests` (with `--extras analysis`) | 255 passed |
| `GeecsWebTheme/tests` (the cross-surface template guards walk the logbook's templates) | 178 passed |

New pinning test: `GEECS-DataPortal/tests/test_logbook_link.py` (absolute URL verbatim, path form carries the proxy prefix, no flag = no link, `/log` is nobody's route on the portal, other-experiment run = no link, no default experiment = no link). The logbook's existing 240 tests were converted from `include_router(..., prefix="/log")` to `create_app(...)` and now exercise the theme/static mounts and the app's own routes.

Local smoke, not a test: `geecs-logbook --experiment Undulator --port 8417` with `STATE_DIRECTORY` set created `logbook.db` + `attachments/`, and `/health` (`writable: true`), `/static/nav.js`, `/theme/kit.css`, `/theme/kit.html`, `/month/2026-09`, `/day/2026-09-13` all answered 200, `/` 307; with `X-Forwarded-Prefix: /log` every asset and step link in the rendered page carried `/log/...`. `deploy/render_units.sh deploy/site.env.example …` renders all seven templates with no unfilled placeholder.

## Hardware / host verification — OWED

Nothing here touches scan execution or devices. The deployment itself is owed on the worker host, in this order (runbook § Moving the entries):

1. pull `portal-checkout` to the merge commit; `poetry install` in `GeecsLogbook/`, `poetry install --extras analysis` in `GEECS-DataPortal/`
2. edit `/etc/geecs/site.env`: `GEECS_PORTAL_EXTRA_ARGS=--config-editor --logbook-url http://192.168.6.14:8400`, add `GEECS_LOGBOOK_EXTRA_ARGS=`; re-render both units
3. stop the portal; `.backup` the database; enable+stop `geecs-logbook` (creates `/var/lib/geecs-logbook`); move `logbook.db` + `attachments/` there; `chown`; start the logbook
4. expected: `curl :8400/health` → `"writable": true, "version": "0.10.0"`; `/day/2026-09-12` shows the six existing entries (Sam's delete tombstones) and an attachment link serves; then start the portal on its new unit and confirm the run page's "log ↗" points at `:8400/day/...#ScanNNN`
5. `scripts/fleet_status.sh` shows a `Logbook … geecs-logbook 0.10.0 (writable)` row

Then ping the scanner session: arc PR 5's Caddy routes `/log` → `:8400`, and `--logbook-url` becomes `/log`.


skbarber:

Reviewed at `1806bbe7` (two commits over `origin/master`). Suites run from the worktree: GeecsLogbook 240 passed / 1 skipped, GEECS-DataPortal 255 passed, GeecsWebTheme 178 passed, root `tests/test_render_units_sh.py` 4 passed; `poetry check --lock` clean in both packages; neither lock carries a `/Users/` or worktree path. I also drove the logbook app under `X-Forwarded-Prefix: /log` by hand (day page, month page, static, theme, entry POST, attachment upload, `/health`) — every `href`/`src`/`data-*`/API base carried `/log`. Findings ranked; nothing here is a correctness bug in the shipped code paths.

### Medium — the logbook lost all prefix coverage, and the suite would stay green if `root` stopped being supplied

Before this PR every logbook test ran the router under `include_router(..., prefix="/log")`, so each `href="/log/…"` assertion incidentally pinned prefix carriage. Now every test hits the root and `grep -rn Forwarded GeecsLogbook/tests` returns nothing. The deployment this PR describes is `/log` behind the front door (`GeecsLogbook/deploy/DEPLOYMENT.md` § Behind a reverse proxy: "Nothing in the page is absolute"), and that claim is now unpinned in the package that makes it.

Concrete would-not-bite scenario: replace `make_templates(...)` in `GeecsLogbook/geecs_logbook/app.py:238` with a plain `Jinja2Templates(directory=…)` (no `root` context processor). Jinja renders an undefined `root` as `""`, so `{{ root }}/static/scanlog.css` becomes `/static/scanlog.css` — which is exactly what every existing test asserts at root. The suite passes; the deployed page under `/log` 404s on its stylesheet and scripts. `GeecsWebTheme/tests/test_web.py` proves the middleware and `root` generically but not the logbook's own URL builders (`routes/_common.py:228-240` — `attachment_base`, `api_base`, the month link — and the `data-api`/`data-prev`/`data-next` attributes the JS reads).

Fix: one test class sending `X-Forwarded-Prefix: /log` to `/day/…` and `/month/…` asserting the `/log/static/…`, `/log/theme/…`, `/log/api`, `/log/day/…`, `/log/month/…` forms (my probe's output is the oracle), plus the attachment `src` after an upload.

### Low — the redirects are the one thing on the page that is not prefix-safe

`GeecsLogbook/geecs_logbook/routes/day.py:50` and `:55` (`GET /`, `GET /today`) return `RedirectResponse(url=f"day/…")`; `/month/today` in `routes/month.py` returns a relative `YYYY-MM#day-…` the same way. Under `X-Forwarded-Prefix: /log` my probe got `307 Location: day/2026-09-13`. A relative Location resolves against the browser's URL, so it is right only when that URL ends in `/`: `/log/` → `/log/day/…`, but `/log` (what a person types, and the shape the fleet-map row's health check `GET /` → 307 describes) → `/day/2026-09-13` at the front door's root, a 404. Pre-existing pattern (it was relative under the portal mount too), but the runbook's "every link, form and fetch carrying the prefix" now rests on it. Fix: `RedirectResponse(url=f"{root_of(request)}/day/…")` — `geecs_web_theme.web.root_of` is already the glue this app imports from.

### Low — the one-time move in the runbook is not WAL-safe

`GeecsLogbook/geecs_logbook/store.py:246` sets `journal_mode=WAL`, so committed rows live in `logbook.db-wal` until a checkpoint. `GeecsLogbook/deploy/DEPLOYMENT.md:106` moves `logbook.db` alone. Connections are per-operation and close (`store.py:239-249`), so after a *clean* portal stop the WAL is normally checkpointed away — but the motivating scenario for this whole PR is the portal being SIGKILLed by `MemoryMax=`; a host whose last portal death was that kill and that has not restarted since holds committed entries only in the `-wal`, and the `mv` drops them (the `.backup` step on line 103 is WAL-aware, so the backup survives; the live copy does not). The recipe also starts the new unit into an empty state directory to create it and then `mv`s over the fresh `logbook.db` it made — harmless after a clean stop, but there is no reason to run the service before the move. Fix: `sudo mv /var/lib/geecs-data-portal/logbook.db* /var/lib/geecs-logbook/` (or `.backup` straight into the new directory), and create the directory with `install -d -o <service user>` instead of a start/stop.

### Low — silent tightening of the run-page link rule, untested in either direction

`GEECS-DataPortal/geecs_portal/app.py:460` changes `if summary.experiment and summary.experiment != default_experiment` to `if summary.experiment != default_experiment`. `RunSummary.experiment` is `""` when the start doc lacks the key (`GEECS-Data-Utils/geecs_data_utils/tiled_catalog.py:281`), which the scanner allows (`GeecsBluesky/geecs_bluesky/plans/run_wrapper.py:204-205` sets `md["experiment"]` only when truthy). Such a run linked before and does not now. Defensible — an unknown experiment is not ours, and the CHANGELOG's "for a run of the portal's experiment" covers it — but `tests/test_logbook_link.py` pins neither outcome for `experiment == ""`; one test would make the choice deliberate rather than incidental.

### Redundancy — no surviving findings

Checked: the web glue is imported from `geecs_web_theme.web`, not copied (`GeecsLogbook/geecs_logbook/app.py:167`); the `$STATE_DIRECTORY` fallback moved rather than forked (the portal's copy is deleted; only `GeecsLogbook/geecs_logbook/__main__.py:62-65` remains); the hand-rolled `/static/{name}` route is replaced by a named mount, not duplicated; `tests/test_scan_log_mount.py` is deleted, not left beside `test_logbook_link.py`. The `importlib.metadata` version idiom now exists in four packages (`GEECS-DataPortal/geecs_portal/app.py:84-89`, `GeecsPvaGateway/geecs_pva_gateway/__init__.py`, `GEECS-Console/geecs_console/version.py`, `GeecsLogbook/geecs_logbook/app.py:178-185`), but they share no runtime dependency that could host it — not worth a move.

### Placement — no surviving findings

The dependency direction is right (logbook → theme `web` extra; theme depends on nothing; the portal no longer imports the logbook and the root `CLAUDE.md` graph says so in both rows). The unit template lives with its service, `deploy/render_units.sh`, `deploy/bootstrap_host.sh` (shared `portal-checkout` handled by the existing `sort -u` clone loop), `scripts/fleet_status.sh` and `tests/test_render_units_sh.py` all learned the new unit. `site.env.example`'s `http://192.168.6.14:8400` is an example value on the allowed list. A repo-wide grep (including `docs/`, `Planning/`, `.claude/skills`) finds no surviving claim that the logbook is served at `:8200/log` outside CHANGELOG history and the intentional migration notes.

skbarber:
## Confirmation review of 5933d627 (same reviewer, continued)

Read from the local HEAD in the same worktree; suites at the fix commit: GeecsLogbook 244 passed / 1 skipped, GEECS-DataPortal 256 passed. Each finding was checked by mutating the code and watching the pinned test fail, then restoring (`git status` clean afterwards).

**(1) Medium — prefix coverage: CONFIRMED FIXED.** `GeecsLogbook/tests/test_nav.py::TestProxyPrefix` drives the day page, the month page, the three redirects and an uploaded attachment's `src` under `X-Forwarded-Prefix: /log`, asserting the `/log/…` forms my probe produced (`test_nav.py`, the `_HDR` class). Mutation (a): replacing `make_templates` in `GeecsLogbook/geecs_logbook/app.py` with a plain `Jinja2Templates` that keeps the globals and filters but drops the `root` processor → `2 failed, 2 passed` (the day and month asset tests fail; the redirect and attachment tests pass because they go through `url_for`/`root_of`, not the template — correct partition). The whole class is green at the fix commit.

**(2) Low — relative redirects: CONFIRMED FIXED.** `routes/day.py:57-59` and `:65-67` build `f"{root_of(request)}/day/…"`; `routes/month.py:87` uses `month_url(request, today, anchor=True)` (`routes/_common.py:238-241`, `url_for(...).path`, which carries the ASGI `root_path`). Mutation (b) — reverting both day redirects to `day/…` → `test_redirects_are_absolute_under_the_prefix` fails; mutation (c) — reverting the month redirect to the relative `YYYY-MM#day-…` → the same test fails independently. The root-level redirect tests were updated to the now-absolute `/day/…` Location, so both shapes are pinned.

**(3) Low — WAL-unsafe move: CONFIRMED FIXED.** `GeecsLogbook/deploy/DEPLOYMENT.md:104-112`: `install -d -o <service user> -g <service user>` creates the directory, `mv …/logbook.db*` carries the `-wal`/`-shm` sidecars, and the unit is first started after the move (`enable --now`, line 112, "adopts the existing directory" — correct: `StateDirectory=` reuses an existing path). No start/stop before the move.

**(4) Low — empty-experiment link: CONFIRMED FIXED.** `GEECS-DataPortal/tests/test_logbook_link.py::test_absent_for_a_run_with_no_experiment_recorded` builds a summary with `experiment=""` against `default_experiment="Undulator"` and asserts `logbook is None`. Mutation (d) — restoring the old `if summary.experiment and summary.experiment != default_experiment` guard at `geecs_portal/app.py:460` → the test fails. The tightening is now a pinned choice.

### New in the fix — two runbook nits, nothing in code

- **Low — the recipe no longer stops a logbook that may already be running.** The section header still says "with both services stopped" (`DEPLOYMENT.md:98`), but the recipe now stops only the portal. The Install section above it (`DEPLOYMENT.md:70`) tells the operator to `enable --now geecs-logbook` — so a host following the runbook top-to-bottom reaches the move with the logbook **running** on a fresh `logbook.db` in `/var/lib/geecs-logbook`, and `mv …/logbook.db*` then replaces a live SQLite database under an open service. The previous recipe's explicit stop was what covered this; one line — `sudo systemctl stop geecs-logbook 2>/dev/null || true` before the `install -d` — restores it, or the header should say the logbook must not yet be installed.
- **Nit — the glob also moves the backup.** `logbook.db*` matches `logbook.db.pre-split` (created on line 103 in the same directory), so the backup lands in `/var/lib/geecs-logbook/`; the closing paragraph (`DEPLOYMENT.md:118-119`) still implies it sits with "the now-empty `/var/lib/geecs-data-portal`". Harmless — it is a stray file in the state directory the service ignores — but the operator told to delete it will look in the wrong place. Writing the backup to a different name/location (`/var/lib/logbook.db.pre-split`, or `.backup` before the glob into a path outside it) fixes both this and the sentence.

No other new surface: `root_of` is imported from `geecs_web_theme.web` (the glue the app already takes from there), `month_url` was an existing helper reused, and no new file, dependency or unit change rides in the commit.

---

**Author's disposition so far.** Round 1's four findings: all **fixed** in 5933d627 and confirmed above. Round 2's two runbook nits: **fixed** in 3a52e43b (the recipe stops `geecs-logbook` before the move, tolerating "not installed"; the `.backup` goes to `/var/backups/logbook.db.pre-split`, outside both state directories, and the closing paragraph says so). Awaiting the same reviewer's confirmation of 3a52e43b and CI on it; the final disposition follows as a separate comment.

skbarber:
## Confirmation review of 3a52e43b (same reviewer, continued)

Docs-only diff (`GeecsLogbook/deploy/DEPLOYMENT.md`, +9/−5), read at the local HEAD.

**(1) Stop a possibly-running logbook before the move: CONFIRMED FIXED.** `DEPLOYMENT.md:103` adds `sudo systemctl stop geecs-logbook 2>/dev/null || true` ahead of the `install -d`; `systemctl stop` on a unit that is not installed exits 5 ("Unit … not loaded"), which the `|| true` absorbs, so the line is safe on both the fresh host and the one that followed the Install section first. The header (`DEPLOYMENT.md:98-99`) now says the logbook must be stopped too and names the case that makes it matter. With every `NotesStore` connection closed per operation (`store.py:239-249`), a clean stop leaves no live database under the `mv`.

**(2) Backup outside the glob's reach: CONFIRMED FIXED.** The `.backup` target is `/var/backups/logbook.db.pre-split` (`DEPLOYMENT.md:106`) — outside both state directories, so `mv /var/lib/geecs-data-portal/logbook.db*` on line 111 cannot carry it along — and the closing paragraph (`DEPLOYMENT.md:121-123`) points at `/var/backups/` as the place to delete it from. `sqlite3 .backup` remains the WAL-aware copy, so the backup is complete even for a portal that last died by `MemoryMax=`. `/var/backups` is a standard directory on the Ubuntu hosts this runbook targets (the store's own doc names Ubuntu 22.04+); it is a fixed system path, not a site value, so nothing needs to move to `site.env`.

**Nothing new to flag.** No code, unit, dependency or test changed; the recipe's step order (stop both → backup → create dir → move `logbook.db*` + `attachments/` → chown → first start → verify → restart portal) is internally consistent, and the earlier full-suite and mutation results (244 / 256 passed; four mutations each caught by their pinned test) stand unchanged since no code moved.

---

## Final disposition

|---|---|---|---|
| 1 | Logbook suite lost all proxy-prefix coverage | Medium | **Fixed** 5933d627 (`TestProxyPrefix`, mutation-proved) — confirmed |
| 2 | `/`, `/today`, `/month/today` redirects relative, wrong under `/log` without a trailing slash | Low | **Fixed** 5933d627 (`root_of` / `month_url`) — confirmed |
| 3 | Runbook move not WAL-safe; unit started before the move | Low | **Fixed** 5933d627 (`logbook.db*`, `install -d`) — confirmed |
| 4 | Empty-experiment run silently lost its link, untested | Low | **Fixed** 5933d627 (pinned as a choice) — confirmed |
| 5 | Recipe no longer stopped an already-running logbook | Low (introduced by #3's fix) | **Fixed** 3a52e43b — confirmed |
| 6 | `logbook.db*` glob swept the backup along | Nit (introduced by #3's fix) | **Fixed** 3a52e43b — confirmed |

Nothing waived. Redundancy and placement: no surviving findings in any round. Three review rounds, each fix commit confirmed by the same reviewer from the code; suites at head: GeecsLogbook 244 passed / 1 skipped, GEECS-DataPortal 256 passed, GeecsWebTheme 178 passed, root 81 passed.

skbarber:
## Disposition of the Codex finding — **fixed** in 50236edd

Confirmed on a live process before changing anything: with `--root-path /log`, `GET /log/static/nav.js` → 200 and `GET /static/nav.js` → 404, exactly as reported. `--root-path` is a prefix-preserving-proxy setting, not a fallback for a stripping proxy that omits the header — and the same wording ("a proxy-sent X-Forwarded-Prefix header overrides this per request") describes the same behaviour on the portal and the scanner, so the semantics are shared and pre-existing.

Chosen fix: **constrain the docs**, not the code. Making the static `root_path` also accept stripped paths would be a change to the theme's shared middleware for every surface, out of this PR's scope. Four places now say the same thing:

- `GeecsLogbook/deploy/DEPLOYMENT.md` § Behind a reverse proxy — rewritten as two proxy shapes: stripping (Caddy `handle_path`, nginx `proxy_pass …/`) ⇒ send `X-Forwarded-Prefix`, no flag; preserving ⇒ `--root-path /log`, and the app then answers the prefixed upstream paths **only** — with the exact failure mode Codex reproduced spelled out.
- `GeecsLogbook/deploy/geecs-logbook.service` comment on `GEECS_LOGBOOK_EXTRA_ARGS` — `--root-path /log` "behind a proxy that forwards the FULL path (prefix-preserving) and sends no X-Forwarded-Prefix; a prefix-stripping proxy must send that header instead".
- `geecs_logbook/__main__.py` `--root-path` help and `geecs_logbook/app.py` `create_app` docstring — the same constraint.

Docs-only diff; no test change (the behaviour itself is what `TestProxyPrefix` pins for the header path). Arc PR 5's Caddy route will be the stripping shape with the header, so the fleet never sets `--root-path`.

skbarber:
## Confirmation review of 50236edd (same reviewer, continued)

Docs/wording-only diff (four files, +30/−12); verified against the code path empirically (Starlette 1.6.0, FastAPI 0.141.1 in the logbook env).

**(a) The live behaviour claim: CONFIRMED, with one refinement the wording slightly overstates.** `create_app("Undulator", root_path="/log")`, no header, upstream port:

| path | status |
|---|---|
| `/static/nav.js` | 404 |
| `/log/static/nav.js` | 200 |
| `/theme/theme.css` | 404 |
| `/log/theme/theme.css` | 200 |
| `/day/2026-09-11` | **200** |
| `/log/day/2026-09-11` | 200 |
| `/health` / `/log/health` | 200 / 200 |

Mechanism: Starlette's router strips `root_path` from the front of `path` only where it matches; a plain `Route` therefore answers both the prefixed and the unprefixed path, but a `Mount` builds its child `root_path` (`/log/static`) and `StaticFiles` then derives the file path from a `path` that does not start with it — so the unprefixed asset URL resolves to a file that does not exist. The page rendered under the flag links `/log/static/…`; behind a **stripping** proxy without the header those become `/static/…` upstream → 404. Result: HTML served, every stylesheet and script lost — exactly the outside reviewer's finding. Also confirmed: with the header present, the header wins and the app then expects the *stripped* path (`HDR /static/nav.js` → 200, `HDR /log/static/nav.js` → 404), matching `web.py:95` and the new paragraph's "when the header is present it wins over the flag".

**(b) The four passages: CONFIRMED consistent with each other and with the header-wins rule** — `deploy/DEPLOYMENT.md` § Behind a reverse proxy, `deploy/geecs-logbook.service:38-41`, `__main__.py` `--root-path` help, `app.py` `create_app` docstring all say: flag = prefix-preserving proxy, stripping proxy = header, header overrides. One nit: DEPLOYMENT.md's "answers `/log/static/…`, `/log/day/…` — and **only** those" (and the help/docstring's "prefixed upstream paths only") is stronger than the truth — route paths (`/day/…`, `/health`) still answer unprefixed; only the **mounted assets** are prefixed-only. The paragraph's own next sentence ("serves the HTML and loses every stylesheet and script") depends on that distinction, so the two sentences mildly contradict. An operator who curls `/day/2026-09-11` on :8400 to test the flag gets 200 and doubts the doc. Suggested wording: "the mounted assets (`/static/…`, `/theme/…`) answer at the prefixed path **only**; the page routes answer either, which is why the failure is a styled-less page rather than a 404". Not blocking.

**(c) Docs-only is the sound choice for this PR — with one owed follow-up.** The alternative (having the flag synthesize a prefix on unprefixed paths so it covers both proxy shapes) is a change to the shared middleware in `geecs_web_theme.web`, which would alter the portal too and belongs to its own PR, not this split. The fleet's front door (Caddy, per the web-scanner brief) sends the header, so the corrected wording is the deployed path. However the same latent misdirection stands in the portal today: `GEECS-DataPortal/DEPLOYMENT.md:149-153` calls `--root-path` the mode "for a proxy that cannot send it", and I reproduced the identical behaviour there under a uvicorn-style `root_path` (`/static/plotly-cartesian-3.1.1.min.js` 404, `/portal/static/…` 200, `/theme/theme.css` 404, `/portal/theme/theme.css` 200, `/health` both 200). Out of scope here and rightly so, but it should be recorded as owed. The natural one-copy home for the preserving/stripping rule is the `web.py` module/middleware docstring (`GeecsWebTheme/geecs_web_theme/web.py:89-96`), which already states "header wins" but not the distinction — putting it there means the portal and scanner inherit the correct wording when they adopt the kit (portal-onto-kit is already the planned last step of the arc).

**Nothing new to flag** beyond the "only those" nit; no code, unit directive, dependency or test changed, and the earlier suite and mutation results stand.

---

**Disposition.** (a) confirmed; (b)'s wording nit **fixed** in 76bef005 (the three passages now say the mounted assets are prefixed-only and the page routes answer either way, which is why the failure is a styleless page rather than a 404); (c) accepted — the portal's `DEPLOYMENT.md` `--root-path` paragraph carries the same misdirection and the one-copy home is the `geecs_web_theme.web` middleware docstring: **recorded as an owed follow-up outside this PR**, not fixed here (the portal's wording is out of this PR's scope). Awaiting confirmation of 76bef005 and CI on it.

skbarber:
## Confirmation review of 76bef005 (same reviewer, continued)

Docs/wording-only (three files, +14/−10); no code, unit directive, dependency or test changed, and both edited Python modules still parse (the edits are inside string literals and a docstring).

**Wording vs the measured table: CONFIRMED.** Each of the three passages now states exactly what I measured under `root_path="/log"` with no header:

- `GeecsLogbook/deploy/DEPLOYMENT.md` § Behind a reverse proxy — "the **mounted assets** answer at the prefixed path **only** — `/static/nav.js` on the upstream port is a 404 with the flag set. The page routes (`/day/…`, `/health`) answer either way" ↔ table: `/static/nav.js` 404 / `/log/static/nav.js` 200; `/day/2026-09-11` 200 / `/log/day/2026-09-11` 200; `/health` 200 / `/log/health` 200. The consequence sentence ("serves the HTML and loses every stylesheet and script, rather than failing loudly") now follows from the distinction instead of contradicting it.
- `__main__.py` `--root-path` help — "the /static and /theme mounts answer at the prefixed path only" — the two mounts are the two that 404'd unprefixed (`/theme/theme.css` 404 / `/log/theme/theme.css` 200).
- `app.py` `create_app` docstring — the same, with "(plain routes answer either way)".

A grep for the old over-statements ("only those", "paths only", "prefixed upstream") across the four files finds none; the unit comment (`geecs-logbook.service:38-41`) never carried the "only" claim and is unchanged, and it agrees with the three (preserving proxy → flag; stripping proxy → header).

**Nothing new to flag.** The header-wins statement in the DEPLOYMENT.md paragraph is untouched and still matches the probe (`HDR /static/nav.js` 200, `HDR /log/static/nav.js` 404). The earlier suite results (GeecsLogbook 244 passed / 1 skipped, GEECS-DataPortal 256 passed) and the four mutation checks stand, since nothing outside prose moved after 5933d627.

---

## Final disposition (supersedes the earlier table)

| # | Finding | Source | Severity | Disposition |
|---|---|---|---|---|
| 1 | Logbook suite lost all proxy-prefix coverage | round 1 | Medium | **Fixed** 5933d627 (`TestProxyPrefix`, mutation-proved) — confirmed |
| 2 | `/`, `/today`, `/month/today` redirects relative | round 1 | Low | **Fixed** 5933d627 — confirmed |
| 3 | Runbook move not WAL-safe; unit started before the move | round 1 | Low | **Fixed** 5933d627 — confirmed |
| 4 | Empty-experiment run silently lost its link, untested | round 1 | Low | **Fixed** 5933d627 — confirmed |
| 5 | Recipe no longer stopped an already-running logbook | round 2 | Low | **Fixed** 3a52e43b — confirmed |
| 6 | `logbook.db*` glob swept the backup along | round 2 | Nit | **Fixed** 3a52e43b — confirmed |
| 7 | `--root-path` documented as a fallback for a stripping proxy without the header | Codex | Medium (docs) | **Fixed** 50236edd (docs constrained to prefix-preserving proxies; behaviour verified live) — confirmed |
| 8 | "only those" overstated: routes answer unprefixed, only mounts don't | round 4 | Nit | **Fixed** 76bef005 — confirmed |
| — | The portal's `DEPLOYMENT.md` carries the same `--root-path` misdirection; one-copy home = `geecs_web_theme.web` middleware docstring | round 4 | Owed | **Not in this PR** (portal wording is out of scope); recorded as an owed follow-up |

Nothing waived. Head 76bef005; suites unchanged since 5933d627 (GeecsLogbook 244 / 1 skipped, GEECS-DataPortal 256, GeecsWebTheme 178, root 81). Ready for the maintainer once CI on 76bef005 is green.

PR #885, opened 2026-09-14
GeecsLogbook 0.10.2: deploy docs corrected from the live logbook-split deploy

skbarber:
## What

Docs-only. The logbook split (#877) was deployed on the reference host on 2026-09-13 by hand; three things the runbook and the fleet map said turned out wrong or misleading in practice, and this records what actually works:

| Concern | Files | +/- |
|---|---|---|
| **No `sqlite3` CLI on the host.** The runbook's `sudo sqlite3 … ".backup …"` failed as one quiet line and the move ran on without a backup. Both backup steps (the nightly note and the one-time move) now use Python's `sqlite3` module, which every host has, plus a copy of `attachments/` and an `ls` that proves both exist before the move. | `GeecsLogbook/deploy/DEPLOYMENT.md` | |
| **`poetry` is not on the login PATH either.** `bash -lc poetry` → command not found on the reference host, so the fleet map's gotcha 4 ("remote commands use `bash -lc`") was wrong; the runbook's bare `poetry install` now reads `GEECS_POETRY` from `site.env`, and gotcha 4 says so. | `GeecsLogbook/deploy/DEPLOYMENT.md`, `docs/platform/fleet_map.md` | |
| **Verification on a tombstone-only day.** The runbook said "the day page shows the entries"; on the host every row of the only day was a delete tombstone, so the page was rightly empty and looked like a failed move. The verification step now counts rows through the change feed (tombstones included) and says why the page can be empty. Records the deploy (6 rows, all tombstones; logbook 0.10.1 / portal 0.27.1). | `GeecsLogbook/deploy/DEPLOYMENT.md`, fleet map snapshot note | |
| New fleet-map bootstrap gotcha 8 (the missing `sqlite3` CLI), patch bump + changelog per the docs-only convention | `docs/platform/fleet_map.md`, `GeecsLogbook/pyproject.toml`, `CHANGELOG.md` | |

Whole diff: 4 files, +61 / −13.

## Facility-literal check

None added. `/var/backups`, `/var/lib/geecs-logbook`, `/etc/geecs/site.env` are fixed system paths; the host is named only as "the reference host".

## Tests

No code changed; the logbook suite is unaffected. `pre-commit` clean on the four files.

## Hardware / host verification

This PR *is* the record of the host verification: the deploy was done and checked live (both units active, `/health` writable on :8400, portal 0.27.1 on :8200 with `--logbook-url`, 6 rows moved and confirmed through the change feed, backups taken with the Python form, old state directory removed).


skbarber:

Reviewed `origin/master...HEAD` (one commit, 480cd7e2) against the logbook's routes/store/tests, `deploy/site_env_lib.sh`, `site.env.example`, the sibling runbooks and `site_profile.md`. Docs-only, so "correctness" = does the runbook do what it says when executed as written. Findings ranked; each survived my attempt to refute it.

### 1. High — the "count that proves the move" always prints `2`

`GeecsLogbook/deploy/DEPLOYMENT.md:138` pipes `GET /api/entries?since=…` into `len(json.load(sys.stdin))`. That route returns a `ChangeFeed` **object**, not a list — `{"entries": [...], "next_cursor": null}` (`GeecsLogbook/geecs_logbook/routes/entries.py:86-96`, `:124-166`), and the package's own tests pin that shape (`GeecsLogbook/tests/test_nav.py:234-236`). `len()` of that dict is the number of keys. Reproduced with the runbook's exact one-liner against a 6-entry response:

```
$ printf '{"entries":[…6 rows…],"next_cursor":null}' | python3 -c 'import json,sys; print(len(json.load(sys.stdin)), "entries incl. tombstones")'
2 entries incl. tombstones
```

Concrete failure: the `mv logbook.db*` glob misses (typo, wrong source dir) and the new unit starts on an empty state directory → the "proof" still says `2 entries` → operator starts the portal and `rmdir`s the old directory per L147-149. The recorded "6 rows moved" (L143-145) cannot have come from this command as written, so the runbook records a check that was not the check run.

Two more gaps in the same step: the default page is `limit=500` (`entries.py:133`), so "every row" is only true when `next_cursor` is `null`; and an after-count alone proves nothing — the number that proves the move is *source count == destination count*. Suggested shape: have the backup step print `SELECT COUNT(*) FROM entries` from the connection it already opens (L123), then verify with `d=json.load(sys.stdin); print(len(d["entries"]), d["next_cursor"])` and compare.

### 2. Medium — the nightly backup, run as root against the live database, can break the service's next save

`DEPLOYMENT.md:97` runs `sudo python3 … sqlite3.connect('/var/lib/geecs-logbook/logbook.db')` while `geecs-logbook` is up. The store opens a connection **per operation and closes it** (`GeecsLogbook/geecs_logbook/store.py:238-250`), and SQLite deletes `-wal`/`-shm` when the last connection closes — so between requests those files do not exist, and root's backup connection creates `logbook.db-shm` as `root:root 0644`. A save arriving in that window: the service user cannot open the shm read-write → `sqlite3.OperationalError` on the write route (the "irreplaceable text" path this split exists to protect). The window is short per run, but this is a job scheduled forever. The one-time migration block (L117-140) is immune — both services are stopped and `chown -R` follows — so the fix is only for the nightly form: run it as the service user (`sudo -u <service user> python3 …`) into a directory that user owns (`/var/backups` is root-writable only; one `sudo install -d -o <service user> /var/backups/geecs-logbook`). `?mode=ro` does not avoid this — WAL readers still need the shm.

### 3. Low/Medium — the new gotcha 4 rule is contradicted by the same package and every sibling runbook

`docs/platform/fleet_map.md:241-247` now says *any* hand-run or remote command should read `GEECS_POETRY` from `site.env`. Unchanged in this PR: `GeecsLogbook/deploy/geecs-logbook.service:10` (the Upgrade header: `(cd GeecsLogbook && poetry install)` — the path that runs most often), `GEECS-DataPortal/DEPLOYMENT.md:182` (same clone; the runbook itself says a pull there is a deploy of both), `GeecsBluesky/qserver/deploy/DEPLOYMENT.md:63`, `GeecsCAGateway/DEPLOYMENT.md:14`, and `docs/platform/site_profile.md:142` still summarises the gotcha as "login-shell PATH". Either scope the rule ("on the reference host's service account") in one place, or the sibling snippets have to follow; a reader of the unit header gets the pre-diff advice.

### 4. Low — hand-rolled `sed` parser duplicates `deploy/site_env_lib.sh:load_site_env`

`DEPLOYMENT.md:62`: `"$(sed -n 's/^GEECS_POETRY=//p' /etc/geecs/site.env)" install`. The repo already has the `site.env` reader (`deploy/site_env_lib.sh:7-19`) that strips whitespace and quotes; `site.env.example:14-16` permits quoted values, and a quoted `GEECS_POETRY="/home/…/poetry"` makes the sed form `"/home/…"` with literal quotes → command not found; a missing key yields `"" install` → `bash: : command not found`. `(. deploy/site_env_lib.sh; load_site_env /etc/geecs/site.env; "$GEECS_POETRY" install)` reuses the parser and runs from the checkout root — where the very next snippet (`deploy/render_units.sh …`, L73) already assumes you are.

### 5. Low — the "not even in a login shell" observation names no account

Ubuntu's stock `~/.profile` adds `$HOME/.local/bin` to `PATH` when the directory exists — exactly where `GEECS_POETRY` points (`deploy/site.env.example:51`), and gotcha 1 (`fleet_map.md:231-233`) still says "`~/.local/bin/poetry` per the runbooks". `bash -lc poetry` fails in the *service account's* login shell only if its `.profile` was replaced or a `.bash_profile` shadows it; it fails trivially from any *other* account, since poetry lives under `geecs`'s home. If the 2026-09-13 observation was made from the operator's login (or via `sudo -u` without the target `HOME`), the pre-diff text was correct and the new gotcha is a misdiagnosis. Name the account in the gotcha; this is a question, not a confirmed bug.

### 6. Low / placement — gotcha 8 is not a bootstrap gotcha, and the fact is now stated four times

`fleet_map.md:227-230` introduces the list as "the gotchas [`bootstrap_host.sh`] bakes in"; the bootstrap does no backups, so gotcha 8 (`:259-263`) is baked into nothing. The same fact appears twice in the runbook (`DEPLOYMENT.md:91-98` and `:121-123`) and in the CHANGELOG. One statement in the runbook plus a pointer would do; if it stays in the fleet map, phrase it as a host fact ("the reference host has no `sqlite3` CLI") rather than a bootstrap gotcha.

### Checked and refuted

- `attachments/` absent on a host with zero uploads would block the "BOTH exist" gate — no: `AttachmentStore.__init__` creates it eagerly (`geecs_logbook/attachments.py:47`).
- `site.env` unreadable by the service account — no: installed `0644` (`deploy/render_units.sh:101`).
- `Connection.backup` missing on the system `python3` — no: present since 3.7 (Ubuntu 22.04 ships 3.10); verified locally.
- `since=2000-01-01T00:00:00%2B00:00` — decodes to an aware datetime; the route requires tz (`entries.py:150-154`). Fine.
- Public-repo hygiene — no account names, hostnames or home paths added; `192.168.6.14` is allowed.
- No shell scripts in the repo rely on `bash -lc` (grep over `*.sh`/`*.py`), so gotcha 4 breaks no tooling — only the sibling docs in finding 3.

---

**Author's disposition (interim).** All six **fixed** in 5da0fb14. (1) The backup step prints the source `count(*)`; the check reads `d["entries"]` and `next_cursor` and the comment says the counts must match — the deploy's real check (run by hand with a shape-tolerant one-liner) did give 6, which the record sentence now states as "6 in the source, 6 on the new port". (2) The nightly form runs as the service user into `/var/backups/geecs-logbook`; the one-time move keeps root, with the reason in a comment. (3)+(5) Gotcha 4 restored to the non-login-shell fact and scoped: the observation was made *as the service account* (`ssh <host> 'bash -lc "command -v poetry"'` → not found, and the deploy itself ran as that account and hit it), so it is named; interactive shells where `poetry` resolves keep the bare form, and the sibling runbooks' snippets are deliberately untouched; the logbook unit header gained a pointer to the fallback; `site_profile.md`'s summary line updated. (4) The install step reads `GEECS_POETRY` through `deploy/site_env_lib.sh` from the checkout root. (6) Gotcha 8 withdrawn; the fact lives once in the runbook's backup section plus a comment at the move step. Same reviewer asked to confirm; CI running.

skbarber:
## Confirmation review of 5da0fb14 (same reviewer, continued)

Fix commit read in full (`git show 5da0fb14`) and each point re-derived from the resulting files, not the commit message.

| # | Severity | Status | Evidence |
|---|---|---|---|
| 1 | High | **Confirmed fixed** | `DEPLOYMENT.md:134` prints `select count(*) from entries` from the still-open source connection; `:151` reads `d["entries"]` and prints `d["next_cursor"]`; the comment states the match-and-null condition and the 500-row page. The one way the counts could legitimately disagree — a NULL `updated_at` excluded by `updated_at > since` — cannot happen: `updated_at` is `NOT NULL` (`store.py:17`) and the migration fill is `COALESCE(edited_at, created_at)` with `created_at NOT NULL`. |
| 2 | Medium | **Confirmed fixed** | `:108-109` creates `/var/backups/geecs-logbook` owned by the service user, then `sudo -u <svc> python3 …` into it; the `-shm` rationale matches `store.py:238-250`. The one-time move keeps root with the justification comment at `:133` — correct, `chown -R` is at `:144`. |
| 3 | Low/Med | **Confirmed** | Unit header pointer added; the runbook names the bare `poetry install` in the unit header and sibling runbooks as valid where poetry resolves, so the docs no longer contradict each other. `site_profile.md:142` updated; `scripts/fleet_status.sh:304` still cites gotcha 4 for the non-login-PATH fact, which the restored wording keeps. |
| 4 | Low | **Confirmed fixed** | The snippet sources `deploy/site_env_lib.sh` from the checkout root; sourcing has no side effects beyond defining functions, and the subshell keeps `GEECS_*` out of the caller's environment. |
| 5 | Low | **Confirmed** | `fleet_map.md:242-247` names the account and the exact probe; the runbook says "observed as that account". The question is answered. |
| 6 | Low | **Confirmed fixed** | Gotcha 8 removed (list ends at 7); no `gotcha 8` reference survives anywhere. The fact lives in the runbook's backup section plus the move step's comment. |

### New in 5da0fb14

**Low — the install snippet is bash-only and the runbook never says which shell the service account has.** `load_site_env` uses `printf -v` and `require_site_keys` uses `${!k}`, both bash-isms, and gotcha 4 shows the case being driven remotely, where the command runs under that account's login shell. Nothing in the repo pins that shell; `useradd`'s default is dash. Under dash: `printf: -v: invalid option`, `GEECS_POETRY` empty, `: command not found`. It cannot silently do the wrong thing, only fail confusingly.

**Nit — `DEPLOYMENT.md:112-113`** opens a paragraph with a lone `The` left over from the first commit's edit. Cosmetic only.

### Checked and refuted this round

- Malformed `site.env`: `load_site_env` returns mid-file and the subshell proceeds with a partial env, but an unset `GEECS_POETRY` produces the same visible `command not found`, never a wrong install.
- Second nightly run on the same day: `s.backup(d)` replaces the destination's pages atomically.
- Quoting in the two new `python3 -c` lines: both execute as written.

---

## Final disposition

|---|---|---|---|
| 1 | Verification one-liner `len()`s a `ChangeFeed` object → always prints 2 | High | **Fixed** 5da0fb14 — confirmed |
| 2 | Nightly root backup can leave a root-owned `-shm` beside the live database | Medium | **Fixed** 5da0fb14 — confirmed |
| 3 | New gotcha 4 rule contradicted by sibling runbooks | Low/Med | **Fixed** 5da0fb14 (scoped; unit header pointer) — confirmed |
| 4 | Hand-rolled `sed` duplicates `site_env_lib.sh:load_site_env` | Low | **Fixed** 5da0fb14 — confirmed |
| 5 | "Not even in a login shell" named no account | Low | **Fixed** 5da0fb14 (account + exact probe named) — confirmed |
| 6 | Gotcha 8 is not a bootstrap gotcha; fact stated four times | Low | **Fixed** 5da0fb14 (withdrawn; stated once + a comment) — confirmed |
| 7 | Install snippet is bash-only, shell unpinned | Low | **Fixed** 5e77fd45 (says bash; gives the `ssh <host> 'bash -c "…"'` form) |
| 8 | Stray lone `The` line | Nit | **Fixed** 5e77fd45 |

Nothing waived. Head 5e77fd45; docs-only, so the earlier suite results stand.

PR #890, opened 2026-09-14
GeecsLogbook 0.11.0: the ops book folds its composer, and notes can cite each other

skbarber:
Two comments on the live logbook, and what they turned into.

> The ops book didn't adopt some things that I like [about] the scan log, which is the close button on an open note. And, is there any way to crosslink to other notes in either the ops log or scan log? could be a nice feature

Bundled because both are the same surface (the composer and the entry
tools) and the second reuses the first's markup. Per-concern breakdown
below so either half is cheap to veto.

---

## 1. The ops book's composer folds behind an affordance, and closes

**~70 LOC** (templates + the editor's opener, minus the month page's
deleted inline handler).

The ops book had no Close button because it had nothing to close: its
single composer was wedged open at the top of the month. The scan log has
folded its composers since they became per-scan — the asymmetry was an
accident of history, same shape as the one that gave the scan log
collapsible scans but not collapsible notes.

Both books now meet **one contract**, which `editor.js` implements once:

| Attribute | On |
|---|---|
| `data-open-composer="X"` | anything that opens composer X |
| `data-compose-host="X"` | the hidden element holding it |
| `data-insert="X"` | an affordance row that folds away while X is open — optional |

Closing stays non-destructive (the form keeps its text in the DOM), which
is what makes Esc safe to wire to it, and it still refuses while the entry
is already in the store — attaching a file autosaves one, so folding the
composer away would take its Discard button with it and silently publish.

**Judgment call, flag it if you disagree:** the ops book's opener sits in
its "New note" heading rather than taking a rule row. On the day page the
rule row *is* the position — the note lands between those two scans. The
ops book's composer has no position; it takes a date, which is what makes
one composer enough for a whole month, so a rule row there would imply a
place the note is not going. Collapsed, the panel is now one header line.

A day heading's "+ note" now opens that composer as well as dating it.
Revealing, dating, focusing and scrolling all happen in `editor.js` so
they cannot come apart — they were two listeners on the same button, and
the scroll ran while the composer was still hidden.

**Not done, deliberately:** the ops book does not grow a composer per day
group. One dated composer is the documented design and a composer per day
would multiply what this change just folded away.

## 2. Notes can cite each other

**~200 LOC** across the route, the renderer, the editor, `nav.js` and CSS.

Three separate things have to hold for a citation to work:

- **One name.** `GET /entry/{id}` redirects to whichever page draws the
  entry — day page for the scans book, month page for the ops book —
  anchored at `#entry-<id>`. A reader citing a note should not have to
  know which book it is in, and the name survives a change of page shape.
  A tombstone is a 404 like any other missing entry.
- **A relative stored form.** A body holds `entry/<id>` and nothing else;
  `render_markdown(entry_base=…)` swaps in the serving route at draw time
  and marks the link `entryref`. Same rule as an attachment, for the same
  reason: the mount prefix and the host are deployment facts and must
  never reach a stored body. The rewrite is keyed on the **id's
  alphabet**, not on the word `entry`, so an ordinary relative link under
  that prefix is left alone.
- **Arrival.** Everything on these pages folds — an entry is a
  `<details>`, so is the scan block around it, and Collapse All is a
  stored per-viewer preference — so a permalink routinely points *into*
  something shut, where the browser scrolls to nothing and the reader sees
  the top of a day. `nav.js` opens the target's ancestors, centres it, and
  `:target` marks it. It also fixes the existing `#ScanNNN` and
  `#day-…` anchors, which had the same problem.

The workflow is: **Link** on any entry copies its permalink; paste that
into any composer and it becomes `[author · 12 Sep](entry/<id>)`.

Two things worth knowing:

- **It does not resolve in the mirrored markdown.** An attachment link
  does, because the file is beside it; an entry reference names a row and
  only the service can turn that into a page. Off-site it reads as a dead
  relative link rather than a wrong one. That is the price of not baking a
  URL into the record, and it is stated in `CLAUDE.md` rather than
  discovered later.
- **The pasted URL's host is not checked.** The same logbook is reached as
  a bare IP, as a name, and through the front door's prefix, and a link
  copied on one is pasted on another. What validates the reference is the
  fetch that follows — an id that is not here comes back 404 and the text
  is pasted unchanged.

`navigator.clipboard` does not exist outside a secure context, and the
logbook is served over plain `http://` at an IP — so the Link tool falls
back to a selection copy rather than failing silently. It is an `<a>`
carrying the real URL, so right-click "copy link address" and cmd-click
keep working whatever the script does.

**Deferred, recorded in `CLAUDE.md`:** backlinks ("notes that link here").
The reverse direction needs an index of what points where, maintained at
every save and edit. Worth it once people are citing enough to lose track;
not for the handful this starts with.

---

## Tests

`./scripts/check.sh` → **rc=0**.

- GeecsLogbook: **272 passed, 1 skipped** (was 244/1 — 28 new: 14 in
  `tests/test_crosslinks.py`, 14 in `tests/test_composer_fold.py`)
- GeecsWebTheme: **179 passed** — its literal-colour guard walks the new
  CSS, which uses only theme tokens

**Every one of the 28 was mutation-checked.** 14 separate breaks of the
code under test — drop the `entry_base` rewrite; rewrite the href but not
the class; match any relative target under `entry/`; send every book to
the day page; resolve a tombstone; render the preview without the base;
stop handing either page a permalink prefix; a composer host that is not
hidden; the ops composer with no Close (the state it was in); "+ note"
that does not carry its day; a composer nothing opens; drop the day page's
rule row; give the ops book one it has no position for — each killed by
its intended assertion, none survived.

## Browser verification

Driven headlessly against a synthetic share and a seeded store (the
package's own `scripts/seed_demo_notes.py` plus ops entries), in Chrome:

- ops composer starts folded; "+ note" on the `Sat 12 Sep` heading opens
  it with the date set to `2026-09-12`; Close folds it; reopening restores
  the half-typed text verbatim
- Link copies `…/entry/4a65b1b45cab`, label flashes "Copied"
- pasting that into the composer yields
  `[demo · 12 Sep](entry/4a65b1b45cab)`
- `/entry/<scans id>` lands on `…/day/2026-09-12#entry-…` with the note
  **and its scan block** opened — including with Collapse All remembered
  as shut, which is the case the reveal exists for
- no page errors (one 404 for `favicon.ico`, pre-existing)

## Hardware verification

**Not applicable** — no scan execution, no devices, no share writes beyond
the existing mirror path, which is untouched.

## Deploy note

The host is on `master`; this is a normal pull + restart of
`geecs-logbook` once merged. No migration: nothing about the stored shape
changed, and no existing body contains an `entry/<id>` link to reinterpret.


skbarber:
## Adversarial review — report and disposition

Fresh-context reviewer on the diff at `8f2858ec`, using the whole repo as context. Eight findings survived its own attempt to refute them. **Six fixed** in `46d8814e`, **two waived**. The reviewer has been sent the fix commit for a confirm/deny; that result follows in a second comment.

Each fix was checked the way this repo checks a test: put the bug back, watch the probe fail. Seven reintroductions, seven caught, none missed.

### Reviewer's report

> **Working tree clean, no lab literals, version bump + CHANGELOG + the three contract docs (root `CLAUDE.md`, package `CLAUDE.md`, `deploy/DEPLOYMENT.md`) all travel with the behaviour.** The scan-folder-creation invariant is untouched. `render_markdown`'s `_ENTRY_REF` regex does survive nh3 — I verified empirically that `link_rel="noopener noreferrer"` appends `rel` after `href`, and that a markdown title, a bracketed author name, multiple refs per body, `?query`, `#frag`, `./entry/`, `ENTRY/` and `&lt;img src="entry/…"&gt;` all behave as the docstring claims.
>
> **1. (Medium) A Save landing before a pasted reference resolves silently drops the citation.** `onPaste` calls `ev.preventDefault()` and hands off to `pasteEntryRef`, which fires `GET /api/entries/<id>` fire-and-forget — the textarea is unchanged until it resolves. Paste a permalink, immediately hit ⌘↩ (the obvious human motion). `save()` → `exclusive()` awaits only `form._uploading`; nothing was registered, so it reads `ta.value` with neither the reference nor the raw URL, PATCHes, and `location.reload()`s the pending write away. The citation is gone with no error. This file already solved exactly this hazard for attachments — *"Kept on the form so a Save that lands mid-upload waits for the link to be inserted... instead of racing it"*. Secondary: if the author types during the flight, `writeAt` writes at stale offsets.
>
> **2. (Medium) `entryIdIn` matches the mount prefix, defeating the case the new doc says it deliberately allows.** The doc says the host is not checked because the logbook is reached through the front door's prefix. But the match anchors on `ENTRY_BASE`, which is prefix-bearing. A permalink copied at `:8400/entry/<id>` and pasted into a composer served under `/log` does not match, so the fallback pastes the absolute URL **including the host** into the stored body — the one thing `render.py` and CLAUDE.md say must never happen. Today's deployment is single-address so it doesn't bite yet.
>
> **3. (Medium/low) `reveal()`'s `block:"center"` scrolls past the heading of any target taller than the viewport.** For a `details.scan` open with its facts table, an entry with an image grid, or a `.daygroup` with five-plus notes, the top ends up above the viewport. Applies to the pre-existing `#ScanNNN` and `#day-…` anchors too. `center` was presumably chosen because `.kit .topbar` is sticky and there is no `scroll-margin-top` anywhere; the standard fix is `scroll-margin-top` plus `block:"nearest"` — which `openComposer` already uses in this same PR.
>
> **4. (Low) The permalink can 302 to a page that does not draw the entry, silently.** Probed against the real fixture: a scan-anchored note whose scan folder isn't on the share is not rendered (`day.html` reaches `entries_block` only from `scan_block`/`between_block`, both driven by `summary.scans`) while the header still counts it. `reveal()` returns silently on a miss, so the reader gets the top of an apparently empty day.
>
> **5. (Low) The fallback copy steals focus and never returns it — the "Copied" flash is invisible to a keyboard user.** `.entry-tools` is `opacity:0` except `:hover`/`:focus-within`. Someone who tabbed to Link and pressed Enter gets the flash at opacity 0 and loses their tab position. This is the branch that actually runs in the lab.
>
> **6. (Low) The redirect is 307, not the 302 the code advertises,** and the new test asserts `in (302, 303, 307)` so it cannot catch the discrepancy.
>
> **7. (Low — placement) The entry-id alphabet now has three independent homes, all failing silently.** `store.py` generates it; `render.py` and `editor.js` each re-declare it; `LogEntry.entry_id` carries no `pattern` (unlike `day`). A `pattern` in `geecs_schemas` is the one home; at minimum a test tying `_ENTRY_REF` to a freshly generated id would bite.
>
> **8. (Low — redundancy) Two small overlaps.** Copy-on-plain-http is now the second implementation (the portal's `run.html` has `copyPlotImage`); they aren't mergeable as-is and `GeecsWebTheme/CLAUDE.md` makes the third surface the forcing function — a note, not a demand. And `day.html`'s `.scanrow` handler is now mostly subsumed by `reveal()`, though not fully (clicking the rail row for the scan you're already on fires no `hashchange`).
>
> **What I checked and could not fault:** `compose_host`/`insert_row` placement; `entry_base` beside `attachment_base` and the `if ctx.writable else ""` guard being load-bearing and correct; `reveal()` genuinely running after the Collapse All preference on both pages; the permalink `<a>` inside `<summary>` not collapsing the entry (activation target is the `<a>`); the new `<a>` excluded from hover-prefetch; no literal colours; `.entryref` covering the preview pane; the re-entrancy guard on `flash`, the `try/catch`es, and `shortDay` avoiding `Date`/`toLocaleDateString`.

### Disposition

| # | Finding | Disposition |
|---|---|---|
| 1 | Paste/save race drops the citation | **Fixed** — `46d8814e` |
| 2 | Prefix mismatch writes the host into the body | **Fixed** — `46d8814e` |
| 3 | `center` hides the heading of a tall target | **Fixed** — `46d8814e` |
| 4 | Permalink lands on a page that can't draw it | **Fixed** — `46d8814e` |
| 5 | Fallback copy steals focus | **Fixed** — `46d8814e` |
| 6 | 307 vs the docstring's 302 | **Fixed** — `46d8814e` |
| 7 | Entry-id alphabet has three homes | **Waived**, with a partial fix |
| 8 | Copy-on-http twice; `.scanrow` overlap | **Waived** (copy) / **Fixed** (comment) |

**On 1**, the fix is the existing `form._uploading` latch, extracted into a `pending(form, work)` helper both writers use. Writing it exposed a second defect the reviewer had not raised: with one shared slot, whichever writer finishes first clears it and the other becomes invisible to `save()`. Both halves of the queue turned out to be load-bearing, and they **fail in opposite paste orders** — a writer must not clear the latch itself (link slow, file fast), and clearing must check the latch is still ours (file fast, link slow). Getting that honest took three attempts: my first two probes passed under every mutation, once because they read rows left by earlier runs, once because a `sleep` in a Playwright route handler blocks the driver, so "submit at 1300 ms" actually happened at 3822 ms with both writes already landed. Both orders are now pinned by probes that fail without the corresponding guard, and both are written into the code.

**On 3**, `scroll-margin-top` is set from the measured `.topbar` height rather than a constant, because the bar wraps to two rows on a narrow window. `block:"start"` rather than `"nearest"`: `nearest` bottom-aligns an element approached from below, which is the same class of problem.

**On 7 (waived)**, a `pattern` on `LogEntry.entry_id` is the right home but it is a cross-package schema change inside a UX PR. Instead `TestTheIdAlphabetHasNotDrifted` ties both matchers — including `editor.js`'s, read off disk — to a **freshly minted** id, so a change to the generator fails loudly rather than silently breaking paste-recognition. The remaining debt is in the package `CLAUDE.md` deferred table. Cheap to veto if you want the schema change now.

**On 8 (waived)**, `GeecsWebTheme/CLAUDE.md` already rules that the third surface is the forcing function for promoting shared behaviour into `kit.js`. This is the second. Recorded in the deferred table so the web console's copy button doesn't become a third copy.

### Verification

`./scripts/check.sh` → **rc=0**. GeecsLogbook **274 passed, 1 skipped** (up 2); GeecsWebTheme **179 passed**.

Browser probes, against a synthetic share and a seeded store, in Chrome — including a second instance of the same store mounted at `/log` for finding 2, because poking `data-entry-base` after load proves nothing (`editor.js` reads it once, at init):

```
F1  OK — the reference survived a same-tick save
F1b OK — the attachment link survived a same-tick save
F1c OK — a save between the two still waited for both
F1d OK — the finished write did not clear the pending one
F2  OK — relative reference, no host, under a different prefix
F3  OK — heading clears the sticky bar  (880 px target, 420 px viewport)
F4  OK — said when it misses, silent otherwise
F5  OK — focus returned, flash visible
```

Reintroducing each bug, to prove the probes bite:

```
CAUGHT  F1  don't queue the pending reference        → the citation was dropped by the save
CAUGHT  F2  anchor the match on ENTRY_BASE           → pasted as a raw URL
CAUGHT  F3  centre the target instead of topping it  → heading is under the topbar
CAUGHT  F4  return silently when the fragment misses → a permalink that landed nowhere said nothing
CAUGHT  F5  don't give focus back                    → focus not returned to the link
CAUGHT      let uploadFiles clear the shared slot    → the slow reference was lost
CAUGHT      clear the slot whoever owns it           → a finished write cleared the pending one
```

These probes are **not** in the repo: the monorepo has no JS test infrastructure and adding it (a browser in CI, playwright as a dev dependency) is a decision to make deliberately, not as a side effect of a UX PR. The invariants they cover are written into `editor.js` and `nav.js` where the next reader will hit them. Happy to land the probe as real infrastructure in a follow-up if that's wanted.

skbarber:
## Review rounds two to four — disposition, and the stop

Three further rounds after the report above, on `c3cd03b0`, `dedeeb50` and `07b6ed10`. **All findings dispositioned; no open findings.** CI green on every round.

The severity curve, stated plainly so the record shows what these rounds were actually about:

| Round | Findings | Worst one |
|---|---|---|
| 1 | 8 | A pasted permalink could write the host into a stored note; a save could silently drop a citation |
| 2 | 2 new + 1 incomplete | Fixing the race left the reference writing at stale caret offsets |
| 3 | 1 new + 1 stale doc line | The caret mode I had just taken on advice was wrong for the commoner motion |
| 4 | — | closed |

### Round two — `c3cd03b0`

Working finding 1 properly **removed** code. Instead of deferring the pasted link behind a latch, the link is written **synchronously** with a placeholder label and only the label — cosmetic — is fetched, applied by finding that exact text again rather than by position. That deletes the queue helper, both guards, and the race they existed for. An upload is again the only thing that defers a textarea write, so `form._uploading` is back to the single slot it was on master.

It also retired a claim from my own round-one commit message: I said the two guards "fail in opposite paste orders"; the reviewer modelled it and showed one was needed in both. Moot now, recorded as wrong.

Also fixed: the topbar measured before the deferred scripts built it, and the "note isn't on this page" message blamed the data share on the **ops** book, which never reads it.

### Round three — `dedeeb50`

Two defects I introduced in round two, plus Codex's comment:

- **The 404 revert had been deleted with the restructure.** Since the match is deliberately host- and depth-agnostic, any foreign URL ending `entry/<hex>` — a permalink to a deleted entry, one from another experiment's logbook — was destroyed and replaced by a dead internal reference. Restored, on 404 only; a network error keeps the placeholder, where the id is fine.
- **Codex P3 (author escaping) — valid, its example not.** `Smith [PhD]` parses fine; CommonMark balances brackets. An *unbalanced* `]` is what ends the label early. **Codex P2 needed nothing** — it read `46d8814e`, and the helper it names was deleted in `c3cd03b0` in favour of exactly the placeholder technique P2 recommends.
- **Reverted my own previous "fix":** the `DOMContentLoaded`/rAF deferral of the topbar measurement. The ordering argument is right, the consequence isn't — those controls are not the bar's tallest element. Cold-loaded at a fragment, at 1380px and at 560px (two rows), inline and deferred place the target identically; at 560px inline reads 2px *larger*. Five lines deleted rather than keep a fix for a bug I could not reproduce.

### Round four — `07b6ed10`

- **The caret.** Round three took `"preserve"` on the reviewer's recommendation, and neither mode is right alone. `"end"` is wrong for someone still typing; `"preserve"` moves a selection starting *after* the replaced range, but the caret the paste leaves sits exactly *at* its end, which collapses to the start. Measured, that is worse than a misplaced caret — the selection spans the reference and the next keystroke **replaces** it: pasting into `"see "` then typing `" next"` gave `"see  next"`, link gone. Fixed by treating "at the end" as "after the end".
- The last stale doc sentence (`CLAUDE.md` still said `reveal()` "scrolls it to centre").
- The CHANGELOG now says what a failed validation does.

### Standing, recorded not fixed

Three, all in the package `CLAUDE.md` deferred table with the reasoning:

1. **Entry-id alphabet has three homes.** A `pattern` on `LogEntry.entry_id` in `geecs_schemas` is the right fix; it is a cross-package change and this is a UX PR. `TestTheIdAlphabetHasNotDrifted` ties both matchers — including `editor.js`'s, read off disk — to a freshly minted id meanwhile.
2. **Copy-on-plain-http exists twice** (here and the portal's `run.html`). `GeecsWebTheme/CLAUDE.md` makes the *third* surface the forcing function for promoting it into `kit.js`.
3. **Whether the topbar measurement must wait for the pickers.** My two widths bracket the band the reviewer was arguing about without entering it. The table records how to settle it — a width sweep on the month page — and says not to add the code back on reasoning alone.

### Verification

`./scripts/check.sh` → **rc=0**. GeecsLogbook **274 passed, 1 skipped**; GeecsWebTheme **179 passed**.

**Fourteen browser probes**, no page errors:

```
F1  the reference survived a same-tick save
F1b the attachment link survived a same-tick save
F1c written at once, typed text intact, label upgraded in place
F1d a same-tick save stored a working reference
F2  relative reference, no host, under a different mount prefix
F3  heading clears the sticky bar (880px target, 420px viewport)
F3b heading clears a two-row topbar (560px)
F4  says when a permalink misses, silent otherwise
F4b the ops book claims no reason it cannot have
F5  focus returned, flash visible
F6  the link survives an author name with an unbalanced bracket
F7  a URL that is not ours is pasted unchanged
F8  a caret moved elsewhere stays where the author left it
F8b typing after paste-and-wait continues after the link
```

Every one was checked by reintroducing its bug and confirming the probe fails. Three probes were caught passing **without testing anything** and rewritten: one read rows left by earlier runs, one had a `sleep` in a Playwright route handler that blocks the driver (so "submit at 1300 ms" happened at 3822 ms with both writes already landed), and one asserted only that the `href` survived when an unbalanced `]` truncates the *label* and leaves the link.

The probes are **not in the repo**: the monorepo has no JS test infrastructure, and adding it (playwright, a browser in CI) is a deliberate decision, not a side effect of a UX PR. The invariants are written into `editor.js` and `nav.js` at the lines that hold them. Happy to land the harness separately if wanted.

### On stopping

An adversarial reviewer will always find something — that is its job — so "the reviewer found something" is not a stopping rule and four rounds of treating it as one is a loop by construction. The remaining items are recorded above rather than chased. This is ready to merge.

PR #897, opened 2026-09-15
GEECS-Core 0.4.1: shortest round-trip float formatting for set commands (#819)

### LINKED ISSUE #819: GEECS-Core: %.12f set formatting sends binary representation error to LabVIEW devices

skbarber:
Fixes #819.

## What

`GeecsUdpClient.set` formatted every float with `%.12f`. At 12 decimals that exposes the binary representation — `40854.24625` went out as `40854.246249999997` — and the LabVIEW side (`Is Value a number.vi`) rejects the string, failing the set in ~0.15 s before any motion. This is the single code path for every float set to every GEECS device (the CA gateway and `GeecsDevice` both forward the raw value to it).

Floats now render through a new `geecs_core.transport._coerce.format_float`, the outbound pair of `coerce_scalar`:

- **shortest round-trip decimal** (Python `repr`): `40854.24625` → `40854.24625`, `40966.0` → `40966.0`, `0.001` → `0.001`
- **never truncated** to a fixed number of decimals — the issue explains why `%.1f` (or any fixed width) is the wrong fix: it silently zeroes DB tolerances like `0.001` and minima like `1e-05`
- **never exponent notation** (LabVIEW may not parse it): `1e-05` → `0.00001`, `1e-07` → `0.0000001`, `1e16` → `10000000000000000.0`
- every finite value keeps a decimal point, as `%.12f` did (the issue flagged "does any parser distinguish `40966` from `40966.0`" — sidestepped by not changing that); `inf`/`nan` render as before

## Other formatters checked (report only, nothing changed)

- `GeecsCAGateway/geecs_ca_gateway/gateway.py` `_make_setter` → `udp.set(geecs_var, value, ...)`: forwards caproto's value unformatted; numpy float64 is a `float` subclass so it takes the same branch. One formatter, one fix.
- `geecs_core/client/geecs_device.py` `set`: forwards to `GeecsUdpClient.set` unformatted.
- `geecs_core/testing/fake_device_server.py` line 98 `{hi:.6f},{lo:.6f}`: the fake's *reply* text mimicking the real device's out-of-range phrasing — not a set path, left alone.

## LOC (one concern)

- `_coerce.py` +34 (`format_float` + docstring), `udp_client.py` ±2, tests +101 (`test_coerce.py` 4 tests incl. 16 parametrized cases, `test_transport.py` 1 wire-bytes test), CHANGELOG/pyproject.

## Tests

GEECS-Core: **110 passed, 5 deselected** (integration), via `./scripts/check.sh` (rc 0, lint clean). `--all` in the fresh worktree fails env-shaped for the 13 untouched packages (no venvs) — CI covers them.

Proved the tests bite: with `%.12f` restored, **19 fail**; with the issue's own proposed `f"{v:.6f}".rstrip("0").rstrip(".")`, **11 fail** (the wire-bytes test, the exponent sweep, and the small-magnitude/`.0` cases).

The wire test captures the datagram on the cmd transport — the fake server alone cannot pin this, since `float("40854.246249999997") == 40854.24625`; only LabVIEW's parser sees the difference.

## Hardware verification (done, 2026-09-14, worker host, lab quiet)

`U_CompAeroTech` / `Position.Axis1` (DB: units µm, range −60000..55000, tolerance 1.0), via `geecs_core.client.GeecsDevice` from a staged copy of this branch (`PYTHONPATH` over the qserver venv); "before" = the host's installed master `geecs_core`. Stage started at **41342.18**.

| step | code | value sent on the wire | result |
|---|---|---|---|
| before | `%.12f` | `40854.246249999997` | **rejected in 0.15 s**: `Is Value a number.vi<ERR> "40854.246249999997" is not a number`; readback unchanged 41342.12 |
| after | this branch | `40854.24625` | **accepted** in 1.25 s, exe returned 40854.24625, readback **40854.1325**, \|Δ\| = 0.114 µm ≤ 1.0 tolerance |
| restore | this branch | `41342.18` | accepted, readback 41342.12125, \|Δ\| = 0.059 µm |

Observation worth recording: not every 12-decimal tail is rejected. `41342.196250000001`, `41342.180009999996` and `41342.180039999999` were all accepted by the same device under the old code (readbacks within 0.06 µm); the issue's `40854.246249999997` reproduces the rejection every time. Whatever `Is Value a number.vi`'s rule is, the shortest-digits string is inside it for all of these. Staged copy and script removed from the host afterwards.

Not verified on other device types (magnets, DG645, cameras) — every value the new formatter produces is a strict subset of what `%.12f` produced digit-wise for the normal range, minus the tail, so the risk is in exponent-range values (<1e-4, ≥1e16), where the old format was sending `0.000000000000`-style strings anyway.

- `format_float` lives in `transport/_coerce.py` next to `coerce_scalar` (inbound/outbound pair) rather than inline in `set`, so it is unit-testable and there is one place to look.
- Integer-valued floats keep `.0` (`40966.0`) rather than the issue's `40966` — conservative, matches the old format's always-a-decimal-point shape.
- Version: **patch** (0.4.0 → 0.4.1) — bug fix, no API change.

Facility-literal check: none added outside CHANGELOG/docstrings (device name in the changelog entry describes the incident).


skbarber:
## Adversarial review (fresh-context subagent, reviewed commit fef8e75c)

Pasted in full; disposition of each finding follows in a separate comment once the reviewer has confirmed the fix commit.

---

**Verdict: sound. One low-severity code gap (pre-existing, but the PR's "one formatter" claim overstates coverage), one low-severity CHANGELOG inaccuracy, no redundancy, placement correct.** Findings below in severity order; everything else I tried to break held.

### What I verified (so the findings are the residue, not the whole picture)

- Suite: 110 passed, 5 deselected in the worktree env.
- Tests bite (via a scratchpad pytest plugin swapping the formatter, no repo edits): old `%.12f` → **19 fail**; the issue's `.6f`/rstrip → **11 fail**; **bare `repr` → 8 fail** — so the exponent-expansion branch is independently pinned, not just the tail removal. The wire-bytes test works because `_exchange_inner` looks up `self._cmd_transport.sendto` at call time (`udp_client.py:~366`), so the attribute patch is seen.
- `format_float` edge cases by hand: `-0.0`→`-0.0`; `9.999e-05`→`0.00009999`; `1e15`→`1000000000000000.0`; `0.1+0.2`→`0.30000000000000004` (exact round trip); `5e-324` → 326-char string, `1e308` → 311 chars — long, but `%.12f` produced *longer* strings for those, so no regression, and all fit a datagram. `Decimal(text)` from a repr string is exact and `format(..., "f")` ignores context precision, so no rounding can creep in.
- Every set funnels to this one line: only `GeecsCAGateway/geecs_ca_gateway/gateway.py:411` and `GEECS-Core/geecs_core/client/geecs_device.py:157` call `GeecsUdpClient.set`; nothing else constructs a `GeecsUdpClient` outside tests. The gateway's `channels.py:cast_value` does `float(value)` before the setter, so caproto/numpy dtypes are a non-issue on that path.
- No other test in the repo pins the old `.000000000000` shape (only the fake server's comment at `fake_device_server.py:94`, which is device-*reply* mimicry). No contract file (`PV_CONTRACT.md`, `DESIGN.md`, `docs/`) states the outbound float format, so nothing travels with this behavior change.

### Findings

**1. Low — the `isinstance(value, float)` gate leaves a hole the PR body says is closed.** `udp_client.py:332` only routes Python `float` (and `np.float64`, a subclass) through `format_float`; `np.float32` and any other `numbers.Real` fall to the `else` branch, `f"{value}"`. Concrete scenario, `GeecsDevice.set` path only (the gateway casts first): `dev.set("Speed.Axis2", np.float32(1e-5))` transmits `9.999999747378752e-06` (verified in the root env) — exponent notation *and* a representation tail, the two things this PR exists to stop. Pre-existing (the legacy API had the same gate), so not a regression — but the PR body's "One formatter, one fix" and the CHANGELOG's "one formatter for every float set to every device" are true only for `float`/`float64`. Cheap closure if wanted: gate on `numbers.Real` minus `Integral` (or `not isinstance(value, (bool, int, str))` → `format_float(float(value))`). If left, the claim in the CHANGELOG should say "Python floats".

**2. Low — CHANGELOG conflates "has a tail" with "rejected".** `GEECS-Core/CHANGELOG.md` 0.4.1: *"LabVIEW devices rejected the string as 'not a number' (86% of five-decimal values in `U_CompAeroTech`'s working range)"*. The 86.3% in issue #819 is the fraction that *format with a tail*; the PR body's own hardware observation is that three of four sampled tails (`41342.196250000001` etc.) were **accepted**. As written a reader infers 86% of sets failed. The test docstring in `test_coerce.py` states it correctly ("grew a garbage tail"); the changelog sentence should match.

**3. Info, not a defect — the residual hardware risk is real, not a formality.** I checked the legacy API at `legacy-scanner-final` (`GEECS-PythonAPI/.../geecs_device.py:421`): it also sent `%.12f`. So there is *no* years-of-field-use history for the shortest form from the Python side; the "not verified on magnets/DG645/cameras" caveat in the PR body is the honest state. The argument that normal-range output is digit-wise a subset of the old output holds; the shape difference (`1.0` vs `1.000000000000`) is the only thing an unusually strict parser could care about, and the Aerotech accepted it.

### Redundancy — none
Searched for competing formatters: `GeecsScanner/geecs_scanner/service/actions.py:21` and `summaries.py:66` (`_fmt`) are display-only (`str`/`:g`), not wire; `fake_device_server.py:98` `.6f` is reply text; no `rstrip("0")`, `Decimal(repr`, or `format(Decimal` elsewhere; the legacy formatter is deleted. `_coerce` is imported only inside GEECS-Core (`udp_client`, `tcp_subscriber`, `fake_device_server`, tests).

### Placement — correct
`transport/_coerce.py` next to `coerce_scalar` (inbound) is the right and only home: transport-layer wire encoding, below every client, no schema/DB knowledge. Nothing here would have to move for a foreseeable feature.

### Nits (not ranked)
- `test_never_exponent_notation_across_magnitudes` ends with `assert math.isfinite(value)`, which is vacuous for the swept inputs.
- In `format_float`, `or not math.isfinite(value)` is dead: `repr` of `inf`/`nan` contains no `e`, so the first clause already returns. Harmless; the docstring's intent is clear.

skbarber:
## Review disposition (fix commit cd62ac26, confirmed by the same reviewer)

| # | Finding | Disposition |
|---|---|---|
| 1 | `isinstance(value, float)` gate lets `np.float32` / other `numbers.Real` fall to `str()` (exponent notation + tail) | **Fixed** in cd62ac26 — `set` now routes every `numbers.Real` that is not `Integral` through `format_float(float(value))`; wire test adds a `Fraction(1, 100000)` case (numpy is not a transport dependency). Reviewer re-derived per type: `np.float64`/`np.float32`/`Fraction` take the formatter, `bool`/`int`/`np.int64`/`str`/`bytes` keep their old branches; with the old gate reinstalled the test fails (`'1/100000' == 1e-05`). |
| 2 | CHANGELOG conflated "grows a tail" (the 86%) with "rejected" | **Fixed** in cd62ac26 — entry now states 86% = tail fraction, rejection reproduced on hardware, "not every tail is rejected, but the fix removes them all". Reviewer confirmed accurate. |
| 3 | Info: no field history for the shortest form from Python; other device types unverified | **Acknowledged** — the PR body's "not verified on magnets/DG645/cameras" caveat stands; nothing to change. |
| nit | dead `or not math.isfinite(value)` clause; vacuous `assert math.isfinite(value)` | **Fixed** in cd62ac26 — both removed; `inf`/`-inf`/`nan` still pass through via the `"e" not in text` early return (reviewer confirmed). |

Reviewer's for-the-record notes, **waived** (pre-existing, unchanged by this PR, and not reals): `decimal.Decimal` is not a `numbers.Real`, so `Decimal('1E-7')` still goes out as `1E-7`; `np.bool_` is neither `bool` nor `Integral`, so it goes out as `True`. Neither is a type any caller in the repo sends; widening the gate to `numbers.Number` would be a separate decision.

No undispositioned findings. Combined suite on cd62ac26: 110 passed, 5 deselected.

PR #898, opened 2026-09-15
CI: install GeecsBluesky's ca extra and assert the mock-CA layer ran (#866)

### LINKED ISSUE #866: CI never runs GeecsBluesky's mock-CA tests (no `ca` extra in unit-tests.yml)

skbarber:
Closes #866.

## What

One concern, two files: `.github/workflows/unit-tests.yml` GeecsBluesky steps (+22 / -2) and the mirrored runner in `scripts/check.sh` (+19 / -1 — its header rule: the runner table changes with CI in the same PR).

1. **Install the `ca` extra** alongside `qserver` and `qs-client`. Every GeecsBluesky test module behind `pytest.importorskip("aioca")` — the device, shot-control, session, and plan tests (29 modules/classes on master) — skipped wholesale without it.
2. **Make the skip an assertion.** The test step now runs pytest with `-rs` (skip reasons in the short summary), tees the output to a file, and fails the step if the summary carries `could not import 'aioca'` / `'p4p'`. `shell: bash` is set explicitly so the step runs under `-eo pipefail` and the tee cannot mask pytest's own exit code.
3. **Spell `not hardware` in the CI `-m` expression** (review finding). The CLI `-m` replaces pyproject's `addopts` marker filter; until now `importorskip("aioca")` was the only thing keeping `hardware`-marked modules off the runner, and this PR removes that gate. Inert on master (no `hardware`-marked test exists), load-bearing the moment the workflow syncs onto `feature/native-bluesky-plans`, whose three `test_*_hardware.py` modules are gated by importorskip alone.
4. **`scripts/check.sh` mirrors the step** (review finding): `run_bluesky_suite` runs the same command and exits non-zero with a pointer to `/env-doctor` when the mock-CA layer skipped, so a local `/check` can no longer report green with 288 tests silently skipped.

No package code changed, so no version bump / CHANGELOG entry.

## Why

A green check on a GeecsBluesky PR said nothing about the detector, shot-control, or plan code paths (found in the #861 review rounds, where CI stayed green while the calibration plans' tests had never run anywhere but a developer machine). Option 1 of the issue is taken; option 2 (a separate path-gated job) is only warranted if the install proves slow — measured below.

## Before / after

Same test tree, GeecsBluesky at master `c49f88fd`:

| | passed | skipped | deselected | notes |
|---|---|---|---|---|
| master CI (run 34856927333, no `ca`) | 468 | 30 | 2 | 29 skips are `could not import 'aioca'`, 1 is `tiled.client` |
| local, `ca` installed | 756 | 1 | 4 | the 1 skip is `tiled.client` (expected — `tiled` is not a CI extra); the 2 extra deselected are the `integration`-marked tests inside `test_scan_request_hardware.py`, which now collects |
| **this PR's CI** (run 34921345029, head 71a46701) | **756** | **1** | **4** | identical to local; the 1 skip is `tiled.client` |

(The issue's 248 / 25 figures were the feature branch's counts; master's are above.)

**Install time, measured on the runner:** the GeecsBluesky install step took 15 s on this PR (02:31:34 → 02:31:49) versus 15 s on master's run 34856927333 — the `ca` extra adds `aioca 2.1` + `epicscorelibs` + `setuptools-dso`, all wheels, about 8 s of downloads inside a step whose cost is dominated by the rest of the env. The real cost is the tests now running: the GeecsBluesky test step went from 13 s to 3 min 06 s (pytest: 185.6 s), and the whole `unit-tests` job from about 4 min to 6 min 38 s. Option 2 of the issue is therefore not warranted by install time; if the +3 min per PR matters, the test duration is the lever, and the step's `--durations` output would say where it goes.

The mock-CA layer is hermetic: the 288 newly-running tests passed on a machine with no lab network and on the ubuntu runner.

## Proof the assertion bites

Running the new step's `grep` against the no-extra output (the exact pytest output the master CI job produces today) matches 29 lines and exits 1:

```
$ grep -cE "could not import '(aioca|p4p)'" pytest-report.txt
29
::error::GeecsBluesky's mock-CA tests skipped …
assertion-on-before rc=1
```

With the extra installed the grep matches nothing and the step exits with pytest's code.

The mirrored `check.sh` gate was exercised end to end: `./scripts/check.sh GeecsBluesky` in an env with the extra → `756 passed, 1 skipped, 4 deselected`, exit 0; the same command with `aioca` stubbed as unimportable on `PYTHONPATH` → `468 passed, 30 skipped, 2 deselected` (today's CI picture), the `check.sh: GeecsBluesky's mock-CA tests skipped — install the ca extra (see /env-doctor)` line, exit 1.

- **`p4p` is in the grep although master's `ca` extra is `["aioca"]` only** (p4p rides the `capture` extra here). No master test importorskips p4p, so it is inert today; on `feature/native-bluesky-plans` `ca = ["aioca", "p4p"]` and two test modules importorskip p4p, so the grep covers that branch when the workflow syncs over.
- **A developer running `/check` without the `ca` extra now gets a red GeecsBluesky suite** instead of a silent skip. That is the issue's point applied locally; the message names the fix. Veto by dropping the grep from `run_bluesky_suite` if a green-with-skips local run is preferred.
- **The check is a grep on pytest's summary, not a pytest plugin.** Simplest thing that provably fails; the matched string is pytest's own `importorskip` reason text (`could not import {modname!r}`), identical in the locked pytest 7.4.4 and the root env's 9.1.1.

## Hardware verification

N/A — CI workflow + local check script only; no scan-execution or device code changed.


skbarber:
## Adversarial review (fresh-context subagent, per `/land`)

Reviewed head `d93253fe` (the initial commit); fixes confirmed by the same reviewer against `71a46701`. Report pasted in full, disposition follows.

---

CI on the head: `unit-tests` green in 6m12s; GeecsBluesky install 12 s (aioca 2.1 + epicscorelibs only), test step 3m03s, `756 passed, 1 skipped (tiled.client), 4 deselected` — matches the PR body's local numbers. The PR body's "CI-measured counts ... filled in below once the run completes" placeholder is still empty.

### Findings (ranked)

**1. MEDIUM — Correctness: the CI `-m` expression drops `not hardware`, and this PR is what removes the only other gate on the hardware modules.**
GeecsBluesky's `pyproject.toml` (line 115) sets `addopts = "-m 'not hardware and not integration'"`; the CI step passes its own `-m "not integration and not fake_server"`. pytest's `-m` is `action="store"` (verified in `_pytest/mark/__init__.py`, and empirically: a scratch suite with a `hardware`-marked test collects it under the CI expression, deselects it under addopts alone). So `hardware` tests are only kept out of CI today because they `importorskip("aioca")` — the exact gate this PR removes. On master no test carries `hardware`, so it is inert here. On `feature/native-bluesky-plans` it is not: `tests/test_806_hardware.py:43`, `tests/test_phase0_hardware.py:53`, `tests/test_phase1_hardware.py:59` are `pytestmark = pytest.mark.hardware`, gated only by `importorskip("aioca")`/`("p4p")`, and that branch's `ca = ["aioca", "p4p"]`. The routine master→feature sync carries this workflow over, at which point those modules collect and run against no gateway on the ubuntu runner (CA connect timeouts → red or very slow job). The PR body's judgment call "the assertion already covers that branch when it lands — no edit needed then" is the wrong conclusion. Fix is one token: `-m "not integration and not fake_server and not hardware"` (no effect on master's counts; the marker is already declared).

**2. LOW–MEDIUM — Repo contract: `scripts/check.sh` is required to change in the same PR.**
`scripts/check.sh` header, line 11: "when CI changes, change the runner table here in the same PR"; `.claude/skills/check/SKILL.md` lines 14–17 restate it ("update the script (not prose) when CI changes"). The GeecsBluesky runner (`scripts/check.sh:243`) is untouched, so a local `/check` on an env without `ca` reports green with the same 288 tests silently skipped — the #866 symptom reproduced locally, by the tool whose stated purpose is to mirror CI. The PR body's disposition ("does not manage extras") is true of the install half, but the assertion half ports as-is. Either port it or amend the check.sh header rule; leaving both as they are contradicts a written contract.

**3. LOW — Diff comment is inaccurate on master.**
The install-step comment says `--extras ca` covers "every module behind importorskip("aioca") / importorskip("p4p") ... aioca and p4p ship wheels". On master `ca = ["aioca"]` (`GeecsBluesky/pyproject.toml:84`); p4p is in `capture` (line 88); no master test importorskips p4p; the CI log shows only aioca + epicscorelibs installed. A reader will conclude `ca` brings p4p. Reword as forward-looking (or drop p4p from the install comment; the grep can stay).

**4. LOW — Placement/simplicity of the assertion (optional).**
The check is expressed against pytest's human-readable `-rs` summary (tee + pipefail + regex). `poetry run python -c "import aioca"` right after the install asserts the same fact with no coupling to pytest output, and the grep has two silent holes the direct import doesn't: `importorskip("aioca", reason=...)` replaces the matched text, and a test that importorskips a `geecs_bluesky` module whose cause is aioca yields `could not import 'geecs_bluesky....': No module named 'aioca'`, which the regex misses. Neither exists today. Also note the same silent-skip hazard remains unasserted for the workflow's four other importorskip-gated extras (DataPortal `analysis`, MCP `analysis-run`, `qserver`, `qs-client`); acceptable for scope, but if grep-on-summary becomes the pattern, one generic `could not import` grep with a tiled allowlist covers them all.

### Refuted / no finding
- Redundancy: no existing skip-guard mechanism anywhere (`.github/`, `scripts/`, root/`tests/`/GeecsBluesky `conftest.py` checked); this is the first. Placement in the workflow is right.
- pytest message text: `could not import {modname!r}` in both the lock's pytest 7.4.4 and root's 9.1.1; `-q` does not suppress `-rs` (guard is `reportchars` only; the CI log shows the SKIPPED line).
- `shell: bash` → `-eo pipefail` claim is correct; pytest's exit code propagates through `tee`.
- Capture tests (`tests/capture/*` importorskip h5py) are not a second #866: h5py is a main dep of GEECS-Data-Utils and rides in; counts confirm.
- No version bump/CHANGELOG needed (no package code); no scan-folder or public-repo-hygiene surface.

---

## Disposition

| # | Finding | Disposition |
|---|---|---|
| 1 | `not hardware` missing from the CI `-m` expression | **Fixed** in 71a46701 — `-m "not integration and not fake_server and not hardware"`, with a comment on why the CLI `-m` must spell it. Reviewer confirmed. |
| 2 | `scripts/check.sh` runner must change in the same PR | **Fixed** in 71a46701 — `run_bluesky_suite` mirrors the step (same `-m`, `-rs`, tee under the script's `set -o pipefail`, grep, non-zero exit with a pointer to `/env-doctor`) and captures pytest's failure explicitly because `run_suite` is called in a `\|\|` context where errexit is off. Reviewer confirmed, including a scratch reproduction of the fall-through without the explicit `\|\| return 1`. Exercised end to end locally: with the extra → 756 passed, exit 0; with `aioca` stubbed unimportable → 468 passed / 30 skipped, the message, exit 1. |
| 3 | Install comment claims `ca` brings p4p | **Fixed** in 71a46701 — the install comment names `aioca` only; the grep and the test-step comment keep p4p as forward cover for the feature branch's `ca = ["aioca", "p4p"]`. Reviewer confirmed. |
| 4 | Grep-on-summary vs a direct `import aioca` check; other extras unasserted | **Waived** — the grep asserts the observable the issue names (the tests ran, not merely that the module imports); the two holes do not exist in the tree today; the other importorskip-gated extras are out of this PR's one-concern scope. Reviewer accepted the waiver. |
| nit | PR body placeholder for CI numbers | **Fixed** — body now carries the CI-measured counts and install time from run 34921345029 (head 71a46701). |

CI on 71a46701: `unit-tests` pass (pull_request run 34921345029, 6m38s; push run 34921341872 also green), `pre-commit` pass, `console-windows` skipped by repo variable.

PR #899, opened 2026-09-15
GeecsScanner 0.6.0: New scan form opens on the mode segment, static hint prose dropped (#895, #896)

### LINKED ISSUE #895: Web scanner: drop the static explanatory hint text under the form controls
Sam, 2026-09-14, after a day of scans on the web scanner: the descriptive prose under the form controls ("presets/ in the configs tree · seeds the form below", "catalog name → device:variable", the four-line trigger-profile note, "strict only · seconds · blank = as fast as the box allows", "goes to ScanInfo and the logbook entry", …) is noise once you have used the page once, and it makes the New scan panel tall and ragged.

**Ruling:** delete the static explanatory hints to start; hover text or a help affordance can come later if anyone misses them.

**Keep** the hints that carry computed state — the point count and direction (`17 points · descending`), the validation `.err` slots, the unit/mode hints that change with the form (`shots-hint` switching between `num` and `shots_per_step`), and the load-failure notes (`actions-note` / `presets-note` when a listing failed). The kit's rule (`kit.css`: ".hint stays — it often carries the unit") applies to those, not to prose.

Where: `templates/console.html` (the `.hint` spans of the submit, devices, actions, calibration panels and the two drawers) and the static strings set from `static/scanner.js` (`presets-note`, `mode-note`'s `NOTES`, `mv-hint`, `devq-hint`).

### LINKED ISSUE #896: Web scanner: the preset picker is optional and should not head the New scan form
Sam, 2026-09-14: the preset picker at the head of the New scan form is fine as a control but wrong as the first thing on the page — presets are optional and will cover a small subset of scans, yet the picker takes a full-width row above the mode segment and reads as "step 1".

**Ask:** move it out of the headline position and shrink it. Two shapes that fit the kit; pick by looking at the rendered page:

1. A compact "Load preset…" select in the panel's header row, right-aligned beside the mode segment, normal field width.
2. Put it in the form footer next to "Save as preset…" — load and save are the same concern, and the footer already holds the estimate and Start.

Either way the form starts at the mode segment. The `preset-name` note that shows which preset seeded the form stays (it is provenance the operator wants).

skbarber:
Closes #895, closes #896. GeecsScanner 0.5.1 → **0.6.0** (minor: the form's shape changes).

## What

**#896 — the preset picker leaves the headline row.** Both placements from the issue were rendered on the demo manager and compared by looking: (1) header-right beside the mode segment, (2) the footer beside "Save as preset…". **Footer chosen**: load, the `preset <name>` provenance note, save and Start sit in one row — the preset concern has one home — and the header keeps its two segments, so the panel opens on the mode choice with nothing wrapping the header at narrower widths. The header variant also split the picker's "Load preset…" from its provenance note at opposite ends of the panel. The picker is a compact select (page-only class `preset-load`, the kit's `--row-h` like the buttons beside it, tokens only) whose first option reads "Load preset…"; it ships with a disabled "Loading presets…" option until the listing arrives.

**#895 — the static explanatory hints are gone.** The prose `.hint` spans under the preset, variable (both axes), trigger-profile, description, calibration and drawer controls; the per-mode `mode-note` paragraph and its `NOTES` table in the script; the success-case prose written into `presets-note`, `actions-note` and `mv-hint`. Kept, per the issue and the kit's rule: the point count + direction (`pts1`/`pts2`), every `.err` slot, `shots-hint` (`num` vs `shots_per_step`), the shot period's unit (`seconds` — reduced from the three-clause sentence), the device search's live "N of M devices" count, and every load-failure / empty-listing note (`presets-note`, `actions-note`, `mv-hint`'s `settablesNote`, `devq-hint`'s unreachable branch). The trigger-profile field's environment-open gotcha (a new or edited profile is refused by the worker until its next environment open) survives as the field's hover text (`title`) — the "hover text" #895 anticipates — and the arc brief's row 5a now says so.

`formShape` / `fillFormFromPreset` / the demo's first-preset default are untouched — how a preset seeds the form has not changed.

- **The picker is a verb, not a mirror of state.** It snaps back to "Load preset…" after a load rather than staying on the chosen name. Reason: in the footer the select would otherwise sit right next to the note saying the same name; and an operator can now re-load the same preset over an edited form (previously re-picking the same option fired no `change`). The `preset-name` note carries the provenance, as the issue asks.
- `mv-hint`'s success string ("4 numeric settables · 2 aliased, listed first · from the GEECS DB") is dropped as prose-with-a-count (fixed once loaded; the option labels already show `alias · name (units)`), while `devq-hint`'s "N of M devices" stays because it changes as you type — live feedback on a search.
- The empty device table said "Pick a preset above." — the row it pointed at is gone; it now reads "No devices — add one.", the wording the script already used. Start's disabled title says "load a preset" to match the picker.
- A preset whose load fails (deleted from the tree since the listing) no longer leaves its name as the form's provenance: `selectPreset`'s `.catch` restores the loaded document's name (or none) and recalculates.
- The `&nbsp;` placeholder hints under start/stop/set-to went too: they only padded row height against hints that no longer exist.

## LOC by concern (approx., from the diff)

| Concern | template | scanner.js | scanner.css | tests |
|---|---|---|---|---|
| #896 picker placement | ~8 | ~14 | +3 / −1 | +14 (new pin) |
| #895 hint prose | ~30 | ~15 | — | +38 (new pin), 1 assertion loosened (`<select id="preset"` prefix, the select now carries `class`/`aria-label`) |
| review fixes (second commit) | ~3 | ~6 | 1 | 3 lines (the prose pin judges `.hint` spans; the loading word is pinned) + 1 line in the arc brief |

No facility literals added. No new colour literals (tokens only; the theme's walk covers `scanner.css`).

## Tests

Both new pins were run against the untouched template first and **fail there** (2 failed), then pass on the change. Run the way CI does (`scripts/check.sh GeecsScanner GeecsWebTheme` + `--lint`, exit codes read unpiped):

- GeecsScanner: **75 passed, 1 skipped** on both commits (the inline-script parse test skips — the page has no inline script)
- GeecsWebTheme (the literal-colour walk over the scanner's CSS/templates): **182 passed**
- lint: OK

## Hardware verification

Not applicable — UI only; verified on the demo manager, screenshots attached in the session.


skbarber:
## Adversarial review (fresh-context subagent, over 37d1536b)

Pasted in full per the land ritual; disposition follows in the next comment once the reviewer has confirmed the fix commit.

---

Checked: the diff, PR body, #895, #896, root/GeecsScanner/GeecsWebTheme `CLAUDE.md`, every reader/writer of the touched ids in `scanner.js`, `kit.css`'s `.field`/`.hint`/`.err`/density rules, the README, `docs/geecs_scanner/overview.md`, and the arc brief. Findings ranked; each survived my attempt to refute it.

### 1. Medium — the picker hardcodes height/padding, which the kit forbids, and mismatches its footer neighbors at the default density

`GeecsScanner/geecs_scanner/static/scanner.css:33` sets `height:26px;padding:0 6px` on `.kit select.preset-load`. The kit's density doctrine at `GeecsWebTheme/geecs_web_theme/static/kit.css:46-50` says: "Nothing else in the system encodes a padding — a surface that hardcodes one stops responding to this." Concrete scenario: at the comfortable density `--row-h` is 34px (`theme.css:76`); the select's row-mates "Save as preset…" and "Start" (`console.html:189-190`) are plain `.btn` at `height:var(--row-h)` (`kit.css:262`), so the picker renders 8px shorter than the buttons beside it; at compact density (`kit.css:53`, 26px) they happen to match. The PR's justification "sized like the topbar's operator box" copies `.kit .who input` (`scanner.css:12`, same hardcode) — but the topbar's neighbors are `.btn.sm`, which the kit itself fixes at 26px (`kit.css:265`), so the copy holds there and not in the footer. Fix: `height:var(--row-h)` (drop the fixed padding), or make the footer buttons `.btn.sm`. Redundancy note: lines 12 and 33 are the same box recipe twice; one rule with two selectors.

### 2. Medium — the preset picker lost its loading state

Old template: `<span class="hint" id="presets-note">Loading presets…</span>`. New (`console.html:187-188`): `presets-note` ships empty and `<select id="preset">` has no options until `loadConfigs` resolves — the whole `Promise.all` (`scanner.js:356-364`), which also waits on `/api/scan-variables` and `/api/configs/trigger_profiles`, neither of which has a `.catch`. Scenario: `/api/scan-variables` is slow or fails → the footer shows an enabled, option-less select with no text beside it; on rejection `renderPresetList` never runs, so the picker stays mute for the session (the banner reports the failure, the picker does not). The same diff kept `actions-note`'s "Loading the action library…" (`console.html:246`), so the two controls now disagree. Kit doctrine: every pane owes `loading · empty · error`. Cheap fix: a disabled `<option>Loading presets…</option>` in the template. The new test asserts `id="presets-note"` exists (`test_page.py:198`) but not that a loading word does, so it passes with the state missing.

### 3. Medium (cross-reference) — the arc brief now points at documentation this diff deletes

`Planning/native_bluesky/10_web_scanner.md:344` (row 5a): "trigger profiles are materialised once at the worker's environment open … documented in the field's hint, not rebuilt live." This diff removes that hint (old `console.html:168`), and I found no other home for the fact that a scan naming a new/edited profile is refused until the worker's next environment open — the surviving "environment open" statements are about shot offsets only (`console.html:274`, `scanner.js:918`; nothing in `docs/geecs_scanner/overview.md`). The deletion follows #895's ruling, so not a veto; but the brief's sentence is now false and an operational gotcha has no user-facing carrier. Suggest a `title` on `#trig` (the page already uses `title` on `.seg` buttons, `console.html:107-108, 132-133` — the "hover text" #895 anticipates), plus a one-word edit to the brief row; the row's "candidate follow-up" preflight check is the real fix.

### 4. Low — contradictory empty states, and the same empty state in two wordings

With zero presets: `#devs` says "No devices — load a preset or add one." (`console.html:176`) while `#preset` is disabled and `presets-note` says "No presets in the configs tree." (`scanner.js:395-396`), and Start's title reads "pick a preset" (`scanner.js:586`) beside a picker that cannot be picked. Separately, `noDevicesNote()` (`scanner.js:466`) replaces the template row with "No devices — add one." after the first preset load or any row removal — two wordings for one state; the divergence predates the PR but the PR rewrote one of the two. One string, owned by `noDevicesNote()`.

Related, out of the diff's stated scope but sharpened by its framing: `valid = ok && !!S.presetDoc` (`scanner.js:576`) means the form cannot Start unless a preset was loaded at some point — a configs tree with no presets can never submit from the page — while #896 says "presets are optional" and `overview.md:45` says "or compose one". Worth an issue, not this PR.

### 5. Low, pre-existing but now the sole carrier — provenance after a failed load

`selectPreset` claims `S.presetName = name` before the fetch (`scanner.js:404`). Scenario: preset B is deleted from the tree after the listing; the operator picks B → 404 → `showError`, but `updateStartGate` keeps writing "preset B" into `#preset-name` (`scanner.js:587`) and `buildPreset` stamps `name: "B"` (`scanner.js:612`) on a form still holding A's values. Before, the select also showed B (equally wrong); the PR body makes the note "the provenance", so it is now the only thing the operator sees. Reset `S.presetName`/restore in the `.catch`.

### Nits
- Empty `.hint` spans remain (`mv-hint` `console.html:227`, `devq-hint` `:308`, `actions-note` once loaded) inside `.field{display:flex;flex-direction:column;gap:4px}` (`kit.css:279`); an empty span is still a flex item, so each adds a 4px gap (`.err` is `display:none` and does not). A kit-level `.hint:empty{display:none}` would settle it; not this PR's.
- Placeholder idiom: the page's other selects use `— pick a variable —` / `— pick an action —` / `— none —` (`scanner.js:760, 834, 377`); the new one is "Load preset…". The PR argues "verb"; it is a look-and-judge call for Sam.

### Verified clean
- **Kit CSS sibling rules:** `~ .err` (`kit.css:550`) is a general-sibling selector and every `.err` still follows its control (`console.html:146,156,161,228,320`); nothing depended on the removed `.hint` spans.
- **Redundancy:** searched the repo for `mode-note|presets-note|preset picker|Pick a preset|seeds the form|Load preset` — hits only in this package, its CHANGELOG and tests. The kit has no compact-select component (its `select` rules are all `.field select`, `kit.css:284-291, 547-551`); the one overlapping recipe is finding 1.
- **Placement:** page-only CSS in `scanner.css` is the home `GeecsScanner/CLAUDE.md` prescribes; two "compact control in a row" instances in one surface are below the kit's "its own small PR against kit.html" bar.
- **Tests:** the new pins' slicing is sound (`test_page.py:158-160`) and per the PR body both fail on the old template. The `change` handler's snap-back is unpinned — inherent, no JS runner in this repo.
- **Docs:** `overview.md:45-55` and brief §3 item 2 describe the form without claiming the picker's position — still accurate. `README.md` "The page arrives in 0.2.0" is stale but pre-existing.
- No facility literals, no colour literals, no scan-folder surface; version/CHANGELOG travel with the change.

skbarber:
## Review disposition

Every finding from the review above went back to the **same reviewer** for confirm/deny after each fix commit (three rounds: 8d294dea, 0e604ebd, d1214ee7). Final verdict from the reviewer: "No open findings from my side — merge-ready pending Sam's look at the rendered footer."

| # | Finding | Disposition |
|---|---|---|
| 1 | Picker hardcoded `height:26px`, shorter than its `.btn` footer neighbours at comfortable density | **Fixed** 8d294dea — `height:var(--row-h)`. `.who input` deliberately stays at 26px (its neighbours are `.btn.sm`). Reviewer confirmed. |
| 2 | Picker lost its loading state | **Fixed** 8d294dea — a disabled `Loading presets…` option ships in the template, pinned by the prose test. Reviewer confirmed. |
| 3 | Arc brief row 5a cited the deleted trigger-profile hint; the environment-open gotcha had no user-facing carrier | **Fixed** 8d294dea — the gotcha is `#trig`'s `title` (the hover text #895 anticipates); the brief row says so. Reviewer confirmed. The reviewer's follow-on note that the pin was a regex over HTML: **fixed** 0e604ebd — `_hint_texts` reads the page through the stdlib `HTMLParser`; the reviewer independently proved it bites (prose re-inserted → fails; nested tags and a prose `title` handled). |
| 4 | Two wordings for the empty device table; Start's title "pick a preset" beside a disabled picker | **Fixed** 8d294dea — one wording ("No devices — add one."), title "load a preset". The `valid = ok && !!S.presetDoc` gate (the form cannot Start without a loaded preset, though #896 calls presets optional) is pre-existing and **deferred to #900** with the reviewer's agreement. |
| 5 | A failed preset load left the failed name as the form's provenance | **Fixed** 8d294dea (restore in the `.catch`); the reviewer found the restore value (`S.presetDoc.name`, the document's inner name, not tied to the file stem on read) introduced a narrow drift → **fixed** 0e604ebd (`S.loadedName`, the listing name that actually loaded); the reviewer found one last edge (a save did not update `S.loadedName`) → **fixed** d1214ee7 (one assignment). Reviewer confirmed at each round. |
| nit | Empty `.hint` spans still cost a 4px flex gap | **Waived** — a kit-level `.hint:empty{display:none}` is GeecsWebTheme's small PR, not this surface's. |
| nit | "Load preset…" vs the page's `— pick a … —` placeholder idiom | **Waived** — the picker is a verb that snaps back after a load, so it reads as an action, not a value slot; flagged in the body as a judgment call for Sam to veto by looking. |

Tests at the head (d1214ee7), exit codes read unpiped: GeecsScanner 75 passed / 1 skipped; GeecsWebTheme 182 passed; lint OK.

Issue #894, opened 2026-09-14
Strict scan fails at its first prepare after a long first move: the file plugin's arming frame never comes while the box is ARMED


skbarber:
## Symptom

Scan002 of 26_0914 (the `00_focuscan` preset: strict `scan` over `U_ModeImagerESP.position_axis_1` −25 → −5, 41 points × 3 shots, `UC_ModeImager`, HTU-NoGas). The stage sat near −5, so the first step was a 20 mm move. The move completed, then the run failed before its first shot. The portal's reason is the bare status repr:

```
TimeoutError("uc_modeimager-hdf-capture didn't match True in 10.0s, last value False")
```

This is the same operator experience as the legacy engine's "a long first move times out the scan", but the mechanism is new.

## Timeline (scan.log + the worker journal)

| time | event |
|---|---|
| 13:00:35.9 | run opens, ScanInfo written; the bracket drives the box **ARMED** |
| 13:00:38.8 | `u_modeimageresp-position_axis_1: moving Position.Axis 1 → -25.0 (tol=0.015)` |
| 13:01:04.4 | move complete (26 s); `uc_modeimager-nonscalar_save_path: native saving on` — the first `prepare` of the step's `take_reading` |
| 13:01:04–13:01:12 | the file plugin's `Capture=1` waits `ARM_TIMEOUT_S = 8 s` for a frame and gets none |
| 13:01:15.5 | `wait_for_value(Capture_RBV, True, timeout=10)` expires → `finished (fail)` |

The plugin's own reason surfaced twenty minutes later, at the start of Scan004, as an un-retrieved task (the stock `ADHDFDataLogic` never awaits the `Capture` put, so its `op.done(error=…)` is orphaned):

```
Task exception was never retrieved ... exception=RemoteError('no frame from UC_ModeImager image within 8 s')
```

## Mechanism

`06_pva_file_plugin.md` §2.3: the plugin acknowledges `Capture=1` only once it has **decoded one frame**, because LabVIEW tells us the geometry only by pushing a frame — "the idle re-push LabVIEW sends at 1 Hz is enough". That holds when the first `prepare` follows the ARMED transition within a few seconds: the last STANDBY frame (or its re-push) is still arriving. Every successful strict scan fits that shape — Scan001 (3 s move), Scan003 and Scan004 the same day (3 s moves) all prepared within ~3 s of ARMED and armed fine.

[...]

Issue #868, opened 2026-09-13
Native branch: a refused motor :SP put leaves no PV name in the scan log (#817's move-side half)


skbarber:
Found by the adversarial review of #867 (the master → feature/native-bluesky-plans sync).

**Gap.** Master's #817 fixed two handlers: the single-shot fire (`plans/single_shot.py`) and the FAILED MOVE line (`plans/step_scan.py`, `move_with_failed_move_pause`), both selecting the cause with `is not None` because a failed `aioca.CANothing` is falsy. The native branch deleted both files. #867 carried the fire half into `plans/strict.py` (`_log_non_frame_failure` / `failure_cause_text`). The move half has no counterpart: on the branch motors move through stock `bps.mv` inside the stock plans, and nothing in `geecs_bluesky` handles a move's `FailedStatus`.

**What the operator sees.** A refused motor `:SP` put (a wrong-catalog PV as in #779, or a nonexistent PV) fails the run; the scan folder's only record is the stop document's `exit_status: reason`, which is the `FailedStatus` repr (`<AsyncStatus ..., done>`). The RunEngine's own `Run aborted` exception log carries the chain, but it fires after the scan log has closed, so it lands in the journal, not the scan log. The PV is not named anywhere in the scan folder.

**Candidate homes.** A RunEngine `msg_hook`/preprocessor on `set` failures is heavy; the natural seam is `CaMotor.set` / `CaSettable.set` logging the CA cause at ERROR before the status fails (the device knows its PV), or the scan-log callback rendering the stop document's exception chain with `str` rather than the repr. Whichever is chosen, `failure_cause_text` in `plans/strict.py` is the one rendering to reuse.

**Pin.** Master's `tests/test_pause_semantics.py::_FalsyMoveError` shape: a falsy cause must appear by `str` in the scan log line.

skbarber:
Verified through the queue after the worker restart (2026-09-14): `mv U_VS1H:Current → 0` on the disconnected device logged `ERROR geecs_bluesky.devices.ca.settable: u_vs1h-current: set Current → 0.0 failed (undulator:u_vs1h:current:SP): CANothing: … Channel write request failed`; the queue history detail carries the cause by `str`, not the status repr.

Issue #854, opened 2026-09-12
GeecsPvaGateway: re-subscribe a live camera subscription that goes quiet past a threshold (the owed half of the boot-order gap; needs a ruling)


skbarber:
## Symptom

After the camera servers' scheduled Windows restart (2026-09-12), the PVA gateway services came back with the machine, but the LabVIEW camera devices are started by hand afterwards. Until the gateway was restarted, image reads and the file plugin's `Capture` on those cameras timed out ("no frame within 8 s" at the first gated arm; the `:connected` view of the device stayed down), although the device's TCP/UDP ports are static and known from the DB.

Found during the phase 2b acceptance (`Planning/native_bluesky/08_gated_batch.md` §5 item 2, PR #850).

## What the gateway does today

The gateway subscribes to each rostered camera at startup. A device that comes up later (or that was restarted uncleanly, as the Windows restart did) is not re-subscribed: the subscription made against a device that was not listening is dead, and nothing re-tries it.

## Asked

- Re-subscribe on a bounded retry when the device is not reachable at startup, and re-subscribe when a live subscription goes quiet past a threshold (the CA gateway's half-open-socket lesson: `cagateway:restart` exists for the same reason).
- Post the device's connected state so the preflight's `gateway_liveness` sees the gap instead of the scan's first arm (relates to #852, the worker-side liveness gate).

## Workaround

Restart the gateway service on that camera server after the LabVIEW devices are up, or start the devices before the gateway. The qserver README's troubleshooting section carries the operator note.

skbarber:
**Re-scoped 2026-09-14.** #901 (GeecsPvaGateway 0.10.0, rolled 9/9) shipped the `<image PV>:connected` state PV (Idle / Disconnected / Connected, MAJOR alarm while down) and the DB endpoint re-resolve at the backoff ceiling. Mechanism note from the PR: the gateway is client-gated, so there was never a startup subscription to retry — the retry loop already existed; what a restart did that the loop did not was learn a new endpoint, and that is now covered.

**What stays open here:** the *re-subscribe when a live subscription goes quiet past a threshold* half. It is not separable from #894: the plugin holds the subscription for the whole capture session and a box ARMED through a long move is ≥ 26 s of legitimate silence inside it. Needs a ruling on the threshold and on whether it may fire mid-session at all. TCP keepalive already resets a dead peer in ~1 min. Leaving open for that decision only.

skbarber:
[...]

Issue #853, opened 2026-09-12
GeecsPvaGateway file plugin: post NumCaptured_RBV = 0 at Capture=1 (areaDetector semantics the stock data logic baselines on)


skbarber:
Found during the phase 2b acceptance (2026-09-12, PR #850, Scan004 of 26_0912).

`file_plugin.py` posts `NumCaptured_RBV` only from `_on_frame` (the session's running count) and from `Rewind`; `_capture_on` never posts a zero. So after a capture session closed at *N*, the PV still reads *N* at the next `Capture=1` until the first frame of the new session is written. areaDetector's NDFileHDF5 resets `NumCaptured_RBV` to 0 at capture start, and ophyd-async's `ADHDFDataLogic` / `StandardDetector._update_prepare_context` baseline `collections_written` on that PV right after the arm — with the stale value the first batch of a gated run counted from *N* (the first step of Scan004 trimmed to 8 = A1's 5 + the quota 3; the later steps, baselined after frames had landed, were exact). Strict carries the same latent race whenever the arming re-push is skipped as stale (no frame written at arm, so no post).

Fix: in `_capture_on`, post `NumCaptured_RBV = 0` (and clear the geometry-independent counters) **before** `Capture_RBV = True`, so a client that baselines on the armed edge reads 0. A worker-side guard rides on #850 meanwhile (`GeecsDetector.zero_count`: a rewind to zero inside the fresh session, which does post the 0), and stays for gateways deployed before this fix.

Deploy = the next fleet roll (share clone pull + `:restart` on the nine boxes).

skbarber:
Verified on hardware 2026-09-14 after the roll, with a PVA monitor on the ModeImager plugin during Scan006 (a 2-shot strict count following Scan005's 3-frame session): `NumCaptured_RBV` 3 → **0 posted at 33.07 s, then `Capture_RBV = True` at 33.07 s** (zero first, then the flip), frames counted 1, 2, then `Capture_RBV = False` at close. `:connected` read Connected during the session and Idle after.

PR #901, opened 2026-09-15
GeecsPvaGateway 0.10.0 + GEECS-Core 0.8.0: file plugin arms on the held frame (#894), NumCaptured_RBV=0 at Capture=1 (#853), :connected + endpoint re-resolve (#854)

skbarber:
## What & why

One PR, one fleet roll: the three GeecsPvaGateway issues found on hardware since the file plugin went live — #894, #853, #854. GeecsPvaGateway **0.9.0 → 0.10.0** (minor: behaviour change + a new PV); GEECS-Core **0.7.0 → 0.8.0** (minor: `pv_naming.CONNECTED_SUFFIX` / `connected_pv`, moved there on review — the roll's `launch.bat` reinstalls GEECS-Core from the share clone on every restart, so it travels with the same roll). No launcher change, no new dependency (`deploy/requirements-fleet.txt` untouched).

**Base branch: `feature/native-bluesky-plans`, not `master`.** The brief said master, but `geecs_pva_gateway/file_plugin.py` does not exist on master (GeecsPvaGateway is 0.6.0 there; 0.7.0–0.9.0 — the plugin, the attribute names, the per-frame scalars — live only on the feature branch, and master has nothing this package lacks there). The fixes cannot be built on master; this targets the branch the fleet's `:hdf1:` PVs come from. Flagging it first so it is cheap to redirect.

### #894 — `Capture=1` arms on the frame the gateway already holds (`file_plugin.py`, `server.py`)

The arm needs *a* decoded frame of the variable for the stream geometry (`ArraySize*_RBV`, `DataType_RBV`), not a fresh push. A box ARMED through a long first move pushes nothing, so the 8 s push-wait failed the run's first `prepare` (Scan002 of 26_0914: a 26 s move, then `hdf-capture didn't match True in 10.0s`).

- `_CameraWorker._publish` keeps the last decoded frame per variable beside the latest-wins slot (`_last_frame`); the plugin gets it as a `last_frame` callable.
- `_capture_on`: with a held frame, post the geometry from it, open the session, `Capture_RBV=1`, complete the put — at once. The held frame is **never written**: the stale watermark is stamped at the arm exactly as before, and the held frame never enters the writer's queue. Only a never-decoded variable still waits for its first push and fails naming the device after `ARM_TIMEOUT_S` — the maintainer-accepted edge. **Its real frequency (review finding 2):** the gateways restart with the camera servers' scheduled Windows restart and nothing holds an image monitor in normal operation, so after every restart cycle each camera is never-decoded until its first plugin session gets a push; the first scan of a cycle whose first step is a long move still fails as Scan002 did. A monitor on the image PV for one gating round-trip in STANDBY seeds the held frame — the #852 preflight's natural job (owed below). Stated so the ruling rests on the number.
- Not read from the DB ROI variables, per the disposition on #894.
- **Judgment call, cheap to veto:** one property moves with this. Before, an unopenable stack (share refused the create) failed the *arming put* with the reason, because the arming frame was written at arm time. With a held-frame arm there is no frame to open on, so the first fresh frame's open failure sets `WriteStatus`/`WriteMessage` and the count never advances — the worker's 3 s shot timeout then names the camera (after the refire quota, so ~N×3 s later than a prepare failure). The raw-session test pins the new surfacing. The alternative (probe-write a file into the run folder at arm) would put a transient file in the scan folder; rejected. Review finding 1 sharpens the cost (the refire gate spends shots on it) — waived with the worker follow-up recorded under Owed.

### #853 — `NumCaptured_RBV = 0` at `Capture=1`, before `Capture_RBV` flips (`file_plugin.py`)

`_reset_session_readbacks()` runs at session creation, before the retain and before any arming frame can be processed: `NumCaptured_RBV`, `ArrayCounter(_RBV)`, `UniqueId_RBV` → 0, `FullFileName_RBV` → `""`. The ordering (zero *then* `Capture_RBV=True`) is pinned by recording the plugin's post sequence. The worker's `GeecsDetector.zero_count` guard stays for gateways rolled before this; nothing worker-side changes.

### #854 — `:connected` per image variable + DB endpoint re-resolve at the backoff ceiling (`server.py`, `config.py`, `__main__.py`, GEECS-Core `pv_naming.py`)

**Mechanism note, honest version.** The issue text says "the gateway subscribes to each rostered camera at startup"; it does not — the subscription is gated per variable (starts with the first PVA client or a plugin session, stops with the last), and while gated on, the existing supervisor already retries connect/subscribe forever with 0.5→30 s backoff. So "a bounded re-subscribe retry at startup" has no startup subscription to attach to; what the existing loop *cannot* do — and what a gateway restart does — is learn a new endpoint from the DB. What ships:

- `<image PV>:connected` — NTEnum `Idle` / `Disconnected` / `Connected` (MAJOR alarm + message while `Disconnected`, timestamped), posted by the supervisor on change: `Connected` after subscribe, `Disconnected` on a failed dial or a drop, `Idle` when gated off (cancelled) or the gateway stops. `Idle` is honest, not "up": gated off means nothing is known — a preflight that wants the verdict for an idle camera holds a monitor on the image PV for one gating round-trip (~1–2 s) and reads this. Per variable because the subscriptions are; deliberately not the bare `<device>:connected` the CA gateway serves over CA for *its* subscription (one flat namespace). Minted in `geecs_core.pv_naming` (`CONNECTED_SUFFIX`, `connected_pv`) beside `HDF_PLUGIN_SUFFIX` — moved there on review finding 3, since its named consumer is the worker-side preflight; `CameraSpec.connected_pv_for` wraps it.
- **Endpoint re-resolve** — the CA gateway's `endpoint_resolver` idiom (#611's adjacent gap, same constants: 10 s off-loop budget, every 10th ceiling cycle ≈ 5 min per down device): once a watched variable's backoff sits at the 30 s ceiling, `GeecsDb.find_device` is re-asked and a moved **port** is redialed promptly. A resolved endpoint on **another host** is logged, never adopted — the served set is host-scoped and a restart re-scopes it. `__main__` passes `GeecsDb.find_device` directly (review finding 4); tests pass a fake.
  **Judgment call:** #854 states the ports are static. The CA gateway's live incident of 2026-08-17 (#611) found a device app re-registering on a different endpoint, and the re-resolve is the one thing a restart does that the retry loop did not, so it is included; ~60 lines, cheap to veto.
- **Owed from #854 (not in this PR):** the "re-subscribe when a live subscription goes quiet past a threshold" half. It is not cleanly separable from #894: the plugin holds the subscription for the whole capture session and #894 shows ≥ 26 s of *legitimate* silence inside it (a box ARMED through a long move; a longer move is longer), so a quiet re-subscribe would fire mid-session — the threshold, and whether it may fire during a session at all, need a ruling. TCP keepalive already resets a dead peer in ~1 min (GEECS-Core 0.3.0), which the supervisor treats as a drop.
- The off-host branch of the re-resolve is log-only and not pinned (indistinguishable from "no resolver" by observation); said here rather than faked.

### Per-concern LOC (from `git diff --numstat`, source only)

| concern | files | ± |
|---|---|---|
| #894 held-frame arm | `file_plugin.py` (arm path, docstrings), `server.py` (`_last_frame`, plugin kwarg) | ~+45/−8 |
| #853 zero at arm | `file_plugin.py` (`_reset_session_readbacks`, call site) | ~+17 |
| #854 `:connected` + re-resolve | `server.py` (+~190/−34 incl. docstrings), `config.py` +11, `__main__.py` +5/−1, GEECS-Core `pv_naming.py` +25 | ~+195 net |
| tests | `test_file_plugin.py` +148/−30, `test_server.py` +89/−3, GEECS-Core `test_pv_naming.py` +10 | |
| docs | GeecsPvaGateway `CHANGELOG.md` +68, `CLAUDE.md` +25/−7, `DEPLOYMENT.md` +8, GEECS-Core `CHANGELOG.md` +10, `docs/geecs_gateway/image_pvs.md` +15, `Planning/native_bluesky/06_pva_file_plugin.md` +19 (§2.3 amended per its own staleness rule) | |

Facility-literal check: no new lab literals (no `192.168.`, no experiment name, no path) outside prose; nothing to relocate. Public-repo hygiene: grep of the added lines clean.

## Checklist

- [x] `poetry version minor` run for GeecsPvaGateway (0.9.0 → 0.10.0) and GEECS-Core (0.7.0 → 0.8.0)
- [x] `CHANGELOG.md` entries added under 0.10.0 and 0.8.0 (Keep a Changelog)
- [x] Tests: exact counts reported below
- [ ] Base branch is `master` — **no: `feature/native-bluesky-plans`, see the note at the top** (the code does not exist on master); master merges are maintainer-only either way
- [x] Public-repo hygiene: no lab accounts/hostnames/home paths
- [x] Adversarial review posted verbatim as a comment; every finding dispositioned in the comment after it (fix commit `3fe82638`, confirmed by the same reviewer)

## Tests

`./scripts/check.sh --base origin/feature/native-bluesky-plans`: lint OK; **GeecsPvaGateway: 51 passed** (was 49 on the branch head); **GEECS-Core `tests/test_pv_naming.py`: 8 passed** (1 new, `test_connected_pv_is_the_image_pv_plus_connected`); `mkdocs build` clean for the touched docs page. 2 new gateway tests, 3 extended:

- `test_file_plugin.py::test_stock_adhdf_data_logic_drives_the_plugin` — extended: after `logic.stop()`, a second stock `prepare_unbounded` with **nothing pushed since** completes within 3 s (#894) and `collections_written` baselines at **0**, not the closed session's 2 (#853); no file in the new run dir.
- `test_file_plugin.py::test_session_semantics_over_raw_pva` — reworked: the held-frame block (a watcher holds the subscription, the camera pushes nothing, the last frame it pushed has a **new geometry** 5×7 → the arm completes < 2 s with `ArraySizeX/Y_RBV = 7/5`, `NumCaptured_RBV` 3 → 0 and the recorded post order has the 0 **before** `Capture_RBV=True`; the held frame is never written — the first fresh push is frame 0 of a (1, 5, 7) stack); the unopenable-stack case in its new form; the accepts-nothing case.
- `test_file_plugin.py::test_never_seen_variable_waits_for_the_first_push` — new: a fresh gateway, nothing ever pushed → the put fails within `ARM_TIMEOUT_S` with `no frame from UC_TestCam image`, `Capture_RBV` stays 0, the retained subscription is released; then two racing `Capture=1` puts arm on the first (stale) push: one session, one connection, geometry from that frame, nothing written.
- `test_file_plugin.py::test_no_plugin_without_h5py` — the provider entries now include `…:image:connected`.
- `test_server.py::test_frames_flow_and_subscription_is_gated` — extended: `:connected` reads `Idle` before any client, `Connected` (severity 0) while watched, `Idle` again after the last client leaves.
- `test_server.py::test_unreachable_device_is_re_resolved_at_the_backoff_ceiling` — new: the DB-held port has no listener; with the backoff bounds shrunk the variable reads `Disconnected` with **severity 2**, the resolver is asked exactly once at the ceiling, the camera is redialed on the resolved port and the state ends `Connected` with severity 0.

**Bite check** (each fix mutated in turn, its tests run, file restored): held frame ignored → 2 failed; zero at arm removed → 2 failed; zero moved *after* `Capture_RBV` → the ordering assertion fails (`assert 13 < 12`); `Connected` never posted → 2 failed; re-resolve disabled → 1 failed; `Disconnected` never posted → 1 failed; never-seen timeout removed → 1 failed. (A first ordering mutant that merely *added* a second zero after the flip did not bite — correctly, the earlier zero was still there — and was replaced by the move.)

## Hardware verification

**OWED.** The lab network is reachable (VPN) but no verification was run: `DEPLOYMENT.md` documents no way to stage one box on a branch short of moving the share clone (the fleet pin — not to be touched) or re-bootstrapping the box from a local clone (removes the service first; not a staging procedure, and the launcher would revert it on the next restart anyway). To verify after the roll, on the UC_ModeImager host, **no scan, trigger box left as-is**:

1. **#894** — `pvput <image PV>:hdf1:FilePath '<UNC run dir that exists>\'`, `pvput …:hdf1:FileName UC_ModeImager`, then `pvput …:hdf1:Capture 1` while the camera has been idle/ARMED for > 30 s: `Capture_RBV` flips to 1 in ≈ 0.1 s (not 8 s, not a timeout), `ArraySizeX/Y_RBV` = the camera's frame, `NumCaptured_RBV` = 0; `pvput …:hdf1:Capture 0`; the run dir has **no** `.h5` (the held frame is never written). Then the real acceptance: the `00_focuscan` preset shape (strict, first step a ≥ 20 mm move) — the first `prepare` completes and the scan runs. Note the never-decoded fallback right after a gateway restart: seed the held frame first with a `pvmonitor` on the image PV in STANDBY (or expect the old 8 s failure on that first arm).
2. **#853** — after any session that wrote N frames, the next `Capture 1` reads `NumCaptured_RBV` 0 immediately (`pvget` right after `Capture_RBV` flips), and a gated run's first batch is exactly the quota (not quota + N).
3. **#854** — `pvget <image PV>:connected` → `Idle`; hold `pvmonitor <image PV>` → `Connected`; stop the camera's LabVIEW device → `Disconnected` (severity MAJOR) within the backoff; start it again on the same port → `Connected` without a gateway restart. The port-move path needs a device that re-registers; log line `endpoint moved … redialing there` is the tell.

## Fleet roll (after merge — nine boxes, executable from here)

Per `GeecsPvaGateway/DEPLOYMENT.md` § Rollout. No launcher change and no new pin, so no per-box stop/copy/start and no `stage_wheels.sh`. GEECS-Core 0.8.0 rides along: the launcher reinstalls the five intra-repo packages from the share clone on every restart.

```bash
# 1. Advance the fleet pin: in the share clone (…\Active Version\GEECS-Plugins),
#    from a machine with the share mounted. Check the checkout first — the plugin
#    line lives on feature/native-bluesky-plans (06 §8 step 1); if the clone is on
#    master this PR is not there until the branch merges to master.
git branch --show-current
git pull
git log --oneline -1          # must show this PR's merge commit

# 2. Restart instances via the :restart PV, ONE host at a time; canary 192.168.6.100 first,
#    then each remaining host in config.ini [pva] addr_list. Never `git pull` mid-restart.
pvput undulator:pvagateway:192_168_6_100:restart 1
poetry -C GeecsPvaGateway run geecs-pva-gateway fleet --experiment Undulator   # wait for that row to read 0.10.0 + a live heartbeat
# repeat for each host token (IP with dots → underscores)

# 3. Done when the fleet probe's version column reads 0.10.0 uniformly (no [WARN] mixed versions)
#    and `pvget undulator:uc_modeimager:image:connected` answers (the new PV exists).
```

Worker side: nothing to restart or configure; `GeecsDetector.zero_count` keeps working alongside the plugin-side zero.

## Owed / follow-ups

- Hardware verification above (three checks) on the UC_ModeImager host after the roll.
- **Worker follow-up (review finding 1, waived here):** when a plugin-backed camera's count times out and its `WriteStatus` reads `Write Error`, raise a non-refirable error carrying `WriteMessage` (the `CONNECTED → GeecsDeviceDownError` rule's shape) — until it lands the strict refire gate spends its quota of shots on a write error before the run fails, and the recorded reason is the frame-drop one. Pre-existing for the (common) stale-arming-frame case; this PR makes it the only case.
- **#852 preflight (review finding 2):** hold a monitor on each preset camera's image PV for one gating round-trip at `stage`, before the bracket arms the box — it reads the `:connected` verdict *and* seeds the gateway's held frame, closing the once-per-restart-cycle recurrence of the never-decoded fallback.
- #854's quiet-threshold re-subscribe: needs a ruling on the threshold and on firing inside a capture session (see the mechanism note).
- #894's "also in this failure" items are not here: the run's recorded reason being the status repr rather than `str` + notes (same family as #868), and the leaked `AsyncStatus` noise.


skbarber:
## Adversarial review (fresh-context agent, read at `75680f91`)

Posted verbatim as the audit record; the disposition of each finding is in the next comment.

---

## Adversarial review of #901 (fresh context, read at `75680f91`)

Checkout read: `.claude/worktrees/pva-arming-roll`. Suite re-run in the package venv: **51 passed in 7.3 s**. Probed p4p 4.2.2 and ophyd-async 0.21.2 from that venv where the diff's claims depended on them. Findings ranked by severity; a short "checked and refuted" list at the end so the dispositions can cite it.

### 1. Medium — the flagged judgment call costs real shots, not just seconds, and the reason never reaches the operator

`GeecsPvaGateway/geecs_pva_gateway/file_plugin.py:641-656`: on the held-frame path nothing is opened at arm, so a stack that cannot be created (share refuses the create, permission, disk full, a directory already named `<device>.h5`) surfaces only as `WriteStatus=Write Error` on the first fresh frame while `NumCaptured_RBV` stays 0. The PR body says the consequence is "the worker's 3 s shot timeout then names the camera (after the refire quota)". Concretely, on the worker side that means:

- `GeecsBluesky/geecs_bluesky/devices/detector.py:859` is the **only** reader of `write_message`, and it runs in `prepare`'s exception path. The trigger/complete timeout paths (`_translate_count_timeout`, `_guarded_complete` at ~880-930) raise a bare `GeecsTriggerTimeoutError("counted no frame …")` without looking at `write_status`.
- `plans/strict.py:115-135`: a `GeecsTriggerTimeoutError` is a *refirable* miss — the gate checks the CA `CONNECTED` PV (which reads `Connected`; the camera is fine), so it **fires again**, up to the quota. Each refire is a physical shot spent on a failure another shot cannot help — the exact class the gate's own docstring reserves `GeecsDeviceDownError` for.
- The run's recorded reason is then "camera X: no frame within 3 s", which is the frame-drop / dead-camera diagnosis #817/#818 worked to stop misattributing.

Honest framing: this gap is **pre-existing** for the stale-arming-frame case (pre-PR, `_on_frame` returned at the stale check before any open attempt, so the old "fail the arming put with the reason" only ever fired when the arming frame was fresh). The PR makes it the only case. Either is acceptable, but the disposition should pick one explicitly:

- (a) keep the PR as is and file the worker follow-up: when a plugin-backed camera's count times out and `write_status == "Write Error"`, raise a non-refirable error carrying `WriteMessage` (mirroring the `CONNECTED → GeecsDeviceDownError` rule) — until it lands, the refire gate burns shots on write errors; or
- (b) veto the judgment call: on the held path, open the stack at arm. The directory is the run folder the worker already claimed (the scan-folder invariant is untouched — the plugin still never `mkdir`s), so the only cost is the §4 "a session that accepts nothing leaves no file" property, which becomes "leaves an empty stack".

### 2. Medium (domain question for the maintainer) — "never decoded" recurs on every gateway restart, not just a "fresh gateway"

`_CameraWorker._last_frame` is filled only by `_publish` (`server.py:432`), i.e. only while a subscription is live *and* the camera pushes. The gateways are NSSM services that come back with the camera servers' scheduled Windows restart (#854's own timeline), and after #892 nothing in the repo holds a PVA monitor on an image PV in normal operation (`grep pva:// GeecsBluesky` finds only the `hdf1:` prefix). So after each restart cycle, **every camera is "never decoded" until its first plugin session receives a push** — and the first scan of that cycle whose first step is a long move fails exactly as Scan002 did (the never-seen fallback is the old 8 s wait; the retry fails the same way because the failed arm decodes nothing either). The #894 disposition accepts "fresh gateway with no frame" as unlikely; the CHANGELOG/PR should state the actual frequency (once per camera per restart cycle, conditional on a long first move) so the ruling is made on that number. Cheap mitigation that stays inside the ruling: the preflight round-trip the PR already recommends for the `:connected` verdict (a monitor on the image PV for ~1-2 s) also seeds `_last_frame` when the box is in STANDBY re-pushing at 1 Hz — make that an explicit requirement of the #852 preflight follow-up (run it at `stage`, before the bracket arms the box). The 06 §2.3 amendment and CHANGELOG say "a fresh gateway, a camera that never acquired", which understates it.

### 3. Medium-low (placement) — `CONNECTED_SUFFIX` belongs in `geecs_core.pv_naming`, next to `HDF_PLUGIN_SUFFIX`

`config.py:50` mints `":connected"` gateway-locally and the PR body defends it as "the same gateway-local composition the identity PVs use". The identity PVs' only consumer is this package's own `fleet` subcommand; `:connected`'s named consumer (#854: "so the preflight's `gateway_liveness` sees the gap") is `GeecsBluesky/geecs_bluesky/qs_client/submit_preflight.py:48`, in a package that per the dependency graph never imports GeecsPvaGateway. `pv_naming.py:102-107` puts `HDF_PLUGIN_SUFFIX` in GEECS-Core for exactly this reason ("the server and the worker's `GeecsHdfIO` both mint it here, so the two sides cannot drift"), and its module docstring forbids hand-assembling a suffix. The first worker-side reader will have to move this constant; move it now (one import change in `config.py`/`server.py`).

### 4. Low (redundancy) — the re-resolve block and its `__main__` wrapper are verbatim copies of the CA gateway's

- `__main__.py:21-31` `_db_endpoint_resolver` == `GeecsCAGateway/geecs_ca_gateway/__main__.py:99-109`. The "light import" buys nothing here: `main()` always calls `PvaGatewayConfig.from_geecs_experiment`, which imports `GeecsDb` before the gateway is built. Passing `GeecsDb.find_device` directly (what the CA gateway's own docstring calls the production wiring) removes the copy.
- `server.py:296-313` `_resolve_endpoint` and the ceiling/holdoff block `365-393` duplicate `GeecsCAGateway/geecs_ca_gateway/gateway.py:~885-960` (same constants, same holdoff dance, same log lines). Both gateways are peers on GEECS-Core; the shared piece (resolver call with budget + holdoff bookkeeping) has a natural home beside `GeecsTcpSubscriber` in `geecs_core.transport`. Not blocking; the PR's own "~60 lines, cheap to veto" applies, and a shared helper would be a GEECS-Core bump.

### 5. Low — every `:connected` post carries `timeStamp = 0`

Probed in the venv: `_connected_value` → `nt.wrap(dict)` → `_annotate(timestamp=None)` leaves `timeStamp` unmarked, so a client reads `secondsPastEpoch 0` (1970) for `Idle`/`Disconnected`/`Connected` alike. Pass `timestamp=time.time()` to `post()` (NTEnum.wrap forwards it). Cosmetic for Phoebus/Archiver; the plugin's existing `_post` has the same gap (pre-existing, not this PR's).

### 6. Low (docs traveling with behavior)

- `docs/geecs_gateway/image_pvs.md:60-65` is the published client-facing PV list for this gateway (it lists the three instance PVs); `:connected` was added to `DEPLOYMENT.md`'s table but not here.
- `file_plugin.py:17-19` still says "One writer thread owns every piece of session state and the file handle, so nothing is shared" — `_last_frame` is now a loop-written, writer-thread-read dict (correctly documented at `server.py:157-159`, contradicted by the plugin's module docstring).

### Checked and refuted (for the disposition table)

- **Frame in the writer's queue at a held-frame arm**: the queue is FIFO; a frame enqueued before the `Capture` command is dispatched first and dropped (`_on_frame`, `session is None`); one enqueued after meets the watermark stamped in `_capture_on`. Identical to pre-PR ordering; the held frame itself never enters the queue.
- **p4p type semantics**: probed — posting a Value wrapped by a *different* `NTEnum()` instance raises `RuntimeError: post() requires the exact type of open()`, so the per-variable `_connected_nt` map is required, not over-engineering; marking `alarm.severity`/`alarm.message` on every post is correct (p4p stores marked fields only).
- **`:connected` name collisions**: 4-component name; the only other 4-component PVs are `<exp>:pvagateway:<host>:{version,heartbeat,restart}`, and `normalize_component` strips `:` from variable/device names, so no DB name can spell `…:connected`. Plugin PVs are 5 components. The guard sees the new entries (`test_no_plugin_without_h5py`).
- **Shutdown ordering**: `GeecsPvaGateway.run`'s `finally` awaits `worker.stop()` (which awaits every cancelled supervisor, so each `finally: Idle` posts on a live server) before `Server.stop()`; the plugin's `_release` scheduled from `plugin.stop` finds no supervisor and posts nothing.
- **Release-then-retain race** (last client leaves as the next arrives in one loop turn; the old task's `finally: Idle` landing after the new task's `Connected`): plausible by reading, but a 20-trial scratch reproduction against the tests' `FakeCamera` never produced it (`0/20`); not reported as a finding.
- **Redial scope**: `find_device(self._spec.device)` by name; `device.name` is the DB key (`get_experiment_devices` joins `device d ON d.name = ed.device`), and an off-host result is refused. It can only ever dial this device's DB endpoint on this host.
- **`FullFileName_RBV` zeroing**: the installed `ADHDFDataLogic.prepare_unbounded` builds the stream URI from `path_info`, never from `FullFileName_RBV`; geometry is read after `Capture_RBV`, which the plugin posts last. Safe.
- **`_last_frame` vs `SharedPV.current()`**: the PV already holds the last posted NTNDArray, but its initial `(1,1)` sentinel would need special-casing; the explicit dict is the cleaner "never decoded" signal.
- Public-repo hygiene, facility literals, scan-folder invariant (the plugin still creates nothing): clean.

skbarber:
## Disposition of the adversarial review (fix commit `3fe82638`; each verdict re-confirmed by the same reviewer against the diff)

| # | Finding | Disposition | Reviewer verdict |
|---|---|---|---|
| 1 | Unopenable stack now surfaces only via the count timeout; the strict refire gate spends shots on it and the reason never reaches the operator | **Waived** (option a): the held-path arm stays as designed; the worker follow-up — a non-refirable error carrying `WriteMessage` when a plugin-backed camera's count times out with `WriteStatus = Write Error` — is recorded under *Owed* in the PR body. Pre-existing for the (common) stale-arming-frame case; this PR makes it the only case | Confirmed (code unchanged at `3fe82638`; the owed line is on the PR) |
| 2 | "Never decoded" recurs once per camera per gateway restart cycle, not just on a fresh gateway | **Fixed** (docs): GeecsPvaGateway `CHANGELOG.md`, `file_plugin.py` docstring, `CLAUDE.md`, `06_pva_file_plugin.md` §2.3 and the PR body state the frequency and name the #852 preflight's image-PV monitor as the seeding mitigation (owed) | Confirmed — "that is the frequency I meant" |
| 3 | `CONNECTED_SUFFIX` belongs in `geecs_core.pv_naming` beside `HDF_PLUGIN_SUFFIX` | **Fixed**: `CONNECTED_SUFFIX` + `connected_pv()` in `GEECS-Core/geecs_core/pv_naming.py`; GEECS-Core 0.7.0 → 0.8.0 + CHANGELOG; `test_pv_naming.py::test_connected_pv_is_the_image_pv_plus_connected`; `config.py` / `server.py` take it from there | Confirmed (no stale importer remains; 8 passed) |
| 4a | `_db_endpoint_resolver` wrapper duplicates the CA gateway's | **Fixed**: `__main__` passes `GeecsDb.find_device` directly | Confirmed |
| 4b | The resolve/holdoff block duplicates the CA gateway's; a shared helper belongs in `geecs_core.transport` | **Waived**: a shared helper refactors the CA gateway too — out of this PR's scope; not filed | Confirmed as a reasonable scope call |
| 5 | Every `:connected` post carried `timeStamp = 0` | **Fixed**: `_connected_value` wraps with `timestamp=time.time()` | Confirmed by re-probe (fresh stamps on both states) |
| 6 | `docs/geecs_gateway/image_pvs.md` lacked `:connected`; the plugin docstring's "nothing is shared" contradicted the held frame | **Fixed**: the docs page gains the `:connected` block; the docstring names the held frame as the one cross-thread read | Confirmed |

Reviewer's "checked and refuted" list (queued-frame ordering at a held arm, p4p exact-type requirement, name collisions, shutdown ordering, release→retain race, redial scope, `FullFileName_RBV` zeroing) stands as recorded in the review comment above.

Process note: for a few minutes #901's description carried another PR's text — a scratchpad file-name collision between two sessions on one machine, republished from a uniquely named file at 02:53:52Z. The reviewer verified the restored body before confirming item 1.

Suites at `3fe82638`: GeecsPvaGateway 51 passed; GEECS-Core `test_pv_naming.py` 8 passed.

skbarber:
Hardware acceptance after the roll is on #894 (Scan005 of 26_0914: 20 mm first move, plugin armed at once, 3 frames in the stack, fleet 0.10.0 ×9).

PR #907, opened 2026-09-15
GeecsScanner 0.7.0 + GeecsBluesky 0.87.1: Start without a preset, a submit queues behind the running plan (#900, #905)

### LINKED ISSUE #900: Web scanner: the New scan form cannot Start without a loaded preset, though presets are optional
Surfaced by the adversarial review of #899 (the #896 change that moved the preset picker out of the headline row on the grounds that presets are optional).

`scanner.js` `updateStartGate`: `valid = ok && !!S.presetDoc` — the form cannot Start unless a preset was loaded at some point in the session. A configs tree with no presets can never submit from the page, and the empty-preset state reads "No presets in the configs tree." beside a Start whose title says "load a preset". Meanwhile #896 says presets cover a small subset of scans and `docs/geecs_scanner/overview.md` says "or compose one".

Ask: decide what an ad-hoc scan needs (a device group at least — `buildPreset` already names it `adhoc`) and gate Start on that instead of on a loaded document. Out of #899's scope; filed so the contradiction has a home.

### LINKED ISSUE #905: Web scanner: a submit while a plan is running is refused and the item removed ("queue start refused: RE Manager is busy")
Observed 2026-09-14 on the live scanner (0.6.0): a `count` was running from the queue and a second `POST /api/submit` (the canonical S1H scan) answered **503 manager_unreachable**: *"queue start refused (Request failed: RE Manager is busy.) — the item was removed again; nothing will run"*.

The submit path adds the item and then calls `queue_start`; while a plan is executing the manager answers *busy* to `queue_start` (the queue is already started), and the scanner treats that as a failure and removes the item. So an operator cannot queue the next scan while one runs — the one thing a queue is for. The Qt console's submit did not have this restriction.

**Fix:** treat a *busy* / *already running* answer to `queue_start` as success when `manager_state` is `executing_queue` (the item is queued behind the running one and will run), and only remove the item when the queue is genuinely stopped and cannot be started. Pin with the stub client: submit while `executing_queue` → 200, item stays queued.

skbarber:
Closes #900, closes #905. Base: `feature/native-bluesky-plans`. GeecsScanner 0.6.0 → **0.7.0** (minor: a behaviour change on the form), GeecsBluesky 0.87.0 → **0.87.1** (patch: a bug in the shared queue client).

## What + why

**#900 — Start without a loaded preset.** `scanner.js` folded "a preset document was loaded at some point" into the form's validity (`recalc`: `valid = ok && !!S.presetDoc`) and `updateStartGate` hovered "load a preset". A configs tree with no presets could never submit from the page, contradicting #896 (presets are optional) and the overview ("or compose one"). Now `valid = ok` — Start depends on form validity (shots ≥ 1, each axis' variable + a positive step) and an idle, reachable manager. `buildPreset` already built the document from the form (`name: "adhoc"`, the table's devices — an empty list is a valid preset per `geecs_schemas.Preset`: "scans with the motors' readbacks only"); `fillFormFromPreset`/`formShape` only *seed* the form and needed nothing; the `preset <name>` provenance note simply stays empty. Nothing in the template changes.

**#905 — a submit while a plan runs queues behind it.** The refusal the issue quotes is produced by **`geecs_bluesky.qs_client.client.ZmqQueueClient._submit_item`**, not by the scanner: the real client adds the item, then calls `queue_start`; while the manager is `executing_queue` the manager answers *RE Manager is busy*, and the client treated that as a start refusal — removed the item and reported "nothing will run" — which the scanner mapped to 503. So the fix lives in the shared client (the one every queue front end uses, GEECS-MCP included; MCP's own "refuse if anything is queued" gate sits above it and is unaffected): on a `queue_start` exception the client reads `status()`; `manager_state` in (`executing_queue`, `starting_queue` — the window before the worker takes the first item, same "busy" answer, review finding 3) → `SubmitResult(ok=True, message="queued behind the running item", item_uid=…)`; any other state → the existing #653 path (best-effort remove, or name the stuck item). **And the page's Start gate no longer closes on a running plan** (review finding 1: every page submit starts at `btn-start`, so the client fix alone was unreachable from the page) — the next scan queues behind it, the existing "Waiting now … Replace it?" dialog covers a third; a *paused* plan still closes it ("resume or stop it first"), because an item added then would be removed by the client. The scanner's idle-only gate for `mv` / `run_action` / the calibration plans (`_require_idle`) is untouched.

## Per-concern LOC

| concern | files | +/− |
|---|---|---|
| #900 Start gate + #905 page gate | `GeecsScanner/geecs_scanner/static/scanner.js` | +9/−4 |
| #900/#905 page test (node harness over idle / running / paused) | `GeecsScanner/tests/test_page.py` | +55 |
| #905 client fix | `GeecsBluesky/geecs_bluesky/qs_client/client.py` | +17/−1 |
| #905 client tests | `GeecsBluesky/tests/qs_client/test_queue_client.py` | +46 |
| #905 scanner test (real client over a stub manager, through `POST /api/submit`; deliberately cross-package) | `GeecsScanner/tests/test_submit_while_running.py` | +101 |
| versions + CHANGELOGs | 4 files | +52/−2 |

1. **Save as preset… follows the same rule** — enabled by a valid form, not a loaded preset (`btn-save-preset.disabled = !valid`, was `!S.presetDoc`). Compose-then-save is the workflow #900 describes; the server still validates the document (`test_save_preset_refuses_an_invalid_document`). Revert to `!S.presetDoc` if Save should stay preset-only.
2. **`executing_queue` and `starting_queue` count as success** (the ruling named the first; the second is the same started queue one request earlier — review finding 3). A *paused* plan (`manager_state == "paused"`) also answers busy; that submit still gets the item removed and a 503, and the page keeps Start closed while paused so the two agree. If pause-then-queue-the-next is wanted, it is one more token client-side plus dropping `paused` from the page gate — not taken here.
5. **A device-less strict count can Start** (review finding 2, a domain question): with no preset and an empty device table, form validity is shots/axes only, per the ruling and the schema ("an empty list scans with the motors' readbacks only"); the shared preflight questions an empty detector list only for gated. If the lab wants that one click away from nobody, the home for the rule is `qs_client/submit_preflight.py` (a question every client sees), mirrored by the page — a follow-up, not this PR.
3. **The fix is in `qs_client`, not `ScannerService.submit`.** The service never sees `queue_start`'s answer (it gets a `SubmitResult`); the ruling's "service layer" is this client. The scanner still maps every non-pending `ok=False` to `manager_unreachable` (503) — a busy-with-stopped-queue refusal is arguably a 409; left as is (out of scope).
4. One extra `status` request on the refusal path only (the success path is unchanged: `queue_get`, `item_add`, `queue_start`).

Facility-literal check: none added. No lab names/hostnames/home paths.

## Tests

Every new test was run against the un-fixed code first and failed there (503 / `ok=False` / `presetDoc` present), then passed with the fix.

- `GeecsBluesky/tests/qs_client/test_queue_client.py::TestQueueStartFailure` +3 (one parametrized over `executing_queue` / `starting_queue`): busy + a started queue → `ok`, `item_uid` set, `item_remove` never called; busy + idle → still removed (the #653 path survives).
- `GeecsScanner/tests/test_submit_while_running.py` (new): the real `ZmqQueueClient` with a stub manager mid-plan (`queue_start` raises *RE Manager is busy*, status `executing_queue`) behind `ScannerService` → `POST /api/submit` **200**, the item stays queued, `/api/status` shows `items_in_queue == 1`. The demo manager never had the bug, hence the real client.
- `GeecsScanner/tests/test_page.py::test_start_gate_needs_a_valid_form_not_a_loaded_preset` (new): `recalc` and `updateStartGate` (extracted from `scanner.js`) contain no `presetDoc`, "load a preset" is gone, and `updateStartGate` runs under **node** over a stub DOM with a valid form and no preset: idle manager → Start enabled, empty hover, Save enabled, provenance note empty; **running** plan → still enabled (fails on the old gate: "a scan is running"); **paused** → disabled with the paused hover.

`./scripts/check.sh --base origin/feature/native-bluesky-plans` (lint + the two affected suites, the CI mapping): lint OK; **GeecsBluesky 493 passed, 4 skipped, 1 deselected**; **GeecsScanner 77 passed, 1 skipped** (the "no inline script" skip, pre-existing). The local GeecsBluesky env lacks the `qs-client` extra so `test_queue_client.py` skips there (pre-existing, CI installs it); run from the scanner's env (which has it): `test_queue_client.py` + the two scanner files = **46 passed, 1 skipped**.

## Hardware verification

**Demo (local, `geecs-scanner --demo`)**: an ad-hoc `count` (no preset name, empty device list) → `POST /api/submit` 200, `planned_shots 3`; a second submit while it ran → 200, `re_state running`, `items_in_queue 1`, the queue panel shows the second `count` waiting. (The demo manager's queue semantics are its own; the real-client path is the pinned test above.)

**OWED post-deploy (live scanner on the worker host, `--experiment`)**: with a `count` running from the queue, Start is enabled on the page; submit the canonical S1H 1D scan from it → expect "Queued: …", the S1H item waiting behind the count, then running when the count ends (Scan N+1 claimed, no "queue start refused" in the scanner journal). Pause the count and check Start greys with "a scan is paused". And: a fresh browser session, no preset loaded, compose a 1D scan → Start enabled, submits. Not done here: the deployed checkout is on 0.6.0 until this merges.


skbarber:
## Adversarial review (fresh-context subagent, commit 13096e36)

Verified first: both new tests bite (with `_EXECUTING_MANAGER_STATE` disabled at runtime the client's `executing_queue` test and the scanner's `test_submit_while_running` fail; the stopped-queue guard passes both ways, as a regression guard should). CI installs the `qs-client` extra, so the client tests run there (a worktree `GeecsBluesky/.venv` without `bluesky-queueserver-api` skips the whole `test_queue_client.py` module silently — pre-existing, `check.sh`'s skip guard only watches aioca/p4p). The manager's `_start_plan` raises the identical "RE Manager is busy." for every non-IDLE state, including `STARTING_QUEUE`. `S.formable` defaults to `true`, so the #900 change does work with an empty configs tree.

### 1. High — the page's Start button is still disabled while a plan runs, so #905's operator symptom is not fixed from the page

`GeecsScanner/geecs_scanner/static/scanner.js` `updateStartGate` (unchanged by the diff):

```js
var busy = !st || !st.connected || st.re_state === "running" || st.re_state === "paused";
btn.disabled = busy || !valid || !S.formable;
btn.title = ... busy ? "a scan is running" : ...
```

Every submit the page makes originates from the `btn-start` click (`submit()` is reached only via that click, `do-submit` and `do-replace`). So with a `count` running, the operator sees Start greyed with "a scan is running" — exactly the #905 story ("an operator cannot queue the next scan while one runs"). The fix in `ZmqQueueClient._submit_item` is reachable only by API callers (curl, a notebook; GEECS-MCP refuses while running by its own gate, `control_tools.py:218`). Consequences:

- The PR's owed hardware step — "submit the canonical S1H 1D scan **from the page** → expect 200 and the S1H item waiting behind the count" — cannot be executed as written; only a `POST /api/submit` can do it. The demo verification in the body was necessarily a POST as well.
- The page is now stricter than its own service and its docs: `docs/geecs_scanner/overview.md` says "The item lands in the queue and, when the worker is idle, runs" and "The page refuses a second waiting item unless you say so", and `test_api.py::test_queue_runs_items_in_order_and_clear_drops_the_waiting` pins a second submit while running → 200. The Start gate is the one place saying otherwise (it dates from the 0.2.0 page commit, before the queue panel existed).

Either drop `running`/`paused` from `busy` in `updateStartGate` (keeping the connected check; the running-item picture is the Now panel's job) and extend the new node harness with `re_state: "running"` → enabled, or scope the PR to the API and stop saying "closes #905" / reword the owed check. If the page gate is opened, note judgment call 2: a submit while **paused** still gets the item removed and a 503 `manager_unreachable` from a manager that is reachable — so either keep `paused` gated on the page or widen the client to `paused` in the same change, not one without the other.

### 2. Medium — the #900 answer lets a strict scan with an empty device table Start; nothing downstream questions it

The issue's ask was "decide what an ad-hoc scan needs (a device group at least — `buildPreset` already names it `adhoc`)". The PR decided: nothing beyond shots/axes. Concrete path: fresh page, no preset, device table reads "No devices — add one." (`noDevicesNote`), mode No-scan, shots 10 → `valid` is true, Start enabled → `count([], num=10, acquisition="strict")`. `submit_preflight.acquisition_refusal` refuses an empty detector list only for **gated** (`no essential triggered device`); strict passes with no question, and no other check (`_check_worker_ready`, `_check_liveness`) looks at the device count. Worker-side nothing refuses either. Result on hardware: a scan number claimed, a folder, ScanInfo and an s-file with no device columns, and ten physical shots fired for nothing. The schema's "an empty list scans with the motors' readbacks only" is a fair reading for a 1D scan (there is a motor readback); for a count there is no readback at all.

This is a domain question for the maintainer, not a code error: is a device-less strict count something the lab wants to be one click away? If not, the home for the rule is the shared preflight (`qs_client/submit_preflight.py`, the seam every client uses) as a question or refusal — not `scanner.js` alone; the page gate can then mirror it (`tableDevices().length` for a count).

### 3. Low — `starting_queue` gets the same "busy" and removes an item that would have run

`bluesky_queueserver/manager/manager.py::_start_plan` sets `MState.STARTING_QUEUE` before the worker takes the plan and raises the identical "RE Manager is busy." for every non-IDLE state. Scenario: client A's `queue_start` is accepted; client B (a notebook, or a second browser) submits inside that window: `queue_get` may already be empty (the item was popped for the worker), the add succeeds, `queue_start` answers busy, `status().manager_state == "starting_queue"` → B's item is removed with "nothing will run" — although the queue is started and would have run it. Narrow, two-client-only, but concrete; the widening is one more token in the state check (`starting_queue` means the same thing as `executing_queue` here). Also worth a line in the docstring: the state read is a second request, so a plan that finishes between the refusal and the read lands on the #653 path — acceptable, the messages there still say what remains queued.

### 4. Low — redundancy/placement: `GeecsScanner/tests/test_submit_while_running.py`

It re-pins the client's behaviour from the scanner package by reaching into `ZmqQueueClient._api` (a GeecsBluesky-private injection point) with a second hand-rolled fake manager (`_BusyManagerAPI` ≈ `GeecsBluesky/tests/qs_client/test_queue_client.py::_FakeManagerAPI`). The scanner's own contribution — an `ok=True` result maps to 200 and `items_in_queue == 1` — is already pinned over the demo by `test_api.py::test_queue_runs_items_in_order_and_clear_drops_the_waiting`, and the client behaviour by the new `TestQueueStartFailure` cases; `GeecsScanner/CLAUDE.md` says the suite runs over `DemoQueueClient`. Defensible as an integration pin, but a future client refactor now breaks a test in the wrong package. Keep or drop is a judgment; if kept, say in its docstring that it is deliberately cross-package.

### Minor wording

After this change the refusal a **third** submit gets while one runs and one waits — "1 item(s) already queued (a failed item returns to the queue front) — clear before resubmitting" — now also describes a legitimately queued-next item. The page's own dialog ("Waiting now: … Replace it with this scan?") is correct; only the API/MCP-facing message misattributes the cause. Not blocking.

### Checked and not a finding

- `S.formable` defaults to `true` (`scanner.js` state init), so an empty configs tree really does reach an enabled Start; the node harness's `formable: true` matches the real default.
- `recalc`'s `valid = ok` covers shots ≥ 1 and, per axis, a positive step and a chosen variable; `buildPreset` with `presetDoc == null` produces a schema-valid document (`trigger_profile: null`, `background` from the mode).
- Save as preset gated on `!valid`: `presetForSave` builds from the form, the name is validated in the dialog, the server validates the document — no new hole; a non-formable loaded preset was re-saveable before and still is (pre-existing).
- Fix placement in `qs_client` is right (the scanner never sees `queue_start`'s answer); `_EXECUTING_MANAGER_STATE` sits beside `_OPENING_MANAGER_STATE`, the only other manager-state constant; no other module hard-codes `executing_queue` outside tests and the demo.
- No facility literals, no scan-folder creation, version bumps and CHANGELOGs travel with the code; CI installs `qs-client` so the client tests are not silent skips there.

skbarber:
## Disposition of the review (fix commit 8f064940, confirmed by the same reviewer)

| # | Finding | Disposition |
|---|---|---|
| 1 | High — the page's Start stayed disabled while a plan runs, so #905 was unreachable from the page | **Fixed** in 8f064940: `updateStartGate` drops `running` from `busy` (null status, disconnected and `paused` still close it; the paused hover reads "a scan is paused — resume or stop it first"). The node harness now runs the gate over idle / running / paused (enabled / enabled / disabled) and fails on the old gate ("a scan is running"). Client and page agree on `paused`: neither queues behind a paused plan. |
| 2 | Medium — a device-less strict count can Start; nothing downstream questions it | **Waived**: the maintainer's ruling is "form validity only", and the schema says an empty device list is a valid preset. Surfaced in the PR body (judgment call 5) as a domain question; if the lab wants the rule, its home is `qs_client/submit_preflight.py` (every client), mirrored by the page — a follow-up. |
| 3 | Low — `starting_queue` answers the same "busy" and got the item removed | **Fixed** in 8f064940: `_QUEUE_STARTED_MANAGER_STATES = ("starting_queue", "executing_queue")`, `in` check, race note in the comment; the client test is parametrized over both states. |
| 4 | Low — the scanner's `test_submit_while_running.py` re-pins client behaviour cross-package | **Fixed** as offered: kept, with a docstring saying it is deliberately cross-package and moves with the `_api` injection point. |
| — | Minor wording — "a failed item returns to the queue front" now also describes a legitimately queued-next item | **Waived**: the text is shared with GEECS-MCP's; the page's own dialog is already right. Its own change. |

### Reviewer's confirmation (verbatim)

Re-derived from `git show 8f064940` in the read-only checkout, not from the summary.

**Finding 1 — CONFIRMED FIXED.** `updateStartGate` now reads `busy = !st || !st.connected || st.re_state === "paused"`; `running` is gone from the gate and the paused hover is "a scan is paused — resume or stop it first". Every page submit still originates from the `btn-start` click, so the #905 path is now reachable from the page. Consistency check on the paused branch: the client (unchanged on this point) removes an item added while `manager_state == "paused"` and the scanner answers 503, so keeping Start closed on paused is the right pairing; page and client agree. Test bites: the commit's node harness over the parent commit's gate gives `running → {start: true, title: "a scan is running"}` and fails the `got["running"] == enabled` assertion; fixed: idle enabled, running enabled, paused disabled with "paused" in the title. Checked for new problems from opening the gate: the click handler's preflight has no idle requirement; the pending-item dialog and `clear_pending` while running are the demo-pinned path; `renderStatus`'s own `running` local (Stop/Pause buttons) is untouched. No leftover "a scan is running" string anywhere.

**Finding 3 — CONFIRMED FIXED.** `_EXECUTING_MANAGER_STATE` is replaced by `_QUEUE_STARTED_MANAGER_STATES = ("starting_queue", "executing_queue")` and the check is `in`; a `None` manager_state (status request failed) is not in the tuple and still lands on the removal path, as before. The client test is parametrized over both states; the `starting_queue` case would fail against the parent commit's equality check by construction. No other module referenced the old constant name.

**Finding 4 — CONFIRMED FIXED as dispositioned.** The docstring now says it is deliberately cross-package, what it adds over the client suite, and that it moves with the `_api` injection point.

**Waived items — noted, no objection to the ruling.** Finding 2: agreed that scanner.js is not the place for that rule. Minor wording: fine.

CHANGELOG wording in both packages matches the code as landed. **Verdict: no surviving findings on 8f064940.**

Issue #906, opened 2026-09-15
CaMotor: a 30 s absolute move budget fails legitimate long stage moves (ModeImager ESP, Scan009 of 26_0914) — bound by readback progress instead


skbarber:
Found 2026-09-14 (Scan009 of 26_0914) while measuring for #854, right after #894's plugin fix removed the *other* long-first-move failure. This is the second mechanism behind "a scan whose first step is a long stage move times out", and it is the one the legacy engine also had (`motor.py` line ~44 cites `U_ModeImagerESP`, Scan034: the move timed out and the scan paused for an operator).

## What happened

Strict `scan` on `U_ModeImagerESP.position_axis_1` from −6 to −25 (19 mm), `UC_ModeImager`, HTU-NoGas:

```
22:51:35.445  u_modeimageresp-position_axis_1: moving Position.Axis 1 → -25.0 (tol=0.015)
22:52:05.447  u_modeimageresp-position_axis_1: set Position.Axis 1 → -25.0 failed (undulator:u_modeimageresp:position_axis_1:SP): TimeoutError: ...
22:52:08.152  scan log: finished (fail): TimeoutError: ca://undulator:u_modeimageresp:position_axis_1:SP: put of np.float64(-25.0) did not complete within 30.0 s
```

The stage **did arrive**: readback −25.005 a minute later. The move simply took longer than `CaMotor`'s `_DEFAULT_MOVE_TIMEOUT = 30.0 s`. Same stage, same day: Scan005's 20 mm move took 18 s, Scan002's 26 s, Scan009's > 30 s (direction and settle differ; the device's `DelayAfterMove` was raised to 800 ms on 09-10). A 30 s absolute budget sits right on this stage's normal range, and any farther start point fails deterministically. The gateway's `:SP` readback still showed −6.0 after the run (the put whose ack arrived after the budget), which is misleading beside a readback of −25.

(#902's logging did its job here: the PV is named in the scan log and the stop reason is text.)

## Why a bigger constant is the wrong fix

Any absolute budget is either too short for the slowest legitimate move or makes a genuinely stuck axis take minutes to fail. The information that distinguishes the two is already polled: the **readback**. A move that is still making progress is not a timeout.

## Proposed

[...]

PR #908, opened 2026-09-15
GeecsCAGateway 0.21.0: a :SP put waits minutes for the device's executed reply (#906)

skbarber:
Closes the **CA gateway half** of #906 (the `CaMotor` half is a separate PR on the native-bluesky branch). Package: GeecsCAGateway 0.20.2 → **0.21.0** (minor: observable put behavior changes; `PV_CONTRACT.md` §2 updated with its pinned tests).

## What

A GEECS set has **two device replies**, and this PR raises the wait for only the second:

| reply | where | budget | this PR |
|---|---|---|---|
| 1. command ACK | command port, `GeecsUdpClient._ACK_TIMEOUT` | 1.5 s | **unchanged** — a dead device (no ACK) or a refusing one (ACK not `accepted`) still fails the `:SP` put in ~1.5 s |
| 2. *executed* reply — the device's verdict once the set converged or failed | cmd_port + 1, the `timeout=` of `udp.set(...)` = `exe_timeout` of `_exchange_inner` | was 30 s | **`_SET_REPLY_CEILING_S = 600.0`** (`gateway.py`; the `GeecsCaGateway(set_timeout_s=…)` default) |

Why 600 s: the ceiling is not a liveness signal (the ACK settles liveness; the reply's own status carries the verdict) — it is only the point past which an acknowledged, still-executing move's reply is dropped unread. It therefore must be **longer than any client's own wait**: the CA client decides how long to wait for a put, and a reply the gateway has already discarded can never reach a client still waiting for it. The #906 design of record gives GeecsBluesky's `CaMotor` a 300 s hard ceiling on its reply wait (bounded by readback progress) in its companion PR — the shipped `_DEFAULT_MOVE_TIMEOUT` is still 30 s — so 600 s is twice that. The caproto put is async, so waiting costs the CA server nothing.

**Flagged consequence (review finding 1, documented in the constant and `PV_CONTRACT.md` §2, number waived per the #906 ruling):** the UDP client serialises exchanges per device, so an acknowledged set whose reply *never* comes — a datagram lost on the wire, or a device that ACKs and then hangs — holds that device's setpoint channel for the ceiling; later `:SP` puts to the same device queue behind it and fail at their own client budgets (the device itself would have rejected them at once). Readbacks are unaffected. A smaller default is one constant if you'd rather bound that blast radius.

**Scan009 of 26_0914** (the incident): a 19 mm ModeImager move answered `no error` at 32 s; the 30 s budget failed the put at 30 s and then the transport discarded the device's clean reply as a stale datagram (`Discarding unsolicited/stale UDP datagram … (no exchange in flight)`).

**Not changed, per the maintainer's ruling on #906:** how an error reply is classified. The device's verdict — `no error` or an error (the check-values error included) — still decides the put; `_parse_exe_response` in GEECS-Core is untouched.

**Late-reply path check (GEECS-Core `udp_client.py`, read, not changed):** the exe future is armed before the command is sent and disarmed only in `_exchange_inner`'s `finally`, i.e. when the exchange ends — success, ACK rejection, or the exe timeout. So for an acknowledged set, the "no exchange in flight" discard is reachable only after the ceiling has fired (or for a duplicate reply after success). Pinned by the second new test below.

## Per-concern LOC (original commit 293 +, 38 −; review fixes a6220651 + f0f2e347 add ~80 lines of prose, a `--set-timeout` validator, and the test tightenings listed in the review comment)

| concern | files | LOC |
|---|---|---|
| the ceiling: constant, comment, docstrings | `gateway.py` | +40 / −14 |
| `--set-timeout SECONDS` CLI flag → `_run(set_timeout_s=…)` → `GeecsCaGateway` | `__main__.py` | +19 / −3 |
| tests (2 gateway, 1 entrypoint; 3 existing extended) | `tests/test_gateway.py`, `tests/test_entrypoint.py` | +178 / −11 |
| contract + docs + changelog + version | `PV_CONTRACT.md`, `CLAUDE.md`, `CHANGELOG.md`, `pyproject.toml` | +56 / −10 |

**Judgment call, cheap to veto:** the `--set-timeout` CLI flag. The gateway's settings pattern for such values is CLI flags (`--log-level`, `--all-variables`), so that is where it went. I did **not** add a `site.env` key / unit-template hole (`GEECS_GATEWAY_EXTRA_ARGS`): `render_units.sh` treats every site.env key as required, so a new key would break existing host profiles, and no site needs a different value today. Drop the flag if you'd rather the constant be the only knob.

**Facility-literal check:** none added (no IPs, experiment names, paths, or timezones outside docstrings/examples).

## Tests

- GeecsCAGateway suite: **139 passed** (was 136; +3 new), `./scripts/check.sh` rc=0 (lint + the package suite the way CI runs it). `--all` not run locally — this worktree has no envs for the other 13 packages and nothing imports the changed modules (GeecsPvaGateway imports only `config.py`, untouched); CI runs the full matrix.
- New pins (`tests/test_gateway.py`), driven end to end through the real `GeecsUdpClient` with fake transports and a **virtual clock** (`loop.time` shifted, so the 32 s elapses in 0.3 s of wall time):
  - `test_acknowledged_set_replying_after_30s_completes_the_put` — ACK at once, `no error` at 32 s → the put completes, `:SP` stores −25.0, no alarm.
  - `test_reply_past_the_ceiling_is_the_only_stale_discard` — the same sequence with `set_timeout_s=30.0` reproduces Scan009 exactly (`no exe response within 30.0s`, `:SP` unstored), and asserts no "no exchange in flight" discard at 29 s and one after the ceiling.
  - `test_entrypoint.py::test_set_timeout_flag` + the flag's flow through `main()` and `_run()`.
  - `test_setpoint_write_uses_move_budget_timeout` floor is now `> 300.0` (the companion PR's CaMotor hard ceiling per the #906 design; not imported — the gateway has no dependency on the worker). **Owed:** the client-side ordering pin when that PR lands.
  - `test_entrypoint.py::test_set_timeout_flag` uses a non-default value (900) and rejects `0`, `-5`, `soon`.
- **Bite proof:** with `_SET_REPLY_CEILING_S` forced to 30.0 the completion test fails with `GeecsConnectionError('… no exe response within 30.0s')` (the Scan009 message) and the floor test fails `30.0 > 300.0`; forced to 300.0 the floor test fails `300.0 > 300.0`. Restored to 600.0: all green. An earlier draft of the virtual-clock test passed vacuously (the clock moved before the client armed its exe timer); that ordering is now fixed and documented in the `_ack` helper.

## Hardware verification

**Not applicable pre-deploy** — nothing in this PR can be exercised against hardware until the gateway host runs it. **OWED after deploy:** one strict scan on `U_ModeImagerESP.position_axis_1` whose first step is a ≥ 19 mm move (the Scan009 shape). Expected: the `:SP` put completes when the device answers (30–45 s), the gateway journal shows **no** `Discarding unsolicited/stale UDP datagram` line for that put, and the `:SP` readback equals the commanded value afterwards (Scan009 left it at the previous setpoint). A move ≥ ~25 mm may still end with the device's own check-values error — that is the device's timeout and the maintainer's ruling keeps it an error; it is not this PR's concern.

### Deploy steps (gateway host, after merge to master)

This deploy also carries **GEECS-Core 0.4.1's float-formatting fix (#897)** — the gateway host is still on GEECS-Core 0.4.0; `geecs-core` is a `develop = true` path dependency, so pulling the checkout and re-installing is enough for both.

```bash
# on the gateway host, as the service account
cd ~/gateway-checkout && git checkout master && git pull
cd GeecsCAGateway && poetry install          # picks up geecs-core 0.4.1 from ../GEECS-Core too
poetry version                                # expect geecs-ca-gateway 0.21.0
# restart = upgrade (DEPLOYMENT.md): the gateway's own restart PV, exit 86 → systemd relaunch
caproto-put <experiment>:cagateway:restart Restart   # or: sudo systemctl restart geecs-ca-gateway
journalctl -u geecs-ca-gateway -n 20                  # "serving N PV(s) across M device(s)"
```

Optional per-host override: append `--set-timeout SECONDS` to the unit's `ExecStart` (keep it above every client's wait; `DEPLOYMENT.md` flag table).

## Follow-ups (not this PR)

- Move the hand-fed `GeecsUdpClient` test double (`_hand_fed_udp` / `_VirtualClock`) into `geecs_core.testing` when the CaMotor companion PR needs the same "reply at 32 s" device (review finding 6, waived here to keep this a one-package PR).
- `GeecsDevice.SET_TIMEOUT_S = 30.0` in GEECS-Core's sync client carries the same "blocks until convergence" doctrine and now disagrees with the gateway (30 vs 600 s) under the same #906 reasoning — reviewer's out-of-scope note.


skbarber:
## Adversarial review (fresh-context subagent, per `/land`) — report and dispositions

Reviewed at ccd99da9 (the original commit); fixes in a6220651 and f0f2e347, each confirmed by the same reviewer. Suite 139 passed throughout.

### Reviewer's report (verbatim)

Read against GEECS-Core `udp_client.py`, caproto 1.3.0's server, GeecsBluesky's CA devices (`motor.py`, `gateway_put.py`, `shot_controller.py`), `geecs_core.testing.fake_device_server`, and the gateway's DEPLOYMENT/PV_CONTRACT/docs pages.

**1. High — the ceiling is also a 10-minute per-device *availability* budget; the PR's "waiting costs the gateway nothing" is only true for caproto.** `GeecsUdpClient._exchange` runs every exchange under `self._lock` (one client per device), and caproto spawns each write as its own task (`WRITE_LOCK_TIMEOUT` is 1 ms, so reads are not stalled but writes pile up). So an ACKed set whose exe reply never arrives — one lost UDP datagram (no retransmit exists), or a device that hangs mid-move — now holds that device's lock for 600 s instead of 30 s, and every later `:SP` put to **any variable of that device** queues behind it and fails at its client's own budget (`GatewaySetpointPut` default 10 s; ShotControl's `CaPutSetter` 10 s; CaMotor 30 s). Concrete scenario: ShotControl fires the trigger box through `CaPutSetter` → gateway `:SP`. One dropped exe reply from the trigger box → every fire for the next 10 minutes times out at 10 s; the scan pauses/fails and stays failed until the ceiling elapses. Same for a multi-axis ESP: axis 1's lost reply blocks axis 2/3 sets for 10 minutes. The only early escape is `_rebuild_udp`, which runs only when the TCP subscription's reconnect backoff hits its ceiling — an unrelated event. The "must exceed every client's wait" argument only needs > 30 s today (see #2); 600 s buys nothing against clients that exist while raising the blast radius 20×. Neither `gateway.py`'s comment nor `PV_CONTRACT.md` §2 mentions the lock. At minimum the contract should say "a set that never replies holds that device's setpoint channel for the ceiling"; a smaller default (still several × the 32 s incident) would keep the fix while bounding this.

**2. Medium — the contract, comment, changelog, and the `> 300.0` pin cite a CaMotor that does not exist anywhere.** `PV_CONTRACT.md` §2, the `gateway.py` block comment, `CHANGELOG.md`, and `test_setpoint_write_uses_move_budget_timeout` all state CaMotor "waits 90 s for the reply … under a 300 s hard ceiling". `_DEFAULT_MOVE_TIMEOUT = 30.0` on `origin/master`, on `origin/feature/native-bluesky-plans`, and on `origin/fix/motor-tolerance-from-db`; no CaMotor-half PR exists yet. The client API contract would merge describing a client behavior that is not written, and the test pins a number borrowed from it — if that PR lands with a different ceiling, three prose sites plus a test drift silently.

**3. Medium-low — `--set-timeout` tests pass if the flag were ignored.** `test_set_timeout_flag` and the extended `main()` flow test pass `--set-timeout 600` and assert `600.0` — which equals `_SET_REPLY_CEILING_S`. If `main()` dropped `set_timeout_s=args.set_timeout`, both assertions still hold.

**4. Low — the 29 s "not discarded" assertion is vacuous.** In `test_reply_past_the_ceiling_is_the_only_stale_discard`, `"no exchange in flight" not in caplog.text` at 29 s cannot fail: no datagram is delivered before 32 s. If `_exchange_inner` disarmed the exe proto right after the ACK, this test passes unchanged. The completion test does catch that bug, but the "only" this test's name/docstring/contract row claims is not pinned by it.

**5. Low — contract files not traveling with the behavior.** `DEPLOYMENT.md`'s CLI flag table lacks `--set-timeout`; `docs/geecs_gateway/client_overview.md` still says "(30 s default budget)"; no floor on `--set-timeout` (`0`/negative makes every put fail immediately after ACK).

**6. Placement (low) — the hand-fed client belongs in `geecs_core.testing`.** `_hand_fed_udp` / `_FakeUdpTransport` / `_VirtualClock` reach into GEECS-Core privates (`_Oneshot`, `_cmd_proto`, `_exe_proto`). `FakeGeecsServer` has no exe-delay knob, so this is not a duplicate — but it is a GEECS-Core test double living in a downstream package, where a rename of `_Oneshot` breaks the gateway's suite rather than Core's. The CaMotor-half PR will want the same "reply at t=32 s" device.

**Redundancy checked, no finding.** `GeecsDevice.SET_TIMEOUT_S = 30.0` (`GEECS-Core/geecs_core/client/geecs_device.py`) is the sibling sync-client constant with the same "blocks until convergence" doctrine — a different client, not a duplicate, though it now disagrees with the gateway (30 vs 600) under the same #906 reasoning; out of scope. No existing `--set-timeout` or site.env knob; GeecsPvaGateway performs no sets. Facility-literal and scan-folder invariants: clean.

### Dispositions

| # | disposition | where |
|---|---|---|
| 1 | **Waived as to the number** — 600 s follows the maintainer's ruling on #906 (ceiling to minutes; the companion CaMotor PR gets a 300 s hard ceiling, and the gateway's must comfortably exceed it so a reply is never discarded while a client is still waiting). **Fixed as to documentation** — the per-device lock consequence is now stated in the constant's comment and `PV_CONTRACT.md` §2, including the reviewer's correction: queued puts the device would have rejected at once instead die at their client budget, and a device that ACKs then never replies (not only a lost datagram) stalls that device's sets for the ceiling; readbacks are unaffected. Cheap to veto: a smaller default is one constant. | a6220651, f0f2e347 |
| 2 | **Fixed** — all four sites cite the #906 design of record (300 s hard ceiling in the companion PR) and state the shipped `_DEFAULT_MOVE_TIMEOUT` is still 30 s; the `> 300.0` pin stays as the contract number. **Owed**: the client-side ordering pin when the companion PR lands. | a6220651 |
| 3 | **Fixed** — tests use 900 (non-default) and add rejection of `0`, `-5`, `soon`. | a6220651 |
| 4 | **Fixed** — at 29 s the test asserts the exe future is still armed (`disarm()` sets it to `None`), which the reviewer confirmed genuinely pins arming. | a6220651 |
| 5 | **Fixed** — `--set-timeout` row in `DEPLOYMENT.md`; `client_overview.md` updated; `_positive_seconds` argparse type rejects `<= 0` and non-numbers. | a6220651 |
| 6 | **Waived** — moving the hand-fed client into `geecs_core.testing` widens this PR to a second package bump; follow-up when the CaMotor companion PR needs the same "reply at 32 s" device (a GEECS-Core patch). | — |

Reviewer's verdict on a6220651: all six dispositions hold. The wording it suggested for #1 landed in f0f2e347 (prose only; 139 passed, `check.sh` rc=0).

PR #909, opened 2026-09-15
GeecsBluesky 0.88.0: CaMotor waits for the device's reply; the readback stall rule is the only client-side timeout (#906)

skbarber:
The worker half of #906 (the CA gateway's `set_timeout_s` is the sibling PR). GeecsBluesky 0.87.0 → **0.88.0** (minor: behaviour change).

## What

`CaMotor._set_and_wait` no longer caps a move at 30 s. The GEECS set's **executed reply is the verdict**: `no error` completes the move (then the existing readback-to-tolerance confirm); an error reply — the device's own check-values timeout included, which the maintainer ruled is a device setting adjusted in LabVIEW and never overridden here — or a put refused inside the command-ACK window fails the move **the moment it lands**, never delayed by the grace. `_DEFAULT_MOVE_TIMEOUT` and the `move_timeout` keyword are gone. Three named bounds replace them (module constants, each a constructor keyword so a test can shrink them):

| bound | value | meaning |
|---|---|---|
| `REPLY_WAIT` | 90 s | the reply is waited for outright (the hexapod answers a finished move slowly). At the threshold with no reply: a readback **within tolerance of the target** completes the move as a lost reply (WARNING with the `:SP` PV); a readback **still moving** keeps waiting |
| `PROGRESS_GRACE` + `STALL_TIMEOUT` | 5 s + 10 s | the readback stall rule — the only client-side timeout left. From the grace on, the readback is polled beside the put purely as a stall detector: no movement beyond the tolerance for 10 s while short of the target → `GeecsMotorTimeoutError` naming PV, target and current, before or after `REPLY_WAIT`. A readback sitting *at* the target is never a stall (it is a device that has not answered yet) |
| `REPLY_CEILING` | 300 s | the put's hard bound — a gateway that never answered; fails as the put's own timeout naming the PV |

One hardening found by the mocks and worth keeping: the stream's arrival and the reply can land in either order, so at the lost-reply decision the put gets one poll interval (0.1 s) to land before the reply is declared lost — otherwise a reply arriving a loop turn after the last readback update past `REPLY_WAIT` logged a spurious "reply lost" WARNING (the move still completed).

**After the reply** the confirm is bounded outright — `PROGRESS_GRACE + STALL_TIMEOUT` (15 s) from the reply — and a NaN readback never counts as progress, so a readback the stall rule cannot see as stalled (a NaN, a supply's ripple wider than its DB tolerance) fails the move instead of holding the scan forever (review finding 1, 7c6893c4). `within_tolerance` in motor.py is now the one tolerance test; `CaConfirmSettable`'s analog match calls it.

Failure path unchanged otherwise: a failed put propagates as the #868/#902 ERROR line, no pause-and-retry. `GeecsMotorTimeoutError` gains `replied` and says which phase stalled ("with no reply from the device" / "after the device replied").

## Why

Scan009 of 26_0914: a 19 mm ModeImager move the device completed with `>>no error` at 32 s, killed at 30 s by our cap (and the gateway's — the sibling PR). The design of record and its amendments are the last comments on #906.

## LOC

- `devices/ca/motor.py` (+~180/−60): the wait, the three constants, `within_tolerance`, the docstrings (the two-reply model: command ACK vs executed reply)
- `devices/ca/confirm.py` (+2/−3): `_matches` over `within_tolerance`
- `exceptions.py` (+23/−12): `GeecsMotorTimeoutError` message/`replied`
- `tests/test_ca_devices.py` (+360): the #906 block + 4 existing motor tests moved off `move_timeout`
- `gateway_put.py`, `tests/test_gateway_put.py` (2 docstring lines each): the `move_timeout` mention
- CHANGELOG + pyproject

## Tests

`GeecsBluesky/tests/test_ca_devices.py`: **36 passed** (11 new; the motor selection runs ~19 s — timings at 1/10 scale, grace 0.5 s / stall 1.0 s / reply wait 9 s). Full local `./scripts/check.sh --all`: see the CI/check comment below.

Each new test was shown to fail under a deliberate break of the code (six breaks, restored after each):

| break | tests that failed |
|---|---|
| a 3 s (scaled 30 s) cap on the put | (a) late `no error` reply; (d) progress past `REPLY_WAIT` |
| progress never resets the stall clock | (a), (d), (e) lost reply, stall-after-progress |
| put failure deferred until after the grace | (c) error reply while advancing; rejected command |
| lost-reply completion removed | (e) |
| stall rule removed | (b) never answered; stall-after-progress (and the pre-existing stuck test hangs) |
| a readback sitting at the target counted as a stall | (e) |
| post-reply bound removed / NaN counted as progress | NaN-after-reply, ripple-after-reply |

The (a)–(f) cases: (a) `no error` at "45 s" with the readback advancing completes on the reply path, no warning; (b) never answers, readback stalled → `GeecsMotorTimeoutError` at grace+stall, ERROR line names the `:SP` PV, exception names device/variable/target/current; (c) the device's check-values error while the stage is moving → the put's exception at once (< grace); a rejected put (ACK-window failure) → at once (< one poll tick), not grace+stall; (d) continuous slow progress past `REPLY_WAIT` with no reply keeps waiting, completes on the reply; (e) no reply ever, readback at the target → completes at `REPLY_WAIT` with the WARNING naming the PV; (f) past `REPLY_WAIT`, still moving, no reply → the ceiling fails the put naming the PV; progress then a stop → fails one stall window after the last step.

## Hardware verification (2026-09-15, lab quiet)

From a staged clone on the worker host (its own venv, the qserver checkout untouched), a bare `make_run_engine(claim=False)` over the real `GeecsNamespace` (`U_ModeImagerESP.position_axis_1` binds as `CaMotor`, DB tolerance 0.015, bounds 90/5/10/300 s), `bps.mv` for a 19 mm move each way:

| move | result |
|---|---|
| −5 → −24 | the **CA gateway refused the put at 30.0 s** (`CANothing: undulator:u_modeimageresp:position_axis_1:SP: Channel write request failed` — its own `_SET_EXE_TIMEOUT = 30 s`, still deployed until the sibling PR) with the readback at −22.8 and still moving; **our timeout did not fire** (the failure is the put's, logged by `_set_logged` with the PV as #868 intends). Readback **−24.0** a few seconds later: the device completed the move the gateway had refused |
| −24 → −5 (restore) | identical: gateway refusal at 30.1 s, readback −7.765 and moving; **−5.0** afterwards. Stage restored |

So on hardware the worker side now outlives the old 30 s (the gateway's clock is the one that fires — the expected gateway-side limit recorded here); the worker-side proof past 30 s is the mock tests until the sibling PR deploys. **OWED** once the gateway PR is live: repeat the −5 → −24 move and see `COMPLETED in ~35 s` with no WARNING (the reply path), and the `:SP` readback following the put.

## Notes for the reviewer / maintainer

- `GeecsCAGateway/geecs_ca_gateway/gateway.py`, `PV_CONTRACT.md` and `tests/test_gateway.py` still *mention* `CaMotor._DEFAULT_MOVE_TIMEOUT` (30 s) as the value the gateway's set budget matches — those are the sibling PR's lines to rewrite (raising `set_timeout_s` past `REPLY_WAIT`); left untouched here on purpose.
- Facility literals: none added outside tests/docstrings (`U_ModeImagerESP`, `Undulator` in the test fixtures follow the file's existing pattern).
- Judgment call, cheap to veto: the 0.1 s "a reply landing now still wins" grace at the lost-reply decision (above). Without it the behaviour is still correct, just a spurious WARNING in a race.
- **Flag for the maintainer (review finding 2, waived as your amendment):** the lost-reply completion cancels *our* caput, not the gateway's UDP exchange — the gateway's per-device lock stays held until the device's reply or the gateway ceiling, so the next point's put queues behind it and a late error reply is discarded. If that outweighs a scan that continues, the alternative is "at target, no reply → keep waiting up to the ceiling" (remove the branch). Details in the review disposition comment.
- `CaPseudoMovable`'s component puts still run under ophyd's 10 s default — pre-existing, filed as #910.


skbarber:
## Local checks (`./scripts/check.sh --all`, worktree of e1336911)

- lint (pre-commit: ruff, ruff-format, pydocstyle, lock-path hooks): passed at commit
- root-tests: 67 passed, 1 skipped, 63 deselected
- **GeecsBluesky** (the only package this diff touches): **502 passed, 4 skipped, 1 deselected** in 203 s — `tests/test_ca_devices.py` alone: 34 passed (9 new)
- the other suites (ImageAnalysis … GeecsWebTheme) errored at *collection* in this fresh worktree — `No module named 'geecs_schemas'` / a 3.10 fallback: no root or per-package venv there, an environment condition, not a code one; none of those packages is in the diff. CI's unit-tests job is the full run and is linked below when it lands.

skbarber:
## Adversarial review (fresh-context subagent, reviewed e1336911; fix confirmed by the same reviewer at 7c6893c4)

Reviewer's report, in full:

> Files checked for overlap: `devices/ca/{motor,settable,confirm,pseudo,gateway_put}.py`, `namespace.py`, `exceptions.py`, GEECS-Core `transport/udp_client.py`, GeecsCAGateway `gateway.py` and the sibling PR #908 diff, both CLAUDE.md files.
>
> **1. Medium — after a `no error` reply, the confirm loop has no bound at all; a NaN or jittering readback hangs the scan instead of failing it.** Once `replied` is True the only exits are `at_target` or the stall rule; the stall rule is defeated by any readback change larger than the tolerance inside each window, and by NaN (`_within(nan, x)` is always False, so every tick counts as progress). Reproduced on the mock backend: NaN readback after the reply — still waiting after 4 s (13× grace+stall); readback flipping 0.0 ↔ 0.02 every 0.13 s — still waiting after 4 s. Lab shape: `GeecsNamespace` builds a `CaMotor` for every numeric settable with a positive DB tolerance, magnet currents included, so a supply whose ripple exceeds its DB tolerance and which the device declares converged sits silently with no ERROR line, where 0.87.0 failed it at 30 s.
>
> **2. Medium-low — the 90 s "lost reply" completion is a third verdict source the design of record does not have, and it leaves the gateway's device lock held.** Cancelling the caput does not cancel the gateway's UDP exchange: `GeecsUdpClient._exchange` holds the per-device lock until the reply or the gateway ceiling. The next scan point's put queues behind it at the gateway, the readback does not move, and `CaMotor` fails that next point at grace+stall with "readback stalled … with no reply" while the stage sits at the previous point — a misattributed failure one step later. Any reply slower than 90 s is discarded; if it is an error, the verdict is lost.
>
> **3. Low — scope gap: `CaPseudoMovable` component puts still die at 10 s** (`pseudo.py:92` builds the put with `timeout=None` → ophyd-async `DEFAULT_TIMEOUT`). Pre-existing.
>
> **4. Low — one-tick race**: if the put fails between the top-of-loop `put.done()` check and the stall `raise`, `GeecsMotorTimeoutError(replied=False)` propagates instead of the device's error reply, and the task's exception is never retrieved.
>
> **5. Low (redundancy) — `CaMotor._within` duplicates `confirm._matches`** (same ULP-slack formula); one module-level helper in motor.py used by both would be the natural home.
>
> Nits: ACK window "1.5 s" in one docstring, "~2 s" in another; the `REPLY_CEILING < gateway ceiling` ordering is docstring-only in two packages; the constants test is a change detector.
>
> Checked and not flagged: no scan-folder writes, no facility literals beyond the file's existing fixtures, version + CHANGELOG travel with the change; `move_timeout` has no callers outside GeecsBluesky; failure propagation and put cancellation correct; the refused-put and error-reply tests fail under the listed breaks; placement right.

## Disposition

| # | disposition |
|---|---|
| 1 | **Fixed** in 7c6893c4 — the post-reply confirm is bounded at `progress_grace + stall_timeout` from the reply (`GeecsMotorTimeoutError(replied=True)`); `within_tolerance` returns False for a non-finite readback and NaN never re-anchors. Pinned by `test_motor_nan_readback_after_the_reply_is_bounded` and `test_motor_ripple_wider_than_tolerance_after_the_reply_is_bounded`, each shown to fail with the bound removed. Reviewer re-ran an independent probe against both commits: hung on e1336911, raises at grace+stall on 7c6893c4 — **confirmed** |
| 2 | **Waived** — the lost-reply completion at `REPLY_WAIT` is the maintainer's explicit amendment to the design (readback within tolerance at the threshold → complete, WARNING with the PV). **Flag for Sam, cheap to veto:** the reviewer's consequence is real — cancelling our caput does not cancel the gateway's UDP exchange, so after a lost-reply completion the gateway's per-device lock stays held until the device's reply or the gateway ceiling, and the *next* point's put queues behind it; a reply that eventually arrives as an error is discarded. If that matters more than a scan that continues, the alternative is "at target, no reply → keep waiting up to the ceiling", which the loop already supports by removing the branch. Reviewer accepted the waiver as a maintainer call |
| 3 | **Waived** — pre-existing, outside this PR; filed as #910 |
| 4 | **Fixed** in 7c6893c4 — a put that finished on the stall tick is raised as the verdict before the stall error; a finished, uncancelled put's exception is retrieved in `finally` — **confirmed** |
| 5 | **Fixed** in 7c6893c4 — one module-level `within_tolerance` in motor.py; `confirm._matches` calls it — **confirmed** |
| nit | ACK window harmonised to 1.5 s — **confirmed**. Constants test kept as the change detector it is (a reinstated cap has to edit it) |

Reviewer's two cosmetic residues after the fix: (a) a `no error` reply landing on the stall tick reported "with no reply from the device" — **fixed** in 6e2a3096 (one line); (b) the post-reply bound reuses the "readback stalled" wording for a readback that rippled rather than stalled — **waived**, wording only; the `replied=True` phase text already distinguishes it.

Test count at head: `tests/test_ca_devices.py` 36 passed (11 new).

skbarber:
## External review (Codex) finding — fixed in `ab630071`, confirmed by a fresh-context reviewer

**Finding (P2):** `within_tolerance(current, anchor, tol)` rejected a non-finite *current* but not a non-finite *anchor*; with a NaN first readback the stall clock restarted at the readback's arrival rather than at the end of the grace, so a stuck axis failed late.

**Fix:** a non-finite readback is never adopted as the anchor; the first *finite* readback becomes the anchor without counting as progress; `within_tolerance` rejects a non-finite value on either side. Regression pin `test_motor_nan_first_readback_then_stalled_fails_at_grace_plus_stall` (NaN → fixed finite short of target, no reply): **fails on the previous head** (2.05 s, scaled), passes on the fix (1.53 s = grace + stall).

**Confirmation (fresh context, arithmetic re-derived by hand):** closes every path — NaN first, NaN after a finite anchor (stall clock keeps running, anchor can never become NaN: both assignment sites are gated on `isfinite`), and the post-reply confirm (already bounded). No regression to readback-at-target, the lost-reply rule, or the ceiling. One nuance: on the previous head the reset happened once (re-anchor on the first finite readback), not on every tick as the finding said — the direction was right, the magnitude one tick. `tests/test_ca_devices.py`: 38 passed. Non-blocking note: the new test's timing window is +0.3 s where siblings use +0.4–0.6; widen to +0.4 if CI ever flakes.

CI on `ab630071`: unit-tests and pre-commit pass (exit codes 0).

Issue #904, opened 2026-09-15
Pseudo scan variables on the native path: pseudo positioners over component user offsets (merge gate)


skbarber:
**Merge-gate item** (every piece in place before `feature/native-bluesky-plans` merges to master). Composite scan variables are scanned regularly on HTU and **cannot be scanned on the branch**: `qs_client/presets.py` refuses a `kind: pseudo` axis and `GeecsNamespace` builds no noun for one. Master can; the branch cannot.

**Design brief: `Planning/native_bluesky/09_pseudo_transform.md`.** §3 was ruled by Sam on 2026-09-14/15 and is closed — the build starts from §5 and does not reopen it. This body is the issue-side summary; the brief is canonical.

## What is decided (§2/§3)

- **Mechanism:** ophyd-async's `Transform` + `DerivedSignalFactory` (0.19.3, installed). A transform declares `derived_to_raw` (our `forward`) **and** `raw_to_derived` (the inverse deferred since #600). The pseudo's readback is computed from the components' live readbacks, so it is defined before any set and `locate()` is real.
- **Vocabulary: frameworks' words only.** A pseudo is a *pseudo positioner* (ophyd/ophyd-async). The steering bump is a pseudo positioner over components read in a *user* coordinate (EPICS user/dial, `.OFF`; ophyd `set_current_position`), the offset held in software because GEECS has no offset field. Never "absolute pseudo" / "offset pseudo" in code or docs; the catalog's `mode: relative` keeps its spelling and means "zero the components' user offsets at scan start".
- **The offset is the primitive, on the component.** Every `CaSettable`/`CaMotor` carries a soft `offset` signal (dial = raw GEECS value); the pseudo transform reads its components' offsets as parameters. This arc zeroes them automatically at stage and adds nothing operator-facing; set-as-aligned + persistence + display is a purely additive follow-on arc.
- **`mode: relative` stays** — it is *not* `rel_scan` over the inverse. A bump is a deviation from today's hand alignment; inverting through one magnet and `rel_scan`-ing would snap the other onto the formula's absolute relation at the first step (the 2b `U_S4H` incident class, doubled). Baseline capture at stage stays, as the components' offsets. All 11 relative corpus formulas are `a*x`, so `set(0)` restores exactly; `f(0) = 0` is pinned at catalog load for every relative entry.
[...]

Issue #879, opened 2026-09-14
Web scanner: pseudo scan variables scannable from the picker


skbarber:
Design door kept open in `Planning/native_bluesky/10_web_scanner.md` §5 (feature/web-scanner). Today `GET /api/configs/scan_variables` returns the catalog with `kind`, and pseudo entries render **disabled** with a refusal text because `expand_preset` refuses `kind: pseudo`.

**Work:** when the pseudo arc (`Planning/native_bluesky/09_pseudo_transform.md`) lands the derived-signal noun, flip the flag: `scannable=True` for pseudo entries, the page changes nowhere else (it never binds variable → device; `expand_preset` does). Per-component confirm, if wanted, is a preflight *question* — the ack dialog is generic over questions.

Related: #855 (rel_* over a pseudo variable needs `locate()`).

skbarber:
Closed by #913 (merged into feature/native-bluesky-plans, GeecsScanner 0.8.0): pseudo entries list as scannable, expand to their namespace noun, and a manual move of one is accepted.

Issue #780, opened 2026-09-04
Curate the GEECS device DB 'tolerance' column so the engine can stop hardcoding 0.005


skbarber:
## Context

While fixing #779 we established that the DB `tolerance` column is consumed in exactly one place: inside GEECS itself, where the LabVIEW blocking set uses it to decide convergence before acking. Nothing on our side reads it — the CA gateway's DB layer loads the column and deliberately drops it (using it as a monitor deadband once froze sub-tolerance motion out of readbacks; see GeecsCAGateway `CLAUDE.md` quirks), and GeecsBluesky's readback-confirming movables use literals: `0.005` in `GeecsSession.motor` / `CaMotor`, `0.05` in the EMQ `confirm` path.

The owner's read (2026-09-03): the DB tolerances were probably never curated well, and at some point we routed around the poorly curated column to make progress. So for now **0.005 (variable units) stays the default motor tolerance** and the new-schema scan-variable catalogs (`scan_variables.yaml`, #779) rely on it for every `kind: motor` axis — positioners and steering supplies alike (supply ripple is reportedly below 0.005 A).

## What is owed

1. **Audit** `devicetype_variable.tolerance` and the per-instance `expt_device_variable` overrides for every settable numeric variable that appears in a scan-variable catalog (Undulator, Thomson): is the value present, non-zero, and physically sensible for that device's readback resolution?
2. **Curate** the column for positioners and supplies (stage encoders, hexapod axes, magnet currents), documenting the semantics: *set-convergence criterion*, never a monitor deadband.
3. **Then** let the engine read it: `GeecsSession.motor` / `build_movable` take the tolerance from a failure-tolerant `GeecsDb.get_device_variables` touchpoint (the `db_runtime.GeecsDbScalarPolicy` pattern), with a pre-claim configuration error when a `kind: motor` variable has no usable tolerance — a curation gap surfaced loudly rather than papered over with a literal.

Until (1)–(2) happen, (3) would just import the uncurated values, so it is deliberately not started.

## Symptom if 0.005 turns out too tight for a supply

A failed-move pause on that axis (never wrong data — the poll waits for arrival), fixable per variable once (3) lands or by a per-variable tolerance knob.

Related: #779 (the readback-confirm default), the pseudo per-component confirmation gap noted there.

skbarber:
[...]

PR #912, opened 2026-09-15
GeecsBluesky 0.89.0 + GEECS-Schemas 0.26.0: pseudo positioners over component user offsets (#904 step 2)

### LINKED ISSUE #910: CaPseudoMovable component puts still die at ophyd's 10 s default (#906 follow-up)
Found by the adversarial review of #909 (the worker half of #906).

`geecs_bluesky/devices/ca/pseudo.py` builds each component's `GatewaySetpointPut(signal=setpoint, timeout=None)` and `_set_and_wait` awaits `put.put(target)` with no per-put timeout, so the put runs under ophyd-async's `DEFAULT_TIMEOUT` (10 s) — shorter than the 30 s cap #906 removed from `CaMotor`. A pseudo axis over a slow stage (ModeImager class) fails at 10 s while the device is still moving.

Pre-existing and outside #909's scope; the fix is the same shape (wait for the device's reply with a readback stall rule per component, or delegate the component moves to `CaMotor` children).

skbarber:
Build step 2 of the pseudo arc (#904, brief `Planning/native_bluesky/09_pseudo_transform.md`; §3 ruled 2026-09-14, not reopened here). Closes #910. Part of the merge gate for `feature/native-bluesky-plans`.

## What

**The primitive: every settable carries a user offset.** `CaSettable.offset` (soft signal, `0.0` by default) and `set_current_position(position)` — ophyd's spelling of the EPICS motor record's `.OFF` (`user = dial + offset`). Nothing moves; the raw GEECS value stays the dial and `set`/`read`/`locate` stay in it. Only the pseudo positioners consume it in this PR; the operator-facing set-as-aligned + persistence + display is the additive follow-on arc.

**`CaPseudoPositioner` replaces `CaPseudoMovable`** (`devices/ca/pseudo.py`). A catalog `kind: pseudo` entry as an ophyd-async `Transform` under a `DerivedSignalFactory` over the components' readback signals, the components' offsets as the transform's parameters:

- `derived_to_raw` = the catalog `forward` formulas; `raw_to_derived` = the inverse — derived by the software for an affine `forward` (`forward_expr.affine_coefficients`, an exact symbolic walk over the whitelisted AST; 24 of the corpus's 26 formulas; the identity component where one exists, else the first), or the entry's physicist-supplied `inverse` (`compile_inverse`; `R56_at_100MeV`).
- The readback is derived from the components' **live** readbacks, so it is defined before any set and `locate()` is real — the case #855 was filed about.
- Component moves go through each component's **own** `set()` (`CaMotor`: the device's reply is the verdict, stall rule per component) — no put budget of this class's own (#910).
- `mode: absolute` = a **plain pseudo positioner** (dial frame; ends where the last step left it; `rel_scan` puts it back through the inverse). `mode: relative` = the components' offsets are **zeroed at every `stage()`** (lazily on the first `set`/`locate` for an unstaged caller, e.g. a manual `mv`), the readback is 0 by construction, each step moves every component by its own formula from its baseline, and **`unstage()` restores the captured baselines** — end of scan and abort alike (a `halt` skips unstage by bluesky contract), formula-independent, riding the components' `set()` so a failed restore fails the plan visibly. `scan` and `rel_scan` over a bump are the same scan.
- **The disagreement check** (§3): `forward(inverse(readbacks))` vs the readbacks, per component within its tolerance (DB tolerance, else `DEFAULT_AGREEMENT_TOLERANCE` = 0.01 in the component's units — #780's magnets). A component off the formula *after this pseudo moved it* raises `PseudoComponentsDisagreeError` before anything moves; before the first move a plain pseudo warns and snaps (today's behaviour), a relative one fails (its deviations were just zeroed).
- `build_pseudo(name, spec, resolve, tolerance=…)` — the one constructor from a `PseudoScanVariable`; every formula checked at build: a relative `forward` must be `0` at `0` (`set(0)` restores), a non-affine `forward` needs the entry's `inverse`, a supplied `inverse` must undo the `forward`s at a probe value, a target must resolve to a `CaSettable`.

**Deleted in the same change** (repo rule: the modernising PR removes the old path): `CaPseudoMovable`, its `absolute`/`relative` branch, per-pseudo baseline capture, `restore_baselines_plan`, the direct `GatewaySetpointPut` fan-out, and its tests. Nothing on this branch built one.

**GEECS-Schemas 0.26.0:** `description` on both scan-variable kinds (free text — for a bump, the geometry and equal-kick assumption behind its coefficients, per the brief's "do now, at zero cost"), and `inverse`'s description now says what the engine reads (components' device names as symbols, or the full target with non-letters as `_`; required only for non-linear relations). No document changes shape; `schema_reference.md` regenerated.

**One deviation from the brief's §2 sketch, on purpose:** `forward_expr.py` is *not* retired. §6's YAML keeps `forward` expressions, so the compiler stays and the software reads `a`/`b` off the expression's AST instead of the catalog carrying coefficients. Recorded in the brief's status line.

- `description` added to the plain `ScanVariable` too, not only the pseudo — one field, same shape; the brief asked for it on the bumps.
- `DEFAULT_AGREEMENT_TOLERANCE = 0.01` (component units) as the fallback where the DB tolerance is 0/NULL; the brief marks this *default* and says refine against real `U_S3H`/`U_S4H` readback noise on hardware — the number in the message will say if it is wrong.
- A never-zeroed relative pseudo's `read()` (before any stage or move — a display) is its identity component's user-frame value with default offsets, i.e. its raw current. `locate()` and `set()` zero first, so no plan ever acts on that number.

## Not in this PR

Namespace nouns + `presets.py` (step 3), `kind: motor` opt-in (step 5), the scanner picker flip (#879) — next PRs. The configs-repo catalog change (bump descriptions, R56 `inverse`) is GEECS-BELLA/GEECS-Plugins-Configs#15; it needs GEECS-Schemas 0.26.0 on the worker before it merges (`description` is a new field under `extra="forbid"`).

## LOC

| concern | lines |
|---|---|
| `devices/ca/pseudo.py` (rewrite) | ~+500 / −215 |
| `forward_expr.py` (affine walk + inverse compiler) | ~+170 |
| `devices/ca/settable.py` (offset + `set_current_position`) | ~+35 |
| `exceptions.py` | ~+12 |
| tests (`test_pseudo_positioner.py` new, `test_forward_expr.py`, schema round-trip) | ~+430 / −140 |
| GEECS-Schemas fields + docs, CHANGELOGs, CLAUDE.md, planning notes | ~+120 |

## Tests

- GeecsBluesky (own env, `scripts/check.sh GeecsBluesky`): **586 passed, 3 skipped (hardware), 1 deselected** — includes the new `test_pseudo_positioner.py` (23) and the extended `test_forward_expr.py` (57).
- GEECS-Schemas (`pytest GEECS-Schemas/tests -m "not integration and not gui"`): **438 passed, 7 deselected**, the schema-reference no-drift guard included; the corpus integration walk passes locally against the configs PR below (all 13 pseudos build).

Bite-probes run and reverted (each made the named tests fail): no restore at unstage → 5 red; disagreement never fails → 2 red; stage does not zero → 2 red; affine inverse reads the wrong component → 2 red.

## Hardware verification

OWED (step 6 of the arc, Sam, on the lab network — after step 3 lands, since nothing submits a pseudo until then):
- a bump scan (`ALine_e_beam_angle_offset_x`, e.g. −0.1 … 0.1 A, 5 points): S3H moves by the value, S4H by −2×, both return to their pre-scan currents at the end (log lines `baselines captured` / `restoring baselines`);
- a `rel_scan` over `R56_at_100MeV` (needs the configs PR's `inverse`): locate reads the current R56 from the chicane, the scan runs about it, the put-back restores it; expect a WARNING at locate if the outer magnet is not at −inner;
- an **aborted** bump scan (Stop mid-way): S3H/S4H back at their pre-scan currents before anything else is staged.
Restore scripts use the pre-scan values read at the top, never a hard-coded zero (2b lesson).


skbarber:
## Adversarial review (fresh-context subagent, on c1228122 + 3b346ec5) — report in full

# Adversarial review — PR #912 (`feat/pseudo-transform` → `feature/native-bluesky-plans`)

Read the full diff, the new `devices/ca/pseudo.py`, `forward_expr.py`, `settable.py`, both test files, the brief `09_pseudo_transform.md`, `preprocessors.connect_on_demand`, `namespace.telemetry`, and ophyd-async 0.19.3's derived-signal backend. PR tests pass locally (80/80 across `test_pseudo_positioner.py` + `test_forward_expr.py`). Two findings below were **reproduced with bite-probes** on the PR code (scratchpad `test_probe_pseudo.py`, not committed). §3 rulings taken as given; only fidelity checked.

## Correctness

### MEDIUM — The disagreement check is compared per component in that component's own tolerance, but the inverse propagates the *reference* component's settle error into every other component's prediction, amplified by `|a_j|`. Reproduced: a false `PseudoComponentsDisagreeError` mid-scan with every motor inside its own DB tolerance.
`pseudo.py:322-355` (`_check_agreement`) + `affine_inverse` (`:145-168`). `value` is read from the identity component (S3H); the predicted dial for S4H is `−2·value − off`. A `CaMotor` completes its `set()` as soon as its readback is within its tolerance, so after step 1 S3H may sit `+ε_ref` off and S4H `+ε_j` off, each legal. The residual the check sees on S4H is `ε_j − a_j·ε_ref` = up to `(1+|a_j|)·tol`, tested against `tol`. Probe: both motors `tolerance=0.005`, mock readbacks land at `target + 0.004` (inside the band), `build_pseudo(..., tolerance=lambda t: 0.005)`, `stage → mv(0.1) → mv(0.2)` with nothing touched by hand → `"U_S4H:Current reads -0.295, formula says -0.307 at 0.104. A component moved under the scan; nothing was moved."` Every relative bump in the corpus has `a = −2` or `−1.5`, so the worst-case residual is 3× the tolerance. The same arithmetic applies to a systematic setpoint-vs-readback offset `e` on both supplies (residual `e·(1−a_j)` = `3e`). The PR's "refine on hardware" note only contemplates readback *noise*, not this propagation. The existing test (`test_relative_disagreement_within_tolerance_is_not_a_disagreement`) uses a mock that lands exactly on target, so it cannot catch it. Cheap fixes: compare against `tol_j + |a_j|·tol_ref` (the coefficients are known for the affine class), or check agreement in the *derived* frame per component (invert through each component and compare values), which is what the "every choice gives the same value" argument in the brief actually assumes.

### MEDIUM — A partial component failure returns the pseudo's status while another component is still moving; the finalize's restore then puts to that component concurrently. Reproduced.
`pseudo.py:408-410` uses bare `asyncio.gather(*(comp.set(...)))`: it raises on the first failure without awaiting or cancelling the rest. Probe: S3H's `:SP` put refused instantly, S4H's readback arrives 300 ms later. Journal: `[S3H bump, S4H bump, S3H restore, S4H restore]` — the S4H restore put was issued before the S4H bump move's readback had landed (its `CaMotor` convergence poll from the first set is now an orphan task that will log its own timeout ERROR later). On hardware this means a second GEECS blocking set is issued to a supply whose first set is still in progress; if LabVIEW refuses the second (busy), the *restore* fails and the pair is left bumped — the exact incident class the brief's §5 lesson names. Fix: `gather(..., return_exceptions=True)`, await all, then raise the first failure (and the same in `_restore_baselines`). Also, because the exception path skips `self._moved = True`, a plain pseudo that half-moved is back in "warn and snap" mode at its next set even though it did move a component.

### MEDIUM — Components are not in the pseudo's device subtree, and `DerivedSignalBackend.connect` is a documented no-op, so nothing connects them on the pseudo's behalf.
`pseudo.py:216-217` keeps components in a plain list (correct — they are the namespace device's children and cannot be re-parented), and `CaPseudoPositioner` does not override `connect()`. `connect_on_demand` (`preprocessors.py:100-103`) connects "the touched object's subtree"; `DerivedSignalBackend.connect` says "Assume that the underlying signals are already connected" (ophyd-async `_derived_signal_backend.py:283-285`). Today it works only because `install_telemetry` connects every scalar-only device at RE build — but that same function **drops any member that fails to connect** with a warning ("left out of the baseline"). Scenario: a magnet's gateway PV is briefly unreachable at worker start → dropped from telemetry → a direct `scan` on `U_S3H.current` later works (connect_on_demand touches it) but a bump scan over it fails at `stage()` with `NotConnected`/hang inside `set_current_position`'s `get_value()`, with no message naming the component. The tests connect each component explicitly (`Bench.__init__`) so they cannot see it. This is the class's to own before step 3 wires namespace nouns: forward `connect()` to the components (`gather(comp.connect(...))`) or yield `ensure_connected` for them.

### MEDIUM — After a restore that failed (or was skipped by `halt`), the next `stage()` re-zeroes with the leftover bump baked in, silently.
`unstage()` (`:289-295`) clears `_zeroed` in `finally` even when `_restore_baselines` raised; `stage()` (`:271-276`) zeroes unconditionally. Scenario: restore refused on S4H (the finding above, or a plain LV error) → plan fails visibly (good) → the operator queues the next bump scan → `stage()` captures the *bumped* S4H current as today's alignment, and the end-of-scan restore now "restores" to the bumped value. The brief §6 names this edge for the hardware acceptance, but only for the halt path; the code adds the failed-restore path and drops the one piece of state that would let the device refuse or recover (the components' `offset` signals still hold the true baselines at that moment; the pseudo just forgets). A device-local `_restore_pending` flag (set at `stage`, cleared only on a *successful* restore) discriminates this from the deliberate manual-`mv`-left-in-place case that `test_a_later_stage_takes_the_current_positions_as_the_new_alignment` pins, and `stage()` can then refuse (or restore first) while one is pending.

### LOW — `affine_inverse` refuses an entry whose first component is a constant when no identity component exists, instead of picking the next non-constant one.
`pseudo.py:155-163`: `next(identity, 0)` then `if a == 0.0: raise`. Verified: `affine_inverse(['probe','jet'], [(0.0, 8.5), (2.0, 0.0)])` → `GeecsConfigurationError: component 'probe' does not depend on the scanned value`. A pseudo that parks one device at a constant while scanning another (`forward: 8.5`) is a legitimate shape; the fix is `next(i for i,(a,b) in enumerate(coefficients) if a != 0.0)` and raise only if none.

### LOW — `_check_inverse` passes silently when both probes fall outside the relation's domain.
`pseudo.py:543-548`: a `GeecsConfigurationError` at both 1.0 and 2.5 → `continue` twice → no check performed, no message. An `inverse` for a formula only defined on, say, negative values is accepted unverified. Cheap: track whether any probe ran and refuse/warn if none did.

### LOW — Brief text this PR edits still describes the *previous* mechanism in two places.
`09_pseudo_transform.md` §2 last paragraph ("can therefore be replaced by a stock end-of-scan set-to-zero") and §5 step 4 ("`restore_baselines_plan` (replaced by set-to-zero through the inverse …) and most of `forward_expr.py`") — the code restores the captured dial baselines formula-independently in `unstage()` and keeps `forward_expr.py`. The new status line corrects only the `forward_expr` half. §6 also says "set(0) restores — also on abort, before unstage" whereas the restore is *in* unstage. Given the brief's own staleness rule, worth one more edit.

### LOW — Schema/docs say the inverse may use "each component's device name"; the engine offers the bare device name only when it is unique among the entry's targets (`inverse_symbols`, `:436-437`).
`scan_variables.py` `inverse` description, CHANGELOG, `schema_reference.md`. A physicist writing `U_ESP302_02` for a two-axis entry on the same device gets "unknown name" with no hint about the uniqueness rule. Also "non-letters replaced by '_'" vs the code keeping digits — say "non-alphanumerics".

## Redundancy
Checked: `within_tolerance` (reused from `motor.py`), `safe_name` / `geecs_core.pv_naming.normalize_component` (reused for keys), `geecs_schemas.restricted_expr` (reused by `compile_inverse`), `GeecsScanner/service/scanner.py` + `qs_client/presets.py` (still refuse pseudos — untouched, correct for step 2), `GEECS-MCP/scans/read_tools.py` (schema-only, unaffected). One overlap worth naming:

- **LOW** — `inverse_symbols` invents a *second* identifier-mangling rule (case-preserving, no run-collapsing) beside `normalize_component` (lowercase, run-collapsing), and a second way of binding device variables to expression symbols beside `geecs_schemas.derived_channels.DerivedInput.symbol` (an explicit per-input `symbol` field, validated with `isidentifier()` and a reserved-name check). The gateway's derived channels solved the same "expression over several device variables" problem with declared symbols; the pseudo's `PseudoComponent` could carry an optional `symbol` the same way, which would also remove the uniqueness caveat above. Not blocking — the corpus has one `inverse`.

## Placement
- **LOW** — `affine_coefficients` and the three catalog-validity rules in `build_pseudo` (relative ⇒ `f(0)=0`; non-affine ⇒ `inverse` required; `inverse` must undo `forward`) are engine-free AST/arithmetic over the schema's own `restricted_expr`. Their natural home is beside `restricted_expr` in GEECS-Schemas (as a `PseudoScanVariable` model validator), so the catalog is refused at edit/load time everywhere the schema is loaded, not only where the engine builds a device. Every current consumer (scanner, MCP) already depends on GeecsBluesky, so nothing has to move *today*; flagging because the picker flip (#879) will want "is this pseudo buildable" client-side and will otherwise reach into `geecs_bluesky.devices.ca.pseudo` for a pure config question.
- `CaSettable.offset` / `set_current_position` on the component: correct home per §3 (the primitive lives on the component). `PseudoComponentsDisagreeError` in `exceptions.py`: correct.

## Verified clean
Version bumps + CHANGELOGs (both packages), `schema_reference.md` regenerated, `GeecsBluesky/CLAUDE.md` and `01a_device_layer_audit.md` updated, no remaining references to `CaPseudoMovable`/`restore_baselines_plan` outside historical CHANGELOG entries, no scan-folder creation, no lab hostnames/paths, `offset` correctly excluded from `read()`, the affine walk's `Mult`/`Div`/`UnaryOp` algebra is right for the whole corpus table, the `rel_scan`/`scan` equivalence over a bump holds as claimed (walked the wrapper order: reset_positions → relative_set → stage → run, both capture at the first `set`, i.e. after zeroing).

---

## Disposition (fix commit 655e9de8; the reviewer's confirm/deny of each fix follows in a further comment)

| finding | disposition |
|---|---|
| MEDIUM tolerance propagation | **fixed** — `_allowance()`: the inverse's own uncertainty measured numerically (each component shifted by its tolerance, re-inverted), each prediction's allowance = its tolerance + its movement under that value shift; the reviewer's +0.004 settle probe is now `test_agreement_allows_what_the_inverse_propagates` |
| MEDIUM partial failure / concurrent restore | **fixed** — `_move_all()` waits for every component, then raises the first failure; `_moved` set before the moves; `test_partial_failure_waits_for_the_other_component_before_failing` (bites with a bare gather) |
| MEDIUM components not connected | **fixed** — `CaPseudoPositioner.connect()` also connects its components (an already-mocked one is left alone: mock connects are not cached); `test_pseudo_connect_connects_its_components` |
| MEDIUM failed restore → restage bakes the bump in | **fixed** — `_restore_pending` (set at stage, cleared only by a completed restore or a successful relative `set(0)`), `_zeroed` kept after a failed restore, `stage()` raises `PseudoRestorePendingError` naming the owed baselines; `test_failed_restore_refuses_the_next_stage_until_moved_back_to_zero` |
| LOW affine_inverse constant-first | **fixed** — identity, else first non-constant; `test_affine_inverse_skips_a_constant_component` |
| LOW _check_inverse silent | **fixed** — third probe (−1.0) and a WARNING when no probe ran |
| LOW brief text | **fixed** — §2, §5 step 4, §6 |
| LOW schema wording | **fixed** — uniqueness rule stated, "non-alphanumerics"; reference regenerated |
| LOW second mangling rule / `symbol` field | **waived** — the corpus has one `inverse`; an explicit `symbol` on `PseudoComponent` (the derived-channel shape) is recorded for the follow-on if a second non-affine entry appears |
| LOW validity rules belong in GEECS-Schemas | **waived** — every consumer already depends on GeecsBluesky and the rules need the expression compiler's whitelist; revisit if a schema-only consumer needs "is this pseudo buildable" |

skbarber:
## Reviewer confirmation of the fix commit 655e9de8 (same fresh-context reviewer, independent re-read + full suite)

Ran the PR test files (85 passed), the full GeecsBluesky suite (591 passed, 3 skipped, 1 deselected), the GEECS-Schemas reference/scan-variable tests incl. the no-drift guard (21 passed), and re-ran the reviewer's own two original probes against the fixed code.

1. **CONFIRMED** — `_allowance()` yields S4H allowance = 0.005 + |f_S4H(v ± 0.005) − pred| = 0.015 for the angle bump; the original settle-at-+0.004 probe now runs to completion. Note: the worst-case residual `tol_j + |a_j|·tol_ref` equals the allowance exactly, so it is tight, not generous.
2. **CONFIRMED** — `_move_all()` awaits all, raises the first failure; `_moved` set before the moves; the restore uses it too. The probe's event log now shows S4H's bump readback landing *before* its restore put.
3. **CONFIRMED** — `connect()` gathers the components' connects (already-mocked components skipped in mock mode; real connects cached by ophyd-async). `test_pseudo_connect_connects_its_components` connects only the pseudo over two never-connected motors and reads the derived readback.
4. **CONFIRMED** — `_restore_pending` set at stage, cleared only by a completed restore or a successful relative `set(0)`; `_zeroed` survives a failed restore so `mv 0` restores; `stage()` raises `PseudoRestorePendingError` naming the owed baselines; the `halt` path is covered by the same flag; the deliberate manual-`mv` case still passes.
5. **CONFIRMED** — `affine_inverse`: identity, else first non-constant, raise only when none.
6. **CONFIRMED** — `_check_inverse` probes (1.0, 2.5, −1.0) and warns "taken on trust" when none ran; R56 still verifies at 1.0/2.5.
7. **CONFIRMED** — brief §2 / §5 step 4 / §6 corrected.
8. **CONFIRMED** — schema `inverse` description and the regenerated reference state the uniqueness rule and "non-alphanumeric".

Observation, not a finding: when `stage()` raises `PseudoRestorePendingError` inside a stock plan, `stage_wrapper`'s finalize still calls `unstage()`, which retries the restore — a benign automatic retry that either recovers or fails loudly with the device's own error.

All findings dispositioned (8 fixed, 2 waived). Ready to merge once CI is green.

PR #913, opened 2026-09-15
GeecsBluesky 0.90.0: pseudo scan variables as namespace nouns; presets stop refusing them (#904 step 3)

skbarber:
**Stacked on #912 — merge #912 first** (this PR will be retargeted to `feature/native-bluesky-plans` before #912 merges). Build step 3 of the pseudo arc (#904). After this, the branch can scan composite variables again — the merge-gate capability #904 is about.

## What

- **`GeecsNamespace.add_pseudos(catalog.variables)`** binds every catalog `kind: pseudo` entry as a `CaPseudoPositioner` (from #912) over the Movable children the roster already holds for its targets — one object per component, so the user offset a bump zeroes is the child's own — under the catalog name as an identifier (`ALine_e_beam_angle_offset_x`), with each target's DB tolerance as its agreement tolerance. Targets resolve through `get_settable` (so a read-only or native-save variable is refused with its existing message). An entry that cannot be built — a formula that does not compile, an unserved target, a relative `forward` not `0` at `0`, a name colliding with a device — is **ERROR-logged and skipped**: one bad catalog line never keeps the worker from opening, and the submit preflight then reports the missing reference before anything is claimed. Pseudos stay out of the telemetry baseline (their components are already in it).
- **The startup profile** calls it after the roster is built, best-effort like the shot offsets (an unreadable catalog costs the pseudos, not the worker; WARNING with the traceback).
- **`qs_client.presets`**: a catalog pseudo expands to its namespace binding and is recorded as a reference the preflight checks against the manager's device tree. The "pseudo axes are not scannable through the namespace yet (phase 3)" refusal is gone.

## Not in this PR

`kind: motor` opt-in (step 5, next PR), hardware acceptance (step 6, Sam). The web scanner picker flip (#879) is folded in as commit dd069d5d (GeecsScanner 0.8.0) — the scanner tests pin the refusal this PR removes, so it cannot be deferred without a red mainline (review HIGH).

## LOC

| concern | lines |
|---|---|
| `namespace.py` (`add_pseudos`, `_settable_for`, `_db_tolerance`, telemetry exclusion, module doc) | ~+95 |
| `qserver/startup/startup.py` | ~+12 |
| `qs_client/presets.py` + `submit_preflight.py` (doc + the expansion) | ~+15 / −15 |
| tests (`test_namespace.py` 2 new, `test_presets.py` reworked 2, `test_qserver_startup.py` extended) | ~+130 |
| CHANGELOG, CLAUDE.md, brief status | ~+40 |

## Tests

- `scripts/check.sh GeecsBluesky GeecsScanner` (own envs): GeecsBluesky **596 passed, 3 skipped (hardware), 1 deselected**; GeecsScanner **77 passed, 1 skipped**.
- The new namespace test builds two pseudos over a magnet roster and checks the components ARE the roster's children (identity), the DB-vs-default tolerance choice, export and telemetry exclusion; the broken-entry test feeds four bad entries and one good one and asserts each bad one's ERROR line and that the device the bad name collided with survived.

## Hardware verification

OWED — step 6 of the arc, Sam on the lab network, after this and #912 are deployed on the worker and GEECS-BELLA/GEECS-Plugins-Configs#15 is merged: a bump scan over `ALine_e_beam_angle_offset_x`, a `rel_scan` over `R56_at_100MeV`, and an **aborted** bump scan proving S3H/S4H return to their pre-scan currents before anything else is staged. Expected log lines: `baselines captured (components zeroed)` at stage, `restoring baselines` at unstage; the scan's s-file column header is the catalog name.


skbarber:
## Adversarial review (fresh-context subagent, on ad91d816 — the pre-rebase head) — report in full

# Review of PR #913 (`feat/pseudo-nouns`, own diff only)

Verified before writing: the queueserver manager discovers a namespace pseudo (`existing_plans_and_devices_from_nspace` lists `ALine_e_beam_angle_offset_x`) and `validate_plan` accepts `scan([...], "ALine_e_beam_angle_offset_x", -1, 1, 5)`; a complete mock bump scan through `make_run_engine(mock=True, telemetry=ns.telemetry())` over a namespace-built pseudo records readbacks −1/0/1 and restores both magnet baselines at unstage. Preflight `references` path, CHANGELOG/version bump, telemetry exclusion, the broken-entry test's assertions (it would bite), hygiene: no issues. Placement is right (namespace owns the binding; `presets.py` shares `identifier_name`).

## HIGH — CI is red: GeecsScanner pins the refusal this PR removes

`gh pr checks 913`: unit-tests **fail** on both runs; the failed step is `Run GeecsScanner tests`:
```
FAILED tests/test_api.py::test_preflight_refuses_a_pseudo_axis - assert (None)
FAILED tests/test_verbs.py::test_move_accepts_device_colon_variable_and_refuses_pseudo - assert 409 == 400
2 failed, 75 passed
```
Both call straight into `geecs_bluesky.qs_client.presets.scan_variable_reference` / `expand_preset`, which now return a noun instead of raising. Beyond the tests, `GeecsScanner/geecs_scanner/service/scanner.py:254-272` still answers `scannable: false` with reason "pseudo axes are not scannable through the namespace yet", and `move()`'s docstring (line 409) says "a pseudo entry is refused there" — both false the moment 0.90.0 is on the worker, and the scanner's own `move` will queue a pseudo `mv` while its picker still greys it out. The PR defers the picker flip to #879, but the two tests and the wire-contract text cannot be deferred without a red mainline. Note `scripts/check.sh` in changed-mode scopes to the touched package only (no dependents), which is why local checks didn't surface it; the PR body's "Tests: COUNTS_B" is an unfilled placeholder.

## MEDIUM — `connect_on_demand` cannot connect a pseudo's components; a pseudo scan depends on the telemetry connect at environment open having succeeded

The pseudo's only child is the derived `readback`; `DerivedSignalBackend.connect` is a documented no-op ("Assume that the underlying signals are already connected", ophyd_async/core/_derived_signal_backend.py:283). `stage_wrapper` stages the pseudo (its root), so `connect_on_demand` connects the pseudo and nothing else. Reproduced in mock (script in the scratchpad): with the components' devices left out of the telemetry set, `RE(bp.scan([], bump, -1, 1, 3))` fails at `stage` with `NotImplementedError('No PV has been set as connect() has not been called')` (from `_zero_components → CaSettable.set_current_position → readback.get_value()`); with them in the set the same scan runs end-to-end.

On the worker this works only because `install_telemetry` (run_engine.py:130) connects every scalar device at environment open — and **drops the ones that fail** (its docstring names the case: a device added to the DB after the gateway started; a gateway restart at open is another). Such a component stays unconnected for the life of the environment: a direct `scan(..., U_S3H.current, ...)` self-heals through `connect_on_demand`, a bump scan over the same magnet fails every time with that opaque error until the environment is reopened. The natural fix is `CaPseudoPositioner.connect()` also connecting its components (or `add_pseudos` giving the pseudo a connect that does), plus one test that drives a namespace pseudo through the RE with `connect_on_demand` — the two new tests inspect the object graph only, and `test_pseudo_positioner.py` connects components by hand.

## LOW — the collision check is exact-case, the lookup is not

`add_pseudos` refuses only `identifier_name(friendly) in self._devices`. A catalog key differing from a device name only by case (`u_s3h` vs `U_S3H`) binds without error, overwrites `_by_geecs_name["u_s3h"]`, gives two devices the same ophyd name `u_s3h` (duplicate event-key prefix), and `get_settable("u_s3h", "Current")` — the lower-case spelling `_lookup` deliberately accepts for action plans — now raises a raw `AttributeError: 'CaPseudoPositioner' object has no attribute 'current'` (reproduced; `_child_for_step` converts only `KeyError`). The DB build has the same shape but MySQL's case-insensitive names make it unreachable there; YAML keys are case-sensitive. Check `safe_name(friendly)` against existing ophyd names / `_by_geecs_name` as well.

## LOW — a catalog key that is a plan name (or `RE`) is silently shadowed

`add_pseudos({"count": …})` binds and `export_into` accepts it (nothing is in globals yet); `startup.py` then does `globals().update(bind_plans(...))` and `RE = …`, silently replacing the noun, and `__all__` lists both. The ERROR-log promise does not fire. DB names are `U_`/`UC_`-prefixed; catalog keys are free text from the web editor. Cheap fix: refuse `GEECS_PLAN_NAMES | {"RE"}` in `add_pseudos` (or pass a reserved set to `export_into`).

## LOW (redundancy) — `_db_tolerance` re-derives what the resolved child already carries

`_movable` binds a `CaMotor` iff the DB tolerance > 0 and stores it as `CaMotor._tolerance` (motor.py:179); otherwise a `CaSettable`, for which `build_pseudo` uses the default anyway. `tolerance=lambda t: getattr(self._settable_for(t), "_tolerance", None)` gives the same answer for every buildable case without a second walk over the roster rows. Fine to keep if the second reader is intentional; then say so.

## LOW (docs/process)

- The brief (`09_pseudo_transform.md`) already says "Step 3 landed" while the PR is open and CI red.
- `GeecsScanner/CLAUDE.md:48-49` ("when the pseudo arc lands, the flag flips") and `README.md:13` become stale with this merge — belongs with the HIGH item's scanner touch.

## Attempted and refuted

- Startup's broad `except Exception` around `add_pseudos` swallowing all pseudos on one bad entry: could not construct an escaping exception — `forward_expr` wraps arithmetic errors in `GeecsConfigurationError`, `_child_for_step` converts `KeyError`, `add_pseudos` catches `TypeError/ValueError`.
- Preflight rejecting the noun: `allowed_device_names` includes top-level devices; `is_device` accepts a `StandardReadable`.
- `_by_geecs_name` exact-case duplicates between two pseudos: caught by the `ns_name in self._devices` check.

---

## Disposition (commits dd069d5d + 814fdbf6, after the rebase onto the fixed #912; the reviewer's confirm/deny follows in a further comment)

| finding | disposition |
|---|---|
| HIGH scanner pins the refusal | **fixed** — the picker flip is folded into this PR (dd069d5d, GeecsScanner 0.8.0, closes #879): `scannable: true` for pseudos, `move()` doc corrected, the three tests reworked, the demo device tree carries the demo pseudo noun, scanner CLAUDE.md/README updated |
| MEDIUM components not connected on demand | **fixed** — in #912's 655e9de8 (`CaPseudoPositioner.connect()` connects its components); pinned here by `test_namespace_pseudo_scans_through_connect_on_demand` (namespace-built pseudo, `connect_on_demand`, nothing connected up front, a full `bp.scan`) |
| LOW case-insensitive collision | **fixed** — checked against `_by_geecs_name` and existing ophyd names; `test_add_pseudos_refuses_case_clashes_and_plan_names` |
| LOW plan-name / `RE` shadowing | **fixed** — refused against `GEECS_PLAN_NAMES ∪ {RE}`; same test |
| LOW `_db_tolerance` redundancy | **waived** — deliberate: the agreement tolerance is the DB fact, not the child's `_tolerance`, because step 5's `kind: motor` opt-in builds a `CaMotor` with the class default where the DB says 0 and the agreement fallback must not inherit that guess |
| LOW docs | **fixed** — brief now says "Step 3 is PR #913"; scanner docs in dd069d5d |

skbarber:
## Reviewer confirmation of the fix commits dd069d5d + 814fdbf6 (same fresh-context reviewer, independent re-read + suites)

Suites run in the PR's worktree: GeecsBluesky `test_namespace.py test_qserver_startup.py qs_client/test_presets.py test_pseudo_positioner.py` → 77 passed; GeecsScanner full `tests` (importing this branch's `geecs_bluesky`) → 77 passed, 1 skipped.

1. **CONFIRMED** (HIGH, scanner pins the refusal) — dd069d5d drops the `scannable=not pseudo` + reason text, corrects `move()`'s docstring, the demo's `allowed_device_names()` carries each catalog pseudo's noun, the three reworked tests pass against this branch; the two previously-failing CI tests are gone.
2. **CONFIRMED** (MEDIUM, components not connected on demand) — `CaPseudoPositioner.connect()` (#912's 655e9de8) connects every component; the reviewer's repro now gets past `stage`, and `test_namespace_pseudo_scans_through_connect_on_demand` **bites**: with the override undone at runtime it fails with exactly the original `NotImplementedError('No PV has been set as connect() has not been called')`.
3. **CONFIRMED** (LOW, case-insensitive collision) — `add_pseudos` checks `_by_geecs_name` and existing ophyd names; the `u_s3h`-over-`U_S3H` repro now binds nothing and `get_settable("u_s3h", "Current")` returns the settable; the new test asserted the opposite of the pre-fix behaviour.
4. **CONFIRMED** (LOW, plan-name/RE shadowing) — `reserved = {*GEECS_PLAN_NAMES, "RE"}`; `count`/`RE` refused with the "plan name" message and pinned by the test.
5. **WAIVER ACCEPTED** (LOW, `_db_tolerance`) — rationale holds; the suggested one-line comment is in (the follow-up commit after the rebase onto the fixed #912).
6. **CONFIRMED** (LOW, docs) — brief and scanner CLAUDE.md/README updated.

No new findings. Observation only: the ophyd-name clash scan is O(devices) per pseudo at environment open — negligible.

All findings dispositioned (5 fixed, 1 waived). Rebased onto #912's head; ready to merge after #912.

PR #914, opened 2026-09-15
GeecsBluesky 0.91.0: the catalog's kind: motor opt-in binds a CaMotor where the DB tolerance is 0 (#904 step 5)

skbarber:
**Stacked on #913 (on #912) — merge those first**; will be retargeted to `feature/native-bluesky-plans` before #913 merges. Build step 5 of the pseudo arc (#904), the deferred item from `01_device_namespace.md`.

## What

Today `GeecsNamespace._movable` decides motor-vs-settable purely from a positive DB tolerance and ignores the scan-variable catalog. The corpus marks the steering magnets (`S1H`…`VS4V`) `kind: motor`, and #780 says their DB tolerance is 0 — so a catalog entry declaring a real positioner silently got a plain setpoint with no convergence confirmation.

- `namespace.motor_targets(catalog.variables)` → the `"Device:Variable"` targets of plain entries declared `kind: motor` (pseudo components are not opted in here; per-component kind stays a later addition).
- `GeecsNamespace(..., motor_targets=…)` / `from_experiment(..., motor_targets=…)`: such a target binds a `CaMotor` even where the DB tolerance is `0`/NULL, at the class default tolerance (`DEFAULT_TOLERANCE`, 0.005 in the variable's units), and the build **WARNs** naming the DB row to curate. A positive DB tolerance still wins its own value; the catalog never downgrades (`kind: setpoint` is the schema default and would otherwise strip every motor).
- The startup profile reads the catalog once (best-effort, like the shot offsets) for the opt-ins and the pseudo nouns.

## Judgment call (cheap to veto)

The opt-in's tolerance is the class default (0.005) rather than a refusal or a catalog field: the schema deliberately keeps tolerances below the configs (device facts), and #780 is the DB curation that makes the default go away. The WARNING says which row.

## LOC

| concern | lines |
|---|---|
| `namespace.py` (`motor_targets`, the `motor_targets` parameter, `_movable`) | ~+45 |
| `qserver/startup/startup.py` (read the catalog once) | ~+12 / −10 |
| tests (3 new in `test_namespace.py`, the startup export test extended) | ~+50 |
| CHANGELOG, CLAUDE.md, planning notes | ~+25 |

## Tests

- `test_namespace.py` + `qs_client/test_presets.py`: 45 passed; `test_qserver_startup.py`: 7 passed (own env). Full `scripts/check.sh GeecsBluesky` count follows in a comment.

## Hardware verification

OWED with the arc's acceptance (Sam): with the corpus catalog on the worker, the S3H/S4H moves of a bump scan now go through `CaMotor` (reply wait + readback-within-0.005 A confirm, WARNING at environment open naming the `Current` rows with DB tolerance 0). Expect no behaviour change on stages that already carry a DB tolerance.


skbarber:
Full `scripts/check.sh GeecsBluesky` on 3820c57b (own env): 599 passed, 3 skipped, 1 deselected, 4 warnings.

skbarber:
## Adversarial review (fresh-context subagent, on 3820c57b) — report in full

# Review of PR #914 (`feat/kind-motor-optin`, own diff only)

Verified locally: `tests/test_namespace.py` + `tests/test_qserver_startup.py` + `qs_client/test_presets.py` → 52 passed in `GeecsBluesky/.venv`. The new tests bite (they assert concrete types + tolerance values on a roster whose S3H row is `tolerance=0.0`; the startup test pins the `from_experiment(motor_targets=…)` wiring through the profile). The corpus (`GEECS-Plugins-Configs/.../Undulator/scan_devices/scan_variables.yaml`) does mark S1H…VS4V `kind: motor`; the EMQ entries are `kind: setpoint` + `confirm`, so the opt-in never touches them. No public-repo-hygiene or scan-folder concerns.

## MEDIUM — the constant's own contract now says the opposite of the PR

`GeecsBluesky/geecs_bluesky/devices/ca/motor.py:86-90`, the comment that owns `DEFAULT_TOLERANCE`:

> Move-completion tolerance of a bare `CaMotor(...)` — tests and ad-hoc devices. In a scan the motor comes from `GeecsNamespace`, which passes the variable's positive DB `tolerance`; **a non-positive one binds a plain setpoint (`CaSettable`), so this default is never used for a DB device.**

After this PR that default is exactly what 16 production magnets run on, and the WARNING the PR emits points a curator at #780 whose text calls 0.005 the deliberate default. The PR updated CLAUDE.md and the namespace docstrings but not the module that defines the number — the next reader of `motor.py` (the file the #906 review lived in) is told the opt-in cannot happen. Same drift, lower stakes, in the planning notes the PR did touch: `Planning/native_bluesky/01_device_namespace.md:29` (the table row still reads "`CaMotor` if the DB gives a positive tolerance, else `CaSettable`" — only the line-137 bullet got the 2026-09-15 note) and `Planning/native_bluesky/03_clean_room_rebuild.md:728` ("Also still deferred … the catalog's `kind: motor` opt-in"). One-line fixes each.

## LOW — an opt-in the namespace cannot honour is silently dropped

`_movable` consults `self._motor_targets` only when it reaches a served, settable, float-typed row. A `kind: motor` target that never gets there — a typo in `target`, a variable the gateway does not serve, a `choice`-typed or non-settable row, a device the roster skipped — produces no motor **and no message**; the set is never diffed against what was bound. The namespace already has the precedent for the analogous document-names-something-I-did-not-build case: `_log_drain_offsets` (`namespace.py` ~line 400) warns about a stale offset precisely because "the failure it causes downstream is silent". Scenario: the catalog says `S3H: target: U_S3H:Curent` (or the DB row is typed `choice`); the operator sees the WARNING for the other 15 magnets, assumes S3H is covered, and S3H scans as a plain setpoint. The unserved-variables preflight catches the typo at submit but not the vartype/settable cases. Cheap fix: after the build loop, warn for `self._motor_targets − {bound CaMotor targets}` the way the offsets are reported.

## LOW — the Tiled event key of every opted-in magnet changes

`CaMotor` reads its readback as `position` (`_readback_attr="position"`), `CaSettable` as `readback`, so `u_s3h-current-readback` becomes `u_s3h-current-position` for S1H…VS4V between 0.90.0 and 0.91.0. The s-file header is unaffected (`CaSettable._column_headers` uses the GEECS `Device Variable` name), `EVENT_SCHEMA.md:54` already documents both suffixes, and I found no repo consumer that keys on the suffix (checked `GEECS-Data-Utils/geecs_data_utils`, `GEECS-DataPortal/geecs_portal`, `GEECS-MCP/geecs_mcp`, `GeecsScanner/geecs_scanner`, `ScanAnalysis/scan_analysis`). Still, the PR body's "expect no behaviour change" covers only stages; a CHANGELOG line saying the magnets' run columns are renamed would save someone reading a run by key.

## LOW — two literals now stand in for the same missing DB fact

For a magnet whose DB tolerance is 0 the component's move confirms within `DEFAULT_TOLERANCE` = 0.005 (`motor.py`) while the pseudo over it agrees within `DEFAULT_AGREEMENT_TOLERANCE` = 0.01 (`pseudo.py:104`). The ordering is sane (agreement ≥ confirm) and #780 endorses 0.005 for supplies, so not a defect — but the WARNING names only one of the two literals #780's curation is meant to retire. Worth knowing for the owed hardware acceptance: with the device having replied `no error`, a readback that sits >5 mA from the setpoint fails the move as `GeecsMotorTimeoutError` after `confirm_budget` = 15 s (`_set_and_wait`), on every step — a loud failure, not a hang, exactly the symptom #780 predicts.

## Redundancy — none found

`motor_targets()` is a new query; the other `kind`-keyed catalog walkers (`namespace.add_pseudos`, `qs_client/presets.py:95 scan_variable_reference`, `GeecsScanner/.../service/scanner.py:262`, `GEECS-MCP/.../read_tools.py:254`) all filter on `pseudo`, none derives motor targets. `ScanVariables` in `GEECS-Schemas/geecs_schemas/scan_variables.py` has no helper methods at all, so there is nothing there to reuse either.

## Placement — defensible, one note

`motor_targets()` is a pure catalog query that could live on the `ScanVariables` model in GEECS-Schemas (`targets_of_kind("motor")`), but the lower-casing is the namespace's own comparison convention and the schema package currently carries no methods; leaving it beside `_movable` is fine. The namespace now takes the same catalog through two seams (`motor_targets=` at construction, `add_pseudos(catalog)` after) because motor binding happens inside `__init__`; a single `catalog=` constructor argument that did both would be one seam, but this is not worth reopening in this PR.

## Not findings (refuted)

- Case handling: both sides lower-case, and the test passes mixed case in. Targets with spaces compare against the DB row name verbatim, the served set's spelling.
- Duplicate targets with mixed kinds (the corpus has `DonDQ12` and `DonD Q12 Gap` on one target) resolve to motor via set union — consistent with "never downgrades".
- `_db_tolerance` (base branch) deliberately reads the roster row — already covered by #913's review.
- Startup fallback: a failed catalog read yields `{}` and `add_pseudos({})` runs harmlessly; the hermetic path is untouched.
- `test_catalog_kind_setpoint_never_downgrades_a_db_motor` passes `motor_targets=set()` rather than a `kind: setpoint` catalog, but with `test_motor_targets_reads_plain_kind_motor_entries_only` the property is pinned; a naming nit only.

---

## Disposition (fix commit 54aea964; the reviewer's confirm/deny follows in a further comment)

| finding | disposition |
|---|---|
| MEDIUM `DEFAULT_TOLERANCE` contract + two planning lines | **fixed** — the constant's comment now names the opt-in as its second user (and the agreement fallback above it); `01_device_namespace.md` table row and `03_clean_room_rebuild.md` "still deferred" line corrected |
| LOW unhonoured opt-in silently dropped | **fixed** — `_log_unhonoured_motor_targets()` after the build loop warns for every `kind: motor` target no `CaMotor` was bound for, the drain-offsets shape; `test_unhonoured_kind_motor_targets_are_warned_about` (misspelled, unserved, read-only listed; a honoured one not) |
| LOW event-key rename | **fixed** — CHANGELOG states the `-position` suffix for opted-in variables and that the s-file header is unchanged |
| LOW two literals for one missing fact | **fixed** (documented) — CHANGELOG and the `DEFAULT_TOLERANCE` comment name both literals as what #780's curation retires |
| Placement note (one `catalog=` seam) | **recorded**, not changed in this PR |

skbarber:
## Reviewer confirmation of the fix commit 54aea964 (same fresh-context reviewer, independent re-read + suites)

Ran `tests/test_namespace.py` + `tests/test_qserver_startup.py` + `tests/test_ca_devices.py` in the PR's env: 79 passed.

1. **CONFIRMED** — `motor.py`'s `DEFAULT_TOLERANCE` comment now states the opt-in as the constant's second user (WARNED at build, naming the row to curate) and names `pseudo.DEFAULT_AGREEMENT_TOLERANCE` as sitting above it; the `01_device_namespace.md` table row and the `03_clean_room_rebuild.md` "still deferred" line are corrected.
2. **CONFIRMED** — `_bound_motors` is initialised before the build loop, both `CaMotor` branches of `_movable` record the target, `_log_unhonoured_motor_targets()` WARNs the set difference right after the drain-offsets report; the new test feeds a misspelled, an unknown-device and a read-only target plus one honoured, and asserts exactly one warning line listing the first three and not the fourth — it would fail if the set difference or the recording were wrong.
3. **CONFIRMED** — CHANGELOG 0.91.0 states the `-position` event-key suffix for opted-in variables and that the s-file header is unchanged, matching `CaSettable._column_headers`.
4. **CONFIRMED** — CHANGELOG and the `motor.py` comment name both literals #780's curation retires.

Placement note recorded, not changed. No new findings. All findings dispositioned (4 fixed). Ready to merge once CI is green.

PR #917, opened 2026-09-15
GEECS-DataPortal 0.28.0: send a plot to the scan's log entry

skbarber:
## What

A **Send plot to this scan's log entry** button on the Plot tab's modebar, and the route behind it (`POST /api/run/{uid}/logbook`, `geecs_portal/logbook_send.py`). The portal's third write verb.

## Why

The request was "the Copy Image to Clipboard button saves instead of copying". It isn't a portal bug and it isn't fixable in the portal: browsers expose clipboard **image** writes on secure contexts only (`https`, or a `localhost` host), so on the plain-HTTP deployment `navigator.clipboard` does not exist and the button's documented download fallback fires every time. No page-side workaround exists — the legacy `execCommand("copy")` path carries `text/plain` and `text/html` alone, and an `` flavour would not even paste into our own logbook, whose composer uploads what arrives as a clipboard *file* (`GeecsLogbook/geecs_logbook/static/editor.js:557`).

Sending sidesteps the whole question: the browser hands the PNG to the portal and the portal calls the logbook **server-to-server**, so nothing depends on the page's origin, the logbook needs no CORS headers, and this package still never imports it (three URLs, per the root CLAUDE.md edge).

## How

One send is four calls — the logbook stores an attachment only *against an entry*, the same problem its composer solves with `ensureEntry`:

1. `POST /api/entries` — create, on this day and scan, in the scans book (skipped when appending)
2. `POST /api/entries/{id}/attachments` — the PNG; the reply names the link, because the store claims the free filename
3. `GET /api/entries/{id}` — the version the upload just moved
4. `PATCH /api/entries/{id}` — the body with the image paragraph appended

Two shapes are load-bearing rather than incidental, and the CLAUDE.md note says so:

- **One entry per scan, appended.** Consecutive image paragraphs are what the logbook's renderer turns into a figure grid (`geecs_logbook/render.py:81` — "the plot-table LogMaker's `gdoc_slot` numbering used to fake"). The page remembers the entry it made, per scan, in the browser, and offers "a new entry" when that isn't wanted; an entry deleted in the meantime starts a fresh one instead of failing.
- **The portal link sits above the images.** A paragraph between two images splits the run and kills the grid. Because the Plot tab's whole state is in its URL, that one link restores the exact analysis, not just the scan.

A 409 (someone saving the entry mid-send) is re-read and retried once — safe *only* because this is an append, never a replace.

Sending needs an **absolute** `--logbook-url`. A path-shaped base (`/log`) describes the browser's front door and names no host this process can dial, so it links but cannot send, and the button is hidden (`logbook_send` in the page context and in `GET /api/run/{uid}`).

Sent plots render at a fixed **720×460** unless an explicit display width/height says otherwise. The note scales the bitmap into its column, so apparent text size is export-width ÷ column-width: a 1200px-wide plot in a three-up grid has unreadable axis labels, a narrow one reads fine.

## Also in this PR (second concern, ~25 LOC — cheap to veto separately)

The clipboard button itself, in the same template and the same subject area:

- **Honesty** — its tooltip now reads "Download plot PNG — clipboard copy needs an https or localhost page" when the API is absent, so it stops looking broken; a real failure names the error (`NotAllowedError` is almost always an unfocused window).
- **A real bug** — the PNG is now rendered *inside* the `ClipboardItem` as a promise value rather than awaited before the write. Safari expires the click's user activation across an `await` and rejects the write that follows, so copy failed there even on a secure page.

## LOC by concern

| Concern | Lines |
|---|---|
| `logbook_send.py` (the conversation) | ~310 |
| `app.py` (route, model, gate) | ~105 |
| `run.html` (button, dialog, send flow) | ~140 |
| `tests/test_logbook_send.py` | ~420 |
| CLAUDE.md / DEPLOYMENT.md / CHANGELOG | ~150 |
| clipboard honesty + gesture fix | ~25 |

## Tests

`./scripts/check.sh GEECS-DataPortal` → **291 passed** (35 new). Lint clean (`--lint`, after the auto-fixers re-staged).

`--all` could not run the other packages' suites in this fresh worktree — they fail at *collection* with `ModuleNotFoundError` (`geecs_data_utils`, `bluesky`, `caproto`, …) because their venvs aren't installed here. Environment, not code; the diff is portal-only plus two CLAUDE.md lines, and CI runs the rest.

New tests were checked for bite by mutation:

| Mutation | Test that caught it |
|---|---|
| single newline instead of a blank line between images | `test_append_separates_paragraphs_with_a_blank_line`, `test_two_images_land_in_one_entry_as_a_run` |
| drop the 409 retry | `test_a_conflict_is_retried_against_the_fresh_body` |
| let a path-shaped `--logbook-url` send | `test_a_path_shaped_url_links_but_cannot_send` |

The fake logbook in the test module rejects a stale `expected_version`, so a client that patched with the version it read *before* uploading would fail there.

## Live verification (no hardware involved)

Run against a **real GeecsLogbook 0.11.0** on loopback (throwaway SQLite store, since removed), not only the fake:

- two sends → one entry, `appended=True` on the second, attachments `plot.png` and `plot-2.png`
- the stored `body_md` came back as the portal link paragraph followed by the two image paragraphs
- the logbook's own `/api/preview` rendered it as `<div class="figgrid"><figure><figure></div>` — the grid, confirmed end to end

**Owner-tested 2026-09-15**: the maintainer drove the finished button in a browser — the PR's portal running locally against the real lab Tiled and the real logbook at `:8400` — and accepted it. So the send path, the dialog, the entry and the grid are all exercised end to end against production services; what is *not* exercised is this code running as the deployed `geecs-data-portal` unit, which is the same code behind the same flag.

Superseded note: Not verified: the button and dialog in a browser. No Chrome extension was connected this session, so the front end is unexercised beyond a JS syntax check and the template assertions. **OWED: one click on the deployed portal** — send a plot from a scan's Plot tab, confirm the entry opens and the image is there, then send a second and confirm the grid.

## Deployment note

Nothing to change for hosts that already pass an absolute `--logbook-url` (the current shape: `http://<host>:8400`). `httpx` moves from a dev dependency to a runtime one — the one call this process makes out of itself — so the portal's env needs a reinstall, not just a restart.


skbarber:
## Adversarial review — full report and dispositions

A fresh-context reviewer was given the diff and the standard three-lens brief (correctness / redundancy / placement), with no access to the PR body's framing. Two rounds: nine findings, then a re-review of the fix commit. Both reports are reproduced in full below, followed by the disposition of every finding.

### Disposition summary

| # | Finding | Disposition |
|---|---|---|
| 1 | A non-JSON peer reply surfaced as a 400 "bad image" | **Fixed** — `722a51a3`; reviewer re-ran their reproducer, now 502 |
| 2 | `GeecsLogbook/CLAUDE.md` + fleet map still said "only links here" | **Fixed** — `722a51a3` |
| 3 | "One entry per scan" is really per-browser | **Fixed (docs)** — `722a51a3` + `5f222d37`. Behaviour deliberately unchanged; rationale below |
| 4 | `test_the_patch_uses_the_version_from_after_the_upload` did not bite | **Fixed** — `722a51a3`; reviewer built their own mutation and confirmed it fails now |
| 5 | `--logbook-url` serves two audiences; docs described one | **Fixed** — `722a51a3` |
| 6 | Whitespace-only author mapped to 502 | **Fixed** — `722a51a3` |
| 7 | `entry` was an unvalidated path segment | **Fixed** — `722a51a3` |
| 8 | A logbook client belongs in the logbook's package behind an extra | **Waived** — reasoning below |
| 9 | The send route stat-ed the SMB share for a folder it discarded | **Fixed** — `722a51a3` |

**#3, why the behaviour was not changed.** The entry to append to is remembered per browser, so two people plotting the same scan from two machines get two entries and no grid. The alternative the reviewer identified — asking `GET /api/day/{day}/entries?book=scans` — was put to the maintainer and not taken: it would append a plot into whatever note an operator happened to be writing. All three documents that stated it (`GEECS-DataPortal/CLAUDE.md`, the CHANGELOG, and `logbook_send.py`'s own module docstring) now say "per browser" and record the rejected alternative so it isn't re-derived as an oversight.

**#8, waived.** The module encodes the logbook's write protocol inside the consumer, where the repo's precedent (`geecs_bluesky.qs_client`) is a client shipped from the service's own package behind an extra. Accepted as-is for one writer: the current root `CLAUDE.md` edge forbids the portal importing GeecsLogbook, and this PR amends that edge explicitly rather than evading it. The trigger for moving it is a **second** Python writer — the logbook's owed doc-export phase, ScanAnalysis publishing summary figures, or GEECS-MCP — at which point `geecs_logbook.client` behind a `client` extra is the home, and the `_IMAGE_RUN` rendering rule in particular belongs beside `render.py`, which enforces it. Cheap to veto: say so and it moves now.

The reviewer also recorded one item explicitly *not* raised as a finding: `author` accepts an embedded newline, which `mirror.render` would write into the markdown mirror's `key: value` front matter. Not actionable here — the logbook's own `EntryCreate.author` has the same tolerance, so this PR adds no reach a lab-network client did not already have against an unauthenticated `POST /api/entries`. If ever fixed it belongs in `geecs_schemas.log_entry`.

### Verification beyond the suite

- The four-call conversation was driven against a **real GeecsLogbook 0.11.0** on loopback (throwaway store, since removed): two sends produced one entry with `plot.png` and `plot-2.png`, and the logbook's own `/api/preview` rendered the stored body as `<div class="figgrid">…</div>` — the grid, end to end.
- The maintainer drove the finished button in a browser against the real lab Tiled and the real logbook, and accepted it.
- New tests were checked by mutation rather than assumed to bite: the paragraph separator, the 409 retry, the send gate, the post-upload version assertion, and both validation patterns each fail the intended test when the code is broken.

**297 passed** in GEECS-DataPortal (41 in `tests/test_logbook_send.py`).

---

## Round 1 report (verbatim)

Adversarial review of PR #917 (`feat/portal-send-to-logbook`, GEECS-DataPortal 0.28.0 — "send a plot to the scan's log entry").

Verified first: full portal suite green (291 passed) in the worktree's own env; no `mkdir`/`ScanPaths(read_mode=False)` anywhere in the diff (scan-folder invariant holds, and `TestReadOnlyDoctrineUnchanged` pins it); no lab hostnames/home paths added (`logbook.example:8400`, `<host>`); version bump + CHANGELOG present and accurate; `httpx` dev→main move is correctly relocked (portal has its own env, root lock rightly untouched); the diff's assumptions about the logbook contract check out against `GeecsLogbook/geecs_logbook/routes/entries.py` (`EntryCreate` fields, `LogEntry.entry_id`/`version`, 409 dict detail), `routes/attachments.py` (404-on-missing-entry, the store-claimed filename in `link`, 20 MiB cap), `mirror.attachment_link` and `render.py` (`_IMAGE_RUN` really does want *consecutive `<p></p>` paragraphs*, so the blank-line separator and the link-above-the-images shape are correct, and `nh3` neutralises any `javascript:` `source_url`).

Findings, ranked.

**1. (Medium — proved by probe) A peer reply the portal can't decode is reported to the user as a 400 "bad image".**

`app.py:1287-1291` wraps `decode_png_data_url()` **and** `send_plot()` in one `except ValueError → 400`. `logbook_send._checked()` calls `response.json()` on every 2xx, and `json.JSONDecodeError` is a `ValueError`. So any non-JSON success reply from the logbook side lands as a client error.

Concrete scenario: `--logbook-url` points at a front door / wrong service that answers `200 text/html` (a proxy maintenance page, a mis-set port). I ran it with a mock transport returning `200 <html>…`:

```
status: 400
detail: {'detail': 'Expecting value: line 1 column 1 (char 0)'}
```

The browser then flashes `could not send: Expecting value: line 1 column 1 (char 0)`. This contradicts both the route's own docstring ladder ("a malformed image → 400 … the logbook refusing → … else 502") and the test class name `TestRouteErrors: Every failure lands as itself, never as a 500`. Same shape for a `KeyError` on `created["entry_id"]` / `uploaded["link"]` / `current["version"]` under portal↔logbook version skew (the fleet has run services on different branches before) — that one surfaces as a portal **500** with a traceback.

Fix: decode outside the send try-block (its own `try/except ValueError → 400`), and have `_checked` turn a non-JSON/missing-key body into `LogbookRefused(502, …)`.

**2. (Medium) `GeecsLogbook/CLAUDE.md` still says the portal only links here — a contract file that did not travel with the behaviour.**

`GeecsLogbook/CLAUDE.md:476`: *"that mount and the portal's `log` extra are gone, and the portal only \*links\* here (`--logbook-url`)"*. After this PR the portal creates entries, uploads attachments and patches bodies in that service. The PR updates root `CLAUDE.md`, `GEECS-DataPortal/CLAUDE.md` and `GEECS-DataPortal/DEPLOYMENT.md` but leaves the peer's own doc asserting the opposite — the next session reading the logbook package will get the wrong picture of who writes to its store.

Adjacent, same class: `docs/platform/fleet_map.md:113` labels the edge `portal -. "run page → scan card (--logbook-url)" .-> logbook` and line 149 describes the portal as *"links to the logbook (`--logbook-url`), no longer hosts it"*. Both now understate a write edge between two services. (Host/port/health are unchanged, so DEPLOYMENT.md's own "update the fleet map in the same PR" trigger doesn't strictly fire — but the arrow label is the thing that's now wrong.)

**3. (Medium) "one entry per scan" is a browser-local promise, and the logbook already exposes the API that would make it real.**

`GEECS-DataPortal/CLAUDE.md` (new text): *"the image is appended to ONE entry per scan"*; `logbook_send.py` module docstring: *"A second plot sent to the same scan joins the first **in one entry**"*. The entry identity actually lives in `localStorage["geecs-portal:logentry:" + UID]` (run.html). So:

- Operator A sends a plot from the control-room PC, operator B sends another from a laptop → **two entries for the same scan, no figure grid** — precisely the layout the docs say this design exists to produce.
- Same for a second browser, a private window, or cleared site data.

The CHANGELOG gets this right ("The page remembers the entry it made (per scan, in the browser)"); the two architecture docs do not. Worth noting that `GET /api/day/{day}/entries?book=scans` exists (`GeecsLogbook/geecs_logbook/routes/day.py:110`, backed by `store.for_day`, and `LogEntry.scan` is right there) — the portal could ask the logbook "does this scan already have an entry?" instead of trusting the browser. That may be deliberately rejected (appending a plot into an operator's unrelated note is its own hazard), but then the docs should say "per browser", not "per scan".

**4. (Medium-low — proved by mutation) `test_the_patch_uses_the_version_from_after_the_upload` does not bite.**

The test docstring claims *"Reading the version before uploading would 409 — the fake checks"*, but `_append_image`'s retry loop re-reads on every attempt, so the stale-version bug self-heals on attempt 2 and the assertion (`body_md.count("![") == 1`) still holds. I built a mutated copy of `logbook_send.py` that captures the version *before* the upload and patches with it, ran it against the PR's own `FakeLogbook`:

```
assertion body_md.count('![') == 1 -> True
BODY: '[Plotted in the data portal](…)\n\n![jet_pressure vs shot](attachments/e0001/plot-1.png)\n'
```

i.e. the named bug is present and the test passes. To make it bite, the fake needs to record PATCH attempts and the test assert exactly one (no 409 on the happy path). (`test_a_conflict_is_retried_against_the_fresh_body` *does* bite — I checked that one separately.)

**5. (Low-medium) `--logbook-url` now carries two meanings and the docs describe only one of them.**

The new DEPLOYMENT.md text tells the admin *"the portal has to dial the logbook itself"* and `app.py`'s comment says sending needs *"an address the PORTAL can reach, not one the browser can"*. But the same string is still the base for the browser-facing run-page link, **and** `run_logbook_send` returns `url = f"{logbook_send_base}/entry/{id}"`, which `run.html` hands to `window.open`. So `--logbook-url http://localhost:8400` (the natural reading of the new guidance on a single-host deployment, and the form `GeecsLogbook/deploy/DEPLOYMENT.md:160` uses in its own curl example) satisfies the send gate, passes every test, and gives every operator a dead link and a dead "Open the entry" button. One sentence — "the absolute URL must be reachable from both the portal process and the operators' browsers" — closes it.

**6. (Low) A whitespace-only `author` is misclassified as the peer's fault.**

`PlotToLogbook.author` is `min_length=1`, so `"   "` validates; `app.py` then does `payload.author.strip()` → `""` → the logbook's `EntryCreate` (`author: min_length=1`) returns **422**, which is not in `(409, 413, 415)` and so maps to **502 bad gateway**. The UI can't produce it (the JS trims first), but the route's contract says 502 means "the peer's own trouble", and this is ours. `Field(min_length=1, pattern=r"\S")` or stripping before validation fixes it.

**7. (Low) `entry` is an unvalidated path segment.**

`PlotToLogbook.entry: str = Field("", max_length=64)` goes straight into `f"{base}/api/entries/{entry_id}/attachments"` and the GET/PATCH URLs. A value containing `/` or `..` addresses other paths under the logbook's base. Bounded and low-value (the logbook API is unauthenticated on the lab network anyway, so a caller who can reach the portal can usually reach the logbook directly, and the route requires a valid run uid) — but the ids are `uuid4().hex[:12]`-shaped, so `pattern=r"^[0-9a-f]{0,64}$"` is free.

**8. (Low, placement) A Python logbook client in the consumer, against the repo's own `qs_client` precedent.**

`logbook_send.py` encodes the logbook's *write protocol* — entry-create field names, the attachment reply's `link`, the optimistic-concurrency `expected_version` dance, the "images must be consecutive paragraphs" rendering rule — inside GEECS-DataPortal. The same knowledge already exists once, in `GeecsLogbook/geecs_logbook/static/editor.js` (`ensureEntry` + `uploadFiles` + `noteVersion`, lines 368-458), and the repo's established answer for exactly this shape is a client module shipped **from the service's own package behind an extra**: `geecs_bluesky.qs_client`, which GEECS-MCP and GeecsScanner both consume (root CLAUDE.md dependency graph). I don't think this blocks the PR — the current root CLAUDE.md edge explicitly forbids the portal importing GeecsLogbook, and the PR amends that edge honestly rather than sneaking past it — but this is the piece a future feature moves. The moment there's a second Python writer (the logbook's owed LogMaker doc-export phase, ScanAnalysis publishing summary figures, or GEECS-MCP), `geecs_logbook.client` behind a `client` extra is the home, and the rendering rule in particular belongs next to `render.py`, which is the module that enforces it.

I searched for prior art before concluding there's none to reuse: no other Python caller of `/api/entries` anywhere in the repo (only `GeecsLogbook`'s own tests and docs), `logbook_send.py` is the only `import httpx` in `geecs_portal/`, and there is no existing service-client helper in the portal. So the module is new code, not a duplicate — the finding is about where it will have to live later.

**9. (Low) The send route does share I/O it doesn't use.**

`run_logbook_send` calls `_resolved_folder(detail, day)` and discards the folder; only `run_day` is used. `resolve_scan_folder` does `is_dir()` stats on the SMB share (`GEECS-Data-Utils/geecs_data_utils/tiled_catalog.py:540-555`). `_run_day(detail, day)` alone is the value wanted, and it keeps a logbook write from touching the scans mount at all — which is also what `TestReadOnlyDoctrineUnchanged` claims ("no filesystem path is ever handed to it") but doesn't currently check.

**Nits (no action needed, recorded only):** `confirmSendToLogbook` uses `plotHost()`, whose contract is "prepare the div for `Plotly.react`" (it wipes `innerHTML` when a `.plotmsg` is present) rather than "get the graph div" — unreachable today because `showPlotError` purges the modebar, but a plain `document.getElementById("plotdiv")` is what's meant. And `TIMEOUT_SECONDS`' docstring ("short enough that a wedged logbook does not hold a portal worker thread") is per-call: worst case is 6 calls × 15 s ≈ 90 s of threadpool hold.

**Redundancy verdict:** nothing here duplicates existing code. Files checked: `GeecsLogbook/geecs_logbook/routes/{entries,attachments,day}.py`, `render.py`, `mirror.py`, `store.py`, `static/editor.js`, `GeecsBluesky/geecs_bluesky/qs_client/`, `GEECS-DataPortal/geecs_portal/*.py`, and a repo-wide grep for `/api/entries` clients. The one near-duplicate the repo already knows about — `editor.js`'s `copyText` vs `run.html`'s `copyPlotImage`, logged as deferred debt in `GeecsLogbook/CLAUDE.md:460` with `kit.js` named as the eventual home on the *third* surface — is touched by this PR but not made worse: it stays two copies, not three.

---

## Round 2 report on the fix commit `722a51a3` (verbatim)

Re-reviewed `722a51a3` by re-reading the code (not the summary). **8 of my 9 findings are confirmed fixed; 1 (#3) is fixed in two of three places.** One new observation, below the bar for action, is noted and explicitly not a work order.

Full suite re-run in the worktree env: **297 passed**.

### Your three specific questions

**(a) Does `_field` cover every reply-key read? — Yes, complete.**

I audited every read of a decoded peer body in `logbook_send.py` rather than trusting the call sites named in the message. Grepping for subscripts and `.get(` on reply dicts across the whole module gives exactly five reads, and all five are safe:

- `created["entry_id"]` → `_field` (line 308) ✓
- `uploaded["link"]` → `_field` (line 361) ✓
- `current["version"]` → `_field` (line 389) ✓
- `current.get("body_md", "")` (line 387) — defaulted, and now wrapped in `str()`, so a peer returning `body_md: null` or a number no longer poisons `appended_body` ✓ (that `str()` wasn't in your summary; it's a real extra catch)
- `_detail_of`: `body.get("detail")` guarded by `isinstance(body, dict)`, the nested `detail.get("message")` guarded by its own `isinstance` — and its `response.json()` already had a `ValueError` arm ✓

The only remaining `["` in the file is line 190, the caption bracket-sanitiser, which isn't a reply. Nothing in `app.py` reads a reply key either — it reads `SendResult` dataclass attributes. **No gap.**

Small note, not a defect: `_checked(patch)` at line 393 discards its return value but still parses the body, so a PATCH answering 200 with non-JSON becomes a 502 rather than being ignored. That's the conservative direction; leave it.

**(b) Is `^[0-9a-f]*$` right? — Yes, on both counts, and I verified rather than reasoned.**

*The alphabet:* `GeecsLogbook/geecs_logbook/store.py:515` generates `entry_id=uuid.uuid4().hex[:12]` — the only producer in the package (I grepped for others; the two other hits are `url_for` placeholders). Corroborated independently by `render.py`'s `_ENTRY_REF = r'<a href="entry/([0-9a-f]{6,64})"'`, which anchors to the same alphabet and the same 64 ceiling your `max_length` uses. So `[0-9a-f]` with `max_length=64` matches the logbook's own tolerance — deliberately looser than the exact 12 chars, which is right: pinning `{12}` would break the day the logbook widens ids, and `render.py` already says it tolerates 6–64.

*The empty string:* `*` rather than `+` is **required**, not merely permissive — `run.html`'s `confirmSendToLogbook` posts `entry: target === "same" ? remembered(ENTRY_KEY) : ""`, so `""` is an explicitly supplied value that goes through validation (pydantic doesn't validate unsupplied defaults, but this one is supplied). `+` would 422 every "a new entry" send.

I probed the actual model under the installed pydantic 2.13.5:

```
ACCEPT ''  'e1'  'abcdef123456'  64×'a'
reject 'ABCDEF123456'  'e1/attachments'  '../../api/entries'  'e1%2Fx'  'e1.x'  65×'a'
reject 'abcdef\n'  '\nabcdef'  'abcdef\nx/y'
```

The newline cases are the one place `^…$` could have leaked: Python's `re` lets `$` match before a trailing newline, so `"abcdef123456\n"` would have slipped through as a URL path segment. Pydantic v2 defaults to the Rust regex engine, where `$` is end-of-text — so it's rejected. Confirmed empirically, not assumed.

`author`'s `pattern=r"\S"` also behaves as intended (`"   "`, `"\t"`, `"\n"` rejected; `" a "` accepted — pydantic patterns are search-semantics, which is what an unanchored `\S` needs).

**(c) Does the spy pin what it claims? — Yes, and it bites. I checked the one way it could have been vacuous.**

The failure mode I went looking for: `app.py` does `from geecs_data_utils.tiled_catalog import resolve_scan_folder`, so patching the *source* module would leave `app.py`'s own binding untouched and the spy would never fire — a test that passes for the wrong reason. It doesn't do that: it patches `app_module.resolve_scan_folder`, and `_resolved_folder` looks the name up as a module global at call time, so the patch lands. `monkeypatch.setattr` runs *before* `_client()`, and the positive control (`client.get(f"/run/{_UID}")` → `assert calls`) proves the spy is wired on the very run that asserted `calls == []`. That control is what makes it non-vacuous, and it's the right shape.

I then mutated the route back to `_resolved_folder(detail, day)` and re-ran: **FAILED** `TestTheSendTouchesNoShare::test_no_scan_folder_is_resolved`. It bites.

Can it pass while a folder is resolved by some other path? In principle the spy covers only `resolve_scan_folder` — but I read the route body, and the only other things it calls are `_load_run` (catalog, not share), `_run_day` (pure: reads `summary.start_time` and parses the `day` string — `app.py:193-213`, no filesystem), `_logbook_sendable` (pure), `decode_png_data_url`, and `send_plot` (stubbed, and separately pinned as URL-only by `test_the_send_touches_no_scan_folder`). So the class name "touches no share" is honest for the route as written, and any future share access would have to be a *new* call the spy doesn't know about — the usual limit of a spy, not a flaw here.

### Confirmations of the rest

**#1 — confirmed fixed.** I re-ran my original reproducer (mock transport answering `200 <html>maintenance</html>`) against the fixed code: it now returns **502 / "the logbook answered 200 but not JSON"** instead of **400 / "Expecting value: line 1 column 1 (char 0)"**. The decode now sits in its own `try/except ValueError → 400` above the conversation, so the two arms can't cross. `_checked`'s non-dict guard additionally catches a 200 returning a JSON *array* or bare string, which I hadn't raised.

**#4 — confirmed fixed, verified independently.** I built my own mutation (attempt 1 patches with `version - 1`) against the current file and ran the suite: `test_the_patch_uses_the_version_from_after_the_upload` **FAILS** (`patch_attempts == 2`), while `test_a_conflict_is_retried_against_the_fresh_body` still passes — so the new assertion is catching the stale-version bug specifically, not just restating the retry test.

**#6, #7 — confirmed fixed** (evidence in (b) above; both now 422 at the portal boundary, `assert not stub.calls` proves nothing is forwarded).

**#9 — confirmed fixed** (evidence in (c)).

**#2 — confirmed fixed, and better than I asked for.** `GeecsLogbook/CLAUDE.md` now describes the write edge *and* names the two of its own rules that shape the caller ("an attachment needs an entry to exist first, and `render.py`'s `_IMAGE_RUN` grids only *consecutive* image paragraphs… Changing either is a change to that caller"). That's the part that makes the doc load-bearing rather than decorative — a future change to `_IMAGE_RUN` now has a pointer to who breaks. `fleet_map.md` fixed in both places (arrow label and the portal's service row, the latter now also carrying the both-audiences reachability constraint).

**#5 — confirmed fixed.** DEPLOYMENT.md names `http://localhost:8400` as the trap explicitly and says why it passes every gate. Good — the failure mode is invisible to tests, so naming it is the only defence.

### The one thing still open

**#3 is fixed in `GEECS-DataPortal/CLAUDE.md` and the CHANGELOG, but not in `logbook_send.py` itself.**

`GEECS-DataPortal/geecs_portal/logbook_send.py:36-45`, unchanged by the fix commit, still reads:

> *Why append rather than replace*
> **A second plot sent to the same scan joins the first *in one entry*** …

That is the per-scan claim, in the module that implements the per-browser behaviour — and it now sits one file away from the new `CLAUDE.md` sentence that says *"Do not upgrade the doc to claim 'one entry per scan' without changing that."* So the PR's own documents contradict each other, and the contradiction is in the file a future reader opens *first* when they want to understand this design. The heading is fine and the rest of the paragraph is fine; it's one clause. Suggested: "A second plot sent from the same browser joins the first *in one entry*…" — and the `entry_id` parameter docstring below it ("Append to this entry instead of creating one") is already correct, so nothing else in the module needs touching.

Severity: low, but it's the same class of finding as #2 (which you accepted and fixed well), and it's a one-word fix, so I'd rather flag it than let the third copy drift.

### Not a work order

`author` accepts an embedded newline (`"a\nb"` validates — `\S` is a search). `mirror.render` writes `author: {value}` into the markdown mirror's `key: value` front matter, so a newline there would split the front matter of the mirrored file. I'm **not** raising this as a finding: the logbook's own `EntryCreate.author` has no such constraint either, so its composer's API can do exactly the same thing, and this PR adds no reach a lab-network client didn't already have against an unauthenticated `POST /api/entries`. If it's ever worth fixing it belongs in `geecs_schemas.log_entry`/`EntryCreate`, not here. Recorded so it isn't re-derived; explicitly not a request.

No other new problems introduced. Nothing else survives refutation.

---

That last open item is fixed in `5f222d37`.

skbarber:
## Second reviewer — two findings, both confirmed and fixed (`e6986eb7`)

A second independent reviewer (Codex) read the diff after the round-2 fixes. Both findings were verified against the code before acting; both were real, and the first is the most serious thing found on this PR.

### 1. An append that could erase the note it appends to — **fixed**

> `GEECS-DataPortal/geecs_portal/logbook_send.py:387` — `_append_image()` defaults missing `body_md` to `""`. If the logbook GET response is version-skewed or malformed but still includes `version`, the portal will PATCH the entry with only the new image and erase existing note text/images. This should use `_field(current, "body_md")` and reject unexpected/missing body content as a 502, like the other required fields.

Confirmed. `_append_image` is the read half of a read-modify-write, and its entire safety argument — restated in the module docstring and in this PR's description — is that the body it PATCHes back is the body it just read. `.get("body_md", "")` breaks that argument silently: it turns *"the body did not come back"* into *"the body is empty"*, and the PATCH on the next line makes that true. The asymmetry was backwards — `version` was guarded by `_field` while the one field whose loss destroys human text was defaulted.

Fixed exactly as suggested. Worth recording why demanding the key is safe: an entry legitimately **has** an empty body (one created with no `source_url` starts with `body_md: ""`), and `geecs_schemas.log_entry` declares `body_md: str = Field("")` — so an empty body arrives as a *present* key with an empty value, which `_field` accepts. Only a missing key or a non-string value now raises, and a non-string raises rather than being `str()`-ed, which would have written the literal word `None` over a paragraph.

Pinned by `TestAnAppendNeverErasesText`, three cases: an entry with an empty body still appends normally; a reply without `body_md` refuses with 502 **and the prose is still in the store afterwards** (the assertion is on the surviving text, not just the status); a null `body_md` refuses rather than stringifying. Verified by mutation — restoring `.get("body_md", "")` fails two of the three.

### 2. The send dropped the page's `?day=` — **fixed**

> `GEECS-DataPortal/geecs_portal/templates/run.html:607` — The new logbook POST drops the page's `day` query. `_run_day()` intentionally requires `day` when a run has no usable `start_time`, and the page can still render `logbook_send=True` using that query. The button will then appear but POST `day=""` and get 404. Existing `/api/run/...` JS calls add `DAY`; this fetch should do the same.

Confirmed, and the reasoning is exactly right. `_run_day` (`app.py:193`) deliberately refuses to fall back to today — scan numbers restart daily, so a bookmarked link must not resolve a different scan's same-numbered folder — which means a run with `start_time == 0` resolves *only* through the explicit param. The page context computes `logbook_send` with the day it was rendered with, so the button showed; the POST sent none, so every send 404'd. A button that is always visible and never works.

The fetch now builds a `URL` and sets `day` from `DAY`, matching the four other `/api/run` call sites in the template.

Pinned three ways: the page-level `logbook_send=True` plus a successful send *with* the param, the 404 without it, and a template assertion that the send fetch carries `searchParams.set("day", DAY)` — the last because the template is the only place this can regress. Verified by mutation.

### State

**303 passed** in GEECS-DataPortal (47 in `tests/test_logbook_send.py`), lint clean.

Neither finding was reachable from the maintainer's browser test: #1 needs a skewed peer reply, #2 needs a run whose start document has no usable time. That is the case for having had two reviewers and a hand test rather than any one of them.

PR #918, opened 2026-09-15
GeecsBluesky 0.91.1: pseudo lifecycle lines visible, a scan point at 0 stays checked (#915); the pseudo arc brief folded into CLAUDE.md and deleted

### LINKED ISSUE #915: Pseudo positioners: stage/unstage log lines are invisible, and a scan point at exactly 0 takes the restore bypass
Two small follow-ups from the pseudo arc's hardware acceptance (#904, 2026-09-15, Scans 5–9 of 26_0915). Neither affected the runs; both are observability/tidiness.

1. **The capture and restore log lines land nowhere.** `CaPseudoPositioner` logs `baselines captured (components zeroed)` at `stage()` and `restoring baselines` at `unstage()` at INFO, but stage precedes `open_run` and unstage follows `close_run`, so the per-run `scan.log` never sees them, and the worker's root logger is at WARNING outside a run, so the journal does not either. The acceptance had to prove the restore from the magnet readbacks alone. Options: have `make_run_engine` set the `geecs_bluesky` logger to INFO (independent of the root), or widen the scan-log handler's window to the staged bracket.

2. **A scan point at exactly 0 takes the recovery-bypass path.** `_set_and_wait` skips the agreement check and logs `restoring the owed baselines` whenever a relative pseudo is set to 0 while `_restore_pending` is true — which is every mid-scan point at 0 (Scan 5's third point), not only the unstaged `mv <pseudo> 0` recovery gesture it was written for. The move is identical either way; only the check is skipped and the message misleads. Gate the bypass on "not currently staged".

Also observed, for the record: steering-magnet readback scatter across a restore was ≤0.25 mA (S4H −0.09925 → −0.09901 → −0.09878 across three runs) against a DB tolerance of 0.01 A, so `DEFAULT_AGREEMENT_TOLERANCE` = 0.01 has an order of magnitude of margin on these supplies.

skbarber:
Closes #915. Two concerns, both housekeeping after the pseudo arc's hardware acceptance (#904):

## 1. The two #915 nits (GeecsBluesky 0.91.1)

- **Lifecycle lines visible.** `make_run_engine` sets the `geecs_bluesky` logger to INFO. The pseudo positioner's `baselines captured` (at stage) and `restoring baselines` (at unstage), and a manual `mv`'s moves, happen outside the run where the root logger sits at WARNING (the scan log lowers it only for the run's duration), so during the acceptance the restore had to be proven from magnet readbacks alone. A logger's level is checked once at the emitting logger, so this reaches the root's handlers (the journal) whatever the root's level; it does not change what the scan log captures. Test: `test_make_run_engine_lets_the_package_speak_at_info`.
- **A scan point at exactly 0 is an ordinary, checked step.** The agreement-check bypass for the recovery gesture (`mv <pseudo> 0` after a failed or skipped restore) now requires the pseudo to be unstaged (`_staged`, set in `stage()`, cleared in `unstage()`'s finally). Scan005 of 26_0915 had taken the bypass at its middle point — same move, but the check was skipped and the log line misled. Test: `test_a_scan_point_at_zero_is_an_ordinary_checked_step` (staged, hand move, then `mv 0` must fail the disagreement check; probe-verified to pass without the gate). The recovery test (`test_failed_restore_refuses_the_next_stage_until_moved_back_to_zero`) still passes: the recovery move is unstaged.

## 2. The executed brief, per the Planning rule

`Planning/native_bluesky/09_pseudo_transform.md` (416 lines) is deleted; the arc is done and hardware-accepted. What was load-bearing is folded into `GeecsBluesky/CLAUDE.md` as "Pseudo positioner rulings" (the bump-is-a-deviation ruling, no reference component, the disagreement fail/warn split and the recovery gesture, the catalog/maths split with geometry left to the physicists, the vocabulary, the follow-on arc). The plan of record's §8 pseudo section is reduced to a done-pointer; the module docstrings that cited the brief now cite CLAUDE.md; `01a` and `10` pointers repointed. CHANGELOG mentions stay (historical record).

## LOC

| concern | lines |
|---|---|
| nits (`pseudo.py`, `run_engine.py`, two tests) | ~+45 / −10 |
| CLAUDE.md fold + brief deletion + planning pointers | ~+40 / −470 |

## Tests

`scripts/check.sh GeecsBluesky`: COUNTS (posted below when the run finishes); `test_pseudo_positioner.py` + `test_run_engine.py`: 33 passed.

## Hardware verification

N/A for the logging change (a log level). For the bypass gate: the behaviour on hardware is unchanged for every run of the acceptance (the same moves), only the check now runs at a 0 point; a rerun of a bump scan through 0 would show no `restoring the owed baselines` line mid-scan. Not required.


skbarber:
Full `scripts/check.sh GeecsBluesky` on 860ff63a (own env): 602 passed, 3 skipped, 1 deselected, 4 warnings.

skbarber:
## Adversarial review (fresh-context subagent, on 860ff63a) — report in full

# Review of PR #918 (`fix/pseudo-nits-and-brief-cleanup` → `feature/native-bluesky-plans`), head `860ff63a`

Method: read the full diff and the deleted brief (`git show origin/feature/native-bluesky-plans:Planning/native_bluesky/09_pseudo_transform.md`), `pseudo.py`, `run_engine.py`, `scan_log.py`, the worker startup profile, bluesky's `RunEngine._run` exit path and the queueserver worker's logging setup in the venv; grepped the whole repo for the deleted file and every `setLevel`/`basicConfig` site; ran three probe scripts in the scratchpad against the test module's `Bench` (nothing in the checkout modified).

## HIGH — a staged scan point at 0 still clears `_restore_pending`, so the refuse gate is absent for the commonest bump-scan shape

`pseudo.py` `_set_and_wait` gates the *agreement-check bypass* on `not self._staged` (the PR's fix), but the bookkeeping at the end of the same method is untouched:

```python
if self._relative and value == 0.0:
    self._restore_pending = False   # "the recovery gesture after a failed or skipped restore"
```

That fires on a *staged* mid-scan point at 0 too. Probe (`scratchpad/probe_918b.py`, scenario A2): `bp.scan([], bump, -0.1, 0.1, 3)` on `ALine_e_beam_angle_offset_x` with baselines S3H 0.35 / S4H −0.099, the S4H put refused only on the restore at unstage:

```
A2: scan failed: RuntimeError LabVIEW: set refused
A2: S4H after failed restore = -0.299 (baseline -0.099)
A2: flags: _restore_pending= False _zeroed= True _staged= False
A2: NEXT STAGE ACCEPTED -> S4H baseline now captured as -0.299 (should be -0.099)
```

The next `stage()` re-zeroes with the leftover bump baked in — the exact incident class the refuse gate (#904) exists for, and the class the PR's own CLAUDE.md fold now states as a do-not-reopen ruling ("A restore that failed or was skipped … makes the next `stage()` refuse"). The behaviour is pre-existing (before the PR the mid-scan 0 took the bypass and cleared the flag the same way), but this PR is titled "a scan point at 0 stays checked", touches exactly this line, and Scan005 of 26_0915 — the PR's own motivating example — is a through-zero scan. It fixes half of the nit. `test_a_scan_point_at_zero_is_an_ordinary_checked_step` cannot catch it (it asserts the check ran, not the bookkeeping), and `test_failed_restore_refuses_the_next_stage_until_moved_back_to_zero` deliberately scans 0.05–0.1, never visiting 0.

Fix is one condition: clear `_restore_pending` only on the `restoring` path (i.e. `not self._staged`), plus a test in the shape of A2 (symmetric scan through 0, restore refused, `bps.stage` must raise `PseudoRestorePendingError`). Break-the-code check: with the clear left as is, that test fails; with it gated, it passes and the existing recovery test still passes (the recovery `mv 0` is unstaged, so `restoring` is True there).

## MEDIUM — "a halt skips unstage" is false for this device, and the fold writes it into CLAUDE.md as a ruling

`stage()`/`unstage()` docstrings (pre-existing), the CHANGELOG entry, and the new CLAUDE.md bullet all say a `halt` skips `unstage`, so the recovery gesture covers "failed or skipped (halt)". Bluesky's `RunEngine._run` `finally` (venv `bluesky/run_engine.py:1779`, "in case we were interrupted between 'stage' and 'unstage'") unstages every leftover staged object on **every** exit path, `PlanHalt` included — it just does not await the returned `AsyncStatus`. Probe (`probe_918c.py`): pause after `mv(bump, 0.1)`, then `RE.halt()`:

```
right after halt(): S4H = -0.299 _staged= True _restore_pending= True
1 s after halt():   S4H = -0.099 _staged= False _restore_pending= False
journal tail: [('U_S3H:Current', 0.35), ('U_S4H:Current', -0.099)]
```

The restore runs on halt, fire-and-forget. Two consequences the fold should carry instead of the wrong premise: (a) the recovery gesture is only ever for a *failed* restore; (b) a restore that fails during a halt is unobserved by the RE (no plan failure, only the loop's unhandled-task path), which is the one case where the operator must read the journal — the very lines nit 1 of this PR makes visible. Since the fold is explicitly "do not reopen", it should not enshrine a claim the code contradicts. (This also disposes of my original worry that `_staged` could stick True after a halt: it cannot, because the sweep clears it.)

## LOW (placement) — a process logging policy inside a library factory

`make_run_engine` now mutates the global `geecs_bluesky` logger level on every construction. Its one production caller is `qserver/startup/startup.py:165`, the worker's process entry — the natural home for process-wide logging policy (the readiness unit already configures its own in `qserver_ready.py:398`). As placed, every test that builds an RE (`test_run_action.py`, `test_connect_on_demand.py`, `test_run_engine.py`, the hardware suites) flips the level session-wide, and the new test leaves it at INFO after resetting to NOTSET. I checked the caplog assertions in `GeecsBluesky/tests` — all content-filtered, so no order-dependent failure today; it is a latent one. Redundancy checked: `scan_log.plan_report_sink` is the repo's existing "lift one logger to INFO" helper (scoped, restoring) — a different scope, not a duplicate; `_QuietNoisyLoggers` and the scan log's root save/restore are unaffected (level is checked at the emitting logger; `close()` restores the root only). The claim that the lines reach the journal holds: the queueserver worker does `basicConfig(level=max(WARNING, log_level))`, whose stderr handler is level NOTSET. Side effect worth knowing: `settable.py:127` ("setting X → Y", every component set outside a run) and the namespace's startup INFO lines now reach the journal too — modest, and arguably wanted.

## LOW (docs) — pointers left stale by this same PR

- `03_clean_room_rebuild.md` §8: "what it left open is #915 (observability nits)" — this PR closes #915.
- `10_web_scanner.md` §5: "Both are owed on this branch (the pseudo arc #904, done 2026-09-15; …" — owed and done in one sentence.
- `01a_device_layer_audit.md` `CaPseudoPositioner` row: the cells no longer match the columns (`Long-lived fitness` reads "Replaced by the pseudo arc…", `Verdict` reads "the namespace registers one per catalog entry").

No reference to `09_pseudo_transform.md` survives anywhere outside CHANGELOGs (grepped md/py/yml/json/html across the repo). The "Planning rule" the PR cites is `CONTRIBUTING.md` §"Planning/ is development scratch" (delete-when-executed, load-bearing content to the package CLAUDE.md first) — the PR follows it.

## LOW (fold completeness) — two things the brief kept "so a later session does not redo the survey"

The fold keeps the rulings well. Dropped: (1) the precedent research behind the named follow-on arc — spec gave every motor a user offset regardless of hardware, Sardana has `Offset`/`Sign` at the pool level with pseudo motors on top, EPICS confined the idea to the motor record — the session that builds set-as-aligned/persistence will re-derive it; (2) the "two meanings of relative" distinction (scan-choice `rel_scan` vs the catalog's `mode: relative` definition; R56 can be `rel_scan`ned but has no relative definition, the bump is the mirror), of which only "`scan` and `rel_scan` coincide" survives. Two lines in the follow-on bullet would keep both. Not blocking.

## Checked and clean

- Correct that `_staged` (not `_moved`) is needed for the bypass gate: a scan whose first point is 0 has `_moved` False.
- `locate()` untouched by `_staged`; abort-time unstage clears it in `finally`; the failed-restore recovery test's `mv 0` is unstaged and still takes the bypass (probe scenarios H/H2 confirm the gesture is accepted after a halt as well, once the sweep has run).
- Both new tests would fail with the respective line removed (probe-verified by the author for the gate; the level test trivially).
- Scan-folder invariant, public-repo hygiene: nothing in the diff touches either.

---

## Disposition (fix commit 3624e863; the reviewer's confirm/deny follows in a further comment)

| finding | disposition |
|---|---|
| HIGH staged 0 clears the owed restore | **fixed** — `_restore_pending` is cleared only on the `restoring` path (unstaged recovery move); new test `test_a_scan_through_zero_still_owes_its_restore` in the reviewer's A2 shape (symmetric scan through 0, restore refused → next `stage()` raises `PseudoRestorePendingError`), probe-verified to fail with the clear ungated; the existing recovery test still passes |
| MEDIUM "halt skips unstage" is false | **fixed** — module doc, `stage`/`unstage` docstrings, `PseudoRestorePendingError` docstring, CHANGELOG and the CLAUDE.md ruling now say the restore runs on every exit path (unawaited on a halt, so a failure there shows only in the journal) and the recovery is for a *failed* restore |
| LOW logging policy in a library factory | **fixed** — the `setLevel` moved from `make_run_engine` to the worker's process entry (`qserver/startup/startup.py`), pinned by the startup export test instead of a run-engine test (no session-wide flip from tests) |
| LOW stale pointers | **fixed** — `03` §8, `10` §5, and the `01a` table row (cells now match the columns) |
| LOW fold completeness | **fixed** — the offset precedent (spec / Sardana / EPICS motor record) and the two meanings of "relative" added to the CLAUDE.md rulings |

skbarber:
Full `scripts/check.sh GeecsBluesky` on 3624e863 (own env): 602 passed, 3 skipped, 1 deselected, 4 warnings.

skbarber:
## Reviewer confirmation of the fix commit 3624e863 (same fresh-context reviewer, independent re-read + suites)

1. **HIGH (staged 0 point cleared the owed restore) — CONFIRMED.** `_restore_pending` is cleared under `if restoring:` only. The reviewer overlaid the pre-fix `pseudo.py` onto a scratch copy and ran `test_a_scan_through_zero_still_owes_its_restore` against both: pre-fix `DID NOT RAISE FailedStatus` (the second stage was accepted); fixed: passed. The existing recovery test still passes (40/40 across the three suites).
2. **MEDIUM ("halt skips unstage") — CONFIRMED.** Module docstring, `stage`/`unstage` docstrings, `PseudoRestorePendingError`, CHANGELOG and the CLAUDE.md ruling now match the probe (restore lands ~1 s after `halt()`); surviving "halt skips" wording is historical CHANGELOG text and a true statement about plan finalizers in `plans/calibration.py`.
3. **LOW placement — CONFIRMED.** `run_engine.py` no longer touches logging; the profile sets the level before the RE is built; the startup export test asserts it after `runpy.run_path`. Nit taken: the test now resets the level to NOTSET first (dce102e7).
4. **LOW stale pointers — CONFIRMED.**
5. **LOW fold completeness — CONFIRMED.**

## Codex review (P2) — disposition

**P2 `stage()` could leave the recovery wedged if the parent stage failed after the offsets were captured** (`_staged` set before `await super().stage()`; a failure left it True, so the unstaged `mv <pseudo> 0` was not recognised and the next `stage()` kept refusing). Inside a RunEngine plan the abort sweep unstages the object and clears the flag, but a headless caller would wedge. **Fixed in dce102e7:** `_staged = True` is set only after the parent stage succeeds; `_restore_pending` stays owed from the zeroing on. Test `test_a_parent_stage_failure_after_zeroing_leaves_the_recovery_open` forces `StandardReadable.stage` to raise for the pseudo, asserts `_staged` False / `_restore_pending` True, then `mv 0` recovers and the next scan runs — probe-verified to fail with the pre-fix order.

All findings dispositioned (reviewer: 5 fixed; Codex: 1 fixed). Full suite and CI on dce102e7 follow.

skbarber:
Full `scripts/check.sh GeecsBluesky` on dce102e7 (own env): 603 passed, 3 skipped, 1 deselected, 4 warnings.

PR #920, opened 2026-09-16
Native optimization: schemas, strict worker plan and scanner Optimize mode

### LINKED ISSUE #880: Web scanner: Optimize mode (the greyed slot in the mode segment)
Design door kept open in `Planning/native_bluesky/10_web_scanner.md` §5. Optimization is **paused, not dropped** (Sam, 2026-09-13). The page shows Optimize greyed with a tooltip; `GET /api/configs/optimizer_configs` already lists configs with nothing behind them.

**Work, when the optimize plan exists on the native path** (`03_clean_room_rebuild.md` §8): one more name in `GEECS_PLAN_NAMES` with its kwargs (`optimizer_config`, `max_iterations`); one more form section keyed by plan name; progress reads `planned_total` from the start doc (adaptive totals render as "of ≤ N"); the Now panel's live values are a server-chosen list so an objective readout can appear. The API is plan-name generic — no mode enum to change.

Related: #744 (scalar-only optimization via the PV producer), #807.

skbarber:
Adds native optimization to the Bluesky worker and scanner: an `OptimizerConfig` selects measurements and an Xopt generator, the strict plan acquires one bin per iteration, and the scanner displays measured values and offers an explicit **Set to best** after successful completion. Replaces the legacy evaluator/configuration loop. Closes #880.

Targets `feature/native-bluesky-plans`. This is the complete optimization arc developed on `codex/native-optimization`; it includes the schema, worker and client changes together.

### Behavior and design decisions

- Analyzer specs declare their emitted scalars, checked against synthetic-frame execution. Ordered derived measurements support dotted symbols and an explicit trusted `python:` callable.
- GEECS-Schemas uses GEST's actual VOCS model, as approved, instead of maintaining a mirror. Heavy Xopt/ImageAnalysis dependencies remain in the worker's `optimize` extra.
- Live camera measurements use bounded PVA buffers and timestamp matching. The plan preserves strict acquisition, pause/refire semantics, actual readbacks, and in-folder Xopt dumps.
- Relative pseudo variables restore on unstage. Set to best shows and confirms exact physical targets from a successful run, keeps them visible after queue acceptance, records the operator, and expires after 15 minutes or a subsequent service-submitted move/action. The 15-minute limit is a revisitable judgment call; out-of-band gateway writes are not observed. Seed reuse across relative zeroing frames is refused. Observables-only BAX has no arbitrary best point.
- Worker and scanner share a Tiled-safe event-column codec. Unsupported column lengths or case collisions fail before scan claim.
- Six keeper configs are validated fixtures. The resolver lists only compatible native documents and the UI disables Optimize when none exist. Live corpus regeneration, including LEGACY headers on the HiResMagCam pair, remains a separate rollout step; the real corpus acceptance check currently **fails at 0/6 native configs** and must pass before rollout. Listing failures and unavailable-config reasons are shown in the scanner.
- All clients share optimizer preset expansion, required-device merging and defaults. Preset saves retain the authored document. Analysis measurements use the existing shared analysis-config root.
- Core owns the wire-format epoch constant; Data Utils independently owns the file-format constant and has **no intra-repo dependency**. A Bluesky test pins their equality. The Core/MySQL dependency added during the first review was reverted, along with the weakened architecture rules; all six affected offline-package locks are free of it.

### Scope (+ / − lines, including tests and package docs)

| Concern | Added | Removed |
|---|---:|---:|
| Deployment, CI and architecture guidance | 24 | 7 |
| Epoch contracts and dependency boundary | 111 | 13 |
| Dependency locks | 302 | 627 |
| MCP legacy retirement | 15 | 45 |
| Schemas | 939 | 739 |
| Worker and legacy retirement | 2,089 | 4,614 |
| Scanner | 828 | 26 |
| Analyzer declarations and parity | 76 | 1 |
| Brief and implementation evidence | 918 | 0 |
| Generated schema documentation | 444 | 363 |

Also updates root architecture guidance, bootstrap and CI worker installs to include the optimize extra. No new runtime facility literals.

### Validation

**3,407 tests passed across 15 suites; full-repository lint passed** (`scripts/check.sh --all`). Counts: root 83; ImageAnalysis 292; ScanAnalysis 192; Data Utils 400; Schemas 457; Bluesky 654; CA gateway 140; PVA gateway 51; Core 125; Data Portal 305; Log Triage 58; MCP 102; Logbook 274; scanner 92; shared theme 182. Hardware tests remain gated. Separately, the real optimizer-corpus integration gate correctly fails at **0/6 native configs**; deployment remains blocked on that migration.

The **11 codec/SQL archival tests** also passed in an isolated environment with Tiled 0.2.14. That server-dependent SQL test is intentionally skipped in the normal client-only worker environment. After the archival fix, **10 native optimization tests and 6 scanner optimization/boundary tests** passed before the full suites. Browser demo verified config loading, device locks, progress and Set to best. Independent adversarial reviews and dispositions are recorded in a PR comment.

Both GitHub lint runs and both unit-test runs passed on the reviewed head `1696f0b6`. The final reviewer confirmed all findings closed.

### Hardware verification

**Passed, beam off:** Scan010 of 26_0915, UID `53a0af7b-7101-4eab-b9a3-3f8ce92f0b42`: BAX simulation, **3 iterations × 2 HTU-NoGas shots**, real S1V/EMQ moves, finite synthetic observations, six correctly binned scalar rows and three Xopt dump rows. Both original magnet setpoints restored to **zero**.

The run exposed the `%` column-name incompatibility with Tiled SQL. Fixed here; replay against isolated Tiled 0.2.14 archived **6 primary, 3 optimization and 2 baseline rows**, plus successful stop. Original encoding reproduced the failure. No further hardware shots were taken. The original live Tiled entry remains partial; captured documents and restoration evidence are preserved inside Scan010.

**OWED:** TopViewMax with beam (**10 iterations × 5 shots**, five valid frames per iteration), live PVA timestamp join and cleanup, objective/image comparison, RSS, scanner submission and explicit Set to best, relative-pseudo hardware workflow, and real-camera BAX acceptance. Three iterations are a smoke test, not convergence evidence. No deployed service or live config tree was changed.

Full evidence: `Planning/native_bluesky/12_optimization_implementation.md`.


skbarber:
### Independent adversarial review and dispositions

A separate reviewer agent assessed correctness, redundancy and placement during implementation. Five concrete findings were fixed in `6d3b88b9` and returned to the reviewer for verification:

1. Partial/refired acquisition rows could displace the successful sample quota — evaluation now selects the latest complete successful quota.
2. Aliases or pseudo variables could overlap physical motors — overlap is rejected before scan claim.
3. Seed observations outside the current bounds could become the reported best — only in-bounds feasible candidates qualify.
4. Relative pseudo seeds could reuse a different zeroing frame — that seed reuse is refused.
5. Full camera descriptor reads could occur before prepare — descriptor access follows the device lifecycle.

The final implementation review reported no surviving P1/P2 findings. Hardware testing subsequently found the Tiled `%` identifier incompatibility; `33753b63` fixes it and adds SQL archival regression coverage.

The same independent reviewer examined that follow-up diff and reported:

> No surviving concrete findings.
>
> The codec is reversible and collision-safe, including literal escape markers and Unicode. The 60-character and case-insensitive uniqueness checks run before scan claim. Sharing this lightweight transport helper between worker and scanner fits the package boundaries and removes duplicated encoding logic.

No findings were waived. The later `b95ec760` change adds the optimize extra to CI and its environment-setup instructions; it does not change runtime behavior. Beam-on acceptance remains explicitly owed in the PR body.

skbarber:
## Adversarial review — `b95ec760` (five fresh-context reviewers, findings re-verified against the head)

Reviewed per `/land` § "The adversarial review step": five independent fresh-context reviewers over disjoint scopes (schemas · the optimize plan core · worker periphery, deletions and deploy · the scanner console · cross-cutting claims, locks and invariants), each given the verbatim three-lens brief and a P1/P2 severity floor. Every finding below was re-checked against `b95ec760` before being reported here; where I did not re-check personally it says so.

**CI note:** both `unit-tests` runs are against `b95ec760` itself and both are green, including the job that now installs `--extras optimize`. The head is covered.

### What survived attack

The plan core is solid. One reviewer's whole job was to break `plans/optimize.py`, and it failed on every axis it tried: the rewind fence around `driver.tell` (`Msg("rewindable", None, False)` empties `_msg_cache`, so pause/resume cannot double-count, and `trigger_and_read`'s internal `rewindable_wrapper(plan, True)` does not undo it); partial-bin and refire filtering (`deque(maxlen=shots)` + the all-timestamps-finite filter + the bin filter); the ±1 ms PVA frame join (no double epoch conversion — the CA gateway serves the raw LabVIEW value per `PV_CONTRACT.md` §3 while the PVA gateway serves it converted, and the code matches that split); stale-frame reuse (`frame_at` refuses ≠1 match; `_on_update` copies the array); NaN/inf never reaching `add_data`; `driver.best` label-vs-position indexing; relative-pseudo restore on end-of-scan, abort **and** halt, with `best_move:*` recording absolute dial positions; resource cleanup ordering; and the scan-folder invariant (`XoptDriver.dump` raises `FileNotFoundError` rather than creating a parent; no `mkdir(parents=True)`, no `read_mode=False`). The column codec round-trips injectively over `:` `.` `-` `~` `%` `--` and unicode, with the 60-char and casefold guards firing before scan claim. Several of the new tests genuinely bite rather than asserting numbers they construct.

Everything below is at the edges: deployment, config plumbing, and the surfaces around the plan.

---

## P1 — fix before merge

**P1-1 · The worker bootstrap never installs the `optimize` extra.**
`deploy/bootstrap_host.sh:68` maps `qserver → "ca tiled qserver"`, and `GeecsBluesky/qserver/deploy/DEPLOYMENT.md:63` still prints that same install line — while the new acceptance section appended by this PR (DEPLOYMENT.md:316) says *"Install the worker's `optimize` extra alongside `ca tiled qserver`"*. Three-way disagreement. A re-bootstrapped host (or a second GEECS site) binds `optimize` fine — the plan's Xopt import is lazy, so `geecs-qserver-ready` passes and the scanner's dropdown works — and the failure lands on the operator at submit time: `ModuleNotFoundError: xopt`, a failed item at the queue front, every later submission blocked until a human clears it. One word in `extras_of()` plus line 63. **CONFIRMED.**

**P1-2 · The scan-analysis configs path gets a second, positional home, and the deploy doc edited in this PR describes a knob the code never reads.**
`GeecsBluesky/geecs_bluesky/config_resolver.py:375` — `analysis_config_dir = self._root.parents[2] / "scan_analysis_configs"` — derives the analysis tree by walking three levels up from the *scanner* configs root, bypassing `[Paths] scan_analysis_configs_path` / `SCAN_ANALYSIS_CONFIG_DIR`, which `geecs_data_utils/config_roots.py:62,107-119` already owns and which ImageAnalysis and ScanAnalysis already read. `measurements.py:193` passes the derived path as `load_diagnostic(config_dir=…)`, overriding that default; `qs_client/presets.py:141` and `GeecsScanner .../service/scanner.py:413` reach it through `diagnostic_device`.

Meanwhile the DEPLOYMENT.md hunk in this same PR (lines 96-101) tells the operator the config key is "required by optimizer configs with diagnostic measurements". Setting it changes nothing.

Two concrete failures at any host where the two trees are not exact siblings — a staging checkout, `GEECS_SCANNER_CONFIG_DIR` pointed elsewhere, or another GEECS facility: (1) every diagnostic-based optimizer config dies at preflight with `diagnostic 'X': expected one document, found 0` and the doc sends the operator to a setting with no effect; (2) worse, both trees exist but differ, and the optimizer silently compiles its measurement from a *different* analyzer document than ScanAnalysis and the portal use — one diagnostic id, two definitions, no error. Root CLAUDE.md § "Facility values have one home". Found independently by two reviewers; the reference layout has the two trees as exact siblings, which is why the smoke test passed. **CONFIRMED.**

**P1-3 · Optimize goes live in the UI against a corpus that cannot load, and the corpus test is knowingly red.**
The live configs repo (`@ ad4e595`) holds **11** optimizer configs, all legacy dialect, none carrying a `# LEGACY` header. `optimizer_config.py:243-248` refuses any document with `evaluator` or `device_requirements` — i.e. all 11 — while `ConfigsRepoResolver.list_optimizer_configs` still lists all 11 and `console.html:136` un-greys Optimize mode in this same PR. Between merge/deploy and the separate configs-repo commit, an operator can select any listed config and gets a validation error.

Separately, `GEECS-Schemas/tests/test_corpus_integration.py:134-145` skips `# LEGACY`-headed files and asserts `validated >= 6` — a corpus state that does not exist yet, so that test fails today against the sibling configs checkout. It is `integration`-marked, so CI deselects it and CI green does not cover it. The PR body discloses the rollout step but not that the UI ships live-and-broken or that a test is red. **CONFIRMED** (corpus inspected directly).

**P1-4 · "Set to best" has no staleness gate and no success gate — both paths end in an unintended magnet move.**
`GeecsScanner/geecs_scanner/service/streams.py:120,203`, `service/scanner.py:436-448`, `static/scanner.js:721`.

`ProgressCache._optimization` is reset **only** in the `start` branch. `stop` sets `finished = bool(self._optimization.run_uid)` and then reads `exit_status` on the very next line for the progress state — deliberately not for `finished`.

- *Aborted run accepted as best:* a 40-iteration run loses the laser at iteration 2 and the operator stops it. `exit_status="abort"`, `finished=True`, `best_moves` holds iteration 2's best — for a `random` generator, essentially one arbitrary sample. The panel gives no sign the run aborted, and `set_optimization_best` refuses only with *"the selected optimization is not the latest completed run"*, which this run satisfies. Click moves the magnets to that point under the label "best".
- *No staleness at all:* `OptimizationOut` carries no timestamp, and nothing but a new `start` clears the record — not an `mv`, not a `run_action` (both are non-run items that open no run), not a gateway write, not twelve hours. A fresh page load the next morning always receives the first `optimization` frame, renders `config · iteration N / M` with no time and no scan number, and enables Set to best. `_require_idle` passes, `run_uid` matches (the browser echoes the same stale cache), magnets move to last night's optimum.

Gate on `exit_status == "success"`, put a timestamp and a bound on `OptimizationOut`, and invalidate the record on any `mv`/`run_action` the service itself queues — it already sees every one in `_queue_item`. Also worth noting: `SetBestIn` carries no `operator` field, so unlike `MoveIn`/`VerbIn` this multi-device hardware move leaves no operator string in the request. **CONFIRMED.**

**P1-5 · `"optimization" in data` rejects a key that is merely present-and-null, killing plain step scans.**
`GEECS-Schemas/geecs_schemas/scan_request.py:605`. The guard is membership, not `data.get("optimization") is not None`. Any client round-tripping `ScanRequest.model_dump()` from geecs-schemas ≤ 0.26 emits `"optimization": None` (SchemaModel sets no `exclude_none`), as does any JSON-Schema form builder materialising the previous artifact's declared `"default": null`. That is a plain step scan, and it now dies at `model_validate` with *"legacy optimization requests are retired"*. It also contradicts the module's own convention for removed fields fifteen lines below, in the same validator: for `trigger_variant`, "an unset one is dropped wherever it sits; a set one gets the v3 remedy". One-word fix. **CONFIRMED.**

---

## P2 — should fix

**P2-1 · `mode: "optimize"` is half-removed; the published contract and the MCP tool description still advertise it, and GEECS-MCP records none of it.**
`ScanRequestMode.OPTIMIZE` survives (`scan_request.py:90`) and so does its description in the regenerated `docs/geecs_schemas/scan_request.schema.json:190-195` — the artifact `test_schema_export.py` documents as what OSPREY's plan panel and other JSON-Schema clients build forms from. A client composes a schema-valid `{"mode": "optimize", …}` and gets `invalid_request: legacy optimization requests are retired… Planning/native_bluesky/11_optimization.md` — an internal repo path surfaced as a user-facing refusal. Residues: `GEECS-MCP/geecs_mcp/scans/control_tools.py:223` (the docstring an agent reads) still states "optimize submissions need an explicit max_iterations" as live policy; the `shots is None` branch at `:107-114` is now unreachable because the optimize arm of `planned_shots()` was deleted; and `scan_request.py:678-679` still tells noscan users their variables belong "inside the 'optimization' block". GEECS-MCP has **no `poetry version` bump and no CHANGELOG entry** despite its observable behaviour changing (its test was rewritten to assert the new refusal) — against root CLAUDE.md § Release & Versioning, and the one place an operator would look says nothing. **CONFIRMED.**

**P2-2 · `optimize` is filed in `NON_SCAN_PLAN_NAMES`, whose contract it violates on every count — and the diff already carves it back out twice.**
`GeecsBluesky/geecs_bluesky/plan_names.py:26-37`: the tuple's own docstring says these plans "open no run, claim no scan number and write no scan data, so a preset … still cannot express them". `optimize` opens a run (`bpp.run_wrapper`), goes through the claim preprocessor, writes `xopt_dump.yaml` into the scan folder, and *is* preset-nameable. The carve-outs are both inside this PR, in two packages: `qs_client/presets.py:66` (`n not in NON_SCAN_PLAN_NAMES or n == "optimize"`) and `GeecsScanner .../service/demo.py:666` (`and item["name"] != "optimize"`). Collateral: `tests/test_plan_registry.py:76` skips every `NON_SCAN_PLAN_NAMES` entry, so `optimize` is the one bound plan whose signature contract is pinned nowhere, and the refusal hint at `presets.py:191-200` still offers "(… is a queue item of its own)" for a name now inside `PRESET_PLAN_NAMES`. The next consumer that reads the list at face value — the portal or the MCP deciding whether a queued item produces a `scans/ScanNNN/` folder — mis-handles optimize, invisibly, because the constant's name says the opposite of the truth. **CONFIRMED.**

**P2-3 · Optimizer preset resolution lives in one client, so the same preset expands differently depending on who submits it — and the save path now persists the transformed document.**
`prepare_optimizer_preset` (`qs_client/presets.py:128`) resolves the `OptimizerConfig`, merges required devices as `essential=True, save_images=True`, refuses non-strict acquisition and fills the `shots_per_step`/`max_iterations` defaults. Its only caller is `GeecsScanner .../service/scanner.py:706`. `client.submit_preset` (`qs_client/client.py:733`) and `submit_preflight._preflight:147` both call `expand_preset` with the default `required_devices=frozenset()`. Concretely, the preset in `GeecsScanner/tests/test_optimization.py:10-18` expands through the scanner to `args[0] == ["U_BCaveICT"]`; through `submit_preset` the same document expands to `detectors=[]`, `non_essential=["U_BCaveICT"]` and is either refused client-side or queued and killed worker-side at `optimize.py:229` after it reached the queue. `expand_preset` already takes `required_devices` — that is the natural home.

Second half: because the transform now sits inside the *validator*, `save_preset` (`scanner.py:623` → `641`) hands the **prepared** model to `resolver.write_preset`, so the YAML committed to the configs repo is not the document that was POSTed — optimizer devices merged in with forced `essential`/`save_images`, `acquisition` stripped. If the optimizer config later swaps its diagnostic device, the stale one stays in the file as essential and image-saving, and `lockOptimizerDevices` only removes rows it added in the current session. A draft also can no longer be saved at all when the configs tree is unreadable or the config was renamed. `GeecsScanner/CLAUDE.md:133-135` documents the prepare step as running "before both preflight and submission" and says nothing about save. **CONFIRMED** (code path); the config-drift consequence is a reasoned inference.

**P2-4 · GEECS-Schemas gains an eagerly-imported third-party runtime dependency, on packages that gain nothing from it.**
`geecs_schemas/__init__.py:157` does a module-level `from .optimizer_config import …` (for `SCHEMA_REGISTRY`), and `optimizer_config.py:12` does `from gest_api.vocs import ContinuousVariable, VOCS`. So every importer of `geecs_schemas` now imports `gest_api` — GeecsLogbook, GeecsCAGateway, ImageAnalysis, ScanAnalysis, GeecsScanner — none of which will construct an `OptimizerConfig`. The locks confirm it (`GeecsLogbook/poetry.lock:410-415`: `optional = false, groups = ["main"]`). This is the one invariant the package exists to hold, and the diff resolves the conflict by editing the three documents that stated it (root `CLAUDE.md`, the README, the module docstring) to match the code. The pin is `>=0.1,<0.2` against a package with a single 0.1 release, so a production IOC now carries an upgrade coupling to a pre-1.0 optimization-standard package. Every other import in this diff is carefully deferred; this is the exception. Making `SCHEMA_REGISTRY` lazy keeps both properties. Flagged as a judgment call — the PR body says the GEST model was "as approved" — but the documentation moved to the code rather than the other way round. **CONFIRMED.**

**P2-5 · The one test of the magnet-moving path asserts the plan name and nothing else.**
`GeecsScanner/tests/test_optimization.py:63` ends at `assert result.plan == "mv"`, with the recording `manager` fixture in the signature and unused. A regression that passed the optimizer-space value instead of `best_moves`, dropped a component, mangled the reference through `scan_variable_reference`, or emitted `["U_S1H:Current", 0.25]` instead of the namespace reference `U_S1H.current` leaves this suite green — for the one code path in the package that moves magnets. Assert the recorded item's args. Secondary: the two `pytest.raises(…, match=…)` use `match="completed"` and `match="latest"`, but `set_optimization_best` raises one message containing both words, so they do not distinguish the two refusals. **CONFIRMED.**

**P2-6 · The six optimization role prefixes are duplicated across worker and scanner; the shared module carries only the escaping.**
`optimization_events.py` — imported by the scanner precisely to avoid duplicating this — exports only `optimization_column` / `optimization_name`. The role vocabulary (`proposal`, `measured`, `output`, `n_valid_shots`, `best`, `best_move`) is written as literals on both sides: `plans/optimize.py:296-302,373-397` and `GeecsScanner .../service/streams.py:172-180`. Rename one role worker-side and the scanner's reduction silently yields empty dicts — empty panel, Set to best disabled — with **no test failing**, because both suites hardcode the literals too. The role names belong beside `optimization_column` in the module both packages already import. **CONFIRMED.**

**P2-7 · Deleted modules are still described as live in two package docstrings.**
`optimization/generators/__init__.py:22-24,29-31,43-46,60-63` still documents the Cheetah generator, its `bayes_cheetah` factory name and a worked `>>> config = {"name": "bayes_cheetah"}` example that now raises `ValueError: Unsupported or undefined generator name`; `cheetah_generator.py` and the factory entry are deleted by this PR, and the file is docstring-only so nothing caught it. `optimization/inspection/__init__.py:7` still cites `BaseOptimizer.seed_from_dumps` as "production"; the seeder is now `XoptDriver.__init__`'s `seed_dumps` loop. A repo-wide grep of every deleted symbol found **no dangling imports** — these two docstrings and the historical `Planning/` briefs are the only surviving references. *(Reported by the deletion sweep; not re-checked line by line.)*

**P2-8 · A fourth home for `LABVIEW_EPOCH_OFFSET`, and one package now imports two of them.**
The constant already exists as public API in `geecs_data_utils.io.scan_stack` (consumed by the portal, ScanAnalysis, `GeecsBluesky/geecs_bluesky/callbacks.py:641` and `GeecsPvaGateway/tests/test_diff.py:12`) and privately in `GeecsCAGateway/geecs_ca_gateway/gateway.py:55`. This PR adds `geecs_core.db.variable_types:33` and switches the PVA server onto it while that package's own tests still import the data-utils one — and inside GeecsBluesky, `measurements.py:27` now imports from core while `callbacks.py:641` imports from data-utils: same number, same meaning, same package. GeecsCAGateway, the one consumer that actually justifies a core-level home (it depends on core but not data-utils), was not migrated. Not a behaviour change — same literal, and the PVA launcher reinstalls both intra-repo packages together from one clone, so a rolling restart is safe. **CONFIRMED.**

---

## P3 — noted, below the floor

- `qserver/startup/startup.py:236-246` warms `xopt`/`torch` in a **daemon thread** at the end of the worker profile. That is the shape of #778 (a background thread importing into the bluesky/numpy graph concurrently with the main thread → `_DeadlockError` / partially-initialized-module on the *other* thread), whose recorded fix was to import eagerly on the main thread before spawning. Rated PLAUSIBLE, not confirmed — no cyclic overlap was demonstrated. A synchronous import costs nothing here, since `geecs-qserver-ready` already gates readiness on the environment opening.
- `GEECS-MCP/poetry.lock` is relocked with no MCP pyproject change and picks up `p4p 4.2.2`, `pvxslibs`, `ply` and `nose2` in the `main` group (verified: absent from the base lock). It is a stale lock catching up to the `ca` extra rather than a new coupling, but a reinstall of the MCP host now pulls a compiled EPICS package it did not before, unrecorded. Related: `GeecsScanner/poetry.lock:900` still records the optimize extra as `gest-api (>=0.1)` while the pyproject now says `>=0.1,<0.2`.
- `Planning/native_bluesky/11_optimization.md:674-693` still instructs a future agent to land "four PRs, each landed per `/land`" with per-PR version bumps that do not match what shipped (§4.7 is titled "The four keepers" and lists six), and `12_optimization_implementation.md:88-90` asserts "No PR has been opened and no branch has been pushed". The rulings did graduate correctly into `GeecsBluesky/CLAUDE.md`, `GeecsScanner/CLAUDE.md` and `EVENT_SCHEMA.md`, which is the convention — the briefs themselves are now debt that will mislead the next agent that reads them.
- `console.html:147` gives `#optimizer-note` `class="hint"` outside a `.field`. The kit's only `.hint` rule is `.kit .field .hint`, and `scanner.css` adds none, so the required-devices disclosure renders at full body size and full contrast — louder than the labelled fields above it. Every other `.hint` in the file is a `<span>` inside a `.field`.
- `GEECS-Schemas/CHANGELOG.md` jumps 0.26.0 → 0.28.0 with no 0.27.0 entry (the Planning doc's milestone table names 0.27.0).
- Cleared on inspection, recorded so they are not re-derived: `restricted_expr.py`'s dotted-symbol support never emits `LOAD_ATTR` (every accepted `Attribute` is rewritten to a collision-free local before `compile()`, and the alias loop survives a measurement literally named `_geecs_symbol_0`); every `emitted_scalars()` declaration matches its analyzer, including the three the parity test skips; `GENERATOR_NAMES` is a hand-copy of `PREDEFINED_GENERATORS` but `test_native_driver.py:17` pins the two sets equal; `PythonDerived` is not browser-reachable (the config editor does not walk `SCHEMA_REGISTRY`); intra-repo `file://` URLs in every touched lock are the normalized form, with no worktree or home paths anywhere in the diff including the new `Planning/` documents; and version bumps + CHANGELOG entries are complete for every package whose source changed except GEECS-MCP (P2-1).

skbarber:
## Claude review dispositions

Addressed the review of `b95ec760` in `4aa610ab`. Thanks for the concrete failure cases: the surviving bugs were in deployment, configuration and client behavior around the plan.

### P1

1. **Fixed — bootstrap extra.** Both the qserver bootstrap and primary install instructions include `optimize`; a regression checks the bootstrap's actual extras function.
2. **Fixed — analysis root.** The resolver uses the existing `scan_analysis_config` manager, honoring `SCAN_ANALYSIS_CONFIG_DIR` and `[Paths] scan_analysis_configs_path`. Tests place scanner and analysis trees independently and cover both configuration mechanisms.
3. **Fixed availability; corpus migration deferred to rollout.** Listing validates native documents and excludes retired/invalid files. Optimize is disabled when none are compatible. A malformed YAML file also cannot hide valid neighboring configs. The integration check validates deployed native documents and reports an explicit migration-pending skip when none exist. The real local corpus produced that skip. Regenerating the six keeper configs and marking the HiResMagCam pair LEGACY remains the separately documented configs-repository rollout, not a completed acceptance claim.
4. **Fixed — Set to best.** The service requires `exit_status == success`, expires offers 15 minutes after completion, and invalidates them after a successful service submission of `mv`, `run_action`, or Set to best itself. The UI shows scan number, completion time, status and invalidation/expiry; the request carries an operator recorded with the physical targets in the service log. Tests cover abort/failure, expiry and subsequent moves/actions. The 15-minute limit is a revisitable judgment call. Direct gateway writes made outside this service remain unobservable and are explicitly documented as a limitation.
5. **Fixed — null legacy field.** Plain requests containing `optimization: null` validate; non-null legacy optimization and `mode: optimize` still receive the migration remedy.

### P2

1. **Fixed — retired public contract.** Removed `ScanRequestMode.OPTIMIZE`, regenerated JSON Schema, removed unreachable MCP budget logic and stale instructions, and bumped/documented GEECS-MCP 0.8.10 and GEECS-Schemas 0.28.1.
2. **Fixed — plan classification.** Added `NATIVE_SCAN_PLAN_NAMES`; optimize is no longer classified as a non-scan plan. Removed the two carve-outs and explicitly pinned its bound signature.
3. **Fixed — shared expansion and authored saves.** `expand_preset(..., resolver=...)` owns optimizer resolution, required-device merging and defaults. Direct submission, preflight and scanner use it; required devices are included in liveness checks. Saving validates the authored preset without looking up the optimizer or persisting runtime expansion. Regressions compare the shared paths and preserve a draft referencing an unavailable config.
4. **Retained by owner decision — direct GEST dependency.** The owner explicitly approved adding GEST to GEECS-Schemas in this task, replacing the proposed mirror. Eager import/upgrade coupling is real, but this is an intentional boundary change, not an accidental violation of the superseded pydantic-only rule. Keeping the real VOCS model and the shared registry is the chosen design; the `<0.2` constraint and documentation remain. A lazy registry is not introduced in this PR.
5. **Fixed — physical move assertions.** Tests inspect the recorded queue args, including both physical components of a best point with a deliberately different optimizer-space value. They also distinguish wrong-run, incomplete/failed and expired refusals.
6. **Fixed — role vocabulary.** Worker and scanner now use `OptimizationRole` beside the shared column codec.
7. **Fixed — obsolete docstrings.** Removed the Cheetah factory examples and replaced the deleted BaseOptimizer production reference with XoptDriver.
8. **Fixed — epoch ownership.** Core owns the constant. CA gateway, worker and PVA tests import it; Data Utils preserves its old public export as an alias, with a regression pin. This adds the explicit downward `GEECS-Data-Utils → GEECS-Core` dependency, including Core's MySQL connector installation dependency. Importing the timestamp contract itself is pure and performs no DB/network access. Dependency/architecture docs and all affected locks were updated.

### P3

- Numerical warm-up now runs synchronously on the worker profile thread before readiness.
- Locks were regenerated; Scanner carries the bounded GEST requirement, and MCP's changelog records the existing CA extra's PVA dependencies.
- Historical briefs identify the superseded four-PR sequence, six keepers, actual PR state and unreleased 0.27 milestone.
- The optimizer hint sits inside a kit `.field`.
- The schema changelog explains that 0.27 was an unreleased draft, rather than inventing a release entry.

### Verification

Final local verification: **3,391 passed across 15 suites**. Counts: root 83; ImageAnalysis 292; ScanAnalysis 192; Data Utils 398; Schemas 452; Bluesky 650; CA gateway 140; PVA gateway 51; Core 125; Data Portal 305; Log Triage 58; MCP 102; Logbook 274; scanner 87; shared theme 182. The first full run applied formatting fixes and found the existing portal environment missing the new Core dependency. After installing from the updated portal lock, all 305 portal tests passed; final lint passed separately. Hardware/integration tests remain gated or deselected. The real optimizer-corpus check reports **1 migration-pending skip**, not native-corpus acceptance.

A fresh-context independent reviewer checked this follow-up diff for correctness, redundancy and placement. It found the malformed-YAML listing failure above; the fix was returned to that same reviewer, which confirmed it and reported **no surviving code findings**. No hardware was contacted or additional shots taken for these review fixes. The original beam-on and corpus rollout acceptance remain owed, as listed in the PR body.

skbarber:
## Round 2 — confirm/deny on `4aa610ab`

Each round-1 finding went back to **the same reviewer that raised it**, with its own context, given only the fix commit and the disposition text (per `/land`: "a fix must go back to the same reviewer for a confirm/deny before it is dispositioned as fixed"). Load-bearing verdicts re-verified against `4aa610ab` before posting. **This is the final review round** — remaining items are listed once, ranked, and are the owner's call.

### Scorecard

| Round-1 finding | Verdict |
|---|---|
| P1-1 bootstrap `optimize` extra | **CONFIRMED FIXED** |
| P1-2 analysis-config root | **CONFIRMED FIXED** — the strongest fix in the commit |
| P1-3 corpus / Optimize live-and-broken | **PARTIAL** — UI hazard closed, see N2 and N3 |
| P1-4 Set to best staleness + abort | **CONFIRMED FIXED** — design better than specified |
| P1-5 `optimization: null` | **CONFIRMED FIXED** |
| P2-1 retired `optimize` contract | **PARTIAL** — enum fixed, prose residue ships |
| P2-2 plan classification | **CONFIRMED FIXED** |
| P2-3 single-client optimizer expansion + authored saves | **CONFIRMED FIXED** — all three entry points verified |
| P2-4 eager `gest_api` import | **WAIVED** — premises verified, owner's design call |
| P2-5 weak hardware-move test | **CONFIRMED FIXED** |
| P2-6 role vocabulary | **CONFIRMED FIXED** |
| P2-7 stale docstrings | **CONFIRMED FIXED** |
| P2-8 epoch constant | **behaviourally fixed — see N1, the fix costs more than the finding** |
| P3 warm-import thread, hint styling, briefs, locks, 0.27 | **FIXED** (briefs finding formally withdrawn) |

Several fixes landed at a **better seam than proposed**: `expand_preset(..., resolver=…)` owning optimizer resolution for all three entry points rather than patching the scanner; `QueueItem.devices` extending the preflight's liveness probe to optimizer-required devices (beyond what was asked); the bootstrap regression test extracting and executing the *real* `extras_of()` line rather than asserting on a copy; and the Set-to-best expiry computed server-side on every read, on the wall clock, with `completed_at = min(scanner_clock, worker_stamp)` so every skew direction shortens the window. Restart drops the offer rather than inheriting a stale one. That is careful work.

---

## N1 (P1) — the epoch fix inverted the repo's foundational layer to deduplicate one integer

`GEECS-Data-Utils/pyproject.toml:28` now declares `geecs-core = {path = "../GEECS-Core", develop = true}` as a **non-optional main** dependency so that `io/scan_stack.py:41` can re-export `LABVIEW_EPOCH_OFFSET`. Core declares `mysql-connector-python = "^9.7.0"` as a hard dependency, so the regenerated locks put **`geecs-core` and `mysql-connector-python` into the non-optional `main` group of GEECS-DataPortal, GEECS-LogTriage, GeecsLogbook, ImageAnalysis and ScanAnalysis** — verified in all five lock diffs. `GeecsLogbook`, whose stated boundary is "GEECS-Data-Utils (**ScanPaths only** — it reads scan folders and nothing else)", and `ImageAnalysis`, which analysts install on laptops, both now install a MySQL driver to obtain `2_082_844_800`.

Three policy documents were edited to match the code rather than the other way round:

- root `CLAUDE.md:150` — `GEECS-Data-Utils → (no intra-repo deps — foundational data layer)` → `→ GEECS-Core (…)`;
- root `CLAUDE.md:249` — "everything depends on it and **it depends on nothing else in the repo**" → "it depends only on `GEECS-Core`";
- `GEECS-Core/DESIGN.md:73` — the clause **"data-utils must stay dependency-free"** was *deleted* from the rationale for keeping the two packages' config reads separate, leaving only the network-access half. That clause is the load-bearing half of the argument the file exists to make; the next person proposing to unify the paths half now finds the prohibition already conceded, with a precedent.

Two further costs. It ships as **`0.32.0 → 0.32.1`, a patch**, where the repo's table reserves patch for "bug fix, no behaviour or API change" — this is a new mandatory dependency under five packages. And it is **no longer a worker-only deploy**: portal, logbook, MCP and qserver checkouts all need `poetry install` from updated locks. The author's own evidence confirms it — `12_optimization_implementation.md` records that the first full run "exposed the portal environment's missing newly declared Core dependency", i.e. the edge broke a deployed service's environment until relocked.

Stated fairly, the defence holds as far as it goes: `geecs_core/db/geecs_db.py:125` imports `mysql.connector` inside the connection helper, not at module scope, so `import geecs_data_utils` costs no MySQL import. The regression is in the **installed closure**, the **stated architecture** and the **semver label** — not in import side effects.

The owner's ruling on the direction of this edge, recorded here so the next reader has it: GEECS-Core is the *access* library — wire protocol, DB, PV naming. The epoch offset earns a home there because Core's own consumers read LabVIEW stamps off the wire. GEECS-Data-Utils reads **files**, and the fact that s-files, TDMS and native filenames carry LabVIEW seconds is a property of the file format it already owns — not something it should ask the network layer about.

**Recommended: revert the edge.** Core keeps its constant for Core's consumers; data-utils keeps its own literal; the cross-package equality pin already added at `test_scan_stack.py:230` is the only part of the fix that had to exist, and it works without any dependency. Then restore the three documents, and re-lock the six packages. A constant that has not changed since 1904 does not earn an architectural edge — the original finding said explicitly it was "not a behaviour change … a duplication a consolidation PR has to undo", and rated it P2 for that reason.

## N2 (P2) — the corpus check's failure floor became a green skip, so it can no longer fail for the thing it watched

`GEECS-Schemas/tests/test_corpus_integration.py:144-155` drops `assert validated >= 6` and adds a `continue` for retired-dialect documents plus a terminal `pytest.skip("optimizer corpus migration pending…")` when none validate. All 11 deployed configs still carry `evaluator:`, so `validated == 0` and the test **skips**. Zero native configs → skip; one → pass; six → pass. It still catches a *malformed* native document, but it can no longer catch the rollout never happening, being reverted, or landing partially — the one step still owed, with no issue tracking it. At `b95ec760` the same run failed loudly, which was the only automated signal that not a single deployed config is loadable by the new code. The repo has an explicit precedent against this substitution, in `test_schema_export.py`'s own docstring: *"Existence is asserted, not skipped on … must fail CI, not silently green-skip (#730 review)."* Bounded — the test is `integration`-marked and CI deselects it either way, so the loss is local-only. `skip XOR (validated >= 6)`, or keeping the floor behind an opt-in marker, restores the bite.

## N3 (P2) — an excluded optimizer config vanishes silently, and the console states a reason that is false

`list_optimizer_configs` now validates every document and drops failures, recording the reason only as a server-side `logger.warning`; `scanner.js:374-376` sets `optimizeButton.disabled = !names.length` with `title = "No compatible optimizer configs available"`. Two distinct real failures now render identically as "there aren't any": (a) someone introduces a schema error in one config — before this commit the operator selected it and read the exact error in `#optimizer-note`; now the file simply is not in the dropdown; (b) `/api/configs/optimizer_configs` returns 500 (an unreachable configs tree — `list_configs` wraps resolver failures as `internal_error`), which the JS `.catch` flattens to `{names: []}`, discarding the message. In both cases the diagnosis is reachable only by ssh-ing to the host and reading the service log — and the intended post-rollout steady state has two `# LEGACY` files permanently invisible with no on-screen explanation. The other listings in `loadConfigs` keep `detail: e.message` for exactly this reason, and the kit already has the vocabulary (`degraded` chip, the pane's `stale`/error state). An `unavailable: [{name, reason}]` field on the listing, or one line beside `optimizer-note`, closes it.

## N4 (P2) — invalidating the offer on queue-accept makes the target numbers unrecoverable when the move fails

`scanner.py:469-472` now invalidates the offer on a successful **queue add** ("best settings have already been applied"). But `submit_plan` returning `ok` means the manager accepted the item, not that the move succeeded — the console has a whole `FAILED_MOVE_LOG_PREFIX` path precisely because worker-side moves fail (a device refuses, a component out of range). When that happens the offer is already dead: `invalidated_reason` is set, the button is disabled, and there is no retry. The operator's only recovery is to re-type the targets into the Move panel — and **those numbers were never rendered into the page**. `scanner.js:725` still puts `best_moves` only into the button's `title` attribute; the table above shows `measured`/`outputs`/`best`, which are optimizer-space values, so for a pseudo variable the operator reads a pseudo value on screen and moves physical component currents they never saw. Either invalidate on the item's completion rather than its acceptance, or render `best_moves` in the table.

**A correction owed on this one:** the underlying "the physical targets live only in a `title=` tooltip, and this is the one consequential action in the console outside the kit's rung-3 confirm pattern" was raised in round 1 by the scanner reviewer and **I dropped it when consolidating the report** — it never appeared in the posted P1/P2 list, so nothing was silently ignored on the author's side. Raising it now, with N4 as the reason it matters more than it did before.

## N5 (P2) — `optimize` survives in the published contract's prose after being removed from the enum

`ScanRequestMode.OPTIMIZE` is gone and well pinned (`test_published_modes_exclude_retired_optimization`, plus the no-drift export test), and `git grep ScanRequestMode.OPTIMIZE` is clean repo-wide. The removal is also safe: the `data.get("mode") == "optimize"` guard runs in the `mode="before"` validator, ahead of enum coercion, so a stored `mode: "optimize"` still gets the migration remedy rather than a bare pydantic enum error. But three prose residues in `GEECS-Schemas/geecs_schemas/scan_request.py` ship into **both** published artifacts:

- `:523` the `mode` **field description** — *"'step' sweeps one or more axes, 'noscan' collects shots without moving anything, **'optimize' lets an algorithm pick the settings**"* → `scan_request.schema.json:210`, `schema_reference.md:162`;
- `:476` the class docstring — "Fill in the mode (sweep / stand still / **optimize**)" → `scan_request.schema.json:200`;
- `:533` the `axes` description — "Leave empty for noscan **and optimize**" → `scan_request.schema.json:213`.

The artifact contradicts itself inside one file: `mode`'s `enum` offers two values while its own `description` enumerates three. Neither the no-drift test nor the new enum test can catch this — the artifact is a faithful render of stale model prose. The stated audience is "OSPREY's plan panel and other JSON-Schema-aware consumers"; an LLM client weights a description at least as heavily as an enum, composes `{"mode": "optimize"}`, and is refused. Three one-line edits plus a regen.

---

### Merge gate

Both `unit-tests` runs at `4aa610ab` are **pending** at the time of writing; `pre-commit` passed twice. The "3,391 passed across 15 suites" figure is a local run — consistent in shape with what the diff touches (6 packages' code, 11 packages' locks; the jump from round 1's counts is accounted for by exactly the packages the Data-Utils edge disturbed), but not a substitute for green CI, and N1 has put a new mandatory dependency under five more packages since the last green run.

### Suggested disposition

- **N1** — revert the dependency edge and restore the three documents; keep the equality pin. The only item recommended as blocking.
- **N2, N3, N5** — cheap and worth doing in this PR (a test condition, a listing field or one line of UI text, three description edits + regen).
- **N4** — either the invalidate-on-completion change or rendering `best_moves` in the table; the latter also closes the dropped round-1 finding.
- **P2-4 (GEST)** — waiver stands, premises verified.
- Beam-on acceptance and the configs-repo rollout remain owed as already recorded in the PR body.

skbarber:
## Round 2 dispositions — `1696f0b6`

**N1 was my mistake.** I introduced an architectural dependency to deduplicate an integer, then weakened the documents that prohibited that dependency. The previous P2-8 disposition's claim that this was an appropriate downward dependency was wrong. Import-side purity did not justify the installed dependency closure.

- **N1 — fixed.** Data Utils has no intra-repo dependency again. Its file-format epoch constant is independent of Core's wire-format constant. The equality check lives in Bluesky, which already depends on both, and compares values rather than object identity. Root `CLAUDE.md` and Core `DESIGN.md` restore the original layering rules. Regenerated locks contain neither Core nor MySQL in **Data Utils, Data Portal, Log Triage, Logbook, ImageAnalysis, or ScanAnalysis**. New boundary tests reject intra-repo runtime dependencies, check the lock closure, and import the file reader in a subprocess that refuses Core/MySQL imports. The GEST owner-approved decision remains separate and unchanged.
- **N2 — fixed acceptance gate; migration still owed.** The corpus check again requires at least six native keeper configs. It fails on zero or partial migration rather than skipping; regressions exercise 0, 1, 5 and 6 native documents through the actual walk. The real local corpus currently **fails with 0 native configs**, which is the correct, visible rollout blocker. No configs repository was modified. This supersedes the previous comment's migration-pending-skip disposition.
- **N3 — fixed.** The resolver exposes usable names and unavailable names mapped to validation reasons. Its optimizer listing uses strict filesystem-error handling; errors from an unresolved/missing root or directory traversal reach the API. Other listing behavior is unchanged. The scanner shows unavailable names/reasons and preserves endpoint errors visibly, even with Optimize disabled; a failed listing is not described as an empty library. Tests cover filesystem failures, API diagnostics and actual JS rendering.
- **N4 — fixed by retaining visible targets.** The panel displays the exact recorded physical targets in a separate table, including after queue acceptance invalidates the offer. A kit confirmation dialog shows those same targets, defaults focus to Cancel and captures the selected run UID. The service now says the move **was queued**, never that it has already been applied. The conservative invalidation policy remains; if a move fails, its target values stay readable for operator recovery. UI tests distinguish those physical numbers from a deliberately different pseudo-space best value; an offline browser check verified the dialog and persistence after queueing.
- **N5 — fixed.** Removed the three remaining optimize-mode descriptions from ScanRequest and regenerated both JSON Schema and the reference page. A regression checks the published model/field prose as well as the existing enum checks.

**3,407 tests passed across 15 suites; full-repository lint passed** (`scripts/check.sh --all`). Counts: root 83; ImageAnalysis 292; ScanAnalysis 192; Data Utils 400; Schemas 457; Bluesky 654; CA gateway 140; PVA gateway 51; Core 125; Data Portal 305; Log Triage 58; MCP 102; Logbook 274; scanner 92; shared theme 182. Hardware tests remain gated. Separately, the real optimizer-corpus integration gate correctly fails at **0/6 native configs**; deployment remains blocked on that migration.

A fresh-context reviewer found two issues in this follow-up: strict listing initially still called a tolerant filesystem helper, and the corpus fixture patch used an import form that broke under dotted test-module names. Both fixes went back to that reviewer, which verified them and reported **no surviving findings**. No hardware was contacted and no additional shots were taken. Corpus deployment and beam-on acceptance remain explicit rollout gates, not completed work.

skbarber:
## Round-2 items confirmed closed in `1696f0b6`

Verification pass, not a third review round — the five open items had objectively checkable end states, so this is a diff read against `4aa610ab..1696f0b6` rather than a fresh adversarial sweep. All five verified directly.

**N1 (P1) — closed, exactly as scoped.** `geecs-core` is gone from `GEECS-Data-Utils/pyproject.toml`; `LABVIEW_EPOCH_OFFSET = 2_082_844_800` is restored as a literal with a comment naming it a *file-format* constant; all three documents are back to their pre-`4aa610ab` wording, including the `"data-utils must stay dependency-free"` clause in `GEECS-Core/DESIGN.md:73`. Verified `mysql-connector-python` and `geecs-core` are absent from the lock of every one of the six affected packages (DataPortal, LogTriage, Logbook, ImageAnalysis, ScanAnalysis, Data-Utils).

Two additions beyond what was asked, both good: the equality pin moved to `GeecsBluesky/tests/test_epoch_contract.py` — the package that legitimately depends on both sides — and a new `GEECS-Data-Utils/tests/test_dependency_boundary.py` pins the invariant two ways, asserting no `path` dependency in the pyproject and neither package in the lock, then importing `scan_stack` in a subprocess behind a `MetaPathFinder` that raises on any `geecs_core`/`mysql` import. Re-adding the edge fails both. That is a better guard than the finding asked for, and it is the kind of thing that stops this recurring.

**N2 (P2) — closed, and the fix proves itself.** `assert validated >= 6` is back with a message naming the shortfall. Better: the new `test_optimizer_rollout_requires_all_keepers` is parametrized over `[0, 1, 5, 6]`, builds a synthetic corpus, monkeypatches `CONFIGS` and calls the *real* walk — asserting it raises for 0, 1 and 5 and passes at 6. It is not `integration`-marked, so the floor's bite is now pinned in CI while the real-corpus walk stays deselected and will fail loudly on a local run until the rollout lands.

**N3 (P2) — closed at the listing seam.** `optimizer_config_listing()` returns `(names, {name: reason})`, `ConfigListOut` carries `unavailable`, and the console renders the reasons in `#optimizer-availability`. The two failure modes are now distinguishable: `_list_folder(..., strict=True)` lets an unresolvable configs root propagate instead of reading as an empty listing, and the JS `.catch` keeps `detail: e.message`, so "unreachable configs tree" no longer renders as "there aren't any".

**N4 (P2) — closed both ways, including the round-1 item that never made my posted list.** The physical targets now render in the page body ("Recorded physical targets for Set to best"), and `dlg-set-best` adds the kit's rung-3 confirm via `window.GeecsKit.confirm()`, listing each `name = value` before the move. Invalidation stays on queue-accept, which is fine now that the numbers survive on screen, and its wording is honest about what happened: "best-settings move was queued; check its result before making another move" rather than "already been applied".

**N5 (P2) — closed.** The published `scan_request.schema.json` now has `enum: ["step", "noscan"]` with a `mode` description that matches it, and no `optimize` residue in the class description or the `axes` description. The only remaining occurrences of the string in `scan_request.py` are the two migration-guard lines, which is correct.

**Waived, unchanged:** the eager `gest_api` import in GEECS-Schemas (P2-4) — owner decision, premises verified in round 2.

### Remaining before merge

- Both `unit-tests` runs at `1696f0b6` are **pending**; `pre-commit` passed twice. Nothing here should merge on a local count.
- Still owed as already recorded in the PR body: **beam-on acceptance** (TopViewMax, 10 × 5 with live PVA join, RSS, scanner submission and an explicit Set to best) and the **configs-repo rollout** of the six keepers plus the two `# LEGACY` headers — which the restored N2 floor will now fail on until it happens.

No open findings from either review round.

PR #921, opened 2026-09-16
Add Sweep payload and shared trajectory expansion

skbarber:
## What & why

The scan composer needs one validated trajectory description that preview and execution can expand identically. This adds `Sweep` to `geecs_schemas` and hardware-free expansion in `geecs_bluesky.trajectory`, using NumPy spacing and Bluesky's own pattern generators.

- Axis sweeps support one to many range/list/log axes, equal-length correlated motion, Cartesian grids and snaking. Five correlated lists of ten values produce ten steps.
- Typed spiral, Fermat spiral, square spiral and x2x variants carry the pattern parameters. Relative flags stay per-axis; expansion preserves offsets for the future execution plan.
- Validation refuses unequal correlated counts, duplicate axes or resolved aliases, malformed payloads and unrepresentable positions.

Foundation only: no deployed registrations, queue submission or scanner UI behavior changes. Plan replacement is a separate PR with its own hardware session. Count remains distinct. Version bumps: GEECS-Schemas 0.29.0 and GeecsBluesky 0.93.0.

Base: `feature/native-bluesky-plans`, at merged optimization #920 (`05c0fc23`), as requested by the owner. The owner will merge this PR.

Scope breakdown: 409 lines of model/exports/expansion, 401 test lines, and package release/instruction updates. Planning records the agreed design and preserves two visual references (1,149 HTML lines); they are references only, not production frontend code. A dated amendment reconciles the older rebuild plan with the agreed count/sweep/optimize destination.

## Tests

`./scripts/check.sh --all` passed, including all-file pre-commit checks. Tests ran sequentially with localhost sockets available; hardware/integration exclusions were preserved.

| Suite | Result |
| --- | --- |
| Root | 83 passed, 63 deselected |
| ImageAnalysis | 292 passed, 5 skipped, 10 deselected |
| ScanAnalysis | 192 passed, 2 deselected |
| GEECS-Data-Utils | 400 passed, 4 skipped |
| GEECS-Schemas | 494 passed, 7 deselected |
| GeecsBluesky | 703 passed, 4 skipped, 1 deselected |
| GeecsCAGateway | 140 passed |
| GeecsPvaGateway | 51 passed |
| GEECS-Core | 125 passed, 5 deselected |
| GEECS-DataPortal | 305 passed |
| GEECS-LogTriage | 58 passed, 1 skipped |
| GEECS-MCP | 102 passed, 1 skipped |
| GeecsLogbook | 274 passed, 1 skipped |
[...]

PR #922, opened 2026-09-16
Replace moving stock scan verbs with Sweep execution

skbarber:
Replaces the worker's moving stock scan registrations with one `sweep` plan backed by the shared Sweep payload and Bluesky `scan_nd`. Count and Optimize remain distinct. Catalog and direct variable references resolve once, relative axes restore their staged baseline on completion/abort, and new runs retain trajectory metadata plus legacy ScanInfo projections.

Stack: merge #921 first. This PR targets `codex/scan-composer-foundation`; the complete preview/composer is child PR #923. Retarget children before deleting any parent branch. Owner merges into the native Bluesky integration branch.

The registry/permissions cutover, client expansion, metadata readers and small compatibility adapter for the existing scanner form travel together so that no stock-moving submission path remains active at cutover. Historical run readers remain. MCP changes are explicitly out of scope.

Scope including tests/docs: Bluesky worker/client/metadata cutover +574/-277 lines; Data Utils classification +37/-1; scanner adapter/demo/summaries +69/-22; delivery record +7/-4. Versions: GeecsBluesky 0.94.0, GEECS-Data-Utils 0.33.0, GeecsScanner 0.10.0.

## Validation

Combined stack checks: Bluesky 718 passed, 3 hardware skips, 1 deselected; Data Utils 404 passed, 4 skipped. Scanner execution-parent suite: 92 passed, 1 skipped. The full repository check was run on the child containing this exact execution commit; final totals are recorded on the child PR. Commit hooks pass.

RunEngine tests cover staged relative baselines, mixed frames, completion, abort, restoration failure, early validation, nested JSON QueueServer submission, acquisition callbacks and historical/new classification. Independent adversarial review completed; report and disposition follow in a comment.

## Hardware verification

**OWED — dedicated Sweep cutover session; #920 optimization acceptance does not cover this change. Do not deploy this cutover independently of the preset corpus update.**

- Strict range: 3 positions × 2 shots = 6 events/rows in 3 bins; verify commanded positions, first-axis ScanInfo, `Plan = sweep` and 1-D classification.
- Correlated lists: equal three-entry lists, including a repeated value, × 2 shots = 6 events in 3 ordered bins. Verify every axis; unequal counts must be refused before motion.
- Snake grid: 2 × 3 positions × 2 shots = 12 events in 6 bins; verify traversal and GRID classification.
- Verify relative physical and pseudo axes restore on completion and graceful abort, capture baselines after staging, and report restoration failure as failure. A hard process kill cannot promise restoration.
- Exercise gated capture, image/scalar correspondence, submission provenance, Count/no-scan behavior and an old scan's continued readability/classification.
- Migrate or retire moving stock presets in the configs repository at the coordinated cutover; preserve optimization configs. This PR does not change the deployed configs corpus.


skbarber:
Independent adversarial review completed through correctness, redundancy and placement. The reviewer examined registry bindings, acquisition wrappers, shared trajectory expansion, client reference resolution, metadata callbacks, scanner submission and historical readers.

Findings and disposition (fixed in cfb76772, then confirmed by the same reviewer):

- **P1:** The existing scanner form still emitted retired moving verbs. Fixed with a narrow Sweep adapter and migrated demo preset; the complete composer is the child PR.
- **P2:** Plot hints assumed device names were hinted fields. Fixed to use actual motor hint fields and preserve Bluesky's fallback for unhinted motors; both cases are tested.
- **P2:** Callback tests still called removed stock registrations. Migrated them to bound Sweep execution so acquisition and file assertions exercise the new path.

Final reviewer verdict: **no surviving findings**. Hardware acceptance remains owed as stated in the PR body.

skbarber:
## Review of #922 — Replace moving stock scan verbs with Sweep execution

Read every hunk and opened the surrounding code (registry/strict binder, `trajectory.py`, the `Sweep` schema, `CaSettable.locate`, the pseudo positioner's stage/unstage, bluesky 1.15's `stub_wrapper` / `relative_set_wrapper` / `reset_positions_wrapper` / `scan_nd`). Ran the full GeecsBluesky suite on the PR head in a throwaway worktree: **717 passed, 3 skipped, 1 deselected**. Probed the plan directly with descending, snaked-grid, single-axis-product, log, and square-spiral payloads.

### The core of the change holds up

- The wrapper nesting is right for the stated invariant: `stage → open_run → [relative_set inside reset_positions] → reset → close_run → unstage`. Relative baselines are captured after stage and restored before close/unstage, and the restore `set` does not pass back through `relative_set_wrapper` — swapping the two would re-offset it.
- The eager-validation trick (`sweep` deliberately *not* a generator function) does run `Sweep.model_validate` + `sweep_to_cycler` before `liveness_gate` and before the box moves. Confirmed the box sees no puts on a bad payload.
- `CaSettable.locate` / `CaPseudoPositioner.locate` keep bluesky off the `obj.position`-is-a-Signal path.
- The `np.float64` that `outer_list_product` leaks into `extents` for snaked grids is harmless — `np.float64` subclasses `float`, so JSON and msgpack both accept it.

---

### Findings (6)

**1. medium — `GeecsBluesky/geecs_bluesky/qs_client/presets.py:242`**

`Sweep.model_validate(kwargs.get("sweep"))` raises `pydantic.ValidationError`, not `GeecsConfigurationError`, but `QueueClient.submit_preset` (`client.py:742`) catches only `GeecsConfigurationError`, and `expand_preset`'s own `Raises:` section promises only that type.

A preset whose `plan.name == "sweep"` is missing `kwargs.sweep`, has unequal zip counts, or has two catalog aliases resolving to one axis (the case `test_alias_collision_is_rejected` pins as a bare `ValueError`) therefore escapes `submit_preset` as an unhandled exception instead of returning `SubmitResult(ok=False, message=…)`. The scanner survives because `_expand` / `run_submit_preflight` both catch `Exception`, but any other caller of the public client API gets a traceback.

Fix: wrap the validate — and the re-validate two lines below — in `GeecsConfigurationError`.

**2. low — `GeecsScanner/geecs_scanner/static/scanner.js:446`**

`formShape` now requires `a.num > 1`, but the form can submit `num === 1`: `points(a,b,s)` returns `Math.floor(0/s)+1 === 1` when start equals stop, and `axis().ok` is still true (only `step > 0` and a chosen variable are checked), so Start is enabled. Save that as a preset and reload it, and `formShape` returns `null` — Start is disabled with "no form for it yet" for a preset the form itself produced.

The note at line 461 compounds it: it reports `args.length`, which is always 0 for `sweep`, so every unsupported sweep reads "runs sweep with 0 argument(s)".

**3. low — `GeecsScanner/geecs_scanner/service/summaries.py:131`**

The sweep branch builds `axes` from axis names only, so the queue/history row loses the range that every stock verb showed. Verified output:

```
sweep U_S1H.current · 7 steps · 10 shots/step
```

where the retired `scan` gave:

```
scan U_S1H.current 2.0 → 5.0 · 7 steps · 10 shots/step
```

Two queued sweeps over the same variable with different ranges are now indistinguishable in the queue view — the one place an operator checks before hitting start. `RangeAxis` carries `start`/`stop` and `ListAxis` its point count, so `_axis()` could still be used.

**4. low — `GEECS-Data-Utils/geecs_data_utils/tiled_schema.py:540`**

The classifier keys GRID on `combine == "product"` alone, but `AxisSweep` explicitly allows a one-axis product ("A single-axis product is allowed (the Cartesian product of one list)"). Such a run has one entry in `motors` and `shape == [n]` yet classifies as GRID, sending a genuinely 1-D scan down the grid path in the portal/analysis. Gate on `len(trajectory.get("axes") or []) > 1` as well.

**5. low — `GeecsBluesky/geecs_bluesky/plans/sweep.py:79`**

`snaking` is `[False]` for every non-grid sweep, so a correlated (zip) sweep over N axes, or an x2x/spiral pair, emits `motors` of length N beside `snaking` of length 1; stock `grid_scan` keeps the two parallel (`tuple(snake for …chunk_args)`). Anything that zips them — bluesky's own `BestEffortCallback` grid logic is the obvious consumer — silently drops axes. Related and smaller: a product grid omits the `hints["gridding"] = "rectilinear"` that stock `grid_scan` sets. No in-repo consumer reads either today, so this is future-proofing rather than a live break.

**6. low — `GeecsBluesky/tests/qs_client/test_presets.py:138`**

`test_expansion_records_its_references_and_leaves_literal_strings_alone` was deleted and replaced by `test_expansion_records_axis_references_and_preserves_order`, which only covers the new sweep path. `_resolve` is still live for `count` / `optimize` args and kwargs, and the contract the old test pinned — a literal string argument (an enum value such as `"on"`) must not be rewritten into a device reference or appended to `references`, where the preflight would then fail it against the manager's device tree — is now untested in either direction.

---

### Nits, not findings

- `qserver/user_group_permissions.yaml:41` still describes the operator regex as "the stock scan verbs bound strict".
- `GeecsBluesky/CLAUDE.md`'s "Deployed acquisition" paragraph still says "the stock parameters minus the hook", which no longer describes `sweep`.

### One to confirm rather than fix

`sweep_first_axis` for a `spiral` / `spiral_fermat` produces a degenerate legacy ScanInfo triple — a square-spiral probe gave `[-1.0, -1.0, 2.0]`, i.e. start equals end. `EVENT_SCHEMA.md` does call the projection lossy, so this was treated as accepted rather than a defect; flagging in case it should read differently.

skbarber:
Implemented the review corrections in ec9ae283 and carried them into the stacked UI branch in acec59b7. Original Claude reviewer recheck is still requested; this is the implementation report, not a claim that the external reviewer has accepted the changes.

1. **Validation error contract — implemented.** Initial and post-alias Sweep validation now raise `GeecsConfigurationError`. Public `QueueClient.submit_preset` regressions cover missing payload, unequal correlated lists, and alias collisions; each returns a refusal before any manager calls.
2. **One-point preset reload — implemented.** The temporary form accepts a one-point range, including a step larger than the range. It reloads the single visited start position with a valid unused step. Unsupported Sweep messages explain the form limitation. The full composer in #923 retains its native point-count inputs, with tests for one-point preset round trips and previews using equal and distinct bounds.
3. **Queue summaries — implemented.** Range endpoints, logarithmic exponents, list lengths, and relative-axis labels are retained.
4. **One-axis product classification — implemented.** GRID requires more than one product axis; a one-axis product remains 1D.
5. **Metadata — flags implemented; plotting-hint suggestion waived.** `snaking` has one entry per motor for correlated sweeps and patterns as well as products. Deliberately did not add `hints.gridding="rectilinear"`: Bluesky LiveGrid assumes uniformly spaced cells and one primary event per cell. Nonuniform lists/logs violate the coordinate assumption, and `shots_per_step > 1` exceeds the grid's event indexing. A regression executes a six-position nonuniform/descending grid with twelve primary events and verifies that no incompatible hint is emitted. This is an owner-vetoable choice, recorded in EVENT_SCHEMA.md.
6. **Generic resolver coverage — restored.** Tests cover catalog/device references alongside literal enum strings and lists, independently of the Sweep-specific path.

Both documentation nits are corrected. The legacy ScanInfo projection remains intentionally lossy; EVENT_SCHEMA.md now explicitly explains that equal first/last values do not imply a motionless pattern and that the full Sweep payload is authoritative.

The independent execution reviewer rechecked the corrections, found the distinct-bound singleton and LiveGrid issues during follow-up, and confirmed **no surviving findings** after those were corrected. Original external Claude confirmation remains pending.

Validation:
- Full GeecsBluesky suite: **726 passed, 3 skipped, 1 deselected**. After the final edge-case refinements, targeted execution/client tests: **69 passed** (including the new multi-shot grid regression).
- #922 Scanner: **97 passed, 1 skipped**; DataUtils: **405 passed, 4 skipped**.
- #923 Scanner after carrying the fixes forward: **115 passed, 1 skipped**.
- Required commit hooks passed on both commits. All four CI checks passed on each updated head (#922 ec9ae283 and #923 acec59b7).

Hardware verification is still **OWED** as described in the PR body; no live scans or deployment were performed for this review update. Neither PR has been merged.

PR #923, opened 2026-09-16
GeecsScanner 0.11: inline Sweep composer and Python trajectory preview

skbarber:
Adds the complete inline trajectory composer to New scan: Count / Sweep / Optimize, shared capture/device controls, Axis Sweeps and Patterns, editable presets, and previews generated by the same Python expansion code used by execution. The fresh page starts unconfigured. The fixed two-axis form and its browser point arithmetic are removed.

Axis Sweeps supports one-to-many mixed linear/list/log axes, correlated or grid traversal, snake, and per-axis relative offsets. Lists accept numeric comma/tab/space/newline input, retain repetitions and show parsed counts. Patterns includes spiral, square spiral, Fermat and X2X. Per-axis plots, two-axis XY, a deferred points table and the full payload use server results. Styling uses the shared GeecsWebTheme kit/tokens.

Stack: merge foundation #921, then execution #922, then this PR. This PR targets `codex/scan-composer-execution`; retarget children before deleting parent branches. Review the full running app before merging the stack. **Deployment is blocked on the coordinated preset/worker/scanner cutover below.**

The preview API bounds raw requests before parsing and limits expanded coordinates and concurrency. Ordinary axis sweeps (up to eight axes) and X2X expand on the threadpool; all spirals and larger axis sets use disposable processes. This is a computation dispatch threshold, not an executable axis ceiling. Square spirals and large axis sets need isolation because their construction cost can exceed their output count. The service package imports lazily to avoid loading the queue client in the child.

Superseded previews are aborted; disconnects cancel isolated children. Busy previews retry twice, then offer manual refresh. Preview success is optional: Start still runs normal preflight validation, and Save validates the Sweep schema before writing. Malformed preset loading clears prior trajectory state and disables Start/Save while loading the new capture fields. Background metadata applies only to Count.

Budget values (256 KiB raw input, 250,000 coordinates, 256 MiB monitored child RSS, eight seconds per child, two concurrent previews) are engineering choices open to adjustment, not operator shot caps. Responses disclose sampling; submission preserves the complete original payload.

Version: GeecsScanner 0.11.0, with changelog and architecture documentation. No hardware deployment or configs repository mutation in this PR. MCP remains out of scope.

Diff by concern: browser/templates **+374/-141**; service/API **+282/-6**; tests **+646/-29**; docs/changelog **+121/-7**; dependency/version files **+7/-5**.

## Validation

Before the review corrections, `scripts/check.sh --all` passed all 15 local suites: **3,527 passed, 16 skipped, 88 deselected**. After the latest corrections: Scanner **140 passed, 1 skipped**; WebTheme **182 passed**; required commit hooks passed. Additional tests cover oversized/chunked/incorrect-length requests, malformed preset replacement, bounded retry/cancellation, cheap ordinary previews, isolated high-axis and thin-square trajectories, child protocol errors, schema-invalid preset overwrite refusal, background mode changes, and deferred table rendering.

Actual browser checks against the demo manager include range/list/grid/pattern previews, five axes, repeated positions, unequal-list rejection, preset round trips, Count/Optimize controls, preflight and simulated acquisition. Latest browser verification checked list input with mixed delimiters, preserved repeat/order, zero table rows while collapsed, rows built on opening, and a clean console. Three-point local previews measured 21 ms initially and 2 ms on two repeat requests. These are local observations, not deployed-host guarantees.

Independent adversarial review of the corrections is complete. The reviewer identified square-spiral/high-axis construction costs and missing Save schema validation; fixes went back to the same reviewer, who confirmed **no surviving findings**. Original external Claude recheck of #923 remains pending. Detailed disposition is in the review follow-up comment.

## Hardware verification and corpus cutover

**OWED — owner UI acceptance and real-worker validation, including #922's separate dedicated hardware session.** Verify strict/gated capture, device saving/essential controls, preview-versus-commanded positions, preset round trips and relative restoration. The local demo runs the actual UI and Python preview but simulates acquisition.

**OWED — paired configs migration before deployment acceptance.** At configs baseline `ad4e595`, the deployed Undulator corpus contains eight `scan` presets and five `count` presets. A conversion patch for all eight moving presets has been prepared and schema/coordinate-validated without changing live configs. The planning brief records the exact files, position/shot counts and the cutover/rollback checklist. Those facility names are documentation of the reviewed corpus, not runtime defaults.

Stop submissions and drain the queue, record all three revisions, recheck corpus changes, migrate the eight presets to Sweep while preserving capture metadata and all optimization configs, then deploy the matching worker and scanner before reopening submission. Confirm all eight load/edit/preflight, the count presets still work, and the hardware session passes. Roll back configs, worker and scanner together on failure. Do not deploy the retired-verb worker against the old preset corpus or the new corpus against the old worker.


skbarber:
Independent adversarial review completed through correctness, redundancy and placement. The reviewer checked the shared schema/trajectory modules, scanner service boundaries, existing settables, theme helpers and repository subprocess/parser/chart implementations.

Findings and disposition (fixed in 84f65ba0, then confirmed by the same reviewer):

- **P2:** Preserved `kwargs.trigger_profile` could override the visible selection. The form now loads the effective value into the selector and removes the hidden override when saving/submitting. Regression coverage checks edits, explicit null, unrelated keyword preservation and Sweep list retention.
- **P2:** Relocking introduced a local checkout URL. The repository's existing lock scrubber restored the canonical `file:///GEECS-Plugins/ImageAnalysis` URL.
- **P3:** The version bump initially lacked its changelog entry. The 0.11.0 entry now records the composer, preview API and removal of the fixed form.

The same reviewer confirmed all three fixes: **no surviving findings**, including no surviving redundancy or placement findings. Hardware and owner acceptance remain owed as stated in the PR body.

skbarber:
Stack update: acec59b7 carries the #922 execution-review corrections from ec9ae283. See the [finding-by-finding response](https://github.com/GEECS-BELLA/GEECS-Plugins/pull/922#issuecomment-5703563021).

The full composer and its layout are unchanged. The temporary form fix from #922 is superseded here by the native Sweep composer; its merge conflict was resolved by retaining the full composer. Additional tests cover one-point preset round trips and server previews with both equal and distinct range bounds. Shared queue-summary, validation, metadata, and historical-classification fixes are inherited from the parent.

Scanner suite: **115 passed, 1 skipped**. Commit hooks passed; all four updated-head CI checks passed. The separate external review of #923 remains pending. Hardware verification remains owed; neither PR was merged or deployed.

skbarber:
## Review of #923 — `codex/scan-composer-ui` → `codex/scan-composer-execution`

Reviewed at `84f65ba0` (815/151 over 15 files). Two independent passes: a full read of the diff plus surrounding context, and a separate fresh-context adversarial pass. Findings below are merged and ordered by what to act on first. Severity floor: behaviour on the production path, or merge sequencing. Nits are labelled and are not work orders.

I have not run the suite locally or driven the app in a browser — the PR's own validation covers both, and the hardware acceptance it marks OWED is the right call.

---

### 1. The stack disables Start on all 8 existing `scan` presets — sequence the corpus migration with it

Stack-level, inherited from #922, not this PR's code; raising it here because it is the merge-window risk.

The deployed Undulator corpus (`GEECS-Plugins-Configs` @ `ad4e595`) holds 5 `count` presets and 8 `scan` presets. On `feature/native-bluesky-plans` today, `formShape` maps `scan`/4-args and `grid_scan`/8-args to a form. After this stack it accepts only `count|sweep|optimize`, so those 8 land on `S.formable = false` → `btn-start.disabled` (`scanner.js:584`) with the tooltip "Retired or unsupported preset plan: scan". They become neither editable nor runnable from the console.

The PR body acknowledges the corpus is unmigrated. `Planning/native_bluesky/12_scan_composer.md` says "Preserve valid preset submission when replacing the form, independent of editability" — that is not preserved. Either migrate the corpus in the same window as the cutover, or restore an unformable-but-submittable path.

### 2. A malformed sweep preset leaves a submittable mixed-state form

`scanner.js:433` / `scanner.js:445`, `sweep-composer.js:160`

The rewritten `formShape` returns `plan.name` for any zero-arg `sweep` plan without checking that `kwargs.sweep` exists — refusing to guess was the old version's whole purpose. `fillFormFromPreset` then calls `composer.load(kw.sweep)` unguarded, which dereferences `payload.trajectory`. Three reachable throws:

- `kwargs.sweep` absent (hand-edited or partially migrated document)
- `trajectory.kind` the composer doesn't know → `patterns[kind].forEach` in `render()`
- a list axis whose `positions` is not an array → `a.positions.join(", ")`

The throw is swallowed by `selectPreset`'s `.catch`, which surfaces a raw JS message and resets nothing. By that point `setMode`, `setAcq` and the background flag are already the new preset's, while shots, period, trigger profile, description and **the entire device table** still hold the previously loaded preset's values — under the new preset's name in `#preset-name`. `recalc()` never runs, so `valid` keeps its prior value; if the operator had a valid trajectory in the composer beforehand, `composer.result()` is still non-null and **Start stays enabled on that mixed document**.

Fix: guard `formShape` with `plan.kwargs && plan.kwargs.sweep && plan.kwargs.sweep.trajectory`, and make `composer.load` defensive with a fallback to `S.formable = false`.

### 3. The 256 KiB input budget is enforced after the parse it is meant to bound

`service/trajectory.py:51`, `service/scanner.py`

`ScannerService.trajectory` runs `Sweep.model_validate(payload)` on an unbounded request dict; `preview()` then does `sweep.model_dump_json()` and only then compares against `MAX_INPUT_BYTES`. The body is fully materialised three times in the shared web process first — Starlette reads and `json.loads` it for `payload: dict[str, object]`, pydantic builds the model, then it is re-serialised. Uvicorn applies no request-size limit. A single POST with a `list` axis of ~10M `positions` validates 10M floats inside the scanner service before any budget applies; `MAX_VALUES` comes later still.

That is precisely the failure mode the disposable subprocess and `MAX_RSS` exist to prevent, and it lands on the shared worker host. Move the size check onto the raw body (a `Request`/`Content-Length` guard or a bounded read) ahead of `Sweep.model_validate`.

### 4. Every preview costs a cold interpreter (~2 s measured), and Sweep submission now depends on one succeeding

`service/trajectory.py:73`, `sweep-composer.js` (`requestPreview`), `scanner.js` (`recalc`)

`recalc()` gates Sweep on `composer.result()` — a successful `/api/trajectory` round trip. A hardware-free convenience is now on the critical path to starting a scan.

Measured cost of the child's dominant imports (`geecs_bluesky.qs_client`, `config_resolver`, `numpy`, `psutil`, `bluesky.plan_patterns`) in the root env: 1.58 / 1.72 / 2.23 s against a 0.48 s bare interpreter. That is a floor, not the figure: launching as `-m geecs_scanner.service.trajectory` makes runpy import the `geecs_scanner.service` package `__init__` first, which pulls `ScannerService` and its whole module graph, then re-executes `trajectory.py` a second time as `__main__` (also emitting runpy's "found in sys.modules" RuntimeWarning). The deployed worker box is slower than the machine measured on.

Consequences:

- `deadline` starts at `Popen`, so ~2 s of the 8 s budget is startup. On a loaded host a trivial 3-point linear sweep can exhaust it and report *"Pattern exceeded the preview time budget. Use a larger radial step or fewer points."* — advice unrelated to the cause.
- Superseded previews are never cancelled. The 300 ms debounce and the `generation` guard only suppress *rendering*; the server runs each one to completion, holding one of the two `_SLOTS` for its full lifetime.
- The 409 (`policy_refusal`, "Preview is busy") path retries via `setTimeout(requestPreview, 500)` with no ceiling and no message change — `#sweep-message` stays on "Calculating trajectory…". One fast typist, or two operators with the console open, can hold the composer in a silent permanent retry loop.

Worth reconsidering the shape: the isolation is only load-bearing for `spiral` and `spiral_fermat`, whose point count is not algebraic. `kind: "axes"` has an exact `n_steps()`, and `spiral_square` / `x2x` are `x_num*y_num` and `num`. Options, roughly in increasing order of change: start the deadline after the child signals readiness; run the child as a plain script path so it does not drag in the service package; cap the 409 retries and say "preview queue busy"; expand the bounded kinds in-process and keep the subprocess for the two spirals; or pre-warm and reuse a worker process.

### 5. A background flag silently follows a count preset into a Sweep submission

`scanner.js:607`

`background: name === "count" ? $("background").checked : !!doc.background`. Load a background count preset, switch to Sweep, submit — the run is stamped `ScanMode 'background'` in `ScanInfo` and in the run document (`geecs_schemas/preset.py:146`), which is what the logbook and analysis read. Metadata only, so no hardware consequence.

The old expression carried the flag the same way, but Background was a visible pressed mode button. This PR demotes it to a checkbox inside `#count-options`, which `setMode` hides whenever the mode is not `count` — so the state is now unobservable at the moment it takes effect. Either clear `doc.background` when the mode leaves `count`, or keep the checkbox visible.

### 6. `json.loads(output)` is unguarded

`service/trajectory.py:105`

If the child exits 0 having written anything to stdout besides its single JSON line — a library that prints on import — `json.JSONDecodeError` escapes `preview()` as a bare `ValueError`, so the route returns a 500 with no operator-facing message instead of the taxonomy's 400. Wrap it and raise `ScannerError("invalid_request", …)`, or frame the child's reply (read only the last line).

### 7. Child stderr is surfaced verbatim to the browser

`service/trajectory.py`

`"Trajectory calculation failed: " + error.strip()[-400:]` puts the tail of a Python traceback — absolute host paths included — into the operator's UI. Fixed message in the response, traceback to the log.

### 8. The points table is rebuilt eagerly inside a collapsed `<details>`

`trajectory-view.js:30`

Up to ~2,000 sampled rows × (1 + N axes) cells built with individual `createElement` calls, plus one SVG chart per axis and the XY chart — roughly 12,000 DOM nodes for a 5-axis sweep — reconstructed after every debounced keystroke, even though the table lives inside a disclosure nobody has opened. Defer the table build until the `<details>` opens.

### 9. The node test in `test_trajectory.py` is unguarded

`tests/test_trajectory.py::test_numeric_list_parser_accepts_delimiters_and_rejects_expressions` calls `subprocess.run(["node", ...], check=True)` directly, while every node test in `test_page.py` goes through `_need_node()` / `node_available()`. On a box without node it errors instead of skipping.

---

### Nits (not work orders)

- `policy_refusal` → 409 for "Preview is busy" is a capacity refusal, not the "not now / not like this" the `errors.py` docstring defines. It works — the JS retry matches — but it blurs the vocabulary.
- `composer.variables(S.variables.filter(scannable).concat(S.settables))` can list the same `Device:Variable` twice when a catalog overlay names a real settable.
- `numeric(..., integer)` accepts `x_num`/`y_num` of 1 while `SquareSpiralSweep` requires ≥ 2. The server rejects it with a readable pydantic message, so this is only a missed client-side hint.

### Checked and clear

Recorded so the floor is visible.

- Duplicate axis names, unequal `zip` counts, empty position lists and zero-axis sweeps are all rejected by the `Sweep` schema with operator-legible messages, so the corresponding `_expand` edge cases are unreachable. `sweep_to_cycler` already raises on an empty pattern, so `indices[-1]` cannot hit an empty list.
- All five schema trajectory kinds are covered by the composer's four patterns plus the axes tab — no gap.
- The repeated `communicate(timeout=…)` loop is correct: `input=None` on retries is required, and `_save_input` retains the payload across them. The trailing `communicate()` after a completed call does not hang.
- Semaphore acquire/release placement is right, and the child is always killed.
- Sampling retains both endpoints — verified at 3000 points → 1501 indices, last = 2999.
- No stale references survive to the removed `var1` / `step1` / `pts1` DOM ids; `defer` ordering makes `window.GEECS_SWEEP` available before `scanner.js` runs.
- The `continue` added to `test_every_literal_data_state_is_a_kit_state` exempts only the two new scripts, which write no `data-state` values — the invariant still bites for `scanner.js`.
- `psutil` is in `poetry.lock`; `bluesky` is a non-optional dep of `geecs-bluesky`, so the child import resolves. `.tscroll` / `.disc` are real kit classes.

One note for hardware acceptance, not a bug: preview resolves axes with `str` while execution resolves through the namespace, so two aliases of one movable will preview cleanly and fail at execution. That is documented behaviour in `sweep_to_cycler` — flagging it only so it is not mistaken for a preview defect during the session.

skbarber:
Implemented the #923 review corrections in **50ba3e29**. Original external Claude recheck is requested; the independently reviewed implementation is ready, while live corpus cutover/hardware acceptance remain explicitly owed.

1. **Preset corpus — accepted deployment gate, not a completed migration.** The brief and PR body now specify the eight moving presets at configs baseline `ad4e595`, their position/shot counts, preservation rules, paired deployment ordering, acceptance checks and three-revision rollback. A patch is prepared locally; all eight validate as Preset/Sweep and preserve the original ordered linear coordinates and non-plan bytes. `git apply --check` passes. Live configs remain unchanged so current workers are not broken ahead of cutover. No fallback can safely submit a verb removed by #922.
2. **Malformed preset state — implemented.** Missing trajectories, unknown patterns and malformed position lists clear old preview/trajectory state and disable both Start and Save. Capture controls finish loading from the new document instead of retaining the old device table. Tests start from a valid loaded form, try each malformed case, and verify new capture fields, disabled gates, and no stale preview revival.
3. **Input budget — implemented before parsing.** The route checks Content-Length and independently counts streamed bytes before `json.loads` or Sweep validation. Oversized requests with omitted or false length headers are covered; malformed JSON uses the error taxonomy.
4. **Preview latency and submission dependency — implemented.** Up to eight ordinary axes and X2X expand without a subprocess. All spirals and larger axis sets retain process isolation; exact coordinate counts do not bound square-spiral clipping or recursive Cycler construction costs. Lazy service exports avoid importing the queue-client graph in the child. Superseded fetches abort and disconnected requests cancel children; capacity retries stop after two and show an explicit retry control. Start uses normal preflight regardless of preview availability. Save now validates nested Sweep schema without numerical expansion. Local three-point preview requests measured 21 ms, then 2 ms twice; these are not deployment-host guarantees.
5. **Hidden background flag — implemented.** Only Count can write `background: true`; tests cover Count → Sweep/Optimize.
6. **Child response decoding — implemented.** Malformed JSON, wrong response shape and invalid response fields become readable `invalid_request` errors.
7. **Child diagnostics — implemented.** Nonzero-exit stderr is logged server-side; the browser receives a fixed explanation without host paths.
8. **Collapsed table work — implemented.** No table rows are constructed until the disclosure opens. Repeated previews replace the pending data; open tables render the latest response.
9. **Node availability — implemented.** The parser test and new component tests use the shared Node availability guard.

The independent reviewer of these corrections found the expensive square-spiral/high-axis cases and the Save validation gap described above. Each fix was returned to that reviewer, who confirmed **no surviving correctness, redundancy or placement findings**. The external Claude reviewer has not yet rechecked this commit.

Validation: **140 Scanner tests passed, 1 skipped; 182 shared-theme tests passed**. Required commit hooks passed. Browser verification confirmed preset loading, mixed-delimiter lists with repeats, lazy table construction and no console errors. All four CI checks passed on head 50ba3e29 (both lint checks and both full test jobs). The real local demo is updated at the existing localhost URL.

No PR merged, no hardware run, and no live configs changed. The corpus/hardware deployment gate in item 1 remains **OWED**, not waived or represented as completed.

skbarber:
## Re-review at `50ba3e29` — findings 2–9 addressed, verified

Confirmed against the amended head, not the reviewed one. Each fix checked for its own defects; none found.

| # | Finding | Fix | Verified |
|---|---|---|---|
| 2 | Malformed sweep preset → submittable mixed-state form | `formShape` guards `kwargs.sweep.trajectory`; `load()` resets, validates shape, rethrows; caller sets `S.formable = false` | `btn-start` **and** `btn-save-preset` now both gate on `!S.formable` |
| 3 | Input budget after the parse | Raw body streamed with a running cap + `Content-Length` pre-check, JSON parsed only after | Chunked bodies fall through to the streaming cap |
| 4 | ~2 s subprocess on every preview | In-process fast path for `x2x` and `axes` (≤ 8); lazy `service/__init__` `__getattr__` so the child no longer drags `ScannerService`/qs_client; `AbortController` + `is_disconnected` cancellation; 409 retries capped at 2; timeout message no longer misattributes | Fast path is exactly pre-bounded — `n_steps()` returns `num` for x2x and `counts[0]`/`prod` for axes, both checked against `MAX_VALUES` before the semaphore. Keeping `spiral_square` isolated despite its exact count is the right call (bluesky clips to the bounding box, so thin grids cost more than their coordinate count) |
| 5 | Background flag followed a count preset into a Sweep | `background: name === "count" && $("background").checked` | Drops the flag for non-count plans. Corpus-safe: exactly one `background: true` preset exists (`Undulator/presets/HasoBackground.yaml`) and it is a `count` |
| 6 | `json.loads` unguarded | Wrapped with an `isinstance` check, `ValidationError` caught, logged | The `{"error": …}` branch still reaches its `raise` after the try |
| 7 | stderr tail leaked to the browser | Fixed message; traceback to `_LOG.error` | |
| 8 | Table built eagerly inside a collapsed `<details>` | Deferred to `ontoggle`, `built` latch, cleared by `invalidate()` | `ontoggle` (not `addEventListener`) means no handler accumulation across previews |
| 9 | Unguarded node test | `node_available()` from `geecs_web_theme.testing` | |

Also checked: `api(path, opts)` forwards `signal` to `fetch`, so an abort rejects as `AbortError` and is filtered in `.catch`. `_validate_preset`'s new server-side `Sweep.model_validate` sits on the three **write** paths only (`preflight`, `submit`, `save_preset`) — the read path is untouched, so a legacy preset still loads and shows as unformable rather than failing to load.

Worth noting as a deliberate behaviour change, not a defect: preview no longer gates Start. `recalc` now only requires the form to *build*, with the estimate line explaining that Start runs preflight validation. That is the right call — it takes a hardware-free convenience off the critical path — and the payload is still validated server-side before it can reach the worker.

One cosmetic drift: `trajectory-view.js` now requires a `<details>` ancestor (`tableRoot.closest("details").open`), while `CLAUDE.md` still says it "only renders server results and can be reused independently of the form". Not reachable in this app — worth a word in the docstring next time the file is touched, nothing more.

CI green: pre-commit and unit-tests pass on both runs.

**Signed off from review — no blocking findings remain at `50ba3e29`.** Finding 1 is unchanged and is not a code issue: the 8 `scan` presets in the deployed Undulator corpus still lose their form and their Start button under this stack. That is a sequencing decision for the merge window, and it is the one thing left to settle before the cutover.

PR #924, opened 2026-09-17
GEECS-MCP 0.9.0: delete the write verbs the native-Bluesky rebuild broke (#727)

skbarber:
## What & why

The native-Bluesky rebuild removed the `geecs_bluesky.qs_client` methods that every GEECS-MCP write verb stood on. `submit_scan`, `submit_action`, `describe_action` and `move_variable` are all gone from the client (its submission surface is now `submit_plan`/`submit_preset` over the `count`/`sweep`/`optimize` plans), and `run_submit_preflight` takes a preset rather than a `ScanRequest`.

So on the integration branch, five MCP tools were broken:

| Tool | Called | State |
|---|---|---|
| `submit_scan` | `client.submit_scan` | gone → `AttributeError` |
| `run_action` | `client.submit_action` | gone → `AttributeError` |
| `describe_action` | `client.describe_action` | gone → `AttributeError` |
| `move_scan_variable` | `client.move_variable` | gone → `AttributeError` |
| `validate_scan_request` | `run_submit_preflight(ScanRequest, …)` | signature now takes a preset |

**This PR deletes all five rather than rewiring them**, per the owner's ruling (2026-09-16): GEECS-MCP was spun up as an experiment, not an operator surface, so it gates no client-seam change. Scans are submitted from `GeecsScanner`, the web scanner console. #727 remains the tracker if an agent-facing write path is ever wanted back; the shape to copy is then `GeecsScanner/geecs_scanner/service/scanner.py` (`build_submission_record` → `md={"geecs": {"submission": …}}` → `submit_plan`).

### What survives

**14 registered tools, every one calling a verb the protocol still has** — verified by starting the server and listing them:

```
clear_queue          get_scan_analysis    get_scan_figure      get_scan_result
list_analysis_groups list_analyzers       list_scan_configs    pause_scan
resume_scan          run_scan_analysis    scan_history         scan_progress
scan_status          stop_scan
```

Read + observe + halt. The whole analysis domain is untouched, and the ownership etiquette on stop/pause/resume is unchanged.

### Why the suite did not catch this — twice

The review found the identical drift a **second** time, one file over: `list_scan_configs(kind="save_sets")` routed to `ConfigsRepoResolver.list_save_sets`, removed in the same rebuild commit, still defined on `_FakeResolver`. Both are fixed, and the fake-vs-real assertions now live in one module (`tests/test_seam_pins.py`) rather than beside each double. See the review comment for the full disposition.

`tests/test_control_tools.py`'s `_FakeClient` still defined all four removed methods, so 102 tests passed against a seam that no longer existed in production. This PR adds `test_the_fake_client_only_promises_verbs_the_real_client_has`, which asserts every public method on the fake exists on the `QueueClient` protocol.

**Proved it bites:** putting `submit_scan` back on the fake fails the test with `AssertionError: _FakeClient promises ['submit_scan'], which QueueClient does not have`. Restored, 20/20 pass.

### Judgment calls (cheap to veto)

- **Deleted rather than left broken.** The owner said either was fine. Deletion keeps `AttributeError` verbs off master and follows the repo's own "remove old implementations" doctrine; the cost is ~1,050 lines of deletion that a future write path would rewrite from the new seam anyway.
- **Dropped the now-unreachable agent shot cap** (`runtime.max_shots`, `DEFAULT_MAX_SHOTS`, `[mcp] max_shots`) and `_task_error_kind`. No caller survives the removal. `DEPLOYMENT.md` notes the config key is no longer read.
- **`ScanRequest` is now consumed only by GEECS-Schemas itself** (plus one legacy read in the scanner's `summaries.py`). Retiring the document is left out of scope.
- **Deployed profiles naming removed tools are inert but stale.** `DEPLOYMENT.md` says to drop those entries rather than silently tolerating them.

### Scope by concern (+ / −)

| Concern | Added | Removed |
|---|---:|---:|
| Tool removal (`control_tools`, `read_tools`, `tool_names`, `server`, `runtime`) | 63 | 861 |
| Tests (deleted coverage + the new protocol pin) | 42 | 421 |
| Package docs (`README`, `DEPLOYMENT`, `CLAUDE.md`, `CHANGELOG`) | 141 | 92 |
| Published docs (`docs/geecs_mcp/*`) | 69 | 80 |
| Root `CLAUDE.md` (package row + dependency graph) | 14 | 7 |

Version: GEECS-MCP **0.8.10 → 0.9.0** with changelog.

## Tests

`./scripts/check.sh` — exact counts from this branch:

| Suite | Result |
| --- | --- |
| GEECS-MCP | **82 passed, 1 skipped** (76 + 6 new seam pins) |
| Root | 83 passed, 63 deselected |
| ImageAnalysis | 292 passed, 5 skipped, 10 deselected |
| ScanAnalysis | 192 passed, 2 deselected |
| Lint (`--lint`, all hooks) | **Passed** |
| `mkdocs build` | **exit 0, zero ERROR lines** |

**Not run locally:** GEECS-Core, GEECS-DataPortal, GEECS-LogTriage, GeecsBluesky, GeecsCAGateway, GeecsLogbook, GeecsPvaGateway, GeecsScanner, GeecsWebTheme — all env-blocked in this fresh worktree (no package venvs; `ModuleNotFoundError` at collection, not failures). The diff touches no code in any of them, so CI is the check that matters here.

Also verified by hand: the server starts and registers exactly the 14 tools above, with none of the five removed names present.

## Hardware verification

**N/A.** This PR only deletes code paths that could not execute — the four control verbs raised `AttributeError` before reaching hardware, and `validate_scan_request` never moved anything. Nothing that survives changes behavior: `stop_scan`, `pause_scan`, `resume_scan`, `clear_queue` and every read tool are byte-identical apart from the module docstring.

Nothing is deployed by this PR. **GEECS-MCP is not currently operable against the integration branch either way** — before this change its write verbs crashed; after it, they are absent. If the deployed `geecs-mcp` service is running from a checkout on this branch, its permission profile will name five tools the server no longer registers: inert (an unknown permission entry matches nothing), but worth pruning at the next profile touch.

## Checklist

- [x] Version bumped with `poetry version minor`; changelog updated.
- [x] Public-repo hygiene checked — no facility values added to runtime code (the one `osprey-htu-assistant` hit in the diff is a pre-existing `e.g.` example line I rewrapped).
- [x] Base is the integration branch `feature/native-bluesky-plans`, where #920–#923 landed. The master merge remains the maintainer's call.
- [x] Exact check results recorded above, including what did not run and why.
- [x] Adversarial review report + disposition posted as a comment — two rounds, 12 findings, all dispositioned; three waivers, one filed as #925.


skbarber:
## Adversarial review + disposition

Two rounds, both by a fresh-context reviewer given only the diff and the standard brief. Round 1 found 8 findings; round 2 confirmed all 8 dispositions and found 4 more. Every finding below was reproduced before being fixed, and every new test was verified to fail against its own drift.

### Round 1

**1. HIGH — a sixth tool was broken by the same rebuild, and this PR had re-advertised it.** `list_scan_configs(kind="save_sets")` routed to `ConfigsRepoResolver.list_save_sets`, removed alongside the client verbs (presets carry the device group). Reproduced against the real resolver:

```
{"ok": false, "error_kind": "not_found",
 "message": "listing save_sets failed: 'ConfigsRepoResolver' object has no attribute 'list_save_sets'"}
```

An agent reads that as *this experiment has no save sets*, not *this tool is broken*. **FIXED** — dropped from `_CONFIG_KINDS`, `_FakeResolver`, its assertion, and every place that advertised it. `_FakeResolver` had defined the method exactly as `_FakeClient` had: two independent copies of one bug class.

*Deliberately not changed:* `read_tools.py:221` `save_sets=list(summary.save_sets)` reads an archived run's recorded metadata off Tiled's `RunSummary` (field confirmed present) — a different collaborator. The reviewer independently confirmed that leaving it is correct.

**2. MEDIUM — the new pin covered only half the seam.** **PARTIALLY FIXED.** `QueueStatus`'s fields were the unpinned half: both fakes fabricate `status()` as a `SimpleNamespace`, so a renamed field left the suite green and every reader raising `AttributeError` in production. Now pinned. **WAIVED:** signature/return-shape pinning and the client-construction seam → **#925**, because the fix belongs beside the protocol in GeecsBluesky where `GeecsScanner`'s `DemoQueueClient` (production code, currently unpinned) also gets it. The reviewer argued 2c specifically should not block: this server gates nothing, and the two safety-critical shapes are shared with the scanner, which has far more coverage.

**3. MEDIUM — root `CLAUDE.md` contradicted the package's own safety vocabulary.** It called `resume_scan` and `clear_queue` "halt verbs" and the server "READ-ONLY", while `tool_names.py` classifies resume as **Q, not S** because it restarts motion. **FIXED** → "the halt family (stop/pause) + three gated go verbs".

**4. MEDIUM — four published surfaces outside `docs/geecs_mcp/` still sold the submit path.** **FIXED:** `docs/agentic/index.md`; both `docs/geecs_schemas/` admonitions, which said the MCP "is rewired onto" presets when it was deleted instead; and `docs/sites/data_flow/`, whose map drew an OSPREY→queue submission arrow this arc removes. Plus two in-page leftovers the reviewer caught.

**5. LOW — `pyproject.toml`'s published description said "scan submission".** **FIXED.**

**6. LOW — stale contract text in surviving modules.** **FIXED:** `errors.py` prose no longer describes the deleted submit path; the spliced `tool_names.py` comment un-spliced; the `wired` fixture no longer patches `get_experiment` (suite still green, confirming nothing reads it). `task_timeout` **kept** in `ERROR_KINDS` with a note that it has no producer — dropping a published envelope value would make `make_error` raise on a kind agents may already branch on. Reviewer agreed.

**7. LOW — the `geecs-schemas` edge asserted an import that does not exist.** Nothing under `geecs_mcp/` imports it; the listing tools duck-type catalog rows. **FIXED as honest labelling** (both the pyproject comment and the dependency graph now say TEST-ONLY). Not moved to the dev group — the resolver returns real models at runtime, and the reviewer agreed moving it would make the suite's realism depend on a dev-only install.

**8. LOW — the registration tests only asserted one direction.** A tool added without a group entry would ship in neither `allow` nor `ask`/`write_tools`. **FIXED** — the partition is now pinned both ways plus no-overlap.

### Round 2

**NEW-1 — the `save_sets` fix had disarmed a guard test.** `test_list_scan_configs_without_experiment` passed `"save_sets"`, so once that stopped being a kind the kind guard answered first and the test never reached the `resolver is None` branch it exists to protect. **Proven vacuous** by deleting the guard and watching it pass; **FIXED**, and it now fails when the guard goes. The fix for one mislabelled refusal had quietly disarmed the test guarding against another.

**MISS-1 — a fifth `save_sets` advertisement:** the FastMCP `instructions` string, which is the catalog description every connecting agent reads first. The changelog's "four places" was one short. **FIXED.**

**NEW-2 — the new cross-reference pointed at the wrong section.** `[No write path here](#the-safety-model)` — that bullet lives under `#where-it-sits-in-the-architecture`. The anchor resolved, so the build stayed quiet and the reader landed in the wrong place. **FIXED.**

**NEW-3 — two more `data_flow` bullets** in the list already corrected still described a submit cap and acknowledge loop, and that map's GEECS-Schemas panel still named the MCP among the packages importing the models. **FIXED.**

### Tests

`GEECS-MCP/tests/test_seam_pins.py` holds the fake-vs-real assertions in one module rather than beside each double — precisely because the review found the drift a second time, one file over from the first. **Each of the five was verified to fail against its own injected drift** (I checked four by hand, the reviewer checked the rest and split the partition into its three directions):

| drift injected | result |
|---|---|
| a removed client verb back on either fake | FAILS, naming which double |
| a fabricated `status()` field `QueueStatus` lacks | FAILS, naming which double |
| a removed resolver capability back on `_FakeResolver` | FAILS |
| `save_sets` back in `_CONFIG_KINDS` | FAILS |
| a tool registered with no group entry | FAILS |
| a tool in two safety groups | FAILS |
| the `resolver is None` guard deleted | FAILS (after NEW-1) |

Final: **82 passed, 1 skipped**; lint clean; `mkdocs build` exit 0, zero ERROR lines; server registers exactly 14 tools with none of the five removed names.

### Open waivers (cheap to veto)

- Signature/return-shape conformance → **#925**, cross-package.
- `task_timeout` kept in `ERROR_KINDS` without a producer.
- `geecs-schemas` stays a main dependency, labelled test-only.

No undispositioned findings.

PR #928, opened 2026-09-17
Add two-axis Grid analysis to Data Portal

skbarber:
## What & why

Two-axis scans need a spatial view of per-bin results. Add a Grid tab with paired scalar and error maps, independent mean/median and error selectors, and the same shot filters used by Plot and Images. Filtered and unacquired positions stay visible.

The pure Data Utils calculation reconstructs native rectangular/Sweep geometry, including snaking, nonuniform and logarithmic axes, relative coordinates and repeated visits. Nonrectangular trajectories render measured points without interpolation. The portal adds square plotting areas, URL-carried state, bin/image inspection, reproducible notebook code, and the existing send-to-scan-log action for each map.

Scope: one grid-view feature across its data and presentation layers. Higher-dimensional slicing is deferred. ScanInfo and ScanAnalysis first-motor labels are unchanged. No acquisition code or deployment changes.

Review follow-up: Grid, Plot and Images now share Data Utils shot-identity resolution, including renamed s-file columns and sparse legacy shot numbers. The same independent reviewer confirmed both fixes at `faed00e8` (71 targeted tests; no surviving findings).

Status: **ready**. Both JavaScript findings from the follow-up review are fixed in `05e614c8` and confirmed by the reviewer at that head (no surviving findings). `77b0885e` adds the one test that was missing — the fixed validator was pinned, but the inputs reaching it were not, so reverting only the `onchange` attributes left the whole suite green. Final-head CI is green and real-data acceptance is complete; see the hardware-verification section.

Versions: geecs-data-utils 0.34.0; geecs-data-portal 0.29.1.

## Checklist

- [x] Package versions updated with Poetry
- [x] CHANGELOG entries added
- [x] Exact affected-suite results recorded below
- [x] Base is master; maintainer performs the merge
- [x] Public-repo hygiene checked
- [ ] Review complete: original adversarial findings were fixed and independently confirmed; Claude's two subsequent JavaScript findings are implemented and awaiting reviewer confirmation

## Tests

- Data Utils: **445 passed**.
- Data Portal: **335 passed**.
- Web Theme: **182 passed**.
- Changed-file pre-commit hooks passed.
- Headless Chrome: desktop/mobile layouts, square plot dimensions, independent statistics, filter behavior, reload/URL preservation, and cell-to-Images navigation.
- Actual average/error PNG exports verified with intercepted logbook POSTs: correct map/caption, same-entry append, preserved Grid URL; no live logbook writes.
- Full `./scripts/check.sh --all` attempted. Repository-wide pre-commit passed. Additional passing suites: root **83 passed / 63 deselected**; ImageAnalysis **293 passed / 3 skipped / 19 deselected**; ScanAnalysis **182 passed / 1 skipped / 2 deselected**; Schemas **494 passed / 7 deselected**.
- The full local sweep was **not green**: package-isolated environments are unavailable in this worktree; the reused root environment produced missing-module/stale-editable-import collection errors in Bluesky, gateways, Core, Portal, LogTriage, MCP, Logbook, and Scanner. Portal and Theme were separately run successfully in the configured Portal environment against worktree sources (counts above). CI is the clean-environment cross-package gate.

## Hardware verification

**Complete.** Real-data acceptance ran on `77b0885e`: Undulator `26_0917/Scan002`, a 3x4 **snaking** grid on
`U_S1H.current` (range -1 -> +1 A, 3 points) x `U_S1V.current` (list `[-1, -0.4, +0.4, +1]` A, nonuniform),
`combine: product`, `snake: true`, 5 shots/step, strict, trigger profile `HTU-NoGas`, `UC_Amp4_IR_input` with
images. 12 positions, 60 shots, 76 s, `ScanEndInfo = "success"`.

Verified against acquired data, not fixtures:

- **All 12 cell coordinates and bin identities correct**, including the reversed middle row — the observed
  traversal was checked against a computed plain raster and does not match it, so the snaking path is genuinely
  exercised. Readbacks track setpoints to ~1e-4.
- **Nonuniform inner axis** reconstructed: `x_values: [-1.0, -0.4, 0.4, 1.0]`.
- **Filtered cells stay visible** — filtering `MaxCounts` to `[1000, 4000]` gives `pass/total: 43/60`, all 12
  cells still rendered, bins 10-12 as `filtered` (`acquired: 5, passing: 0`), bin 9 recomputing its center from
  the 3 survivors (1937.0 -> 1985.0).
- **Image membership consistent** — `bin-images` returns 12 bins x 5 shots with shot lists identical to the grid
  cells'; 60 image files on disk.
- **Scan-log export confirmed** against the live logbook.

Setpoints were restored to 0.000 A afterwards, per the acceptance convention in #807 (the `sweep` plan
auto-restores relative axes only; these were absolute).

Full detail, including the failed first attempt (`Scan001`, a device-side setpoint-write refusal, no data) and
the exact cell table, is in the hardware-verification comment on this PR.


skbarber:
Fresh-context adversarial review of `2e859677...590db39a` completed through correctness, redundancy and placement. One surviving finding:

**[P2] Preserve the shared shot-identity fallback.** `scan_grid.py` duplicates the portal's `shot_axis_for_frame` but omits `Shotnumber (s-file)`. In a supported union-frame collision, the second bin has a valid scalar but Grid reports no member shots and disables its Images button. Images correctly identifies shot 2. The existing union loader explicitly supports this renamed s-file identity.

**Disposition: fix in progress.** Move the shared identity helper into Data Utils and use the same implementation from Grid, Plot and Images. Add regression coverage for the collision and verify the fix with the original reviewer.

The reviewer found no other surviving issues. Real-data acceptance remains outstanding; this review did not operate hardware.

skbarber:
Adversarial-review disposition at `faed00e8`:

1. **P2 — lost member shots after a native/s-file column collision:** fixed in `53e9c110` by moving `shot_axis_for_frame` into Data Utils and using one resolver for Grid, Plot and Images.
2. **P2 — first fix regressed sparse s-file identities:** fixed in `ea69795f`; explicit native and legacy shot identities take precedence over row positions, and the unambiguous suffixed s-file identity wins over a colliding native column. The former positional test expectation and release notes were corrected in `faed00e8`.

The **same independent reviewer rechecked both fixes**, reproduced the original collision and sparse-file cases, checked the no-identity fallback, and ran **71 targeted tests**. Its conclusion: **no surviving findings**. No findings were waived.

Final affected-suite validation: **Data Utils 445 passed; Portal 322 passed; Web Theme 182 passed**. Changed-file hooks passed. The PR body records the separate full local check attempt and its environment limitations.

The PR remains a draft: final-head CI and real-data acceptance are still outstanding. No production checkout was changed and no hardware was operated.

skbarber:
Fresh-context adversarial review at `faed00e8`.

Branch suites re-run: 36 grid tests and 322 portal tests, all green. The pure calculation and the `/api/grid` endpoint were probed with hand-built edge cases — float/NaN bin columns, fully-filtered runs, bins beyond the plan, relative + logarithmic sweeps, 1xN grids, repeated visits, index/log axes, single-motor point mode — and the snake/visit/relative-anchor arithmetic was re-derived by hand against expected coordinates. The Python side held up on all of them; no findings there.

Both findings are in the page's JavaScript.

## 1. Medium — opening the Grid tab silently restricts the Images tab

`GEECS-DataPortal/geecs_portal/templates/grid_script.html:60`

```js
if (!cells.some(c => String(c.bin) === String(S.gridbin))) S.gridbin = cells.length ? String(cells[0].bin) : "";
```

This runs on **every** successful `/api/grid` response, and `writeState()` then persists it to the URL. So `S.gridbin` is not a user choice — it is a side effect of rendering the tab.

`GEECS-DataPortal/geecs_portal/templates/run.html:1244` treats it as one:

```js
if (!data.bins.length || (S.gridbin && !data.bins.some(b => String(b.bin) === String(S.gridbin)))) {
  grid.innerHTML = `<div class="plotmsg">no bins (check filters / bin column)</div>`;
```

Failure scenario: an operator opens a two-axis run, clicks Grid to glance at the map, then switches to Images in binned view. Instead of the full bin gallery they get only bin 1, with no action of theirs having selected it.

The sharper case: if their `bincfg` uses numeric binning (`bin_width` / `quantile_bins`, producing labels like `3.25`) or a different bin column, no label can ever equal an acquisition-bin number. The guard then fires and the tab reports `no bins (check filters / bin column)` — which misdiagnoses the cause, since the filters and the bin column are both fine.

The `Grid selection: bin N [show all bins]` banner makes this recoverable, but not self-explanatory. Suggested fix: gate the Images filter on a flag set only by `gridOpenImages()`, rather than on `S.gridbin` merely being non-empty.

## 2. Low — clearing a Grid settings number field bakes an invalid config into the URL

`GEECS-DataPortal/geecs_portal/templates/run.html:257`

```html
<input id="grid-min_count" type="number" min="1" value="3" onchange="changeGrid('min_count',Number(this.value))">
```

Clearing the field leaves `this.value === ""`, and `Number("") === 0`, which violates `GridConfig.min_count`'s `ge=1`. The request 400s with a raw pydantic message in `#grid-message` and both maps vanish. Because `changeGrid` writes the value into `S.gridcfg` and calls `writeState()`, the invalid config is also baked into the URL — so a link shared or bookmarked at that moment is broken for the recipient too.

The `min="1"` attribute does not help here: nothing submits a form, so no constraint validation runs.

Lines 255-256 have the same shape — clearing "Upper percentile" yields `0` and trips the `lower >= upper` check.

Suggested fix: guard the coercion (`Number(this.value) || 3`, or skip the update when the field is empty) before writing it into the config.

skbarber:
Both findings from [Claude's review](https://github.com/GEECS-BELLA/GEECS-Plugins/pull/928#issuecomment-5710047979) are addressed in `05e614c8`, ready for Claude's recheck:

1. **Images restriction caused by visiting Grid:** the map inspector keeps `gridbin`; the Images gallery now uses a separate URL-carried `imagebin`, set only by **View this bin's images**. Merely rendering/selecting Grid cells leaves the gallery and its existing numeric/custom binning alone. Applying a different bin definition clears the explicit image selection. “Show all bins” clears only the Images restriction, preserving the map selection.
2. **Empty/invalid numeric edits:** a dedicated numeric handler validates finite values, integer minimum counts, percentile limits and ordered bounds before changing config or URL state. Empty edits restore the last applied value; other invalid edits restore it and give a short explanation. Neither case requests new maps or hides the current ones.

Validation: **335 Portal tests and 182 Web Theme tests passed**, including 13 new executable JavaScript regression cases. The updated proxy/URL-state test also passed. Chrome verified empty/invalid edits preserve the maps and URL, a passive Grid visit leaves Images unrestricted, explicit image selection survives reload, and clearing it does not clear the selected map cell. Changed-file hooks passed.

The reviewed head `faed00e8` passed CI; the new head will run CI again. These fixes are **implemented, pending the original reviewer's confirmation**. No deployment or hardware operation was performed.

skbarber:
Confirmation of the fixes at `05e614c8`. Both findings are resolved. One follow-up pushed as `77b0885e`.

## 1. Grid visit restricting the Images gallery — FIXED

`S.imagebin` is written in exactly four places, all deliberate: URL restore (`run.html:791`), the explicit **View this bin's images** button (`grid_script.html:155`), `applyBinset()` clearing it (`run.html:1570`), and "show all bins" clearing it (`run.html:1233`). `S.gridbin` is no longer read anywhere in the image path — the only remaining reads outside the map inspector are `writeState`/`syncRailLinks`.

A passive Grid visit still auto-selects a map cell into `S.gridbin` (`grid_script.html:81`), but the gallery keys off `S.imagebin`, which stays empty. The "no bins (check filters / bin column)" guard is now gated on `imagebin`, and the button that sets it also sets `bincfg` to the same acquisition-bin column the cell numbers come from — so the numeric-binning misdiagnosis is unreachable without an explicit selection. Clearing either selection leaves the other intact.

## 2. Raw `Number(this.value)` on the Grid number inputs — FIXED

All three number inputs route through `changeGridNumber` (`grid_script.html:37-57`). Empty/whitespace, non-numeric, NaN/Infinity, non-integer `min_count`, `min_count` below 1, out-of-range percentiles and `lower >= upper` all restore the last applied value without writing `S.gridcfg`, without calling `writeState`, and without requesting new maps. The accepted boundaries were checked against the server rather than assumed: `lower=0` and `upper=100` do validate in `GridConfig` (`scan_grid.py:40-46`), and the JS `defaults` object matches the model's defaults exactly.

The Visit control keeps `Number(this.value)`, but it is a `<select>` populated from integers >= 1, so it cannot produce an invalid value. Not a defect.

## No new defects at medium or above.

## Both fixes were verified to bite, and one gap was found

Each fix hunk was reverted in a scratch worktree and the tests re-run:

- Finding 1 reverted: `test_grid_visit_does_not_restrict_image_bins` **failed** with `0 !== 2` — zero figures rendered instead of two, the exact misdiagnosis, with 3.25/5.25 bin labels.
- Finding 2 reverted: **9 of 12** parametrized cases failed, while the 3 genuinely-valid cases still passed. Correct discrimination.

The gap: reverting **only** the HTML wiring — `run.html:255-257` back to `onchange="changeGrid('min_count',Number(this.value))"`, leaving `changeGridNumber` intact — left the entire suite green at **335 passed, exit 0**. `test_numeric_edits_preserve_last_valid_state` calls the validator directly and never reads `run.html`, so it pins the validator but not the fact that the inputs reach it. A future edit could silently reintroduce finding 2 with CI green. (Finding 1's test does read `run.html`, so that side was already covered.)

`77b0885e` closes it: it parses `run.html` with stdlib `HTMLParser` and asserts the three `grid-*` number inputs route through the validator and do not coerce their own value. Test-only, no production code touched; `run.html` is byte-identical.

Verified to bite by the same method — with the wiring reverted, the 335 existing tests still pass and only these three fail:

| | wiring reverted | wiring intact |
|---|---|---|
| before `77b0885e` | 335 passed, exit 0 | 335 passed, exit 0 |
| after `77b0885e` | 335 passed, **3 failed** | 338 passed, exit 0 |

No version bump or CHANGELOG entry: test-only, and the PR already bumped to 0.29.1.

## Suites

Portal **338 passed, 0 failed, exit 0** (run against a detached worktree of the branch using the portal's own venv, since the main checkout is on master and lacks `geecs_data_utils.scan_grid`).

## Still outstanding

Review-wise this is done — no surviving findings, and I would stop here rather than open another round. The remaining gate is the one already recorded in the PR body: **real-data acceptance on a 3x4 snaking grid**. Everything verified so far, on all sides, is synthetic or headless-browser. For a feature whose job is reconstructing scan geometry — snaking, repeated visits, relative anchors — that is the check that has not been done.

skbarber:
## Hardware verification — real-data acceptance

The OWED item in the PR description is discharged, with one leg outstanding (noted at the end). Run on `77b0885e`, against a portal serving this branch and the live share and Tiled.

**Scan:** Undulator `26_0917/Scan002`, plan `sweep`, trigger profile `HTU-NoGas`, strict, 76 s, `ScanEndInfo = "success"`. `UC_Amp4_IR_input` with images — 60 PNGs plus the h5.

**Trajectory:** a 3x4 **snaking** grid, uniform outer axis and nonuniform inner axis, chosen so one scan exercises both claims.

- `U_S1H.current` — range −1 → +1 A, 3 points
- `U_S1V.current` — list `[−1, −0.4, +0.4, +1]` A
- `combine: product`, `snake: true`, 5 shots/step → 12 positions, 60 shots

### Acquisition landed as planned

| bin | S1H | S1V traversal |
|---|---|---|
| 1–4 | −1.0 | −1.0 → −0.4 → 0.4 → 1.0 |
| 5–8 | 0.0 | **1.0 → 0.4 → −0.4 → −1.0** |
| 9–12 | 1.0 | −1.0 → −0.4 → 0.4 → 1.0 |

The middle row is reversed, and the observed order was checked against a computed plain raster and does **not** match it. This is a genuine snaking trajectory, not a raster relabelled.

### Grid reconstruction — correct

`/api/run/{uid}/grid` returns `kind: grid`, 12 cells, `visits: 1`, `pass/total: 60/60`, `bin_column: bin_number`, `x_values: [-1.0, -0.4, 0.4, 1.0]`, `y_values: [-1.0, 0.0, 1.0]`.

All 12 cell coordinates and bin identities are correct **despite** the reversed middle row — e.g. bin 5 at `(x=1.0, y=0.0)` and bin 8 at `(x=-1.0, y=0.0)`, matching the s-file's `(S1H=0.0, S1V=1.0)` and `(S1H=0.0, S1V=-1.0)`. Every `x_readback`/`y_readback` tracks its setpoint to ~1e-4. The snaking and nonuniform-axis reconstruction paths are therefore verified against acquired data rather than synthetic fixtures.

### Filtered cells — correct

Filtering `UC_Amp4_IR_input MaxCounts` to `within [1000, 4000]` gives `pass/total: 43/60`, and **all 12 cells remain visible**:

| bin | acquired | passing | status | center |
|---|---|---|---|---|
| 1–8 | 5 | 5 | ok | unchanged |
| 9 | 5 | 3 | ok | 1937.0 → **1985.0** |
| 10–12 | 5 | 0 | **filtered** | `None` |

Cells whose shots are entirely filtered stay in the map as `filtered` rather than vanishing, and a partially-filtered cell recomputes its center from the survivors.

### Image membership — consistent

`/api/run/{uid}/bin-images` returns 12 bins x 5 shots whose shot lists are identical to the grid cells' (`bin 1 → [1,2,3,4,5]`, `bin 2 → [6,7,8,9,10]`, …). 60 image files on disk.

### Housekeeping

Both setpoints were restored to 0.000 A afterwards (readbacks ~1e-4), per the acceptance convention in #807. Note the `sweep` plan auto-restores **relative** axes only (`reset_positions_wrapper`); these axes were absolute, so the restore was explicit.

`26_0917/Scan001` is a failed shell from a first attempt that died on the first setpoint write — `CANothing: undulator:u_s1h:current:SP: Channel write request failed`, a device-side state, fixed before the retry. It holds `scan.log` and `ScanInfo` only, no data.

### Still outstanding

**The scan-log export leg was not exercised against a real scan.** It writes to the logbook, so it was left alone here; the existing coverage remains the intercepted-POST testing already recorded in the PR description.

### Observation, not a finding

`MaxCounts` increments by exactly +192 per bin across bins 1–7 in acquisition order (2377, 2569, 2761, 2953, 3145, 3337, 3529) — six identical deltas tracking shot number rather than magnet position. That looks like a synthetic source rather than beam. It has no bearing on the geometry under test here, but it means the scalar values in this run should not be read as physics.

Issue #756, opened 2026-09-01
Decide how non-triggered (snapshot-role) devices record non-scalar data


skbarber:
## The gap

A save-set entry with `role: snapshot` (legacy `synchronous: false`) — a device that is **not shot-triggered** — has no path to save its non-scalar data (camera frames, traces) through the engine:

- `images: true` on such an entry is a **silent no-op** for any devicetype. The engine never drove native saving for the snapshot role (`session.snapshot` has no save-images step; `CaSnapshotReadable` sets no `_save_nonscalar_data`, so `session._configure_saving` skips it).
- Since #751 (GeecsBluesky 0.70.1) the capture daemon does not own snapshot-role devices either: they are dropped from `capture_devices` with a WARNING, because `CaSnapshotReadable` has neither a save-control child nor an `acq_timestamp` join column (#702's two broken shapes).
- The device's own save flag is therefore neither commanded nor suppressed. Whatever it was left at out of band is what happens.

#754 covers the *warning* at the DB-free role seam. This issue is the **design question** underneath it: how should a non-triggered device's non-scalar data be recorded at all?

## What "snapshot" means today

The schema describes the role as "slow readbacks that don't produce one value per shot" (`SaveSet` entry `role`). The device reads its scalar variables once per bin via ophyd-async, with no shot identity. Non-scalar data was never part of that contract — the field exists on the entry, the role ignores it.

## Options to decide between

1. **Snapshot role grows non-scalar support** — the other remedy #702 names: a save-control child (so the eager toggle-off `save="off"` reaches it) plus an `acq_timestamp` join column, so the capture daemon can own it and its stack row-joins like a sync camera. Semantics question: one frame per *bin* (matching the scalar cadence) or whatever frames arrive during the bin?
2. **Snapshot role stays scalar-only and the schema says so** — reject or warn on `images: true` with `role: snapshot` at validation time (#754's warning becomes a schema rule), and document "to record a camera's images it must be shot-synchronized". Cheapest; loses the use case entirely.
3. **A distinct per-bin non-scalar role** — e.g. a "reference frame per bin" capture: the daemon grabs one frame at bin boundaries, stamped with the bin index rather than a shot id. Fits the record-vs-live doctrine in #738 (the daemon owns the record), but is a new role in the schema vocabulary and a new daemon mode.

Whichever is chosen, the PV-producer direction in #738 is orthogonal: it covers *live scalars derived from frames*, not recording the frames.

[...]

PR #934, opened 2026-09-17
GeecsBluesky 0.94.2: MagSpecStitcher is a triggered devicetype (native saving for U_BCaveMagSpec)

skbarber:
## What

`looks_triggerable` gains `TRIGGERED_DEVICETYPES`, an allowlist of devicetypes that acquire per shot off *other* devices' frames and therefore carry no trigger-named DB variable of their own. First and only entry: `MagSpecStitcher`. Matched on the whole devicetype, lower-cased, checked after the authoritative `acq_timestamp` DB-row rule and before the trigger-source exclusion.

GeecsBluesky 0.94.1 → 0.94.2 (patch). Hot fix: 19 lines in `namespace.py` (one constant + one `if`, the rest docstring), CHANGELOG, version.

## Why

Scan004 of 26_0917 (`count`, 50 shots, strict) listed `U_BCaveMagSpec` as a detector and wrote no `U_BCaveMagSpec/` folder. The scan log shows "native saving on" for the three magspec cameras and the ICT and nothing for the stitcher; the s-file carries its `DeltaAngle`/`timestamp` columns without an `acq_timestamp` column.

Root cause: the stitcher's DB rows mention no trigger and the DB has no `acq_timestamp` row for it, so the name rule classified it as a snapshot device and the namespace built a `CaSnapshotReadable`. Only `GeecsDetector` drives `LocalSavingPath`/`Save` (both settable on this device); the snapshot class ignores them, and a preset's `save_images: true` is a silent no-op for it.

The stitcher *is* a per-shot acquirer: live 2026-09-17 its gateway `acq_timestamp` advanced every shot with a value identical to `UC_BCaveMagSpecCam1`'s. Its stitched `Image` never reaches remote subscribers; only a PVA gateway local to its host could capture it, and that host runs none today (it is not in `file_plugin_addr_list`), so native PNG saving is this device's only record for now. Under the legacy scanner it saved (`scans/Scan011/U_BCaveMagSpec/*.png` on 26_0501; the legacy save set had it `synchronous: true`, `save_nonscalar_data: true`); no scan since July has the folder.

With the allowlist the stitcher becomes a `GeecsDetector` with native saving on and no HDF plugin (its host serves none, and the image is local-only anyway): the same shape `U_BCaveICT` had in Scan004. Its scalars gain the shot-join column.

Not a #756 case (that issue is about genuinely non-triggered devices). The alternative knob, an `acq_timestamp` row on the devicetype in the DB, was set aside because the LabVIEW program does not implement that variable.

## Tests

No new pinning test (maintainer's call: hot fix, verified live). Existing suites touching the namespace, run with the package env against this branch:

- `tests/test_namespace.py`, `test_connect_on_demand.py`, `test_qserver_startup.py`, `test_run_action.py`, `test_native_optimization.py`, the three `*_hardware.py` files: **63 passed, 3 skipped** (the hardware files skip off-lab), rc 0.

## Hardware verification

**OWED (maintainer, live in production):** a strict `count` with the BCaveMagSpec preset. Expected:

- scan.log gains `u_bcavemagspec-nonscalar_save_path: native saving on → …/ScanNNN/U_BCaveMagSpec`
- `scans/ScanNNN/U_BCaveMagSpec/` (and the `-interpDiv` / `-interpSpec` siblings) populated, one file per shot
- the s-file gains a `U_BCaveMagSpec acq_timestamp` column matching the cameras' stamps
- the stitcher's host is not in `file_plugin_addr_list` (review finding 1); when the PVA rollout bootstraps that box, the first scan tells whether the stitched image publishes to a local subscriber
- watch the per-shot cadence: the stitcher stamps *after* it has all three frames, so the strict wait now includes its lag; any `no shot within` naming `u_bcavemagspec` is the thing to report

## Deploy

Editable install on the worker host: `git checkout fix/stitcher-triggered-devicetype` in the qserver checkout + restart `geecs-qserver` (no dependency change, no reinstall needed; `/health` keeps showing 0.94.0 until the next `poetry install`, cosmetic).


skbarber:
## Adversarial review of #934 @ 9df10370 (fresh-context reviewer, pasted in full)

Scope checked: `namespace.py` (`_build`, `DeviceRoster.from_geecs_db`), `devices/detector.py`, `devices/hdf_plugin.py`, `plans/gated.py` (`shot_clock`, `refuse_native_essentials`), `qs_client/presets.py`, `qs_client/submit_preflight.py`, `assets/{specs,registry}.py`, `GeecsPvaGateway/{config.py,DEPLOYMENT.md}`, `tests/test_namespace.py`, issue #756, root + GeecsBluesky `CLAUDE.md`. Hygiene, scan-folder invariant, version/CHANGELOG: clean. `DeviceRoster.triggered` (the existing per-device override) is never populated by `from_geecs_db`, so the allowlist does not duplicate a live knob.

### 1. P2 — The reclassification opens a plugin-arming path for an image that never publishes
`_build` attaches an HDF plugin iff the DB lists an image variable **and** the device's endpoint is in `[pva] file_plugin_addr_list` (`namespace.py:564-569`); the PVA gateway serves `:hdf1:` PVs for every image-typed row (`GeecsPvaGateway/geecs_pva_gateway/config.py:138`, `GeecsPvaGateway/DEPLOYMENT.md` §"The file plugin"). The stitcher has an `Image` row whose frames never reach subscribers (PR body). Scenario: the magspec box is rolled for the plugin (the #738 rollout adds boxes to that list one by one) → the stitcher, now a `GeecsDetector`, gets `ADHDFDataLogic` armed on `u_bcavemagspec:image:hdf1:` → every `trigger()` waits `exposure_timeout` for a count that never arrives → `GeecsTriggerTimeoutError` on every shot, and gated mode would accept it as "plugin-backed". As a snapshot device it was immune. Fix is one line: skip `plugin_vars` for `TRIGGERED_DEVICETYPES` (or a `LOCAL_IMAGE_DEVICETYPES` note), and add "stitcher's host is not in `file_plugin_addr_list`" to the hardware checklist.
*Refute attempt:* "its host serves none" — true today, but a deployment fact the code does not pin; the allowlist's own docstring says the image is local-only.

### 2. P3 — Behaviour change not in the CHANGELOG: gated presets naming the stitcher are now refused
With `native_save=True` and no plugin, the stitcher now trips `refuse_native_essentials` (`plans/gated.py:164-176`) and the preflight (`qs_client/submit_preflight.py:350-362`) for any gated preset listing it with `save_images: true`. Previously it passed silently (no-op save). Correct behaviour, but a user-visible refusal shipped as a "hot fix" with no note; the fix is a CHANGELOG line ("use `save_images: false` in gated presets").
*Refute attempt:* the BCaveMagSpec preset likely already carries a native-only device (the ICT), making it strict-only already — I could not read the preset (configs repo), so it may be moot.

### 3. P3 — Second spelling of the stitcher's devicetype; the assets registry already encodes it as a native per-shot writer
`assets/specs.py:9` has `MAGSPEC_STITCHER_DEVICE_TYPE = "MagSpecStitcher"` and `assets/registry.py:389-400` declares its native PNG + `-interpSpec`/`-interpDiv` outputs (`supports_device_type()`). The new `frozenset({"magspecstitcher"})` is a hand-typed copy; a future entry needs both tables edited. Cheapest fix: `frozenset({MAGSPEC_STITCHER_DEVICE_TYPE.lower()})` (`specs.py` has no imports; `assets/__init__` pulls only numpy). Deriving the whole allowlist from the registry is *not* recommended in a hot fix (it would widen classification to every registry devicetype).
*Refute attempt:* the registry is read-back-only today (consumers: `assets/readback.py`, `assets/tiled_readback.py`), so it is not a classifier — but the spelling duplication stands.

### 4. P3 — No pinning test, though the one-liner bites
`assert looks_triggerable([row("DeltaAngle")], "MagSpecStitcher")` in `test_looks_triggerable_heuristic` (`tests/test_namespace.py:108`) is False without the allowlist (no `acq_timestamp` row, no `trig` match, not a trigger source) and True with it. Maintainer's call per the PR body; noting that the cost is one line.
*Refute attempt:* the PR explicitly waives it; I confirmed the test would fail on the pre-PR code by tracing the three rules.

Findings I tried and dropped: `shot_clock` picking the late-stamping stitcher as the gated clock (gated mode refuses it as an essential anyway, and as `.scalars` a late clock is benign — members are already cached); exact-match vs substring semantics between the two devicetype sets (deliberate per the docstring, and the registry keys the same exact string); docs/contract drift (no doc restates the rule's internals; `CLAUDE.md:136` names only the function).

---

## Dispositions (author)

| # | Sev | Disposition |
|---|---|---|
| 1 | P2 | **Waived.** The plugin attach is keyed on `file_plugin_addr_list` by design, for every device. The stitcher's host is one of the four boxes the PVA rollout intends to bootstrap, and a local PVA gateway there is exactly how its image would be captured: the image is local-only on the wire, which a local subscriber satisfies. Whether it publishes locally is checked at that roll like every other box, and an arm timeout is loud (the #894 shape), not silent. A code guard would block the intended capture. Hardware checklist gains: the stitcher's host is not in `file_plugin_addr_list` today. |
| 2 | P3 | **Fixed in dc821e8a**: a "Changed" entry under 0.94.2 states the gated refusal and the two ways out (`acquisition='strict'`, or `save_images: false`). The BCaveMagSpec preset has no ICT, so the refusal is real for it. Reviewer confirmation follows in the next comment. |
| 3 | P3 | **Waived for the hot fix.** `namespace.py` imports nothing from `assets/`, and `assets/__init__` pulls in event_model and the Tiled config reader (the reviewer's "only numpy" undercounts it); widening the namespace import surface for one string is not a hot-fix change. Follow-up: give the devicetype constant an import-free home both can share. |
| 4 | P3 | **Waived** on the maintainer's explicit instruction (hot fix, verified live). One-line add on request. |

skbarber:
## Hardware verification

Maintainer ran a scan live on the worker host with the branch checked out (2026-09-17, HTU): the `U_BCaveMagSpec/` folder and its `-interpSpec` / `-interpDiv` siblings are written again. The OWED item in the body is met; the stitcher's host is not in `file_plugin_addr_list` today (review finding 1).

skbarber:
## Review round 2 @ dc821e8a (same reviewer)

- Finding 2: **CONFIRMED** fixed by the CHANGELOG entry; both remedies match the refusal text.
- Finding 1: waiver stands; the reviewer flagged that the PR body's "no file plugin can capture it" contradicted the waiver's "a local PVA subscriber could". The body now says: never reaches *remote* subscribers, only a gateway local to its host could capture it, and that host runs none today. Also added to the hardware checklist.
- Findings 3 and 4: waivers accepted.

Review closed; no undispositioned findings. Hardware-verified by the maintainer (comment above). Ready to merge once CI on dc821e8a is green.

PR #935, opened 2026-09-18
GeecsPvaGateway 0.11.0: compress frame stacks by default (zlib)

skbarber:
## What

One-line behaviour change: the file plugin's `Compression` parameter now defaults to `zlib` instead of areaDetector's `None`, so the HDF5 frame stacks are written with shuffle + gzip level 1.

## Why

The write path's `zlib` branch has been there since #806 (`file_plugin.py:808`) and has never run. Nothing ever puts that PV: stock `ophyd_async` declares the signal on `NDFileHDF5IO` but `ADHDFDataLogic` never sets it, and our `GeecsHdfIO` adds only `Rewind`/`WriteStatus`/`WriteMessage`. So every stack on the share is raw frames, for no reason other than inheriting areaDetector's default.

Reference numbers for these exact filters, measured on real Scan003 frames by the since-deleted central capture daemon (GeecsBluesky 0.65.0): **2.04 MB compressed vs 7.92 MB raw vs ~2.5 MB for the equivalent LabVIEW PNGs**, at 3.9 ms/frame on 600×600 and ~25 ms/frame on 1025×1281. Note those were measured in the *central* single-writer daemon; the plugin is per camera server, so the per-box budget differs and deserves its own measurement if we ever push frame rates (see Hardware verification).

## What does not change

- **Schema** — `geecs-capture/1` unchanged; gzip and shuffle are built-in HDF5 filters and self-describing, so no reader learns anything. `geecs_data_utils.io.scan_stack`, h5py, MATLAB and Tiled all read it transparently, and no reader-side change ships here by design.
- **Chunking** — still one frame per chunk, so per-shot random access still costs one chunk (now plus its decompress).
- **Lossless** — the existing end-to-end pixel-equality assertion runs through the new filters.
- **The escape hatch** — a client that wants raw frames puts `Compression=None` before `Capture=1`; the arm-time validator still accepts exactly `None` or `zlib`.
- **Existing stacks** — untouched. New scans pick this up per box as the gateway is upgraded.

## LOC breakdown

| Concern | Lines |
|---|---|
| The default flip + its comment | 6 |
| Tests (both branches) | 10 |
| `CLAUDE.md` write-contract clause | 8 |
| Version + CHANGELOG | 22 |

No facility literals added.

## Tests

`./scripts/check.sh GeecsPvaGateway` — **51 passed**, rc=0. Repo-wide lint (`--lint`) clean. The other packages' suites were not run locally (the change is confined to this package and a fresh worktree would need ~10 envs); CI covers them.

Both new assertions were proved to bite:

- Reverting the default to `"None"` fails `test_stock_adhdf_data_logic_drives_the_plugin` with `assert None == 'gzip'`. This is the production path — that test drives the plugin through the stock `ADHDFDataLogic`, which never puts the PV, so it pins the default a deployed camera actually writes with.
- Deleting the `Compression=None` put fails `test_session_semantics_over_raw_pva` with `assert 'gzip' is None`, pinning the escape hatch.

## Hardware verification

**OWED**, on the first box upgraded:

1. Run a scan with a plugin-backed camera; confirm `<device>.h5` opens and `f['/entry/data/data'].compression == 'gzip'`.
2. Compare file size against a pre-upgrade stack of the same camera — expect roughly 3–4× smaller, and smaller than the LabVIEW PNGs for the same frames.
3. `geecs-pva-gateway diff` for that camera family passes as before (the comparison is on decoded pixels, so compression must be invisible to it).
4. Read `queue_drops` (plus `frames_received` / `frames_written`) off the stack of a scan on the **largest-frame camera** on that box: `_close_session` already stamps the counters dict onto the file as root attributes, so this needs no instrumentation — `h5py.File(stack).attrs["queue_drops"]`. This is the one number the reference measurement cannot answer for the distributed writer; `0` closes it. The writer thread costs ~24 ms per megapixel with these filters, so the bound is roughly 10 Hz of *new* frames at 4 Mpx — well clear of the 1 Hz fleet, and plausibly offset by ~2x fewer bytes over SMB, but unmeasured on hardware.

Rollout is per camera server (gateway upgrade + restart), not a worker restart.

## Note on scope

`Planning/data_capture/02_nonscalar_over_pva.md:179` sequences `Compression=zlib` behind per-family `geecs-pva-gateway diff` parity, but that gate is about *PNG retirement* (#738) — compression is lossless and independent of it. Flagging it since the brief bundles the two.


skbarber:
## Adversarial review — fresh-context reviewer, two rounds

Round 1 read `61bec999`; round 2 confirmed the fixes on `7381eb5e`. Reports in full below, then the disposition.

### Disposition

| # | Finding | Disposition |
|---|---|---|
| 1 | Per-frame writer cost rises 5–25×; CHANGELOG figures ambiguous | **Partly fixed** (`7381eb5e`, `23c3ee4c`) — premise corrected, measurements recorded, no code. Reviewer confirmed no code was the right call. |
| 2 | `CLAUDE.md` overstated the put-time validation | **Fixed** `7381eb5e` — confirmed against the code by the reviewer. |
| 3 | Lands ahead of the parity gate the plan sequences it behind | **Waived** — maintainer's call, already flagged in the PR body. |
| 4 | "scales with frame AREA, not linearly" self-contradictory (round 2) | **Fixed** `23c3ee4c`. |

On finding 1, one premise was wrong and I pushed back rather than accepting it: the reviewer read "2.04 MB vs 7.92 MB" as a single ~4 Mpx frame, but 600×600 uint16 = 0.72 MB and 7.92 / 0.72 = 11.0 exactly, so those are whole-stack figures for an ~11-frame stack. The reviewer re-derived it, tried three ways to refute it (including noting the exactness is load-bearing only in decimal MB, and that the per-frame reading needs a 2048×1934 sensor nobody ships), and withdrew that part. The surviving half — its own 2048×2048 measurement — is now in the CHANGELOG. Finding 4 is the reviewer catching an error in my fix for finding 1.

Round 4 of this ladder was not opened: finding 4 is a prose correction applied exactly as the reviewer specified.

**Cheaper evidence for the owed hardware item**, from round 2: `_close_session` already stamps the whole counters dict onto the stack as root attributes, so the check needs no instrumentation — one scan on the largest served camera, then `h5py.File(stack).attrs["queue_drops"]` (plus `frames_received` / `frames_written`). `queue_drops == 0` closes finding 1. The PR body's verification section is updated to say so.

---

<details>
<summary><b>Round 1 report (on <code>61bec999</code>) — verbatim</b></summary>

## Review — PR #935, `feat/stack-zlib-compression` (GeecsPvaGateway 0.10.2 → 0.11.0)

One commit, 5 files, 44 insertions: the file plugin's `Compression` PV default flips `None` → `zlib` (the write path's existing `{gzip:1, shuffle}` branch), plus two test assertions, CHANGELOG, CLAUDE.md, version bump.

**Verdict: no correctness defect found.** The mechanism, the escape hatch and every consumer I could find check out. Three findings below, all minor; the bulk of the value here is the refutation list — I tried hard to break this and could not.

### What I verified (each was a candidate finding that died)

- **No client puts the PV.** `ophyd_async 0.21.2 ADHDFDataLogic.prepare_unbounded` (`.venv/.../epics/adcore/_data_logic.py:180-205`) sets `chunk_size_auto, num_extra_dims, lazy_open, swmr_mode, xml_file_name, enable_callbacks`, the path set and `capture` — never `compression`. Repo-wide grep for `Compression` outside the plugin: only `GEECS-Data-Utils/tests/test_io_images.py` (an unrelated IMAQ wire byte-string) and one Planning line. `GeecsHdfIO` (`GeecsBluesky/geecs_bluesky/devices/hdf_plugin.py:45`) adds only `Rewind/WriteStatus/WriteMessage`; `detector.py:695` constructs the stock logic unmodified. Nothing in deploy scripts, fleet screens or docs puts it. **And this is pinned**, not just asserted: `tests/test_file_plugin.py::test_stock_adhdf_data_logic_drives_the_plugin` drives the *real* `ADHDFDataLogic` over real pvAccess, so an ophyd-async upgrade that started putting `Compression` would fail the new assertion. Good coverage choice.
- **Tests bite.** Reverting the default to `"None"` in place → `test_stock_adhdf_data_logic_drives_the_plugin` fails on `frames.compression == "gzip"` (50 passed, 1 failed). The second test's `frames.compression is None` also bites (it would read `"gzip"` if the put were ignored). Full suite: 51 passed. pre-commit (ruff/ruff-format/pydocstyle/…) clean on all five files.
- **The gated rewind path is safe.** `plans/gated.py` calls `rewind_to_step_baseline` / `truncate_to_quota`, i.e. `Rewind` → `frames.resize(n, axis=0)` on a *filtered* dataset, then re-appends at the same indices. I ran 20 append-6/trim-2 cycles with h5py 3.16 in the package venv: content byte-identical to intent, and **no file-space leak** — final size 0.06 MB compressed vs 13.11 MB raw for 80 frames (raw landed exactly at the ideal size, so the freed chunks are reused in both modes). This was my strongest defect hypothesis and it is dead.
- **Every reader is transparent.** `geecs_data_utils/io/scan_stack.py` (h5py only; its docstring's "chunked one frame per chunk" and "a single chunk read" stay true), `geecs_pva_gateway/diff.py` (parity tool, via `read_shot`), `ScanAnalysis` renderers, `ImageAnalysis` `ShotRef`, portal/Tiled (`mimetype application/x-hdf5`, stock HDF5 adapter). No `read_direct_chunk`, `memmap`, byte-offset or file-size assumption anywhere in the monorepo. No second producer of `/entry/data/data` outside tests. Attribute datasets stay uncompressed, so the consolidator's `chunk_shape=(16384,)` is untouched.
- **No mid-write reads.** `StackCheckCallback` (`GeecsBluesky/geecs_bluesky/callbacks.py`) waits for the `finalized` root attribute before opening, and reads stamps + shape only. `scan_stack`'s docstring states stacks are read only after the scan closes. So the "partial chunk → inflate error instead of garbage" concern has no live consumer.
- **No doc drift, no duplication, placement correct.** `README.md`, `DEPLOYMENT.md` (file-plugin section, l.270-300) and `docs/geecs_gateway/image_pvs.md` document no PV defaults. No second PV table or fake plugin to keep in sync (GeecsBluesky tests use `mock+pva` signals and assert nothing here). `config.py` states "the GEECS DB is the source of truth — there is no per-host config file", so `PV_TABLE` is the only home a default could have, and the per-camera override that a future feature would want already exists — it's the PV. `ScanAnalysis`'s renderers use their own gzip-4 for analysis outputs; different artifact, not duplication. Version bump + CHANGELOG per root CLAUDE.md; no scan-folder invariant, hygiene or facility-value exposure.

### Findings

**1 — Low/medium (risk to verify, not a demonstrated bug): the per-frame writer cost rises 5–25×, and the CHANGELOG's timings stop short of the frame size its own ratio comes from.**
The writer is one thread per plugin with a 64-deep queue that also carries *command* puts (`file_plugin.py:466, 511, 538`), so its per-frame cost sets both the drop threshold (`queue_drops`) and the tail latency of `Capture=0`/`Rewind`. Measured in the package venv (h5py 3.16, macOS SSD, gaussian beam + noise uint16, `chunks=(1,H,W)`, flush per frame):

| frame | raw | zlib | bytes/frame |
|---|---|---|---|
| 600×600 | 0.4 ms | 8.2 ms | 0.72 → 0.33 MB |
| 1025×1281 | 1.3 ms | 30.9 ms | 2.63 → 1.19 MB |
| 2048×2048 | 4.4 ms | 103.6 ms | 8.39 → 3.77 MB |

The CHANGELOG quotes 3.9 ms (600×600) and ~25 ms (1025×1281) — consistent with mine — but the compression ratio it cites (2.04 vs 7.92 MB) is a ~4 Mpx frame, the size class where I measure ~100 ms/frame. Concrete scenario: a camera that pushes **new** (non-duplicate, non-stale) frames during an open capture session faster than ~10 Hz at 4 Mpx / ~25 Hz at 1.3 Mpx — e.g. a free-running live view during a scan — now backs the queue up where it previously kept up, costing `queue_drops` and up to a few seconds of `Capture=0` latency in the pathological full-queue case. Honest counterweight, which is why this is not ranked higher: the real write target is the NAS over SMB, and compression cuts bytes-on-the-wire ~2.2×, so the net on the production path is plausibly neutral-to-positive — it is simply unmeasured. Worth one hardware datapoint on the largest served camera (writer CPU + `queue_drops` from the counters) rather than more code. Read side, for completeness: per-shot reopen+read 1.2 → 10.0 ms, whole 20-frame stack 13.8 → 201 ms locally; again a wash or a win over SMB.

**2 — Low (doc inaccuracy in a contract file): `GeecsPvaGateway/CLAUDE.md:145-146` says "a client wanting raw frames puts `None` before `Capture=1`, and only those two values are accepted."**
Put-time acceptance is all eight `COMPRESSIONS` choices (`_unwrap_put` validates against `choices`, `file_plugin.py:225`); the `(None, zlib)` restriction is enforced only in `_capture_on` (`:616`), where a third value fails the *arm* with `WriteStatus=Write Error`. An agent or client author reading that sentence would reasonably expect `pvput Compression Blosc` to be rejected, and it is not. The PR's own code comment gets this right ("A client that wants raw frames puts `None` before `Capture=1`"); the CLAUDE.md clause is the loose one. One-clause fix: "…and only those two survive the arm."

**3 — Informational (sequencing, maintainer's call): your own untracked plan sequences this behind the parity gate.**
`Planning/data_capture/02_nonscalar_over_pva.md:181` (untracked in the main checkout, not on the branch): "the same parity gate every camera family passes before `Compression=zlib` / PNG retirement (#738)". This lands the flip before `geecs-pva-gateway diff` evidence exists for every family. I can't refute it into a real problem — compression is provably lossless (verified round-trip) and orthogonal to PVA-vs-PNG parity, and `diff.py` reads through h5py either way — but it does add a variable to a path still under acceptance, and the deliberate deferral is yours to release, not mine.

### Not findings, recorded so they aren't re-derived
`NumFramesChunks > 1` would turn each append into a decompress/recompress of the whole chunk (the 1 MB default chunk cache can't hold one), but nothing sets it: the PV default is 1 and `ADHDFDataLogic` only raises 0→1. Tiny frames (the 4×6 test image) compress to larger than raw — irrelevant at camera sizes. `ADCompression` is a `StrictEnum` whose members match `COMPRESSIONS` exactly, so ophyd-async connect-time validation is unaffected by the new default value.

</details>

<details>
<summary><b>Round 2 report (on <code>7381eb5e</code>) — verbatim</b></summary>

## Confirmation round on `7381eb5e` (read the commit, not the description)

Docs-only: `CHANGELOG.md` +19/-5, `CLAUDE.md` +4/-1. No code, so the 51-test suite and the pinning test I broke last round are untouched; no re-run needed. Both files are clean for the whitespace/EOF hooks.

### Finding 2 — **fixed, correct.**
New clause: "Any of the eight areaDetector choices is *accepted* by the put; only those two survive the arm, and a third fails `Capture=1` with `WriteStatus=Write Error`." Checked against the code, not your description:
- `_unwrap_put` (`file_plugin.py:316-332`) resolves an enum put by **index** into the choices list and returns the string; every one of the eight resolves. Only an out-of-range index raises. So "accepted by the put" is right.
- `COMPRESSIONS` (`:225`) has exactly 8 members — "eight" is right.
- `_capture_on` (`:616-618`) appends the problem, and the failure path (`:627-632`) calls `_error(message)` → posts `WriteStatus="Write Error"` **and** `WriteMessage`, posts `Capture` false, and fails the put with the reason. So "fails `Capture=1` with `WriteStatus=Write Error`" is right, and if anything understates it (the put itself errors too, which is what ophyd-async sees first). No objection.

### Finding 1 premise — **your reading is right; mine was the weaker one. Here is my refutation attempt, which failed.**
Arithmetic re-derived from scratch: 600×600 uint16 = 720,000 B = 0.72 MB decimal; 7.92 / 0.72 = **11.0 exactly**. 2.04 / 11 = 0.185 MB/frame → a 3.88× ratio, which the entry rounds to nothing but is consistent with "3.9×". The PNG figure corroborates: 2.5 / 11 ≈ 227 KB per 600×600 PNG, i.e. 3.2× off raw — an ordinary number for deflate+predictor on a beam image — and under this reading the sentence's punchline is stack-vs-stack (2.04 MB beats the 2.5 MB of PNGs it replaces), which is the comparison that means something.

Attempts to refute it:
- **Is the integer a coincidence?** Partly — it is not uniquely diagnostic. 600×600 **uint8** = 0.36 MB gives 7.92/0.36 = 22, also exact; any 0.72/2ᵏ divides in. So "exactly 11.0" pins *whole-stack*, not *eleven frames*. It is 11 because the plugin's frames are uint16 (`DataType_RBV` default, the u2 test fixture, GEECS IMAQ 12/16-bit). Either way your conclusion stands. One genuine assumption: the exactness only holds in decimal MB — in MiB, 7.92 / 0.6866 = 11.53, no integer. Given both figures presumably came off the same `getsize/1e6`, decimal is the right reading, but that is the load-bearing assumption, not the integer.
- **Could 7.92 MB be one frame after all?** It would need 3.96 Mpx — 2048×1934, which is not a sensor anyone ships (2048×2048 = 8.39 MB, 2448×2048 = 10.0 MB, 1936×1216 = 4.71 MB). And the only resolutions the sentence names are 600×600 and 1025×1281. The per-frame reading requires an invented sensor; the whole-stack reading lands on a resolution already in the sentence. Refuted.
- **Is 3.9× plausible on real frames when I measured 2.2× synthetic?** Yes, and my own data points that way: holding size fixed at 1025×1281 and only raising noise (σ 30 → 300) dropped the ratio 2.21× → 1.79×. Less noise ⇒ better ratio; a real frame with a genuinely dark background (shuffle putting the near-all-zero high bytes together) reaching 3.9× is credible.

So the part of my finding 1 that claimed the entry "cites a ~4 Mpx frame whose timing isn't quoted" was **wrong** — the entry never cited a 4 Mpx frame; it was ambiguous about units, not internally inconsistent, and its two quoted timings do cover both quoted resolutions. The surviving part of finding 1 was always my own independent 2048×2048 measurement, which the new text now carries.

### One new, small defect in the fix text — non-blocking
"Per-frame write cost … **scales with frame AREA, not linearly**" is self-contradicting as written. Cost *is* linear in area: across my three points the rate is 22.8 / 23.5 / 24.7 ms per megapixel (360 k → 1.31 M → 4.19 M pixels), constant to within 8% over a 12× span. A reader who takes "not linearly" at face value will conclude the numbers below disprove the sentence above them. (Your real-frame pair runs the other way — 3.9 ms at 0.36 Mpx = 10.8 ms/Mpx vs 25 ms at 1.31 Mpx = 19.0 ms/Mpx, i.e. *super*-linear in area — which is presumably content, not scaling, but it is certainly not sub-linear either.) I take the intended meaning to be "not with linear dimension". Suggested: "…scales with frame **area** — about 24 ms per megapixel across the three sizes below — so doubling each dimension quadruples it." Everything else in the new paragraph checks out against my measurements: 8/31/104 ms ✓ (8.2/30.9/103.6), 2.2× ✓ (2.18/2.21/2.23), ~10 Hz at 4 Mpx ✓ (104 ms → 9.6 Hz, and correctly stated as an upper bound, since decode is on the same thread), "single writer thread carries command puts as well as frames" ✓ (`:466` queue, `:511` commands, `:538` frames), `queue_drops` / `Capture=0` latency ✓.

### Finding 1 disposition — **sufficient, and cheaper than you may realise.**
No code was the right call. Worth knowing for whoever does the owed hardware item: the evidence needs no instrumentation at all — `_close_session` (`:886-903`) stamps the whole `_Counters` dict onto the stack file as root attributes and logs it. So the check is one scan on the largest served camera, then `h5py.File(stack).attrs["queue_drops"]` (plus `frames_received`/`frames_written`) and the "session closed" log line. If `queue_drops` is 0 there, finding 1 is closed.

### Finding 3 — waived, noted, no argument from me.

Nothing further from me on this PR.

</details>

skbarber (commented):
No findings. The default zlib compression behavior, explicit uncompressed escape hatch, validation, version bump, and documentation are internally consistent and covered by focused tests. I also confirmed the diff passes whitespace checks. I could not complete the targeted test locally because the temporary environment lacked the p4p dependency; the PR's unit-test and pre-commit checks are green.

skbarber:
## Hardware verification — done, 2026-09-18

Closing out the OWED list in the body. Deployed by advancing the lab's shared *Active Version* clone from `feature/native-bluesky-plans` @ `9b5eb908` (a now-deleted branch, and an ancestor of master — the pull was a fast-forward) to master @ `6e37be31`, then restarting the canary box 192.168.6.100 via its `:restart` PV. `version` went `0.10.0` → `0.11.1`. Nothing under `GeecsPvaGateway/deploy/` had changed since that pin, so no per-box launcher visit was needed.

Evidence is Scan001 of 26_0918, `UC_Amp2_IR_input`, 10 shots, 500×500 uint16.

| # | Item | Result |
|---|---|---|
| 1 | stack opens, `compression == 'gzip'` | ✅ `gzip`, opts `1`, `shuffle` True, writer `geecs-pva-gateway 0.11.1`, `finalized` True |
| 2 | size vs a raw stack | ✅ frames 1,303,005 B for 5,000,000 B raw = **3.82×** |
| 3 | `geecs-pva-gateway diff` parity | ✅ `pass (matched=10 identical=10 png_only=0 stack_only=0)` |
| 4 | `queue_drops` under load | ⚠️ **0**, but see below |

**Item 4 is not really answered.** `queue_drops=0` with `frames_received=17`, `frames_written=10`, `duplicates_dropped=3`, `stale_skipped=4` — the 17→10 is correct behaviour (idle re-pushes deduped on `acq_timestamp`, pre-watermark frames dropped), not loss. But 500×500 at 1 Hz is 0.25 Mpx and ~8 ms of compression per frame; it was never going to stress the single writer thread. The meaningful test is the largest-frame camera on whichever box rolls next. Parking it as low-risk rather than claiming it clear.

**Item 3 is the one that matters** and it is stronger than the size numbers: ten frames pixel-identical between the compressed stack and the LabVIEW PNGs, nothing missing in either direction, through the real parity tool on real data.

## Two things learned that the PR body got wrong

**The "beats the PNGs" claim needs splitting.** The *frames* do — 1.30 MB against 1.66 MB of PNGs. But the file as written was **3.03 MB**, nearly 2× the PNGs, because the 13 uncompressed `NDAttributes` datasets took 1.70 MB to store 1,040 bytes of numbers: their chunk is 16384 f8 slots (128 KiB) and HDF5 commits the whole chunk on first write, so the cost is constant per scan regardless of shot count. Crossover against PNGs is ~48 shots. #941 fixes it by passing the same `filters` to the attribute loop (1,703,936 → 9,191 bytes), and only then is "smaller than the PNGs" true at every shot count.

**These frames were blank** — 27 distinct values, mean 56, max 3895. PNG's per-row predictors are a real structural advantage that shuffle+deflate lacks, so the compression ratio on actual beam images is still unmeasured and could move either way. The `diff` result is unaffected by that; the size claims are not.

PR #936, opened 2026-09-18
Magcam1 analyzer

CurtBerg0137:
## What & why

Added a new analyzer class for the UBcaveMagSpec 1 camera device. Slight modifications to existing functions to expose certain things like the returned vertical position from the bowtie analyzer.

- [x] `poetry version patch|minor` run for every package whose code changed
      (ImageAnalysis 2.3.0 → 2.4.0 minor; GEECS-Schemas 0.29.1 → 0.30.0 minor)
- [x] `CHANGELOG.md` entry added under the new version (Keep a Changelog)
- [x] Tests: exact counts reported below (not "tests pass")
- [x] Base branch is `master` — no `users/<name>` branch exists for this
      developer; merge is maintainer-only and is NOT being requested here
- [x] Public-repo hygiene: no lab accounts/hostnames/home paths
- [ ] Adversarial review posted and every finding dispositioned
      — review RUN, result "no surviving findings"; posting as a PR comment
        immediately after this PR is created (there was no PR to comment on
        at review time). Nothing to disposition.

## Tests

`./scripts/check.sh ImageAnalysis GEECS-Schemas` (root Poetry env, Python 3.11):

- **ImageAnalysis: 318 passed, 5 skipped, 10 deselected, 1 failed**
- **GEECS-Schemas: 497 passed, 7 deselected**

New tests — 24 in `ImageAnalysis/tests/analyzers/test_bcave_magspec_cam1_analyzer.py`:
construction/defaults, beam + bow-tie scalar presence and finiteness,
`count_threshold` actually applied, result shape and render keys, the `ax=`
rendering contract, and the waist-position work — `center_at` directly (exact
column, midpoint, both out-of-span edges, NaN `x0`, no valid columns) plus end
to end (`y0` inside the frame, tracks a 12-row `np.roll` of the synthetic beam
one-for-one, NaN on a blank frame). All use the existing synthetic bow-tie
generator — no data files, CI-safe.

Regression on the shared algorithm: `test_hi_res_mag_cam_analyzer.py` **11 passed**
(`HiResMagCamAnalyzer` behaviourally untouched).

Also run: `pre-commit run --all-files` — all hooks pass (ruff, ruff-format,
pydocstyle, check-ast/json/yaml/toml, both lock-path guards).
`GEECS-Schemas/tests/test_analysis_corpus.py` against the companion config
branch: **2 passed**.

**The 1 failure is pre-existing and unrelated to this diff:**
`test_line_stitcher.py::TestLineStitcherMultiDeviceLoading::test_analyze_image_file_keeps_available_segments_when_sibling_missing`
— a Windows path-separator assertion (expects `Scan001/DevB/...` in a value
that is a backslash path on Windows). Verified failing on a clean `master`
checkout via `git stash`; left unfixed as out of scope.

## Hardware verification

**OWED** — code-complete, not hardware-verified. No `Z:` drive on the authoring
machine, so the camera parameters could not be measured. The companion config
ships deliberately pass-through (full frame, zero background) rather than
carrying values copied from another camera.

Acceptance test, next lab session, on a Z:-mounted machine with a scan
containing `UC_BCaveMagSpecCam1`:

1. **ROI + background** — open the scan in the data portal, add a `roi` around
   the dispersed trace and set `background.constant_level` to the measured dark
   level; confirm `bit_depth` (currently 16) against the camera.
   *Expected:* processed frame shows the trace with a near-zero pedestal.
2. **Noise floor** — re-check `count_threshold` (default 10.0) and
   `min_total_counts` (default 2500.0) against real frames.
   *Expected:* thresholding removes background speckle without eating the
   trace wings; the per-column valid mask spans the visible trace.
3. **`emittance_proxy`** — run over a scan where the beam is known to focus.
   *Expected:* the proxy minimizes near the waist rather than drifting
   monotonically, which would indicate the fit is not finding the bow-tie.
4. **`bowtie_y0`** (the value this PR adds) — compare against the visually
   apparent vertical position of the trace at the waist column.
   *Expected:* agreement within a few pixels. Note it is a row index of the
   **processed** frame, so once step 1 adds an ROI it is relative to the crop,
   not the raw sensor.
5. **Failure path** — confirm `bowtie_y0` is NaN (not a number) on a frame with
   no beam, rather than a fabricated row index.

Until tuned, `priority: 50` in the companion config keeps this behind the
established diagnostics.


skbarber (approved):

skbarber (changes requested):
see comments

skbarber:
Thanks for putting this together, Curtis. Reusing the existing analysis and exposing the per-column centroids is a good direction. I’d like to adjust the structure so we get a reusable tool out of this.
The measurement we’re after is beam position across the dispersed image, rather than position at the fitted waist. You don’t need a bowtie fit to get that—the useful calculation already happens in extract_profile.
Could you update this PR along these lines?
- Move the profile extraction into a separate, camera-independent algorithm. Return the column coordinates, vertical centroids, RMS sizes, integrated intensities, and validity information.
- Have BowtieFitAlgorithm use that shared extractor. Keep bowtie-specific selection, such as the contiguous-region validity rule, in the fitting algorithm.
- Replace the new camera-specific analyzer with a generic profile analyzer that uses the extractor directly, without running a bowtie fit. Camera-specific preprocessing and thresholds should come from configuration.
- Keep HiResMagCamAnalyzer’s existing behavior and outputs intact.
- Make the full profiles accessible to downstream callers, including validity information, and show a small example of retrieving them across shots. Pixel coordinates are fine for the generic algorithm; conversion to energy can happen separately.
Please add synthetic tests for a known centroid curve, widths, intensities, and invalid columns, and retain the existing HiResMagCam regression tests.
This builds on the useful part of your implementation while avoiding a second copy of the HiResMagCam analyzer. It should also make the same measurement available to other cameras without adding another class for each device.

PR #937, opened 2026-09-18
Retire the capture daemon: purge every reference, drop its dead toggle

skbarber:
## What

#806 deleted the central PVA capture daemon's Python and left everything around it standing: its systemd unit template, its runbook, two bring-up probes, a `capture` role threaded through the deploy scripts, a fleet-map row, a node in the data-flow map, and prose in five packages crediting it for work the file plugin now does. On the host the unit has been `disabled`/`inactive` since 2026-09-13.

This removes all of it. **Depends on #935** (also bumps GeecsPvaGateway) — merge that first.

## What was actually still running

The unit is dead, but a daily cron at 06:15 on the worker host still invoked the deleted `geecs-capture-diff` entry point and had failed with `Command not found` every morning for six days. That is a host-side cleanup, not part of this diff; commands handed to the maintainer separately. `~/.local/state/geecs-capture/diff-evidence.jsonl` is real dual-write evidence and should be kept.

## LOC breakdown

| Concern | Files | Lines |
|---|---|---|
| Delete `GeecsBluesky/capture/` (unit, runbook, 2 probes) | 5 | −560 |
| Deploy wiring: the `capture` role + its template | 4 | −12 |
| Fleet tooling: map row, role order, unit→role, version label | 4 | −6 |
| Prose corrections (7 references) | 7 | ±14 |
| `native_image_save` removal + v4 migration + tests | 4 | +96 / −40 |
| Data-flow map rewire | 1 | +12 / −30 |
| Versions + CHANGELOGs (5 packages) | 10 | +70 |
| Regenerated schema artifacts | 4 | ±80 |

No facility literals added.

## The judgment call: removing `native_image_save`

Flagging this explicitly because it is the one thing here that is not mechanical, and it is cheap to veto.

The field existed to skip a camera's LabVIEW per-shot files **because the central daemon was capturing them losslessly instead**, with an engine preflight that refused the scan if the daemon was not monitoring every capture camera. #806 deleted the daemon *and that implementation*, so since then the field has had no reader anywhere. Verified: zero consumers in the monorepo, and zero occurrences in the HTU configs corpus (checked on the share, including `experiment_defaults.yaml`, which sets only `schema_version` and `trigger_profile`).

It cannot be made honest by rewording, because there is no behaviour left to describe. Keeping it is the worse trap: someone sets it before a long run, believes their PNGs stopped, and finds out from the disk. PNG retirement (#738, open) owns the replacement, and its preflight asks the *distributed* plugin a different question — "armed on every camera in the save set?" — so nothing here was reusable.

Two different removals, because the two models differ:

- **`ScanRequest` → format v4**, following the v3 `trigger_variant` precedent already in that file: the before-validator drops an unset value (flat v1 or inside `capture`) and refuses a set one with a remedy naming #738; `schema_version` ≤ 3 normalizes to 4. The two removed fields now share one `_REMOVED_FIELDS` mapping instead of bespoke per-field code.
- **`ExperimentDefaults` → removed outright.** That model has no lifting validator at all and the corpus does not set the field, so a stray value fails loudly as an unknown key — the base class's stated contract. Building migration machinery there for a zero-instance risk would be the fallback-shim pattern this repo avoids.

**Native saving itself is untouched.** It is decided per device by the GEECS DB (`save` + `localsavingpath`) and executed by `LvNativeFileDataLogic`; plugin-backed cameras deliberately keep writing PNGs beside the stack as dual-write parity evidence (`namespace.py:561`). Nothing in this PR changes what lands on disk.

## Data-flow map

The map drew the daemon as a node between the PVA gateway and the NAS. Rewired so the gateway writes the stacks directly, which is both what the code does and a better picture — it now shows images bypassing the central server, the property the distributed design exists for. The daemon's panel is deleted and its substance (lossless session contract, dedupe, single writer thread, counter identity at finalize, the PNG-retirement step) folded into the gateway's panel. Verified: the `INFO` JSON still parses, every node id has a panel, no orphan panels.

## Tests

`./scripts/check.sh` across every affected suite, rc=0:

| Suite | Result |
|---|---|
| GEECS-DataPortal | 338 passed |
| GEECS-Schemas | 497 passed |
| GeecsBluesky | 726 passed, 4 skipped |
| GeecsPvaGateway | 51 passed |
| ScanAnalysis | 192 passed |
| root-tests | 83 passed |

Repo-wide lint clean. Two new tests cover the v4 removal in both positions (flat v1 and inside `capture`) and both verdicts, and were proved to bite: dropping `native_image_save` from `_REMOVED_FIELDS` fails the unset test with `Extra inputs are not permitted` and the set test on the `#738` regex — so each asserts the path it names, not a generic pydantic refusal.

`root-tests` covers the two pinning tests that had to lose a line (`test_render_units_sh.py`'s template list, `test_fleet_table.py`'s role ordering).

## Hardware verification

**Not applicable** — nothing here touches scan execution or devices. The only behaviour change is a schema refusal for a field nothing sends. The host-side cleanup (cron, unit file) is listed above and is the maintainer's to run.


skbarber:
## Adversarial review — fresh-context reviewer, three rounds

Round 1 read `ed4d12fc` (9 findings), round 2 confirmed `ea52ca61` (1 surviving miss), round 3 is the changelog sentence in `a8485210`. Reports in full below.

### Disposition

| # | Finding | Disposition |
|---|---|---|
| 1 | `deploy/README.md` still indexed the deleted template dir | **Fixed** `ea52ca61` (also added the missing `GeecsScanner/deploy/`) |
| 2 | Production copy of the deleted `capture/FORMAT.md` citation survived | **Fixed** `ea52ca61` |
| 3 | "fails loudly as an unknown key" is false — both callers swallow | **Fixed differently** `ea52ca61` — see below |
| 4 | Gateway node subtitle overflowed its box | **Fixed** `ea52ca61` — reverted, the edge label already says it |
| 5 | `ScanRequest.capture` description still advertised the removed field | **Fixed** `ea52ca61` — the only user-facing survivor |
| 6 | `db_runtime` claimed a devicetype blip leaves native saving unaffected | **Fixed** `ea52ca61` — it silently reinstates #934 |
| 7 | Three "the daemon" prose credits | **Fixed** `ea52ca61` (+ GEECS-Data-Utils 0.34.2) |
| 8 | GeecsPvaGateway 0.11.1 skips 0.11.0; Bluesky patch too low | **Half** — Bluesky → 0.95.0; 0.11.1 kept, reviewer withdrew the objection |
| 9 | Nothing retires the already-installed unit | **Fixed** `ea52ca61` — operator step in the CHANGELOG |
| 10 | ScanAnalysis changelog credited only the test copy (round 2) | **Fixed** `a8485210` |

**Findings 1, 2, 5, 7 share one cause worth naming:** the first sweep matched identifiers (`geecs-capture`, `capture daemon`, `geecs_bluesky.capture`) and never the bare prose form (`the daemon`) or the path form (`capture/FORMAT`). Four survivors, one of them user-facing. A regex sweep is not a sweep until the prose form has been grepped too.

**Finding 3 — the reviewer's suggested fix would not have worked, and it agreed on re-derivation.** It proposed "the same drop/refuse before-validator, or correct the changelog". A *refuse* validator is swallowed by the same `except Exception` the finding is about (`plans/registry.py:104`), so the operator gets the identical misleading "no default trigger profile" refusal. The validator therefore **drops** the key unconditionally. The reviewer then re-derived it independently, confirmed `ExperimentDefaults` is validated in exactly one place behind two swallowing callers, withdrew its suggestion, and supplied the better argument now in the changelog: dropping reproduces today's runtime behaviour exactly, whereas refusing would be a new failure mode invented by a cleanup PR.

**Finding 6 is the best catch in the report.** Re-derived rather than taken on trust: with `devicetype=""`, `looks_triggerable` skips both `TRIGGERED_DEVICETYPES` and the `TRIGGER_SOURCE_DEVICETYPES` substring test and falls through to the name rule — so a `magspecstitcher` with no trigger variable of its own returns False and `native_save` goes False. That is #934, reinstated three commits after it was fixed, while the log line told the operator nothing was affected.

**Finding 8, kept over the reviewer's initial objection**, which it then withdrew on a stronger argument than mine: both PRs touch the same `pyproject.toml` version line and insert at the top of the same CHANGELOG, so whichever lands first, the other conflicts and a human resolves it. The wrong numbering cannot ship silently.

**Declined, recorded here so it is a decision and not an oversight:** the reviewer asked whether `resolve_experiment_defaults` should log a warning when a defaults file still carries the dead key, so the site is told to delete the line. Reasonable, but it is new runtime behaviour in a cleanup PR and the key is inert — it belongs with #738, which will touch that path anyway. Raised as a question by the reviewer, not a work order.

### Verification

Independent re-run by the reviewer, rc captured unpiped: GEECS-Data-Utils 441, GeecsBluesky 726, ScanAnalysis 192, GEECS-DataPortal 338, GEECS-Schemas 498, root-tests 131 — matching the author's numbers. Both new tests proved to bite, by the author and again by the reviewer. SVG parity re-checked after every edit: 16 nodes, 16 panels, no dangling `data-n`.

---

<details>
<summary><b>Round 1 report (on <code>ed4d12fc</code>) — verbatim</b></summary>

See the review comment thread below; the full text is reproduced in the following two comments to stay within GitHub's comment size limit.

</details>

skbarber:
**Round 1 review report, verbatim** (read `ed4d12fc`).

---

# Review — PR #937 `chore/retire-capture-daemon` (+249 / −802)

Verified: full GEECS-Schemas suite (504 pass, incl. the `schema_export`/`docgen` no-drift guards), root `tests/` (131 pass, 15 skip), `ScanAnalysis/tests/analyzers/test_stack_shot_mapping.py` (9), `GeecsBluesky/tests/test_assets.py` + `test_db_runtime.py` (18), `ruff check`/`format` clean on the touched Python, headless render of the data-flow page before *and* after, and a script-verified node↔edge↔panel parity check on the SVG.

## Findings, ranked

### 1. `deploy/README.md:29` still lists `GeecsBluesky/capture/deploy/` as a live unit-template home
The PR removes that directory and scrubs it from `deploy/render_units.sh`, `deploy/bootstrap_host.sh` (`SERVICES`, `clone_of`, `pkgdir_of`, `extras_of`, `units_of`, `templates_of`) and `tests/test_render_units_sh.py` — but not from the README whose whole job is to index those lists. The path it names no longer exists after this merge. While in there: the same sentence also omits `GeecsScanner/deploy/`, which *is* in `render_units.sh`'s `TEMPLATES` (pre-existing; the two lists have no cross-check test — `tests/test_render_units_sh.py:29` hand-duplicates the script's array).

### 2. `single_device_scan_analyzer.py:533` still cites the deleted `capture/FORMAT.md`
`ScanAnalysis/scan_analysis/analyzers/common/single_device_scan_analyzer.py:533` — `_map_shots_from_stack`'s docstring names ``GeecsBluesky/geecs_bluesky/capture/FORMAT.md`` as the stack contract. That file went with #806. The PR's ScanAnalysis CHANGELOG entry explicitly claims this citation was fixed — but only the *test* copy (`tests/analyzers/test_stack_shot_mapping.py:1`) was. The production copy sits ~180 lines below line 355, which the PR *did* edit in the same file. One grep for `capture/FORMAT` finds both. (Third surviving copy, historical and correctly left: `GEECS-Data-Utils/CHANGELOG.md:561`.)

### 3. The `ExperimentDefaults` removal's stated justification is false on the deployed path
CHANGELOG: *"a stray value fails loudly as an unknown key — the base class's stated contract."* It does raise `extra_forbidden` — but **both** production callers of `resolve_experiment_defaults()` swallow it:
- `GeecsBluesky/geecs_bluesky/plans/registry.py:102-106` → `except Exception: logger.warning("experiment defaults not loaded: %s", exc); defaults = None`
- `GeecsBluesky/geecs_bluesky/qs_client/submit_preflight.py:209-215` → same shape

Concrete scenario: a site whose `experiment_defaults.yaml` carries `native_image_save: false` — exactly where the removed field's own description told you to put the PNG-deprecation step ("flipping this off is the PNG-deprecation step") — upgrades geecs-schemas. Nothing fails at load. `defaults` becomes `None`, the experiment silently loses its default trigger profile, and every scan that doesn't name one is refused at `ShotControlRegistry.resolve` with *"no trigger_profile given and the experiment has no default (experiment_defaults.yaml)"* — a message pointing at a file whose `trigger_profile:` line is perfectly fine. The real cause is one `logger.warning` in the worker journal. The ScanRequest path, by contrast, raises a remedy naming #738 straight at the operator.

Mitigation confirmed, so this is not a live break: the corpus's only defaults file, `GEECS-Plugins-Configs/scanner_configs/experiments/Undulator/experiment_defaults.yaml` @ `cbcf6be` (today), is `schema_version: 1` + `trigger_profile: HTU-NoGas` — nothing else in that repo mentions the field. But the asymmetry is *not* defensible on the reasoning given; the cheap fix is either the same drop/refuse before-validator on `ExperimentDefaults`, or correcting that changelog sentence so it doesn't become precedent.

### 4. The data-flow SVG's PVA gateway node text now overflows its box
`docs/sites/data_flow/index.html:252` — appending `· writes the HDF5 stacks` makes the third subtitle line render ~60 units past the rect's right border (rect is `x=290 w=210`, so it ends at 500; the text runs to ~562), into the whitespace the deleted capture node left. Confirmed by rendering both revisions headless — master's shorter line fits, the PR's does not. Suggest splitting to a fourth line or shortening to `· writes the HDF5 stacks` on its own line.

Everything else about the map is sound: no dangling `data-n` tokens (every edge token resolves to a node `data-id` and every node appears in an edge), every `data-id` has an INFO panel, the re-pointed `pva-gateway nas` edge lands inside the NAS rect (`y=470`) with its step marker and label repositioned consistently with the `cameras nas` pattern, and the "developing" legend dot is still used by other nodes.

### 5. `scan_request.py:542` still advertises the removed field to users
The `ScanRequest.capture` **field** description was missed: *"...save sets, telemetry and native-image toggles, and the trigger profile."* The `CaptureSettings` class docstring was rewritten ("five settings", telemetry only) but its parent field wasn't. This string ships in the published artifacts — `docs/geecs_schemas/scan_request.schema.json:209` and `docs/geecs_schemas/schema_reference.md:164` — i.e. exactly what external clients (OSPREY's form renderer) fetch by URL. A reader following it writes `native_image_save` and gets the v4 hard refusal. This is the *only* surviving user-facing reference, and it matters more than usual because nothing in the monorepo consumes `ScanRequest` any more (see notes) — the JSON artifact is the whole audience.

### 6. `GeecsDbDeviceTypes` still documents itself in the deleted toggle's terms, and the claim is now wrong
`GeecsBluesky/geecs_bluesky/db_runtime.py:136-181`. Docstring: *"consumers treating 'unknown devicetype' as 'not capture-eligible' therefore fail open: native image saving is never switched off because the DB blipped."* Its warning log: *"no device is capture-eligible this run (native saving unaffected)."*

Keeping the class is right — it is live, not the daemon's. But its only consumer is `namespace.py:279` → `roster.types` → `looks_triggerable(rows, devicetype)` at `namespace.py:511`, which gates `native_save = triggered and NATIVE_SAVE_VARIABLES <= settable_names` at :522. So on a devicetype-read failure native saving *is* affected, in both directions: `TRIGGER_SOURCE_DEVICETYPES` no longer excludes the DG645s (they register as acquirers) and `TRIGGERED_DEVICETYPES` no longer includes `magspecstitcher` — the blip silently reinstates the #934 defect (no `U_BCaveMagSpec/` folder, Scan004 of 26_0917, fixed three commits ago) while the log line tells the operator nothing is affected. This is the last place the daemon's vocabulary is load-bearing, so the sweep that deleted `native_image_save` should own it.

### 7. Prose survivors crediting the deleted daemon
- `GEECS-Data-Utils/geecs_data_utils/io/scan_stack.py:292` — the keep-first dedupe contract is justified by *"the daemon dedupes identical timestamps upstream"*. The guarantor is now `HdfFilePlugin` (`GeecsPvaGateway/geecs_pva_gateway/file_plugin.py:750`, `duplicates_dropped`). GEECS-Data-Utils is untouched by this PR, yet the PR's **own new** INFO link for the PVA gateway ("[Stack format] → scan_stack.py") sends readers straight there. Would need a GEECS-Data-Utils patch bump.
- `GEECS-DataPortal/tests/test_cache.py:193` — "the stop doc lands BEFORE the daemon finalizes the stack".
- `ScanAnalysis/tests/analyzers/test_stack_shot_mapping.py:94` — "The daemon can capture pre-save-window frames" — in the very file whose module docstring the PR did fix.

### 8. GeecsPvaGateway version: 0.10.2 → **0.11.1** for a one-line docstring change
Two issues: it skips 0.11.0 entirely (the CHANGELOG now reads `0.11.1` directly above `0.10.2`), and a docs-only change is a **patch** under the root CLAUDE.md table → `0.10.3`. `/fleet-status` compares deployed package versions, so this mints a phantom 0.11.0 nobody can ever have run — and if the in-flight non-scalar-PVA work claims 0.11.0, they collide. Lower-confidence companion: GeecsBluesky 0.94.3 is a patch for a `### Removed` section that deletes a deployable systemd unit template; minor is the honest digit (the Python is unchanged, so this is arguable).

### 9. Nothing retires the unit already installed on the host
The scripts stop *rendering* `geecs-capture.service`, but `/etc/systemd/system/geecs-capture.service` stays put, and `scripts/fleet_status.sh` discovers by `list-units --all "geecs-*"` — so the stale unit still emits a row, now labelled with its raw unit name (`role_for_unit` falls through to `*)`) and appended after the known roles by `fleet_table.py` with no version (harmless: unknown roles append, no crash). No line in the diff tells the operator to `systemctl disable --now geecs-capture && rm /etc/systemd/system/geecs-capture.service && systemctl daemon-reload`. Worth one sentence in the GeecsBluesky CHANGELOG `Removed` entry.

## Checked and refuted — these are clean

- **`_REMOVED_FIELDS` covers every position the bespoke code did.** Exercised against the branch: flat/nested × set/unset × mixed-with-live-flat-fields × both-fields-set. Refusals and drops match v3 semantics exactly; precedence is removed-field-before-mixed-layout (so a flat unset removed field beside `capture` still gets the drop, not the "mixed layout" verdict); a document setting both gets the `trigger_variant` remedy by dict order, which is the same verdict master gives. `_V1_CAPTURE_FIELDS` dropping the field is correct and matches the v3 precedent. `stale_schema_version(data, 4)` + `lifted["schema_version"] = 4` are consistent, and a v5 stamp survives.
- **Regenerated artifacts match the generators.** The no-drift guards pass; the only non-`native_image_save` deltas in the JSON/MD are the two intentional v4 stamps and the `data_format` description rewrite. No unrelated drift.
- **No over-deletion of live file-plugin vocabulary.** "capture stack", "capture session", "capture-eligible", `diff.py`'s dual-write audit, the 5568 doc proxy (still consumed by `qs_client`, so the `site.env.example` reword to "the doc-stream consumers' addr" is right), and the fleet map's `FilePath/Capture/NumCaptured` PVs are all still-live areaDetector/file-plugin terms and are untouched. `HdfFilePlugin` is real and does dedupe, so every re-attribution in the diff is factually correct.
- **Completeness elsewhere.** `mkdocs.yml`, all of `docs/**/*.md`, `docs/platform/fleet_map.md`, root + package `CLAUDE.md`s, `GeecsBluesky/pyproject.toml` extras (no `capture` extra survives — `bootstrap_host.sh` was passing a nonexistent extra before this PR), and the configs corpus carry no surviving daemon reference. CHANGELOG mentions are historical record, correctly left.
- **Redundancy / placement.** `_REMOVED_FIELDS` is net de-duplication. It correctly stays in `scan_request.py` rather than `_base.py`: `TriggerProfile`'s v2 `variants` removal (`trigger_profile.py:281-305`) is emptiness-based, not `None`-based, so there is no shared abstraction to hoist.
- **Context worth knowing:** nothing in the monorepo imports `ScanRequest` any more outside `geecs_schemas` itself and its tests (`schema_export`, `__init__`, cross-references). The funnel plan it parameterized went with #807; `schema_export.py`'s docstrings still describe it as "the queueserver funnel's one `request` parameter" (pre-existing drift, not this PR). So the v4 migration's only audience is external clients fetching the published JSON — which is why finding **5** is the one that actually reaches a user.

skbarber:
**Round 2 confirmation report, verbatim** (read `ea52ca61`).

---

# Confirmation review — `ea52ca61` (fix commit on #937)

Read the commit, not the summary. Re-verified every accepted fix independently, re-ran the suites, re-rendered the page. **One real miss, one thing to write down, and I withdraw my finding-8 objection.**

## Finding 3 — checked hardest, as asked. Your fix is right and mine was wrong.

Re-derived from scratch: `ExperimentDefaults` is validated in **exactly one place in the entire repo** — `config_resolver.resolve_experiment_defaults()` (`config_resolver.py:594`). `SCHEMA_REGISTRY` has no runtime consumer at all (only GEECS-Schemas' own tests and `docgen`), and no config editor, portal page or scanner form touches the model. Both callers swallow: `plans/registry.py:102-106` and `qs_client/submit_preflight.py:209-215`. So a `ValidationError` genuinely cannot reach a human anywhere today. My "same drop/refuse before-validator" suggestion would have bought a different journal line and the identical operator-facing breakage. Drop is correct.

**Direct answer to your question: no, the silent drop of `false` is not worse.** And the argument in your changelog undersells it. The decisive point is that **the drop reproduces master's runtime behaviour exactly** — master today parses `native_image_save: false` into a field nothing has read since #806. Dropping it is a no-op against shipped behaviour; refusing would be a brand-new failure mode invented by a cleanup PR, and a mis-attributed one at that. A cleanup must not be able to break a scan. An operator who wrote `false` has been wrong since #806 and neither verdict corrects them — but the drop costs them nothing they still have, whereas the refusal costs them the `trigger_profile` line sitting three lines above. That sentence is the one that generalises; worth swapping in.

Two things to add while you're there, neither a blocker:

- **Write the asymmetry down as a rule.** ScanRequest refuses, ExperimentDefaults drops — coherent, because each verdict matches *where the document is validated*: nothing in the monorepo validates `ScanRequest` at all (it is a published JSON artifact whose consumers are external clients that see the `ValidationError` directly), while `ExperimentDefaults` is validated once, behind two `except Exception`s. Right now only the `ExperimentDefaults` docstring explains its own half; a future reader could "unify" these in the wrong direction. One clause in the GEECS-Schemas changelog pins it.
- **Nothing will ever tell the site to delete the stale line**, so it can sit in `experiment_defaults.yaml` misleading a human reader forever. GEECS-Schemas is the wrong place to log — it is a pure vocabulary — but `resolve_experiment_defaults()` is in GeecsBluesky, already has a `logger`, and holds the file path: one `if "native_image_save" in document: logger.warning(...)` before `model_validate`. A question, not a work order.

**Proved the test bites myself**: neutered `_drop_removed_fields` to `return data`, `test_removed_native_image_save_is_ignored_not_refused` failed with `extra_forbidden`, restored the file (`git diff` clean).

## Surviving problem

**`ScanAnalysis/CHANGELOG.md` was not corrected.** The commit message says "only the test copy was fixed, and the changelog wrongly claimed otherwise" — but the file is not in the commit. The 1.23.2 entry still reads *"The stack-mapping **test's** docstring also pointed at `GeecsBluesky/geecs_bluesky/capture/FORMAT.md` ... it now names the read side"*. That is now an understatement: `ea52ca61` also fixed the production docstring at `single_device_scan_analyzer.py:533`. Same unreleased entry, so no re-bump — just the sentence. It is the one claim in the fix commit's message that the diff does not back.

## Finding 8 — argument accepted, keep the hunk

Verified #935 (`feat/stack-zlib-compression`, open, base master) takes GeecsPvaGateway 0.10.2 → 0.11.0, so under the stated order 0.11.1 is exactly the patch and no phantom exists. Stronger than "I will rebase if": **whichever PR lands first, the other conflicts** — both touch `GeecsPvaGateway/pyproject.toml`'s single version line and both insert at the top of the same CHANGELOG. The wrong numbering cannot ship silently; a human resolves it either way. Do not drop the hunk; the constraint is self-enforcing. (Numbering this one 0.10.3 and letting #935 land on top as 0.11.0 would be order-independent, but that is a preference, not a defect.)

## Findings 1, 2, 4, 5, 6, 7, 9 — all correct

- **6** — your re-derivation is exact. I re-read `looks_triggerable` (`namespace.py:174-182`) to confirm the mechanism: with `kind=""`, `"" in TRIGGERED_DEVICETYPES` is False *and* `any("dg645" in "")` is False, so **both** lists stop applying before the name-rule fallback. And `roster.types` has precisely one read site (`namespace.py:511`), so "the one consumer is `looks_triggerable`" is literally accurate, not approximately. New docstring and warning both say the right thing.
- **4** — re-rendered headless: overflow gone, nothing else moved. Reverting rather than re-laying-out was the right call; the violet edge label plus the INFO panel's `does` text carry the fact. Parity re-run: 16 nodes, 16 panels, no dangling `data-n`, no node absent from an edge.
- **1** — `deploy/README.md`'s list now matches `render_units.sh`'s seven templates exactly.
- **Sweep re-run clean**: every remaining `capture daemon` / `the daemon` / `capture-eligible` hit outside CHANGELOGs is a deliberate historical reference naming the deleted thing to explain why. Nothing dangling, nothing over-deleted.
- **9** — the daily cron invoking the deleted `geecs-capture-diff`, failing for six days, is a leftover no repo sweep could have found. Good that the host was probed.

## Independent test run, rc captured unpiped

`./scripts/check.sh GEECS-Data-Utils GeecsBluesky ScanAnalysis GEECS-DataPortal` → **rc=0**: 441+4s / 726+4s+1d / 192+2d / 338 — matching your numbers exactly. GEECS-Schemas via check.sh: 498 passed, 7 deselected (my earlier 505 was the unfiltered `pytest GEECS-Schemas/tests`; no discrepancy). Root `tests/`: 131 passed, 15 skipped. `ruff check` and `ruff format --check` clean on all four touched modules; no line over 88.

**Verdict: ship after the ScanAnalysis CHANGELOG sentence.** Everything else is optional polish, and per the stopping rule I have nothing at or above the severity floor left.

skbarber (commented):
Found one minor cleanup issue:

**[P3] Remove `capture` from the documented `--only` choices**

`deploy/bootstrap_host.sh` still advertises `capture` as a valid `--only` service, even though this PR removes it from `SERVICES`, `clone_of`, `pkgdir_of`, `extras_of`, `units_of`, and the unit template list. As a result, `--only capture` now quietly performs no capture setup, which can mislead an operator following the bootstrap usage guidance. Please remove `capture` from the comment on line 11.

skbarber:
### Maintainer finding — `deploy/bootstrap_host.sh:11`, fixed in `905e4f4d`

Confirmed, and it was drifted in **two** directions, not one: the `--only` usage header advertised the retired `capture` role *and* had never listed `scanner`, which is live. Same shape as the `deploy/README.md` finding earlier in this review — a prose list of services that no test compared against the service table.

Third instance in this PR of one root cause: my sweep matched the identifier forms (`geecs-capture`, `capture daemon`, `geecs_bluesky.capture`) and never the **bare word** in a comma-separated prose list. The reviewer caught the README copy; this one survived both.

**Two fixes, because the reported line hid a second problem.** The quiet no-op the finding names is not specific to `capture` — *any* unknown `--only` name selected nothing and exited 0, so a stale runbook looked like it had worked. Unknown names now exit 2, naming the service and the known set:

```
bootstrap_host.sh: --only: unknown service 'capture'
  known services: gateway portal logbook qserver mcp scanner
```

Flagging the guard as the judgment call — the finding as written only required the header edit. It is three lines, it is what makes the retirement actually enforce itself, and without it the next retired role repeats this exactly. Cheap to veto if you would rather keep the diff to the header.

**Pinned, both proved to bite:**

- `test_usage_only_list_matches_the_service_table` parses the header and `SERVICES=` out of the script and compares them — restoring the old header fails it with `assert [... 'capture', 'mcp'] == [... 'mcp', 'scanner']`.
- `test_unknown_only_service_is_refused_not_silently_skipped` runs the real script with `--only capture` — removing the guard fails it with `assert 0 == 2`, which is the finding's "quiet no-op" as an assertion.

`./scripts/check.sh` rc=0 across all seven suites; root-tests 83 → **85**. No version bump: the change is confined to `deploy/` and `tests/`, which carry no package version, consistent with the other root-level edits in this PR.

PR #938, opened 2026-09-18
Web surfaces: reachable, not just present (config editor, scanner console, logbook)

skbarber:
Seven pieces of UI polish reported from the floor. Bundled into one PR at the
maintainer's request: they are one concern — the surfaces work, but they do
not signpost themselves — even though the files live in three packages.

## Where this came from

A user reported that the analysis config editor "cannot specify the type of
analyzer when creating a new config — instead we get `ref` as a custom string
to fill out". That turned out not to be a bug in the create path: `ref` exists
only on `AnalyzerRef` (`geecs_schemas/analysis/group.py:22`), so they had
created a **group**, not a diagnostic. The diagnostic path works and even
pre-seeds `analyzer = {kind: "beam"}`. They landed in the wrong form because
both sidebar sections ended in an identical, unlabelled `namespace / new id /
[new]` row, and nothing on the resulting page said which kind was open.

The rest of the list came from the same pass over the operator surfaces.

## Per concern

| Concern | Package | LOC (±) |
|---|---|---|
| Collapsible sidebar tree, labelled `new` rows, kind in the title bar | ScanAnalysis (config editor) | 94 |
| Multi-select add-device drawer; filterable move-variable picker | GeecsScanner | ~70 |
| New scan: invariant fields above, plan-swapped editors below | GeecsScanner | ~65 |
| Devices as a folded chip list | GeecsLogbook | 24 |
| Tests | all three | 111 |
| Versions + CHANGELOGs | all three | 42 |

### ScanAnalysis 1.24.0 — config editor
- Sidebar is a collapsible tree: kind → namespace → document, everything
  closed until asked for. Collapse state persists per browser
  (`localStorage`, every access in try/catch); an auto-expansion is never
  written back, so opening a document does not permanently pin its section
  open. Whatever holds the open document expands on render.
- A namespace that hides a file which does not validate carries the count and
  the `bad` styling on its own summary — collapsing must not hide a broken
  config.
- The `new` rows are labelled `new diagnostic` / `new group` and lead their
  section instead of trailing a 50-item list; the form's title bar carries a
  `DIAGNOSTIC` / `GROUP` chip.

### GeecsScanner 0.12.0 — console
- **Add-device drawer is a multi-select.** A click toggles a row
  (`aria-pressed`, the kit's own selection vocabulary — no new CSS),
  shift-click takes the range and skips devices already in the table,
  `Select all shown` takes the filtered list, `Add N devices` commits them in
  one write. Previously each pick added one device and closed the drawer.
- **Move variable picker filters as you type** — an `<input>` over a
  `<datalist>`, the same affordance the sweep composer's axis field already
  had. `Move` is now gated on a name the experiment actually has, so a typed
  fragment cannot be submitted.
- **New scan is split invariant / plan-dependent.** Shots per step, trigger
  profile, shot period, description and the device table sit at the top under
  `Every scan`; everything the mode buttons swap sits below a hairline under
  an eyebrow that names the current mode (`Count · options`, `Sweep ·
  trajectory`, `Optimize · optimizer`).

### GeecsLogbook 0.12.0 — day page
- A scan's saved devices become a full-width list of chips, folded behind a
  `<details>` with the count on the summary once there are more than three.
  Thirteen devices joined into one string inside a 128 px grid cell rendered
  as ~28 lines of names broken mid-token (`UC_BCaveMagSpecCa` / `m1,`); three
  or fewer still read inline, which is how they already looked good.

1. **A rule, not a border.** The request was "maybe a border around the
   capture description and devices". `kit.css` reserves border + fill + radius
   for `.panel` and says explicitly that a second frame inside one competes
   with it, so the split is a hairline plus two eyebrows instead. Easy to make
   heavier if it does not read strongly enough on the floor.
2. **Short device lists stay expanded** (≤ 3). Folding a single device would
   cost a click for nothing.
3. **Sidebar `new` rows lead their section** rather than trailing it, so an
   expanded 34-item namespace cannot bury them again.
4. **`diagnostics`, not `analyzers`, as the sidebar heading** — the rest of
   the UI already says diagnostic. The API kind stays `analyzer`; no route or
   payload changed.

## Reversal of a previous decision

`test_form_starts_at_the_mode_segment` pinned #896's "the body opens on its
mode-specific controls". The reorder reverses that half deliberately; the test
is rewritten as `test_form_opens_on_what_every_scan_needs`, pinning the new
order *and* still pinning #896's other half (the preset picker lives in the
footer, nothing fillable above the fields). Verified it bites: with the two
blocks swapped back it fails, with them in the intended order it passes.

## Tests

`./scripts/check.sh` (the suites CI maps to this diff), all green:

- GeecsWebTheme — 182 passed (this is the suite that walks the other packages'
  CSS/templates for literal colours and unstyled classes)
- GeecsLogbook — 277 passed, 1 skipped
- GeecsScanner — 143 passed, 1 skipped
- ScanAnalysis — 192 passed, 2 deselected

New pinning tests: the selection logic runs under `node` through the repo's
existing `_script_function` harness (toggle, range, skip-already-added), plus
markup-level pins for the filterable picker, the panel order, and the logbook's
folded / inline device list.

Not run locally: `--all`. This worktree has envs for the four packages above
only; CI covers the rest.

## Verification

Every change was rendered and looked at, not just tested — headless Chrome over
CDP against the three surfaces (the scanner in `--demo`, the logbook over a
synthetic share with a 13-device scan, the editor over a throwaway copy of the
configs tree). Confirmed: the tree collapses and remembers, `new diagnostic`
produces a diagnostic form with the `kind` dropdown on `beam`, a shift-click
range selects six devices and the button reads `Add 6 devices`, the new-scan
panel opens on `Every scan`, and thirteen devices fold to one line and expand
to three rows of intact names.

**Hardware verification: not applicable.** Nothing here touches scan
execution, device I/O, or the queue — the diff is templates, page JavaScript
and CSS. One behaviour change is worth a live look when convenient:

- **OWED (live console, no hardware needed):** confirm the move panel's
  variable picker lists the real settables and that `Move` still enables for a
  picked variable. The gate changed from "the field has a value" to "the value
  names a settable the experiment has".


skbarber:
## Adversarial review — round 1, and dispositions

Fresh-context reviewer against `6eb9b66e`. Eleven findings; **eight fixed** in `65617dd0`, **three waived**. Report in full below, dispositions first.

| # | Finding | Disposition |
|---|---|---|
| 1 | Auto-expanded node cannot be collapsed durably — Save re-expands it | **Fixed** — tri-state store (`wantOpen`), `65617dd0` |
| 2 | The sidebar tree was the only untested part | **Fixed** — four node-run tests, `65617dd0` |
| 3 | Kit rung / two `<details>`-persistence idioms | **Waived** — see below |
| 4 | `Collapse all` cannot reach `details.devlist` | **Fixed** — selector + test, `65617dd0` |
| 5 | Fourth chip treatment; duplicated marker rules | **Waived** — see below |
| 6 | Dangling `·` when a settable has an alias but no units | **Fixed** — `65617dd0` |
| 7 | Typed variable matched byte-for-byte only | **Fixed** — `settableMatch`, `65617dd0` |
| 8 | "This plan" eyebrow over nothing at first paint | **Fixed** — empty state + notice moved, `65617dd0` |
| 9 | A vacuous assertion in the rewritten layout test | **Fixed** — now pins #896's real other half, `65617dd0` |
| 10 | Docs describe the control that changed | **Fixed** — `65617dd0` |
| 11 | Esc discards pending picks | **Waived (code), fixed (comment)** |

Finding 1 was a real defect in the headline feature: a set of open keys cannot express "I closed this", so the section holding the open document sprang back open on every `renderSide()` — and Save renders. The fix makes the store tri-state (open / closed / unsaid) behind a `wantOpen()` seam; the test for it fails against the old OR semantics and passes now.

### Waivers (cheap to veto)

**3 — kit rung / two idioms.** The config editor is not a kit surface: its template loads `theme.css` only, no `kit.css`, no `<body class="kit">`, as the reviewer notes. The rule cited ("tracking open state in JS means you are on the wrong rung") is about using `<details>` as an overlay; this is a navigation tree whose collapse state is a per-viewer convenience. Rung 1's "the selection belongs in the URL" already holds here — `open()` writes `#/analyzers/<id>` — and which branches are folded is not the selection. The `click` + `setTimeout` idiom stays rather than the logbook's `toggle` listener because `toggle` cannot distinguish a programmatic expansion from a user's without a guard flag whose correctness depends on whether the browser queues a toggle task for a detached element; with a tri-state store, mistaking one for the other now writes a wrong value rather than a redundant one.

**5 — chip and marker duplication.** Cosmetic, and the two components differ in role: `.tagchip` is accent-coloured and interactive on the month page, these are neutral non-interactive labels inside a fact cell. Adopting `details.disc` for the marker would bring its border and filled background with it, which is wrong inside a fact cell. Aligning the editor's marker idiom means making the editor a kit surface — a larger change than this PR.

**11 — Esc discards pending picks.** The reviewer half-refuted this themselves: Esc meaning cancel is standard, and if Esc cancels then clearing on reopen is the consistent choice. Only the comment recording the decision landed.

### Still untested, deliberately

The kind chip in the form title bar and the `N · M bad` namespace summary are markup with no logic; the four new tests cover the state machine behind the tree.

---

## The report, in full

Scope: 15 files, +410/−96, three packages. Purely browser-side (Jinja templates, page JS, CSS) + three CHANGELOGs and three version bumps. No Python runtime surface, no scan-folder creation path, no lab literals, no colour literals (all six touched files are in `GeecsWebTheme/tests/test_no_literal_colours.py::_SURFACES`, and every new rule uses tokens). `node --check` passes on both changed JS files. Version bumps and CHANGELOG entries are present and correctly sized (all three are minor). Dependency graph untouched.

Eleven surviving findings, ranked.

### 1. Config editor: an auto-expanded tree node cannot be collapsed durably — **Save re-expands it** (correctness, moderate)

`ScanAnalysis/scan_analysis/config_editor/static/editor.js:409` and `:425`

The open decision ORs the remembered set with "holds the open document", while `trackOpen` (`:386`) can only ever *remove* a key from the remembered set (`rememberOpen(key, false)` → `s.delete(key)`). So a node that is open **because** it holds the current document can never be recorded as closed.

Concrete sequence, entirely inside the normal editing loop:

1. Open `htu/beam1`. `open()` (`:455`) calls `renderSide()`; `holdsOpen` is true, so the `diagnostics` section and the `htu` namespace auto-expand.
2. Collapse either one. `rememberOpen` deletes a key that was never there — a no-op.
3. Edit a field, press **Save**. `save()` awaits `loadListing()` (`:551`), which calls `renderSide()` (`:367`). `state.kind` is unchanged, so `holdsOpen` is still true and both nodes **spring back open**.

Same on `duplicate()` (`:616`), on `remove()` (`:565`), on opening a second document of the same kind, and on the public `reloadListing` — which the portal calls at `GEECS-DataPortal/geecs_portal/templates/run.html:905`. The CHANGELOG promises "remembering what was opened (per browser)"; the commonest operator loop silently undoes the user's collapse.

Fix: make the store tri-state (explicitly-open / explicitly-closed / unset) and let `holdsOpen` apply only when unset.

Refutation attempted: could `holdsOpen` be meant to always win? No — then `trackOpen`'s write of `false` would be dead code, and the sidebar would be un-collapsible for the section you are working in, which is the section you are most likely to want out of the way.

### 2. The largest new behaviour in the PR is the only untested part (test coverage, moderate)

GeecsScanner got a node harness for `pickDevice` (three cases, and they do bite — the harness re-runs the real extracted function body). GeecsLogbook got three router tests. The config-editor sidebar tree — collapse defaults, the remembered set, the `N · M bad` namespace summary, the new kind chip — has **no test at all**. `ScanAnalysis/tests/test_config_editor_app.py:51-59` only asserts `editor.js` is served with `cache-control: no-cache`.

The technique is already in this repo: `GeecsScanner/tests/test_page.py:136` `_script_function(source, name)` extracts one top-level function of the page IIFE and runs it under node. `openSet` / `rememberOpen` / `trackOpen` are exactly that shape, and the "bad" summary string is a pure expression. Finding #1 is precisely the class of bug such a test catches.

### 3. Tracking `<details>` open state in JS is the anti-pattern the overlay ladder names by name, and it is the repo's *second* implementation (placement, moderate)

`GeecsWebTheme/CLAUDE.md` overlay-ladder table, rung 0:
> `details.disc` — Native `<details>` — keyboard, no-JS and find-in-page all work. **Tracking open state in JS means you are on the wrong rung.**

Repeated in `GeecsWebTheme/geecs_web_theme/static/kit.css:414-417`. The new `openSet` / `rememberOpen` / `trackOpen` block is 30 lines of exactly that. The prescribed rung for "the user steps through many items" is 1 (`.inspector`, "the selection belongs in the URL") — and the editor already half-implements it: `open()` writes `location.hash = #/analyzers/<id>` (`editor.js:456`) and boot reads it back (`:599`).

Second, the repo already has a `<details>`-open-persistence implementation: `GeecsLogbook/geecs_logbook/static/nav.js:153-162`, which uses the **`toggle` event**. The new code instead listens for `click` on the summary plus `setTimeout(…, 0)`. The stated reason (`editor.js:390-393`) is sound against a naive `toggle` listener, since setting `open` even on a detached element queues a toggle task; but a one-line guard flag gets the same result with the standard event, no polling delay, and no divergence when the `open` state changes without a summary click (Chrome's find-in-page auto-expansion of `<details>`, for one). Either way there are now two idioms for one job.

Caveat I could not refute away in the other direction: `ScanAnalysis/scan_analysis/config_editor/templates/editor.html:13` loads `theme.css` only — no `kit.css`, no `<body class="kit">` — so the kit rule does not bind the editor mechanically. That is why I file this as placement/doctrine rather than a rule violation.

### 4. Logbook: the new `<details>` is a collapsible that "Collapse all" cannot reach, contradicting a comment this PR left standing (correctness of a contract, low-moderate)

`GeecsLogbook/geecs_logbook/templates/day.html:279-283`:
> Collapse All means **every collapsible thing on the page**, which is what it says — an entry that could not be folded while the scan around it could was an accident of history, not a decision.

The new `details.devlist` (day.html:43, nested inside `details.scan`) is not in that selector. Concrete: on a day where a scan saved 12 devices, press **Expand all** — the scan opens, the device list stays folded. The button's own text and the comment both say otherwise. `GeecsLogbook/CLAUDE.md:285-287` repeats the claim ("Everything on these pages folds").

Either add `details.devlist` to `cards()` or amend both statements. Leaving it is precisely the "accident of history" the comment was written to prevent.

### 5. Logbook: a fourth chip treatment in a stylesheet that already has three, and a verbatim copy of a kit rule the page already loads (redundancy, low-moderate)

**(a) The disclosure marker.** `day.html:29` is `<body class="kit">` and `:23` loads `kit.css`. `kit.css:422-429` already defines, for `.kit details.disc > summary`: `list-style:none`, `::-webkit-details-marker{display:none}`, `::before{content:"\25B8"}`, `[open] > summary::before{content:"\25BE"}`. New `scanlog.css:66-69` reproduces all four lines for `.devlist`. And `ScanAnalysis/.../editor.css:11-13` writes a **third** copy for `.ce-side details` — this one as a `rotate(90deg)` transform rather than the glyph swap.

**(b) The chip list.** `scanlog.css:292-296` already has `.tagcloud` + `.tagchip` — "a wrapping list of small labelled pills", used on the month page. New `.devlist ul` / `.devlist li` is the same component at different numbers. `.tagchip` is accent-coloured and interactive so it is not a drop-in, but a neutral variant belongs beside it, not as a private fifth component.

Files checked before concluding duplication: `kit.css`, `kit.html`, `scanlog.css`, `editor.css`, `scanner.css`, `GEECS-DataPortal/geecs_portal/templates/run.html`. Not a duplication: the device drawer's multi-select correctly reuses the kit's `.picklist [aria-pressed="true"]` rule rather than inventing a selection style — that part is clean.

### 6. Scanner: dangling `·` in the datalist label when a settable has an alias but no units (correctness, low)

`GeecsScanner/geecs_scanner/static/scanner.js:855`. `alias="Jet X", units=""` → `"Jet X ·"`. Verified: `node -e` returns `"Jet X ·"`. `.trim()` removes the trailing space, not the separator. `SettableOut` (`service/models.py:344-347`) defaults `alias` and `units` to `""` independently, and per the DB survey aliases without units are common. The sweep composer this is modelled on has no such bug.

### 7. Scanner: the move picker now accepts free text but matches it exactly (correctness, low)

`settableFor` (`scanner.js:845`) is `s.name === name`. With the `<select>` gone, the user can now type or paste, and the gate at `:818` is `!!settableFor(mvName)`. Typing `u_s1h:current`, or pasting `" U_S1H:Current"` off a log line, leaves **Move** disabled with "no settable named …" even though the variable exists. The `<select>` made that state unreachable. A `trim()` plus case-insensitive lookup — writing the canonical name back into the input on a match — closes both.

### 8. Scanner: on first paint the new rule and the "This plan" eyebrow point at nothing (correctness, low)

Boot is `setMode("", true)` (`scanner.js:1198`) — no mode is pressed. So on every page load `#sweep-composer`, `#count-options` and `#optimizer-form` are all `hidden`, and `#scan-plan` renders as an eyebrow reading "This plan" over an empty box, below a horizontal rule that separates it from nothing. When the worker has no optimizer configs, `renderOptimizerAvailability` un-hides `#optimizer-availability` with "No compatible optimizer configs available", which now sits under that eyebrow and reads as if it were the plan's own content. This is the first thing every operator sees.

### 9. Scanner tests: one assertion in the rewritten layout test cannot fail (test quality, low)

`invariants = body.index('id="scan-invariants"')` and `#scan-invariants` is the **first** element inside `<div class="body">`, so `head` is ~30 characters of rendered markup. The assertion is vacuous now and stays vacuous under any future layout that keeps `scan-invariants` first. Its predecessor spanned a real region. The three ordering assertions above it *do* bite; this one does not.

### 10. Docs describing the changed control were not updated (contract travel, low)

`docs/geecs_scanner/overview.md:60-62` still describes a dropdown ("**Pick one**", "alias first"). It is now a type-ahead `<input list>`. (The same page's "New scan" bullet is separately stale — it still names "No-scan / 1D / Grid / Background" against the page's Count / Sweep / Optimize — but that predates this PR.) `GeecsScanner/CLAUDE.md:188` still reads "Capture controls remain shared outside the trajectory tabs" — the eyebrow named **Capture** is what this diff renamed.

### 11. Scanner: Esc / scrim silently discards pending picks (low; half-refuted, recorded for the record)

`closeDrawer()` in `kit.js` fires on Esc and on the scrim. Selections live only in `devPicked`, which nothing persists. Shift-select 18 cameras, press Esc, reopen — gone. My own refutation: Esc meaning "cancel" is standard, and if Esc cancels then wiping on reopen is the *consistent* choice. What survives is only the tension with the kit's rung-2 rule as stated — worth one line in the CHANGELOG or a comment stating the choice, not a code change.

### Things the reviewer checked and found clean

- `el()`'s boolean attribute handling correctly omits `open` when false.
- `trackOpen`'s listener is bound to the node's own summary, so a nested namespace click does not double-fire on the parent.
- `pickDevice`'s range branch uses `devAnchor !== null`, so index 0 works as an anchor; `devShown` indices align with the `forEach` index.
- `renderDeviceList` resets `devShown`/`devAnchor` on every keystroke but keeps `devPicked` — the documented intent, and correct.
- `have[n]` disables already-added rows and the range loop skips `have`, so `addDevices` cannot double-add; the test would fail if the guard were removed.
- Every new class in `console.html` is styled, so `test_page_uses_only_kit_or_page_classes` holds.
- Logbook fixtures: `make_run(1)` creates no device folders, so `<summary>12 saved</summary>` is exact; the `<li>` / `"UC_Amp4_IR_input,"` pair genuinely distinguishes list from comma-join.
- `.fact.devices{grid-column:1/-1}` is sound against `repeat(auto-fit, minmax(128px,1fr))`, and `overflow-wrap:anywhere` keeps a long device name from blowing out min-content width.
- No intra-repo version pins need updating; no `analysis_status` / scan-folder writes anywhere in the diff; no public-repo hygiene breaches.

skbarber:
## Review round 2 — verdicts and the last three fixes

Same reviewer, fresh look at `65617dd0`. **Seven CONFIRMED, one DENIED, both waivers accepted** (finding 3 withdrawn outright — the tri-state store inverted the cost of a wrong guard flag, which makes `click` + `setTimeout` the safer idiom, not merely a divergent one). All three remaining items are closed in `a7628093`.

### 10 — DENIED, and the denial was right

The round-1 docs edit promised that typing an alias narrows the move picker. `settableMatch` consulted only `s.name`, so an operator who read that sentence, typed `Jet X` and stopped got **Move** disabled with the title "no settable named Jet X". The round-1 text was vague-and-wrong; the replacement was a concrete promise the page refused, which is worse.

Closed from the code side rather than by hedging: an alias resolves **when it names exactly one settable**, and an alias the DB has put on two variables names neither — the console refuses rather than guessing which magnet was meant. The POST still sends the canonical `Device:Variable`. Pinned by `test_the_alias_resolves_when_it_names_one_variable`, including the ambiguous case.

### 2 — the storage test did not bite

`test_unreadable_storage_is_not_a_broken_sidebar` seeded nothing, so `getItem` returned `null`, `openState` short-circuited at `raw ? … : {}`, and neither the array guard nor the `catch` was reached — the docstring named three failure modes and covered none. The harness now seeds `localStorage`; the test covers unparseable JSON, the previous format's array, a literal `null`, and that a stale array does not survive the next write. Verified it fails when the array guard is removed.

### 7 — two unreachable-today nits

`renderMoveVars` passes the resolved name to `watchReadback` rather than the raw field text, and the unresolvable branch of the Move handler calls `renderIdleGates()` instead of leaving the button stuck disabled.

### Consequences named, not fixed

- **Permalink into a closed section.** Once a node has been toggled its stored value is definite, so arriving at `#/analyzers/<id>` whose section the user previously closed shows a sidebar that does not reveal where you are. That is the correct resolution of "the user's word beats the auto-rule" and matches how file trees behave; left as is.
- **`Expand all` / `Collapse all` now reach the device list**, so a viewer who has used that button has a stored choice that wins over the per-scan default on every load. Noted in the logbook CHANGELOG.
- **The `N · M bad` namespace summary** stays untested: it lives inside the `section` closure, which the extraction harness cannot reach. The reviewer's correction is accepted — the honest reason is reach, not "markup with no logic".

Stopping the review here: round 2 closed every finding that survived, and the remaining notes are recorded rather than open.

Tests after these fixes: GeecsWebTheme 182 passed; GeecsLogbook 278 passed, 1 skipped; GeecsScanner 145 passed, 1 skipped; ScanAnalysis 196 passed, 2 deselected. Lint clean.

skbarber (commented):
Reviewed the PR. I did not find any blocking issues in the changed logbook, scanner, config-editor, or docs surfaces.

Local verification note: the changed JavaScript files pass `node --check`. I could not run the targeted Python suites in the temporary PR worktree because the available Poetry environment was missing package test/runtime dependencies such as `fastapi`/`pytest`.

PR #940, opened 2026-09-18
CI turnaround: selective legs, cached downloads, no double runs (~10 min → ~3 min)

skbarber:
## What and why

Unit-tests CI had crept to ~10 minutes on every push, which throttles the
whole workflow. Measured breakdown of run `35361075031` (612 s):

| | Install | Tests | Total |
|---|---|---|---|
| **GeecsBluesky** | **129 s** | **225 s** | **354 s (58%)** |
| Everything else (10 pkgs) | 129 s | 124 s | 253 s |

Repo-wide, **258 s (42%) was `poetry install`** re-downloading dependencies
that had not changed.

**Where the creep came from** (`git log` on the workflow): four commits on
2026-09-14/15 added the `ca`, `tiled` and `optimize` extras to GeecsBluesky.
Those were right — #866's whole point was that `importorskip` had been
turning the mock-CA layer into silent skips, so a green check said nothing.
**Nothing here rolls that coverage back.** What was missing is the
infrastructure that makes broad coverage affordable.

## The four changes (one concern: CI turnaround)

| Concern | LOC | Effect |
|---|---|---|
| Selective legs (`scripts/ci_select.py` + tests) | +509 | −40% runner minutes |
| Workflow restructure (matrix, cache, dedup, concurrency) | +325/−130 | wall clock → slowest leg |
| `QS_CONNECT_TIMEOUT` stall fix | +36 | −77 s |
| `check.sh` sync note | +8 | — |

**1. Selective legs.** `scripts/ci_select.py` derives the package graph from
the packages' own `pyproject.toml` path deps — never a hand-maintained map,
so a new dependency widens CI the moment it is declared. A PR runs the
packages it touched **plus their reverse closure**. Simulated over the last
28 merged PRs: **60% of the runner minutes (−40%)**, with leaf and
docs-heavy PRs dropping to 1–2 legs and foundational ones correctly staying
at 8–10 (changing `GEECS-Data-Utils` really does need 10 of 11 legs).

Conservative by construction: infra paths (`.github/`, `scripts/`, root
lock/pyproject, pre-commit config) select everything, and an **unrecognised
top-level path selects everything** so a new package fails safe rather than
being silently skipped.

**2. No more double runs.** `on: [push, pull_request]` fired both events for
every branch push — verified: SHA `905e4f4d` ran as a 610 s push run *and* a
610 s PR run concurrently. Push is now master-only. Added `concurrency`
groups to both workflows so a superseded push stops holding a runner.

**3. Cache Poetry downloads** — `~/.cache/pypoetry/artifacts` + `cache`,
explicitly **not** `virtualenvs`. The wheel cache is where the cost was: the
`optimize` extra pulls `xopt → botorch → torch`, and on Linux torch drags
the multi-GB nvidia CUDA wheels with it, re-fetched every run.

> **Corrected in `9484cd3e`.** The first version of this step cached
> `~/.cache/pypoetry` wholesale while claiming it excluded the virtualenvs.
> It did not: Poetry's default `virtualenvs.path` is `{cache-dir}/virtualenvs`,
> i.e. *inside* that directory. The GeecsBluesky entry saved at **12.02 GB**,
> above GitHub's entire 10 GB per-repo cap. See the review comment below.

**4. A 20 s stall paid four times.** Profiling showed 20.005 s of *pure lock
wait*, zero CPU, in four `test_qserver_startup.py` tests. With no CA gateway
on a runner, `install_telemetry` runs every telemetry object out to its
deadline before dropping it from the baseline — ~80 s spent waiting for a
baseline none of those tests assert on. `QS_CONNECT_TIMEOUT` now bounds it
(**default unchanged at 20.0**); the tests set 0.1. The connect still
happens, still fails, and is still logged — just promptly.

## Also: a coverage gap closed

`GEECS-LogTriage` has **58 passing tests and has never run in CI**, though
`scripts/check.sh` has listed it all along. It gets a leg.

## Safety: what keeps narrowing honest

Selection is only safe because **every push to master still runs the full
matrix**. Declared path deps are the graph, but undeclared coupling exists —
the theme guard is exactly that case (GeecsWebTheme's tests walk *other*
packages' templates; it depends on nobody, so the reverse edge is spelled
out explicitly and pinned by a test). A mis-selection on a PR therefore
surfaces minutes after merge rather than never.

A `gate` job (`if: always()`, `needs: [select, test]`) gives one stable check
name. `master` is not protected today, but a dynamic matrix cannot back a
required status check without this, so it is there if you ever enable it.

## Review

Round 1 found **five silent-skip holes** in the selection logic — all fixed in
`9484cd3e`, each with a test proven to bite. Full report and dispositions in
the comment below. Headline number moved −40% → **−39%** of runner minutes;
that one point bought back correctness on all five.

## Tests

- `tests/test_ci_select.py` — **26 passed** (18 before review). Proved they bite: stubbing
  `dependents_closure` to return only the touched package makes
  `test_foundational_change_pulls_in_dependents` and
  `test_transitive_dependents_are_reached` fail, which is the exact silent
  failure mode that matters.
- `GeecsBluesky` full suite — **726 passed, 4 skipped**, identical to the
  pre-change baseline, so no test was lost to the timeout change.
- `tests/test_qserver_startup.py` — **7 passed, ~90 s → 10.9 s**.

I did not run `check.sh --all` locally: it would need every package env
installed in a fresh worktree. **This PR's own CI is the better check** —
it touches `.github/`, so `ci_select` selects all 11 legs and the full
matrix runs under the new structure.

## Hardware verification

**Not required.** No scan-execution or device behaviour changes. The one
runtime change is `QS_CONNECT_TIMEOUT`, which defaults to the 20.0 s that
was previously hard-coded — worker behaviour is byte-identical unless the
variable is set.

- **Selective CI at all** was my addition to the scope, not asked for. If
  you would rather have only the caching/matrix/timeout work, the
  `ci_select.py` + `tests/` + `select` job come out cleanly and you still
  get ~10 min → ~3 min.
- **`GEECS-LogTriage` leg** is adjacent to the stated concern. It costs
  ~12 s and closes a real gap, but it is a separable line.
- **`pytest-xdist`** inside the GeecsBluesky leg was considered and left
  out: those tests carry RunEngine loop threads and asyncio state, and
  flaky CI would cost more than the ~60 s it might save.


skbarber:
## Adversarial review — round 1, with dispositions

Fresh-context reviewer, run against `9505020d`. **Five findings, all silent-skip class** — ways the new selection could run fewer legs than a change needs and still report green. I verified every one against the tree before fixing it; none were taken on trust.

**All five fixed in `9484cd3e`.** Each fix has a test that fails when the fix is reverted (I checked by reverting each).

| # | Finding | Disposition |
|---|---|---|
| 1 | `git diff --name-only` hides a rename's source path | **Fixed** — `--no-renames`, here and in `check.sh` |
| 2 | Theme-guard glob list missed all 3 GeecsScanner surfaces | **Fixed** — match by extension; new test reads `_SURFACES` itself |
| 3 | `docs/` ignored, but GEECS-Schemas pins `docs/geecs_schemas/` | **Fixed** — carved out to the root leg |
| 4 | `deploy/` + `.claude/` ignored, but 3 root suites validate them | **Fixed** — both, plus `*/deploy/*.service`, select root |
| 5 | Poetry cache swallowed the virtualenvs (12.02 GB entry) | **Fixed** — caches `artifacts` + `cache` only |
| op 1 | `cancel-in-progress` could cancel the master backstop | **Fixed** — `${{ github.ref != 'refs/heads/master' }}` |
| op 2 | Required status check may need repointing at `gate` | **Not actioned** — `master` has no branch protection (API: 404 "Branch not protected"), so no check to repoint. Flagged for the owner if protection is ever enabled. |

Verification of each before fixing:

- **#1** reproduced in a scratch repo: `git mv PkgA/mod.py PkgB/mod.py` → `git diff --name-only` prints `PkgB/mod.py` alone; `--no-renames` prints both.
- **#2** `GeecsWebTheme/tests/test_no_literal_colours.py::_SURFACES` does contain `GeecsScanner/geecs_scanner/{templates/console.html,static/scanner.css,static/scanner.js}`, none matched by the old globs. Since the theme depends on nobody, the dependents walk can never reach it — so this was a genuine regression introduced by the PR, not a pre-existing gap.
- **#3** `test_schema_export.py:93` and `test_schema_reference.py:51` both assert against `docs/geecs_schemas/…`, and the former's docstring says a docs reorg dropping one "must fail CI, not silently green-skip (#730 review)".
- **#4** confirmed `tests/test_render_units_sh.py`, `tests/test_bootstrap_host_sh.py`, `tests/test_skill_frontmatter.py` all validate otherwise-ignored paths; `test_render_units_sh.py:30-36` also walks per-package `*.service` units whose validator lives in root `tests/`.
- **#5** confirmed from the run log: venvs resolve to `/home/runner/.cache/pypoetry/virtualenvs`, and the entry saved at **12.02 GB** — above GitHub's entire 10 GB per-repo cap. The reviewer added a consequence I had missed: a restored venv would defeat the #866 `aioca` guard, because `poetry install` does not remove what a lock dropped.

Impact on the headline number: selection over the last 28 merged PRs goes from 60% → **61% of runner minutes (−39%, was −40%)**. One point of savings bought back correctness on five silent-skip holes.

`tests/test_ci_select.py`: 18 → **26 tests**, all passing.

The fix commit has gone back to the same reviewer for confirm/deny, per `/land`. I'll post the outcome.

---

<details>
<summary><b>Full reviewer report (verbatim)</b></summary>

Reviewed at `9505020d`, worktree `.claude/worktrees/ci-efficiency`, diff vs `origin/master`. I verified the derived dependency graph against every `*/pyproject.toml` (it matches the root `CLAUDE.md` graph exactly — no missing *declared* edge), confirmed all 11 legs are executable (every `OWN_ENV_PKGS` dir has `tests/` and a `dev` group, including the newly-added `GEECS-LogTriage`), confirmed the `gate` job cannot report green on a failed or cancelled leg, and confirmed the `QS_CONNECT_TIMEOUT` change is behaviour-neutral when unset (`GeecsBluesky/geecs_bluesky/run_engine.py:61` already defaults `connect_timeout=20.0`, and `_connect_timeout()` returns 20.0 for unset/empty/unparseable — no finding there).

Five findings survive. All five are silent-skip class.

### 1. HIGH — `git diff --name-only` hides the *source* of a rename, so a module moved between packages never runs the source package's leg

`scripts/ci_select.py:243-249` runs `git diff --name-only <merge_base>`. Rename detection is on by default (`diff.renames=true` since git 2.9) and `--name-only` prints **only the destination path**. Verified empirically in a scratch repo:

```
$ git mv PkgA/mod.py PkgB/mod.py && git diff --name-only HEAD~1
PkgB/mod.py          # PkgA/mod.py is not listed at all
$ git diff --name-status HEAD~1
R100  PkgA/mod.py  PkgB/mod.py
```

Concrete scenario: a PR moves `GeecsBluesky/geecs_bluesky/foo.py` → `GeecsScanner/geecs_scanner/foo.py` (exactly the kind of move this repo does routinely — the geecs-core extraction, the console→scanner migration, and the "modernization PRs delete the old path in the same change" rule in `CLAUDE.md`). `classify()` sees only `GeecsScanner/...` → `dependents_closure("GeecsScanner")` = `{GeecsScanner}` → **legs = `["GeecsScanner"]`**. `GeecsBluesky`'s 44 test files — several of which import the module that just vanished — never run. Gate goes green; master breaks on the next push.

Note that this defeats the *specific* thing the fail-safe branch (`ci_select.py:216-218`) was written for: if a whole package directory were renamed, you'd expect the "unrecognised path" escape hatch, but with rename detection the old name is simply absent from the input.

Fix: `git diff --name-only --no-renames <merge_base>` (a rename then reports as add + delete, and both packages' legs get selected). `scripts/check.sh:125` has the same flag gap; it's pre-existing there and lower stakes (local pre-flight), but they should get the same flag in the same PR since this PR is what promotes the pattern into the CI gate.

### 2. HIGH — the theme guard's glob list misses `GeecsScanner`, so the no-literal-colour rule stops being enforced on the scanner's own assets

`scripts/ci_select.py:100-106` lists five `THEME_GUARDED_GLOBS`. `GeecsWebTheme/tests/test_no_literal_colours.py:80-99` defines `_SURFACES`, the list of surfaces the guard actually walks, and it contains three paths the globs do not cover:

```
GeecsScanner/geecs_scanner/templates/console.html
GeecsScanner/geecs_scanner/static/scanner.css
GeecsScanner/geecs_scanner/static/scanner.js
```

(All three exist on disk.) Concrete scenario: a PR edits only `GeecsScanner/geecs_scanner/static/scanner.css` and adds `color: #ff0000`. `classify()` → `top = "GeecsScanner"` → `dependents_closure` = `{GeecsScanner}` (nothing depends on the scanner) → legs = `["GeecsScanner"]`. `GeecsWebTheme` is a *dependency* of GeecsScanner, not a dependent, so the reverse closure never reaches it. The theme leg doesn't run, `test_surface_uses_only_tokens[GeecsScanner/...scanner.css]` never executes, and the literal lands on master. Before this PR the theme leg ran on every push, so this is a regression introduced here, not a pre-existing gap.

`test_theme_guarded_globs_point_at_real_directories` (`tests/test_ci_select.py:193-200`) only pins that the listed prefixes exist — it cannot catch a *missing* prefix. The test that would bite is one that asserts every entry of `GeecsWebTheme.tests.test_no_literal_colours._SURFACES` outside `GeecsWebTheme/` is matched by some `THEME_GUARDED_GLOBS` prefix. (`scripts/check.sh:129-134` has the same `GeecsScanner` gap in its `case`.)

### 3. HIGH — `docs/` is unconditionally ignored, but `GEECS-Schemas`' published contract artifacts live under `docs/` and have an explicit "must not green-skip" pin

`scripts/ci_select.py:90` ignores everything under `docs/`; `:91` ignores every `.md`. But:

- `GEECS-Schemas/tests/test_schema_export.py:50` asserts each committed `docs/geecs_schemas/<name>.schema.json` equals a fresh render, and its docstring says verbatim: *"a docs reorg that drops or moves one must fail CI, not silently green-skip (#730 review)"*.
- `GEECS-Schemas/tests/test_schema_reference.py:27,51` does the same for `docs/geecs_schemas/schema_reference.md`.

Concrete scenario: a PR that only moves/renames/reorganises `docs/geecs_schemas/` (or hand-edits one `.schema.json`) produces `legs = []` → `any=false` → `test` job skipped → `gate` treats `skipped` as a pass (`unit-tests.yml:216-218`). The test written specifically to make that failure loud now never runs. `tests/test_ci_select.py:168-175` (`test_docs_only_change_selects_nothing`) actively *pins* this behaviour, so the new tests endorse the gap rather than catching it.

Minimum fix: carve `docs/geecs_schemas/` out of `IGNORED_PREFIXES` and map it to the root leg (or, more robustly, treat any non-`.md` file under `docs/` as unrecognised → fail-safe).

### 4. HIGH — `deploy/` and `.claude/` are ignored, but three root-`tests/` suites exist solely to validate them

Same mechanism as #3, three more instances:

| Ignored path | Test that pins it | Where |
|---|---|---|
| `deploy/render_units.sh`, `deploy/site.env.example` | `test_render_units_sh.py` | `tests/test_render_units_sh.py:26-27` |
| `deploy/bootstrap_host.sh` | `test_bootstrap_host_sh.py` | `tests/test_bootstrap_host_sh.py:37` |
| `.claude/skills/*/SKILL.md` (tracked files) | `test_skill_frontmatter.py` | `tests/test_skill_frontmatter.py:13,19` |

Scenario A: a PR edits `deploy/render_units.sh` and breaks its placeholder-validity check → zero legs selected → green. Scenario B: a PR adds a skill whose `name:` doesn't match its directory (the exact failure `test_frontmatter_is_wellformed` exists to catch) → `.claude/` + `.md` both ignore it → zero legs → green.

There is a fourth, subtler variant: `tests/test_render_units_sh.py:30-36` validates the **per-package** unit templates (`GeecsCAGateway/deploy/geecs-ca-gateway.service`, `GeecsBluesky/qserver/deploy/geecs-qserver.service`, …). Those paths are *not* under the ignored `deploy/` prefix, so they select that package's own leg — but the test that validates them lives in root `tests/`, which is **not** selected. So editing `GeecsScanner/deploy/geecs-scanner.service` to reintroduce a hard-coded lab path runs GeecsScanner's suite and skips the site-profile validator entirely.

### 5. MEDIUM-HIGH — the Poetry cache path includes the **virtualenvs**, contradicting the step's own comment and allowing a stale env to green a build

`.github/workflows/unit-tests.yml:97-104` caches `path: ~/.cache/pypoetry`, with the comment stating *"Poetry's wheel/artifact cache, **not the virtualenvs**"*. That is not what the path does. Poetry's default `cache-dir` on Linux is `~/.cache/pypoetry`, and `virtualenvs.path` defaults to `{cache-dir}/virtualenvs`. There is no `poetry.toml` anywhere in the tree and no `POETRY_VIRTUALENVS_*` in the workflow, so **every leg's virtualenv is inside the cached path**.

Two consequences:

1. **Green-when-it-should-be-red.** `poetry install` does *not* remove packages that are no longer required. A PR that removes a dependency from a lock — or drops one of the `--extras` — installs onto a venv that still has it. Most pointedly, the `#866` guard (`grep -E "could not import '(aioca|p4p)'"` → `exit 1`) is defeated: if the `ca` extra silently stops resolving, the restored venv still has `aioca`, the tests still import, the grep finds nothing, and the check that exists to stop exactly this silent skip passes.
2. **It defeats its own stated goal.** With `restore-keys` prefix matching and 11 legs each saving a multi-GB entry, GitHub's 10 GB per-repo cap will evict continuously — the failure mode the comment says it was avoiding.

Fix: narrow the path to `~/.cache/pypoetry/artifacts` + `~/.cache/pypoetry/cache`, or set `POETRY_VIRTUALENVS_IN_PROJECT: true`.

### Two operational notes

- **`cancel-in-progress: true` also applies to `push: master`.** The whole narrowing argument rests on "every push to master runs the whole matrix", but two merges in quick succession cancel the first run's exhaustive backstop. Use `cancel-in-progress: ${{ github.ref != 'refs/heads/master' }}`.
- **Branch protection.** The required status check is presumably still named `unit-tests`; that job no longer exists. Merges will hang pending until the required check is repointed at `gate`.

### Checked and cleared (no finding)

- **Graph derivation** — all 15 packages parsed, every edge matches root `CLAUDE.md`, including optional/extras edges. No package uses PEP 621 `[project]` tables, so the Poetry-only parsing is complete today.
- **`gate` result handling** — `needs.test.result` is the matrix aggregate; `failure`/`cancelled` both fall into `*` → `exit 1`.
- **Repo-root files** — `conftest.py`, `LICENSE.txt`, `image_analysis_configs/`, `apps_script/` all fall into the fail-safe "unrecognised → every leg" branch. Over-selection, the safe direction.
- **`hashFiles(steps.leg.outputs.lock)`** — every leg has a `poetry.lock`, so it never degenerates to the empty hash.
- **Fork PRs / `origin/<base_ref>`** — `fetch-depth: 0` uses `+refs/heads/*:refs/remotes/origin/*`, so `origin/master` exists.
- **`QS_CONNECT_TIMEOUT`** — unset/empty → 20.0, identical to the existing default. No production change.
- **Cross-package test coupling beyond the theme guard** — swept every `parents[2]`/`parents[3]` reference in all 14 test trees. `GeecsBluesky/tests/test_action_compiler.py:48` reads `GEECS-Schemas/tests/fixtures/...`, which *is* covered by a declared edge. `GEECS-LogTriage/tests/test_classifier.py:159` would be an uncovered reverse edge, but that file no longer exists and the test already skips unconditionally — pre-existing dead pin, unrelated to this PR.
- **Leg ↔ workflow parity** — all 11 selectable legs have a `tests/` dir and a `dev` group and are reachable by the `case` dispatch.

</details>

skbarber:
## Adversarial review — round 2 (confirmation): all findings closed

The same reviewer re-ran against the fix commit `9484cd3e` and **confirmed all five findings fixed, with no new silent skip**. It verified by *executing* the classifier rather than reading it, and mutation-tested each fix (reverting it and watching the right test fail). It also ran the full root suite: `111 passed, 63 deselected`.

Per the review stopping rule, round 2 is where this stops.

| # | Verdict |
|---|---|
| 1 Renames | Confirmed fixed; `--no-renames` in both `ci_select.py` and `check.sh`; flag-pin test bites |
| 2 Theme guard | Confirmed fixed — all 18 `_SURFACES` entries select the theme leg, incl. the 3 GeecsScanner surfaces that were the hole |
| 3+4 Root-tested paths | Confirmed fixed; `docs/platform/*.md` still correctly selects nothing |
| 5 Cache path | Confirmed fixed; `virtualenvs/` now outside the cached set |
| op 1 | Confirmed; PR runs still cancel, master runs do not |
| op 2 | Accepted — no branch protection, so `gate` is advisory until it is enabled |

Two things worth recording from its answers:

**The `*.service` fall-through keeps the dependents walk.** `GeecsBluesky/qserver/deploy/geecs-qserver.service` → `['root', 'GEECS-MCP', 'GeecsBluesky', 'GeecsScanner']`. That was the subtle part that could have been lost. And `deploy/render_units.sh` `continue`s before reaching the fail-safe branch, which removing `"deploy/"` from `IGNORED_PREFIXES` made reachable for the first time.

**Extension matching widens slightly, in the safe direction.** Because the theme check now sits above the ignore list, a `.js` under `extras/` picks up the ~19 s theme leg it didn't before. The reviewer's judgment, which I agree with: moving the check below the ignore list to "fix" that would reintroduce the exact ordering fragility we just removed.

### One residual, recorded and deliberately not fixed

`test_real_locks_are_clean` (`tests/test_scrub_lock_paths.py`) walks every committed `<Pkg>/poetry.lock`, but a package lock selects that package's legs, not `root` — so that assertion does not run on a lock-only change.

Not fixed, because **the rule itself is still enforced on every PR by a different route**: `.pre-commit-config.yaml`'s `scrub-lock-paths` plus the `poetry-lock-no-worktree-paths` pygrep backstop, both `files: poetry\.lock$`, carrying character-for-character the same three patterns the test asserts — and `pre-commit.yml` runs on every PR. The only genuinely unrun part is the set-shape assertion (that the lock files still exist), which guards against deleting a lock rather than against a dirty one. A third route to the root leg is not worth it for that.

Reviewer's closing: *"all five findings closed. Nothing survives."*

skbarber:
## CPU-only torch: attempted, reverted, filed as #942

Tried to fold CPU-only torch into this PR (owner's call — same theme). It does not work with Poetry, so it is reverted out (`4d14ac92`) and this PR is back to its CI-only scope.

Three forms, all failing differently:

| Approach | Result |
|---|---|
| Single source, all platforms | **Breaks macOS** — `poetry install --extras optimize` exits 1, "38 wheel(s) were skipped" |
| Platform split (Linux→index, macOS→PyPI) | macOS fine, **CUDA still installs on Linux** (15 nvidia + triton, seen on the runner) |
| Both arms on the index, macOS pinned `==2.10.0` | macOS arm resolved back to PyPI (PEP 440: `==2.10.0` matches `2.10.0+cpu`); lock still carried all 19 |

Root cause: Poetry cannot express *"this transitive dependency, from this index, on this platform only."* Full write-up, including the measured 8.4 GB and the options (CI-only overlay, `uv`, or wait), is in #942.

Worth noting the second form was caught **by CI, not by inspection** — the run at `3723bf08` failed on a guard added specifically because a Mac cannot answer what installs on Linux. Without it this would have merged looking correct and changed nothing. That guard is in #942 for whoever picks this up.

No change to what was reviewed: the diff is the same 9 files the round-2 confirmation signed off on, plus the revert.

PR #941, opened 2026-09-18
GeecsPvaGateway 0.11.2: compress the attribute datasets too

skbarber:
## What

One line: pass the existing `filters` dict to the `NDAttributes` `create_dataset` loop, which it was never passed to. The frames already used it (#935); the attributes beside them did not.

## Why

Found while chasing a real observation on the first scan taken after #935 went live — **"the HDF5 is ~2× bigger than the combined PNGs"** (Scan001 of 26_0918, `UC_Amp2_IR_input`, 10 shots, 500×500 uint16).

The frames were not the problem. Accounting the file by dataset:

| | storage | actual data |
|---|---|---|
| `/entry/data/data` | 1,303,005 B | 10 frames, 3.82× off raw |
| 13 × `NDAttributes/*` | **1,703,936 B** | **1,040 B of numbers** |

An attribute chunk is `ATTRIBUTE_CHUNK` (16384) f8 slots = 128 KiB, and HDF5 commits the whole chunk on the first write, padding the rest with fill. So each attribute dataset costs 128 KiB whether the scan is 1 shot or 16,384 — a **constant** per-scan, per-device cost that happened to exceed the frames here.

Measured effect of the fix on that same data: attribute storage **1,703,936 → 9,191 bytes (185×)**, file **3.03 MB → ~1.31 MB**, which puts it *under* the 1.66 MB of equivalent PNGs. The frames were already winning (1.30 vs 1.66); the padding was burying it.

## Why this is safer than the frame compression it follows

- **Content-independent.** The saving is untouched fill, not image data. Blank frames, real beam, saturated — same win. None of #935's "measured on blank images" caveat applies.
- **`ATTRIBUTE_CHUNK` is deliberately unchanged.** ophyd-async declares `chunk_shape=(16384,)` in the stream resource it hands Tiled (`_data_logic.py:225`, "NDAttributes appear to always be configured with this chunk size"). Shrinking the chunk would have saved less *and* desynced the file from that declaration; compressing it saves more and keeps the contract. Measured alternative for the record: `chunks=(1024,)` uncompressed gives 106,496 B — 11× worse than compressing.
- **One switch for both datasets.** Gated behind the same `Compression=zlib`; a client putting `None` gets raw frames *and* raw attributes.

## Cost

~3 ms/shot for 13 attribute writes in a synthetic loop, on the same single writer thread as the frames. Irrelevant at 1 Hz; noted because it is not free.

## Tests

`./scripts/check.sh GeecsPvaGateway` — **51 passed**, rc=0. Lint clean.

Assertions added to the two existing end-to-end tests rather than a new fixture, so they run through the real `ADHDFDataLogic` path:

- the default path now also asserts the stamp dataset has `chunks == (16384,)` (the contract), `compression == "gzip"`, and `get_storage_size() < 16384 * 8` — the last is the one that would have caught this originally, since it fails on a fully-materialised empty chunk
- the `Compression=None` path asserts the attributes come back raw too

Proved to bite: removing `**filters` fails with `assert None == 'gzip'`.

## Hardware verification

**Not owed.** The observation this fixes came *from* hardware — Scan001 of 26_0918 on the canary box (192.168.6.100, gateway 0.11.1), and the numbers above are that file. Losslessness on that scan was independently confirmed by `geecs-pva-gateway diff`: `pass (matched=10 identical=10 png_only=0 stack_only=0)`. This change touches only the filter applied to fill, so it cannot alter pixel or attribute values.

Rollout is per camera server as usual, and **deliberately not urgent** — it should ride along with the next gateway roll rather than justify one of its own.


skbarber:
## Adversarial review — fresh-context reviewer, three rounds

Round 1 read `4709c994` (verdict: sound, two minor findings + one positive finding), round 2 confirmed `bea21bcd` and **retracted one of its own claims**, round 3 is that correction in `057903c2`.

### Disposition

| # | Finding | Disposition |
|---|---|---|
| 1 | `CLAUDE.md` said only the *frames* are compressed by default | **Fixed** `bea21bcd` |
| 2 | `get_storage_size() < 16384*8` passes vacuously at 0 | **Fixed** `bea21bcd` — now bounded below, `shuffle` added |
| 3 | The entry was silent about write traffic — the larger effect | **Added** `bea21bcd`, independently re-measured |
| 4 | My round-2 fix claimed "the consolidator refuses a mismatch" — false | **Fixed** `057903c2` |

**Finding 3 changed the size of this change and I had it wrong in the PR body.** The entry argued storage only. But HDF5 rewrites every dirty chunk in full on each per-frame `flush()`, so the padding was crossing SMB **on every shot**, not stored once per scan. I verified before repeating it, with my own write-byte counter (13 attributes, `chunks=(16384,)`, attribute writes only):

| | bytes written per frame |
|---|---|
| uncompressed | **1,708,401** |
| gzip1+shuffle | **16,113** |

~106×, and `13 × 131072 = 1,703,936` accounts for essentially all of the uncompressed figure — all thirteen full chunks, every shot. The reviewer's independent harness got 1,798,684, and the 94,748 gap is exactly one gzip'd frame (its harness wrote one per iteration; mine didn't), so the two agree. 106× is the conservative end-to-end number; attribute-payload-only is ~140×. Kept the conservative one.

**Finding 4 is the reviewer correcting itself, on a claim I had explicitly flagged as taken on trust.** I asked it to check `consolidators.py:321`, and it found the call site it had skipped. Verified here against the installed packages:

- `tiled_writer.py:597` only calls `validate()` when the stream resource carries `_validate` — default `False`
- ophyd-async's NDAttribute `StreamResourceInfo` (`_data_logic.py:219-229`) passes `parameters={"dataset": ...}` only
- and the call is `validate(fix_errors=True)`, i.e. warn-and-adopt, with the exception downgraded to a `warn()`

So nothing refuses anything. Wrong in the direction that matters: a future reader would believe lowering `ATTRIBUTE_CHUNK` fails loudly, when it would silently corrupt the layout Tiled registered. The clause now states what is actually load-bearing — ophyd-async hard-codes the declaration, so the file must match it — and says plainly that nothing enforces it.

### What the reviewer tried to break and couldn't

Recorded because these are the failure modes a compressed, append-per-frame, rewind-capable dataset would plausibly have:

- **File-space leak from rewriting compressed chunks** — no. 1000 frames with per-frame flush: 57 KB of all-in overhead; HDF5's free-space manager reuses vacated blocks. Attribute storage 8.5 KB at 10 shots → 10.9 KB at 2000.
- **Rewind/re-append corruption or cliff** — no. 200 cycles of "append 5, resize down 3, re-append": every value bit-exact, file 62,964 B compressed vs 1,731,470 B uncompressed. `Rewind` to 0 then re-append exposes no stale values.
- **Un-finalized file readability** — unchanged by this PR. A mid-write snapshot fails identically with and without attribute compression (`file is already open for write`), a pre-existing property of the writer's superblock flag.
- **Reader side** — every `scan_stack` entry point plus `diff.py` run end-to-end against a compressed, rewound stack, all correct. Nothing depends on contiguous storage.
- **GIL starvation of the event loop** — no evidence: p99 2.60 → 1.97 ms across the change; h5py releases the GIL, and appends run on the dedicated writer thread regardless.

### Deployment

Reviewer and I agree: **no out-of-band roll.** Two refinements from it worth recording, one of which corrects my reasoning in the PR body:

- **The waste is not contingent on frame compression.** The attribute padding has crossed SMB on every shot since the attribute datasets landed (0.8/0.9), independent of `Compression`. All nine boxes pay it today — 0.11.0 did not create it and the non-canary boxes are not insulated from it. So the benefit is real for every box the moment it rolls.
- **The bound is per capturing camera, not per host**: ~1.7 MB/s per camera at 1 Hz, so a four-camera host pushes ~7 MB/s of padding during a scan. Scales with cameras-per-host and attribute count, not shot rate alone.

**What would change the answer:** nonzero `queue_drops` or `append_failures` in the canary's session counters, or `frame queue full: the writer is wedged` in its logs — that is the path where this is a data-loss fix rather than hygiene, since a writer thread spending its per-frame budget pushing 1.7 MB of zeros at SMB is how a 64-deep queue overruns. Scan001 of 26_0918 on the canary reads `queue_drops=0`, so: ride along with the next roll.

Round 4 was not opened. `./scripts/check.sh GeecsPvaGateway` → 51 passed, rc=0 at every fix commit.

PR #944, opened 2026-09-21
native_image_save: the run-level LabVIEW-files switch for plugin-backed cameras (#738)

skbarber:
## What

`native_image_save`, rebuilt as the run-level switch for LabVIEW's per-shot files (PNGs) on plugin-backed cameras — the migration lever PNG retirement (#738) has owed since #937 removed the inert daemon-era field.

- **One value per scan**, on the document the scanner actually submits: `Preset.native_image_save: bool | None`. Unset defers to `ExperimentDefaults.native_image_save: bool = True`, which the worker reads **at every run** — editing `experiment_defaults.yaml` reaches the next scan with no environment reopen.
- **Reach is derived, not chosen**: the switch turns off native saving for **plugin-backed cameras only** (a PVA stream the file plugin captures). A device without a file plugin — a camera on a box not yet rolled, HASO or any proprietary-format device — has no other record and is never touched, whatever the switch says. Rule agreed with the maintainer 2026-09-20: the plugin covers everything a PVA-streaming device saves natively (the non-scalar PVA work delivers the remaining coverage), so no per-devicetype list is needed.
- The controls stay owned when off: a stale `save=on` is still cleared at `stage` (the 26_0828 lesson); `prepare` never switches saving on, creates no device directory and adds no `-nonscalar_save_path` column. The switch is restored in a `finalize_wrapper`, so the namespace's long-lived detectors carry their construction default (dual-write) into the next run. The start document records `native_image_save`; the journal names the cameras switched and the ones kept.
- Scanner: a three-state **LabVIEW files** select among the fields every scan needs (`experiment default` / `on` / `off · plugin stacks only`), round-tripped as the preset field; the queue summary says `no LabVIEW files` when an item switches them off.

Gated runs were already PNG-free for plugin-backed cameras (the fly prepare leaves the native logic out); this matters for strict runs. `geecs-pva-gateway diff` reports a switch-off scan as `capture_only`.

## Why here and not on ScanRequest

`ScanRequest` has no consumer outside GEECS-Schemas; the scanner submits `Preset` documents. `ScanRequest` keeps its v4 refusal with the remedy repointed at `Preset.native_image_save`.

## Per-concern LOC (one concern, three packages)

| Package | Version | Lines | What |
|---|---|---|---|
| GEECS-Schemas | 0.31.0 | +101 / −57 | the two fields, the 0.30.0 drop-validator removed, remedy text, tests, regenerated `schema_reference.md` |
| GeecsBluesky | 0.97.0 | +~410 / −23 | `LvNativeFileDataLogic.enabled` + `GeecsDetector.native_image_save`; `resolve_native_image_save` + `native_image_save_wrapper` (registry); the kwarg on the scan verbs and `optimize`; preset expander; `RESERVED_DEVICE_ATTRIBUTES`; EVENT_SCHEMA + CLAUDE.md; tests |
| GeecsScanner | 0.13.0 | +115 / −2 | the select, the JS round-trip, the summary line, tests |

**Judgment calls, cheap to veto:** (1) the default is read per run rather than at bind time — deliberate, so a defaults edit needs no worker restart; (2) no endpoint shows the experiment default in the form — the trigger-profile picker has the same gap today, so the select's first option reads `experiment default` without saying which; (3) `optimize` takes the switch too — its objective reads live PVA frames, never the PNGs, so nothing pins it on.

**Facility literals:** none added.

## Interaction with the non-scalar PVA arc (`feature/nonscalar-pva`)

No shared code: this PR does not touch the plugin-variable block in `namespace.py` that arc's PR 2 rewrites. The dependency is semantic and is the rule above — the switch assumes the plugin captures everything a plugin-backed device saves natively; that arc's capture-stream declaration is what makes it true for MagSpecCamera (ImageInterp) and the trace devices.

## Tests

New (11): `test_native_image_save_defaults_on_and_is_read`, `test_native_image_save_is_one_value_per_scan` (schemas); `test_native_image_save_off_keeps_the_controls_and_writes_nothing`, `test_native_image_save_needs_the_controls`, `test_native_image_save_off_reaches_plugin_cameras_only_and_restores`, `test_native_image_save_default_comes_from_the_experiment_defaults_per_run`, `test_native_image_save_default_is_on_when_the_defaults_cannot_be_read`, `test_native_image_save_rides_as_the_plans_keyword_only_when_set` (bluesky); `test_native_save_control_sits_with_what_every_scan_needs`, `test_native_save_control_round_trips_the_run_level_switch`, `test_native_image_save_off_is_said_and_unset_is_silent` (scanner).

Bite check: with `prepare_single` ignoring the switch, 3 of the new tests fail; with the wrapper never setting it, 2 fail.

Counts, worktree code (the main checkout's venvs via `PYTHONPATH` for schemas/bluesky; the scanner in its own worktree env):

| Suite | Result |
|---|---|
| GEECS-Schemas `tests/` | 506 passed |
| GeecsBluesky `tests/` (not integration/gui) | 731 passed, 4 skipped, 1 failed → the `RESERVED_DEVICE_ATTRIBUTES` pin, fixed; `test_namespace` + `test_run_action` + `qs_client` re-run: 143 passed |
| GeecsScanner `tests/` | 148 passed, 1 skipped |

`scripts/check.sh --all` was not run from the worktree (no per-package envs there); CI is the backstop.

## Hardware verification

**OWED**, after deploy (worker: pull + `poetry install` + environment reopen for the new plan signature; scanner: pull + restart):

1. A strict scan with `native_image_save: false` on a plugin-backed Point Grey camera: the `<device>/<device>.h5` stack present, **no** PNGs in `ScanNNN/<device>/`, no `-nonscalar_save_path` column for it in the s-file, `geecs-pva-gateway diff` → `capture_only`; the journal line `native saving off (file-plugin stacks only) this run for …`.
2. The same preset with the field unset and `experiment_defaults.yaml` saying `native_image_save: false`, then flipped back to `true` between two scans with **no** worker restart: the first scan PNG-free, the second dual-writing.
3. A scan including a camera on a box not yet on the plugin: it still writes its PNGs.

## Adversarial review

Posted as a comment below once complete.


skbarber:
## Adversarial review of #944 (`feat/native-save-toggle` @ 932d58a1) — correctness / redundancy / placement

*Fresh-context reviewer subagent; report pasted verbatim.*

Read against the code, not the PR body. CI is green on all 11 legs; locally from the worktree: GEECS-Schemas 506 passed, GeecsBluesky (registry, detector, presets, namespace, strict, gated) 123 passed, GeecsScanner 148 passed / 1 skipped. Three scenarios the suite does not cover were run as scratch probes on the worktree code (a gated bound count, a `.scalars`-only plugin camera, pause + `RE.halt()` / `RE.abort()` / `RE.stop()`); the findings below cite what they printed. Nothing blocks the merge; 1 and 2 are the ones worth fixing before hardware verification, the rest are doc/consistency.

### Findings

**1. Low–Medium — the start document records the switch, but EVENT_SCHEMA says "wrote".**
`GeecsBluesky/EVENT_SCHEMA.md:30` defines `native_image_save` as "Whether the plugin-backed cameras also wrote their LabVIEW per-shot files this run"; `plans/registry.py:353` writes `md["native_image_save"] = native_files` unconditionally.
Scenario (probed): `count([UC_Plugin], 2, acquisition="gated")`, switch unset → start doc `native_image_save: True`, no `Scan001/UC_Plugin/` PNG dir, no `-nonscalar_save_path` column (the fly prepare leaves `LvNativeFileDataLogic` out, as the CHANGELOG's own Notes say). Same for `count([UC_Plugin.scalars], 1)` → `True`, nothing written. Any consumer that reads this key to decide where a run's frames live — the exact decision #738 is about — would look for PNGs that were never written.
Fix: reword the EVENT_SCHEMA row and the `md` semantics to "the run's LabVIEW-files switch (the preset's value, else the experiment default); whether a camera actually wrote is the presence of its `<name>-nonscalar_save_path` column". Recording an "effective" value instead would still be wrong for `.scalars` entries, so the wording fix is the honest one.

**2. Low — the journal line names cameras the switch cannot reach, and the docstring's justification is inverted.**
`plans/registry.py:232-263`: `cameras` is built from `[*detectors, *non_essential]` with `.scalars` views mapped to their owner (l. 236) and no regard for the acquisition mode; the INFO line then reports "native saving on this run for X" / "off (file-plugin stacks only) this run for X".
Probed: `count([UC_Plugin.scalars], 1, native_image_save=False)` logs "native saving off (file-plugin stacks only) this run for UC_Plugin" — a view writes neither PNGs nor a stack; a gated run logs "native saving on this run for UC_Plugin" with no PNGs. Non-essential flyers (unbounded prepare) are in the same position. The docstring at l. 216-217 says a view "counts as its owner (the owner's data logics are what a prepare consults)" — the opposite of `GeecsDetectorScalars` (`detector.py:552-553`, and `calibration.py:672-676` relies on exactly that: the view leaves the parent's data logics unprepared). Setting the owner's flag for a view is a harmless no-op, so this is a wrong log line and a wrong sentence, not wrong behaviour.
Fix: build `cameras` from the strict detectors only (skip views and `non_essential`), skip or soften the line when `acquisition == "gated"`, and correct the docstring.

**3. Low — `RE.halt()` leaves the switch stale, and the `previous`-based restore then re-instates the stale value after every later run.**
`plans/registry.py:255-270`: `finalize_wrapper` skips cleanup on `GeneratorExit` (bluesky 1.15.0 source, and `calibration.py:868/1125` already document this contract for their own finalizers).
Probed: pause inside `native_image_save_wrapper(…, False)` → `RE.halt()` → `cam.native_image_save` stays `False`; the next run with the switch unset sets `True` for itself (PNGs written — correct), then restores to the stale `False`. No scan misbehaves (every scan verb sets the switch per run; calibration plans prepare `.scalars` views), `RE.abort()` and `RE.stop()` both restore (probed), and halt is reachable only from the `qserver` CLI (no client calls `re_halt`). But the CHANGELOG/docstring claim that detectors "carry their construction default into the next run" is false after a halt.
Fix: restore to the construction default (`True`) rather than `previous` — nothing else sets this attribute, so there is no per-device value to preserve — or add "except after `RE.halt()`, which skips finalizers by bluesky contract" to the docstring and CHANGELOG.

**4. Low — CHANGELOG operator step contradicts the qserver README.**
`GeecsBluesky/CHANGELOG.md` 0.97.0 Notes: "reopen the worker environment (or reload the allowed lists) after deploying". `qserver/README.md:154-162`: the on-disk `existing_plans_and_devices.yaml` is regenerated only at environment open, and after a deploy "the worker process is equally stale, so `systemctl restart geecs-qserver` is the gesture". A `permissions reload lists` after deploy without a reopen would reload the pre-deploy signatures and the manager would keep refusing `native_image_save`. Drop the parenthetical.

**5. Low (placement / design-of-record consistency) — Rule 2 and the "Do not" list now contradict the mechanism.**
`GeecsBluesky/CLAUDE.md:36-39` ("Nothing outside a device configures it for a run") and `:469-470` ("Configure a device for a run from outside its lifecycle"). `native_image_save_wrapper` sets `GeecsDetector.native_image_save` from the plan before `stage` and restores it after `unstage` — per-run device configuration from the plan side, through a Python attribute rather than a PV. The PR updates the Images paragraph (l. 422-431) to describe the wrapper but leaves the rule and the bullet unqualified, so the file the header calls "the design of record" now argues with itself. Not asking for a redesign (threading the value through `TriggerInfo`/the hooks is heavier for no gain); ask that the rule name the carve-out — a run-level switch a device exposes as a property, set by the plan before stage and restored by finalize, PV writes still the device's own — so the next reader does not "fix" it in either direction.

### Redundancy (no action required in this PR)
- The "saves natively without a plugin" predicate now lives in three places: `namespace.py:398-403` (startup log), `gated.py:164-168` (`refuse_native_essentials`), `registry.py:240-245` (the wrapper's `kept`). The owner-dedup idiom (`getattr(d, "_owner", d)` + a `seen` id set) is in `registry.py:235-239`, `registry.py:337` (`strict_plan`'s `owners`) and `calibration.py:682-689` (`_stamp_views`). A `GeecsDetector.native_only` property and an `owners()` helper would collapse both; follow-up material.
- No prior implementation of the switch, a per-run defaults reader, or a run-level restore exists — checked `config_resolver.py`, `qs_client/submit_preflight.py`, `TriggerProfiles.from_resolver`, `plans/gated.py`, `plans/strict.py`, `plans/calibration.py`, GEECS-MCP, and the scanner service. The fake resolvers in `tests/test_plan_registry.py` (`_Defaults`, `Broken`, `Absent`) sit beside `_FakeResolver` (`tests/qs_client/test_submit_preflight.py:100`) and `StubResolver` (`tests/test_run_action.py:103`) — a pre-existing test-double sprawl, not this PR's.

### Checked and clean
- Removing the 0.30.0 drop-validator flips nothing at deploy: the configs repo's `Undulator/experiment_defaults.yaml` (cbcf6be, 2026-09-17) carries only `schema_version` and `trigger_profile`; nothing in `scanner_configs/` sets the key.
- "Read at every run" is true: `strict_plan.plan` and `optimize` are generator functions, so `resolve_native_image_save` runs at run start; `ConfigsRepoResolver.resolve_experiment_defaults` is uncached (`exists` + YAML load on the host's local checkout); missing → `None` → on; unreadable → warning + on (once per run, not "once" as the docstring says — trivial).
- Wrapper ordering: flags are set when the wrapper is first advanced — after the bracket's ARMED move (a refused move sets nothing), before `stage` (stock plans stage inside; `sweep_plan`'s and `optimize`'s `stage_wrapper` are inside the wrapper) — and restored after `unstage`; abort/stop restore (probed). One RE, one plan at a time, so no concurrent-run case exists.
- `kwargs.pop("native_image_save")` precedes `bind_partial`, so stock signatures never see it; `sweep(detectors, *, sweep, …)` exposes `detectors`; the manager-side `bool | None` annotation follows the working `str | None` / `float | None` precedent; an old worker refuses the kwarg (documented, minus finding 4).
- Scanner: the node harness pins field-wins / kwargs-copy-dropped / `null` when unset; `formShape` ignores kwargs; `summarize_item` speaks only on `False`; the service validates with `Preset` (a `SchemaModel`, unknown fields refused). `ItemSummary.native_image_save` is not surfaced in `models.py`, consistent with `acquisition`/`trigger_profile`.
- Docs: no published JSON schema exists for `Preset`/`ExperimentDefaults` (`docs/geecs_schemas/` has scan_request, capture_settings, etc.), so `schema_reference.md` was the only regeneration owed and the no-drift test passes.
- Scan-folder invariant untouched (`prepare_single`'s parent-must-exist + `mkdir(exist_ok=True)` unchanged). No facility literals added. `RESERVED_DEVICE_ATTRIBUTES` pin (`tests/test_namespace.py:368`) covers the new attribute. Version bumps 0.30→0.31, 0.96→0.97, 0.12→0.13 with CHANGELOG entries.

skbarber:
## Review round 2 of #944 — fix commit 808aacf4 checked against each round-1 finding

*Same fresh-context reviewer, continued; report pasted verbatim.*

Read-only, same worktree; diffed `932d58a1..808aacf4`. Re-ran the round-1 scratch probes (gated bound count, `.scalars`-only plugin camera, pause + `RE.halt()`) on the fixed code, and the Bluesky plan/detector/preset/namespace/strict/gated/optimization suites: 172 passed, 1 skipped. CI had not yet reported on 808aacf4 at the time of this check.

1. **CONFIRMED.** `EVENT_SCHEMA.md:30` now defines `native_image_save` as the run's switch, names the three shapes it cannot reach (view, non-essential stream, gated batch), and points at the `<det>-nonscalar_save_path` column as the record of what was written; the CHANGELOG entry says the same. `md["native_image_save"]` is still written unconditionally (a gated run still records `True` with no PNGs — probe A), which is now what the contract says it means.

2. **CONFIRMED.** `registry.py` `native_image_save_wrapper(plan, detectors, enabled)` keeps only `isinstance(d, GeecsDetector)` with `native_save` (`GeecsDetectorScalars` subclasses `ScalarsView`, not `GeecsDetector`, so views drop out); `strict_plan` passes `detectors` only and applies the wrapper under `if acquisition == "strict"` (`ACQUISITION_MODES` is exactly `("strict", "gated")`, so the gate has no third mode to miss); `optimize.py` passes `detectors` only. Probe: the gated count and both `.scalars`-only runs now emit no registry journal line, and the owner's flag reads `True` throughout. The docstring's view sentence is corrected to the right direction (view leaves the owner's logics unprepared). The new caplog assertion bites: the substring it asserts absent, "native saving off", is verbatim the line the pre-fix code printed for the same run in my round-1 probe.

3. **CONFIRMED.** `restore()` sets `cam.native_image_save = True` and the `previous` capture is gone. Probe C: after a halt the flag stays `False` (bluesky contract, now stated in the docstring and CHANGELOG), the next unset run sets and logs its own value and, on finishing, leaves the flag at `True` — the stale value no longer survives a later run.

4. **CONFIRMED.** CHANGELOG operator step now reads `systemctl restart geecs-qserver` and says `permissions reload lists` alone would reload the pre-deploy signatures — consistent with `qserver/README.md:154-162`.

5. **CONFIRMED.** `CLAUDE.md` rule 2 names the carve-out (a run-level switch the device exposes as a property, flipped before `stage`, restored after, PV writes still the lifecycle's) and the "Do not" bullet cross-references it as the named exception, so the design-of-record no longer contradicts itself.

**New findings introduced by the fix:** none of substance. One wording nit, not a work order: the wrapper docstring's "a halt costs nothing but the next run's journal line" is confusing — after a halt the next scan verb sets, logs and restores its own value, so the halt costs nothing at all; the CHANGELOG's phrasing ("the next scan verb sets the switch for itself anyway") is the accurate one and the docstring could just say that.

---

## Dispositions

| # | Finding | Disposition |
|---|---|---|
| 1 | EVENT_SCHEMA said "wrote"; the key is the switch | **Fixed** in 808aacf4, confirmed above |
| 2 | Journal named views / non-essential / gated cameras the switch cannot reach; docstring inverted | **Fixed** in 808aacf4 (wrapper takes strict detectors only, applied under `acquisition == "strict"`; caplog test), confirmed above |
| 3 | `RE.halt()` left a stale value that `previous`-restore re-instated | **Fixed** in 808aacf4 (restore to the construction default; halt contract stated), confirmed above |
| 4 | Operator step contradicted the qserver README | **Fixed** in 808aacf4, confirmed above |
| 5 | CLAUDE.md rule 2 / "Do not" contradicted the mechanism | **Fixed** in 808aacf4 (carve-out named), confirmed above |
| — | Redundancy: the native-only predicate in three places, the owner-dedup idiom in three | **Waived** — pre-existing idiom sprawl plus one new instance; a `GeecsDetector.native_only` property and an `owners()` helper are follow-up material, not this PR's concern |
| — | Round-2 nit: docstring "costs nothing but the next run's journal line" | **Waived** — wording only; the CHANGELOG carries the accurate sentence. Folded into the next GeecsBluesky touch rather than a third round for one docstring clause |

Severity floor set before review: anything below "wrong behaviour on a real path" is fixed if it is cheap or waived with a reason; two rounds, then stop. Round 2 found no new finding of substance, so this closes the review.

GeecsBluesky/geecs_bluesky/qs_client/presets.py
```diff
@@ -268,6 +270,8 @@ def expand_preset(
         kwargs["non_essential"] = non_essential
     if preset.trigger_profile is not None:
         kwargs.setdefault("trigger_profile", preset.trigger_profile)
+    if preset.native_image_save is not None:
+        kwargs.setdefault("native_image_save", preset.native_image_save)
```
skbarber: [P2] Make the preset field the sole source of truth

`PlanCall.kwargs` accepts `native_image_save`, and `setdefault` lets that value silently override the canonical `Preset.native_image_save` field. For example, a preset with top-level `false` and kwargs `true` expands to `true`. A kwargs-only preset also behaves differently when submitted directly versus loaded through the scanner, which removes the kwargs copy. Reject or remove this key from `plan.kwargs`, then derive it solely from the top-level field.

skbarber: Agreed, and fixed in e80ed667: a `native_image_save` key inside `plan.kwargs` is now **refused** by `expand_preset` (`GeecsConfigurationError`, remedy naming the preset field and the scanner's "LabVIEW files" control) rather than merged, and the top-level field is assigned into the item's kwargs when set. Refusing beats dropping: a hand-written YAML carrying the copy fails at preflight with the reason, instead of running with a value the author did not choose. Pinned by `test_native_image_save_in_plan_kwargs_is_refused_not_merged` (kwargs copy `true` against top-level `false` and against unset; both refused).

Noted, not changed here: `trigger_profile` has the identical `setdefault` shape a few lines up, and the scanner already deletes both copies on save for that reason. Making the trigger profile the sole source of truth the same way is a one-line follow-up outside this PR's concern.

Independent confirmation of the fix commit follows as a PR comment, as with the earlier rounds.

skbarber:
## Review round 3 of #944 — fix commit e80ed667 (Codex finding: kwargs copy of `native_image_save` overrode the preset field)

*Same fresh-context reviewer, continued; report pasted verbatim.*

Read-only, same worktree; diffed `808aacf4..e80ed667`. Local runs on the new head: GeecsBluesky `tests/qs_client` + `test_plan_registry.py` 127 passed; GeecsScanner 148 passed, 1 skipped. CI at check time: Scanner, MCP, CAGateway, Logbook, WebTheme, pre-commit, select passed; GeecsBluesky, root and DataPortal still pending.

**CONFIRMED** — the refusal fires before anything consumes the key. In `expand_preset` (`qs_client/presets.py`) the only steps ahead of the check at l. 275 are: `_resolve` over every kwarg value (a bool passes through untouched — only strings are treated as scan-variable names), the Sweep branch (reads `kwargs["sweep"]` alone), and the `acquisition` check. For an optimize preset, `prepare_optimizer_preset` (l. 142-165) runs first but does `kwargs = dict(preset.plan.kwargs)`, pops only `acquisition`, setdefaults `shots_per_step`/`max_iterations`, and returns the copy via `model_copy` — a `native_image_save` key survives it and is refused at l. 275. No other reader of `plan.kwargs["native_image_save"]` exists (`summarize_item` runs on the expanded item, which a refused preset never becomes; the scanner JS reads `doc.native_image_save` only and deletes the kwargs copy on save). The assignment `kwargs["native_image_save"] = preset.native_image_save` is correct given the refusal has emptied the slot; unset still leaves the key absent.

**CONFIRMED** — operator-facing. `run_submit_preflight` (`submit_preflight.py:145-151`) wraps `expand_preset` in `except Exception: report.refusal = str(exc)`, so the message lands in the preflight report; `ScannerService.submit` raises `ScannerError("invalid_request", report.refusal)` (HTTP 400) and `ScannerService._expand` maps the same exception to `invalid_request` (`scanner.py:779-783`). The `--demo` path (`demo.py:369`) validates through the real `expand_preset` too. A direct `client.submit_preset` (`client.py:741`) raises the `GeecsConfigurationError` to its caller, which is right for a library call.

**CONFIRMED** — the test bites. Pre-fix `expand_preset` had no raise on this path: top-level `False` + kwargs `True` expanded to `True` under `setdefault`, and top-level `None` + kwargs `True` expanded to `True`; both `pytest.raises` blocks in `test_native_image_save_in_plan_kwargs_is_refused_not_merged` would fail on that code. The helpers it uses (`sweep_call()`, `_preset(plan=…)` via `base.update`) exist and the match string is unique to the new message. No preset in the configs repo carries the key (`scanner_configs/` grep is empty), so nothing live is refused at deploy.

**CONFIRMED** — CHANGELOG sentence and the scanner JS comment now describe the behaviour accurately.

**New findings introduced by the fix:** none blocking. Two observations, not work orders (round 3, severity floor applies):
- The two run-level preset fields now have opposite kwargs-copy policies: `trigger_profile` still lets a `plan.kwargs` copy win by `setdefault` (`presets.py:273`), while `native_image_save` refuses one — Codex's "the preset field is the one source of truth" applies to both. The JS comment documents the asymmetry rather than resolving it. Fine to leave; if it is ever unified, refusal is the one to keep.
- The error text in `geecs_bluesky.qs_client` names one client's widget ("the scanner's 'LabVIEW files' control"). `qs_client` is the seam every client uses, so a relabel of that control leaves the message stale with nothing to pin it; "set it at the top level of the preset (`native_image_save`)" alone would say enough.

---

## Dispositions (round 3)

| Finding | Disposition |
|---|---|
| Codex: a `plan.kwargs` copy of `native_image_save` silently overrode the preset field | **Fixed** in e80ed667 (refused with a remedy; pinned by `test_native_image_save_in_plan_kwargs_is_refused_not_merged`), confirmed above |
| Observation: `trigger_profile` keeps the opposite policy (kwargs copy wins) | **Waived** for this PR — pre-existing behaviour the scanner already works around; a one-line follow-up outside this PR's concern. Agreed that refusal is the shape to keep when unified |
| Observation: the error text names the scanner's control label | **Waived** — the preflight report is where an operator most likely reads it, so naming the control helps them; a relabel would touch the scanner's CHANGELOG, where a grep for the label finds this string |

Three rounds total: two by the stopping rule, plus one for the external finding. This closes the review.

skbarber:
## Hardware verification — 2026-09-21, 26_0921 Scans 003–007 (laser off, `HTU-LaserOFF`)

Deployed from `~/qs-checkout` at master `4d915eb0` (this PR's merge): `poetry install` for GeecsBluesky (`ca tiled qserver optimize`) and GeecsScanner, then `geecs-qserver` + `geecs-scanner` restarted. A **CA gateway restart** was also needed: the Point Grey devicetype had gained `UDPtrigger` and `AutoTriggerLevelMaxCounts` rows since the gateway's 2026-09-16 start, and its roster is startup-only, so the worker's fresh namespace refused `UC_ALineEbeam1` at connect (Scan001, nothing shot). After the gateway restart the manager's `count`/`sweep` signatures carry `native_image_save`; the scanner serves 0.13.0 with the `LabVIEW files` control.

All scans: strict `count` on `UC_ALineEbeam1` (plugin host 192.168.7.161), submitted through the scanner's preflight + submit API.

| Scan | Item / defaults | Start doc `native_image_save` | `UC_ALineEbeam1/` | `geecs-pva-gateway diff` |
|---|---|---|---|---|
| 003 | item `false` | `False` | `UC_ALineEbeam1.h5` only — stack `(5, 964, 1288)` uint16, 5 `frame_acq_timestamp`s, **0 PNGs** | `capture_only (stack_only=5)` |
| 004 | item unset, defaults file silent | `True` | stack + **5 PNGs** | `pass (matched=5 identical=5)` |
| 005 | item unset, defaults file `native_image_save: false` (uncommitted edit) | `False` | stack, 0 PNGs | — |
| 006 | item unset, defaults file reverted | `True` | stack, 5 PNGs | — |
| 007 | item `false`, devices `UC_ALineEbeam1` + `UC_Stretcher_MI` (no plugin on its box) | `False` | ALineEbeam1: stack, 0 PNGs; **Stretcher_MI: 3 PNGs, no stack** | — |

005 → 006 had **no worker restart** between them: the defaults file is read at every run, as designed. Journal lines per run: `native saving off (file-plugin stacks only) this run for UC_ALineEbeam1` (003, 005, 007) and `native saving on this run for UC_ALineEbeam1` (004, 006); the scanner's queue rows read `· no LabVIEW files` for 003 and 007. The s-file carries the camera's scalar columns and stamp in every case; it never carries `-nonscalar_save_path` (a companion column), so PNG presence and the diff verdict are the on-disk record.

**Two operational findings, not defects of this PR:**

1. **Laser-off strict scans need a frame at arm time.** The file plugin acknowledges `Capture=1` only after one frame; the PVA gateway's image subscription is gated (`:connected` reads `Idle` until a client monitors), and with the DG645 quiet in ARMED (`Single shot`) no frame comes within the 8 s wait — Scan002 failed there, and a gated count fails the same way because its arm also precedes the box going to SCAN. The #894 held-frame arm only helps once a frame has reached the gateway. Seeded here by holding a PVA monitor on the image and firing one shot (`mv U_DG645_ShotControl.trigger_source "Single shot"`, `mv …trigger_executesingleshot "on"`); every later scan armed instantly. Worth a note in the laser-off procedure, or an `ARMED` state on `HTU-LaserOFF` that lets the box free-run.
2. **`U_BCaveMagSpec` is now plugin-backed** (its host joined `file_plugin_addr_list`) while its stitched image is not on the PVA wire, so the switch **off** would leave it with no record until the non-scalar PVA arc gives it a captured stream. Keep the stitcher out of switch-off presets until then.

Deploy note for the record: `/health` on the scanner kept reporting 0.11.0 until the unit restarted (the metadata-on-disk trap), and the scanner's `poetry.lock` still pins geecs-bluesky 0.94.0 / geecs-schemas 0.29.0 — cosmetic, both are editable path installs and import the checkout's code; a relock can ride the next scanner PR.

PR #945, opened 2026-09-21
GEECS-Core 0.9.0 + GeecsBluesky 0.97.0: capture streams declared per devicetype (non-scalar PVA arc, PR 2)

skbarber:
PR 2 of the non-scalar-over-PVA arc (brief: `Planning/data_capture/02_nonscalar_over_pva.md` §4.3/§4.3b on this base). **Base is `feature/nonscalar-pva`**, the arc's integration branch, per the owner's ruling that the arc lands on `master` once; PR 1 (bootstrapping the three hosts) was deployment only and is recorded in the brief.

## What

The worker picked which variable a plugin-backed camera captures by a positional guess (`primary_image_variable`: `image` if the DB lists it, else the alphabetically first image variable). That rule is right for Point Greys and wrong for everything else in the arc:

- **FROG** resolves to `SpatialImage`, which the device never pushes, so every `prepare` waits out the arm timeout; the variable it does push, `frogTrace`, is not captured.
- **MagSpecCamera** captures `Image` but not `ImageInterp`; analysis wants both.

This PR replaces the guess with a **per-devicetype declaration** in GEECS-Core, `geecs_core.db.device_streams`: `capture`, an allowlist in capture order — nothing is captured unless named. (The serving-side `exclude` polarity from the brief was in the first commit and was **dropped after review**: it had no consumer on this branch and its Point Grey entry would have removed served PVs the documented criteria did not cover. It lands with PR 3, beside the gateway code that reads it.) Names resolve against a device's DB rows case-insensitively and come back in the DB's spelling; a declared name the DB does not list is dropped with a WARNING, never armed. A devicetype with **no entry returns `None`**, and the worker keeps today's guess for it — so every camera type nobody has declared behaves exactly as before.

**One folder per stream** (added in review round 2, Codex's P1): the primary stream keeps `<device>/<device>.h5`; a second stream of the same device writes the sibling `<device>-<variable>/<device>-<variable>.h5`, the layout the LabVIEW-native files use for a device's second output, which `scan_stack.find_stack_file` already resolves. Before the fix both plugins of a device were handed one path, and each gateway writer opens its file `"w"`, so the second truncated the first.

`namespace.capture_streams(rows, devicetype)` consumes it: declared streams, restricted to what the gateway serves today (image-typed variables); a declared `1darray` stream (the magspec lineouts) is logged at INFO and waits for PR 3, with no declaration change then. The detector build and the optimizer's live frame source both go through it.

Entries (all established live 2026-09-16..20, recorded in the module docstring): Point Grey Camera → `image`; MagSpecCamera → `Image`, `ImageInterp`, `interpSpec`, `interpDiv`; MagSpecStitcher → `Image`, `interpSpec` (its `interpDiv` arrives malformed and is left out); FROG → `frogTrace` (the device never pushes `SpatialImage` or the retrieved traces); PicoscopeV2 → captures nothing yet (its channels are armed per instance once arrays are capturable).

## Per-concern LOC

| concern | files | +/− |
|---|---|---|
| The declaration + resolver (GEECS-Core) | `db/device_streams.py` | +155 |
| The shared case-insensitive row index (also used by `scalar_attribute_variables`, same answers) | `db/variable_types.py` | +19/−9 |
| Its parity tests + the recorded fixture + the recorder (`GeecsDb.get_devicetype_variables` + `scripts/record_devicetype_variables.py`) | `tests/test_device_streams.py`, `tests/fixtures/devicetype_variables.json`, `db/geecs_db.py`, `tests/test_geecs_db.py`, `scripts/` | +117, +1355 (JSON: name/type rows of the five devicetypes, read from the DB 2026-09-21), +27, +19, +62 |
| Worker consumes it | `namespace.py`, `optimization/measurements.py`, `devices/detector.py` (docstring), `db_runtime.py` (degraded-path text names the second consumer) | +49/−8, +5/−2, +4/−3, +12/−7 |
| One folder per stream | `devices/hdf_plugin.py` (`PluginPathProvider(variable=)`, `stem`) | +33/−9 |
| Worker tests | `tests/test_namespace.py`, `tests/test_native_optimization.py` (FROG-typed fake roster pins the optimizer wiring), `tests/test_hdf_plugin_detector.py` (two-stream prepare/resource test, secondary provider case) | +127/−2, +21/−4, +101/−0 |
| Docs + versions | both CHANGELOGs, `GeecsBluesky/CLAUDE.md`, `GEECS-Core/DESIGN.md`, pyprojects | +64/−6 |

1. **Point Grey Camera has an entry (`image`) rather than relying on the fallback.** Same answer, pinned by a test against the recorded rows; the fallback is for devicetypes nobody has looked at.
2. **The per-instance gate (Picoscope `Enable.Ch<X>`), the padding ceiling and the `exclude` list from the brief's table are all deferred to PR 3**, when arrays become capturable and the gateway is the consumer — no dead schema here.
3. **`GeecsDb.get_devicetype_variables` is a new public query** (type-level rows, no instance merge) so the fixture recorder is one documented command rather than reverse-engineered SQL.

Facility-literal check: the only `192.168.` literals added are host strings inside tests, the pattern the existing `#806` tests already use.

## Tests

- GEECS-Core: **146 passed, 1 skipped** on f3f36dd4 (18 new in `test_device_streams.py`, 1 in `test_geecs_db.py`); `test_device_streams.py` 23 passed on e46d555d with the folder-name check added.
- GeecsBluesky: **731 passed, 4 skipped** on e46d555d (full suite, local; CI reruns it).
- GeecsPvaGateway (the other consumer of the refactored `scalar_attribute_variables`): 48 passed locally; its 3 local failures are `ModuleNotFoundError: ophyd_async` in the local env and fail identically with the unmodified GEECS-Core — CI's gateway leg is the real check.
- The new tests were shown to bite: misspelling `frogTrace` in the table → 3 failures; adding a table entry with no recorded fixture → 2 failures; making the worker ignore the declaration → the FROG and MagSpec namespace tests fail. Restored, green.
- `scripts/check.sh` fails environment-shaped in a worktree (no Poetry envs there); both suites were run with the main checkout's package envs against this worktree's source (`PYTHONPATH`), markers as CI (`not integration`, `not gui`).

Deploy = worker restart (`geecs-qserver`), then add `192.168.6.73` (the FROG host, bootstrapped in PR 1 but deliberately left out of `[pva] file_plugin_addr_list` until this lands) and restart once more.

- **FROG**: a strict scan including `U_FROG_Grenouille` writes a `u_frog_grenouille` stack of `(N, 576, 768)` uint8 and `prepare` no longer times out. Also the brief's Q3: the trace has real content when the FROG is actually acquiring (every pixel was 0/1 in the no-laser probe).
- **MagSpec camera**: `UC_BCaveMagSpecCam1` writes two stacks in **two folders** — `Image` `(N, 160, 719)` as `UC_BCaveMagSpecCam1/UC_BCaveMagSpecCam1.h5` under the `<name>` stream key, and `ImageInterp` `(N, 189, 1)` as `UC_BCaveMagSpecCam1-ImageInterp/UC_BCaveMagSpecCam1-ImageInterp.h5` under `<name>-imageinterp` — both readable through Tiled and `scan_stack`, and the INFO line names `interpSpec`/`interpDiv` as waiting.
- **Stitcher** (`U_BCaveMagSpec`, host already in the list): one stack on `Image`, as today.
- **Regression**: any Point Grey scan writes exactly the one `image` stack it writes today.

Review record: fresh-context adversarial review (7 findings, all fixed in f3f36dd4 and confirmed), Codex review (1 P1, fixed in e46d555d, verified by a fresh-context verifier) — all as comments below.


skbarber:
## Adversarial review (fresh-context reviewer, commit 0c12352e)

Diff: `feat/capture-stream-declaration` vs `feature/nonscalar-pva` — GEECS-Core 0.9.0 (`geecs_core/db/device_streams.py` + parity test + 1355-line recorded fixture) and GeecsBluesky 0.97.0 (`namespace.capture_streams`, wired into `_build_device` and the optimizer's frame source).

The core change is sound: the FROG now arms on `frogTrace`, MagSpec cameras get two plugins in declared order, undeclared devicetypes keep the old guess, and every behaviour I mutated (case-insensitive resolution, DB-spelling return, table/fixture parity, capture order, the image-only filter, the `None`→default fallback) made a test fail. Findings below are what survived my attempts to refute them.

### 1. P2 — The `exclude` polarity ships with no consumer, and its contract text is false on this branch

`GEECS-Core/geecs_core/db/device_streams.py:1-40, 57-74, 82-85, 175-185`

- `excluded_variables` / `DeviceTypeStreams.exclude` are imported by nothing outside `tests/test_device_streams.py` (grepped the whole repo for `excluded_variables|device_streams|streams_for|DEVICE_TYPE_STREAMS`). `GeecsPvaGateway/geecs_pva_gateway/config.py` still serves `image_variables(rows)` unfiltered.
- The docstring is written in the present tense — "Variables never served over PVA and never captured" — yet on this branch the PVA gateway serves `bakground image` / `processed image` on every Point Grey (`GeecsPvaGateway/tests/test_server.py:353` even exercises `bakground image` as a served variable). A reader of GEECS-Core 0.9.0 is told something the fleet does not do. The CHANGELOG's "the PVA gateway adopts `exclude` ... with array support" is future tense; the module text is not.
- The Point Grey entry contradicts the module's own exclusion criteria. Lines 7-11 say serving is "cheap and gated" and exclusions are for "a variable that pushes one number under an array type, a twin nobody needs, a name the device never publishes". `bakground image` / `processed image` are none of those — they are real images pushed when an operation produces them (the PVA CLAUDE.md's own example of why gating makes them free: "Watching `image` never costs anything for `processed image`"). The real cost that would justify excluding them — one `HdfFilePlugin` PV set + writer thread per served image variable, `GeecsPvaGateway/geecs_pva_gateway/server.py:182-187`, i.e. ~80 extra plugin instances across 40 Point Greys — is not stated anywhere. Concrete scenario on adoption: the next PR implements the declaration verbatim, and two PVs per Point Grey that an operator can monitor today (`<expt>:<cam>:processed_image`) vanish from a namespace the PVA CLAUDE.md calls a contract, with the documented rationale not covering them.

Suggested disposition: either drop `exclude` from this PR and add it with its consumer (the "one change, old path deleted" rule cuts both ways — don't ship a contract ahead of behaviour), or keep it and (a) write the docstring in the future tense / name the pending consumer, (b) state the plugin-per-variable cost as the fourth exclusion category, (c) decide explicitly whether the Point Grey pair is excluded from *serving* or only from *capture* (today it is only ever the latter, via the allowlist).

### 2. P2 — `GeecsDbDeviceTypes`' degraded-path contract is now stale, and the silent regression is the very defect this PR fixes

`GeecsBluesky/geecs_bluesky/db_runtime.py:136-148, 182-188`; `GeecsBluesky/geecs_bluesky/namespace.py:176-177, 593-603`

The docstring says "the one consumer is `looks_triggerable`" and the WARNING an operator reads at environment open enumerates the consequences of an empty types map (trigger sources misclassified, the stitcher not triggered, #934). `capture_streams` is now a second consumer of `roster.types`, and on the degraded path `devicetype == ""` → `streams_for("")` → `None` → `primary_image_variable` → the FROG plugin-backs on `SpatialImage` again and waits out its arm timeout on every `prepare`; a MagSpec camera silently drops `ImageInterp`. Nothing logs this — the degraded run looks like the pre-0.97.0 world. The root CLAUDE.md's "contract files travel with behaviour" applies: add the two consequences to the docstring and the warning text (one sentence each). Bounded because the map is only empty on a DB failure the worker already warns about.

### 3. P3 — The optimizer's devicetype wiring is not pinned

`GeecsBluesky/geecs_bluesky/optimization/measurements.py:207-210`; `GeecsBluesky/tests/test_native_optimization.py:221-227`

Mutation: replaced `namespace.roster.types.get(diag.name, "")` with `""` → `tests/test_native_optimization.py` still passes 10/10. The only test roster has `types={}` and a camera whose sole image variable is `image`, so the declared and default paths give the same answer. A one-line addition (`types={"Camera": "FROG"}` with `frogTrace` + `SpatialImage` rows, asserting the source PV ends in `:frogtrace`) would make the CHANGELOG's "the optimizer's live frame source reads the device's first declared stream" a tested claim.

### 4. P3 — Admission argument overstated; record the rationale where DESIGN.md keeps them

`GEECS-Core/geecs_core/db/device_streams.py:38-40`; `GEECS-Core/DESIGN.md:24-28`

The docstring says "the worker ..., the PVA gateway ... and the CA gateway all need one answer". The CA gateway never touches non-scalars — it drops `_SKIP_VARTYPES` at `GeecsCAGateway/geecs_ca_gateway/config.py:189` and has no exclusion machinery to feed. The real argument is the `scalar_policy` precedent (worker + PVA gateway, neither may import the other), and on this diff only the worker imports it. Placement in GEECS-Core is defensible **if** the PVA PR lands; if it does not, this is a per-devicetype table that belongs beside `TRIGGERED_DEVICETYPES` / `TRIGGER_SOURCE_DEVICETYPES` in `namespace.py`. Either fix the CA-gateway claim and say "pending the PVA gateway's adoption", or record it in DESIGN.md's rule 3 the way `db/alarms.py`'s exception is recorded.

### 5. P3 — The re-record instruction is not executable as written

`GEECS-Core/geecs_core/db/device_streams.py:77-80`; `GEECS-Core/tests/fixtures/devicetype_variables.json` (`_recorded`)

"Adding an entry means recording that devicetype's rows into `tests/fixtures/devicetype_variables.json`" and "Re-record when a devicetype's variables change" — but there is no script, no documented query, and `GeecsDb` has no per-devicetype variables method (`get_experiment_device_variables` is per experiment device and already merged with instance rows, which is not the `devicetype_variable` shape the fixture claims). Checked `scripts/`, `GEECS-Core/`, `.claude/skills/`. A contributor adding `ThorlabsWFS` has to reverse-engineer the SQL from `geecs_db.py`. A 20-line `scripts/record_devicetype_variables.py` (or the SQL in the fixture's `_recorded` note) closes it. The fixture itself scanned clean for lab literals (no hostnames, accounts, home paths).

### 6. P3 — `_resolve` re-implements a pattern already in the same package

`GEECS-Core/geecs_core/db/device_streams.py:122-148` vs `GEECS-Core/geecs_core/db/variable_types.py` (`scalar_attribute_variables`, the `by_lower` block)

Both build `by_lower = {name.lower(): name}` via `setdefault`, look declared names up case-insensitively, return the metadata row's spelling, and skip/warn on misses. Two copies of "match declared names against DB rows case-insensitively, answer in the row's spelling" in one package is the drift the package exists to prevent. A tiny shared helper (`rows_by_lower(rows) -> dict[str, str]`) would serve both.

### 7. P3 — A second, disagreeing per-devicetype stream table already exists

`GeecsBluesky/geecs_bluesky/assets/registry.py:361-401` (`FROG_ASSETS`, `MAGSPEC_CAMERA_ASSETS`, `MAGSPEC_STITCHER_ASSETS`, keyed by the same devicetype strings in `assets/specs.py:7-14`)

It declares MagSpec camera streams as image / `interp_image` / `interpSpec` / `interpDiv`, the stitcher **including** `interpDiv` (which `device_streams` excludes), and the FROG as `Spatial` + `Temporal` (which `device_streams` says the device never pushes). It is dormant — no in-package imports outside `assets/`, and `GeecsBluesky/CLAUDE.md:79` says it goes with PNG retirement (#738) — so this is not a live conflict, but a reader searching for "which streams does a MagSpecCamera have" now finds two tables that disagree. One cross-reference line in `device_streams.py` ("the native-file layout table in `geecs_bluesky.assets.registry` describes what LabVIEW writes, not what the device pushes; it retires with #738") prevents the wrong one being extended.

### Refuted while reviewing (for the record)

- Excluding `EnergyAxis` / `AngleAxis` loses the energy axis of a captured `interpSpec` — no: the MagSpec fixture carries `EnergyMin` / `EnergyMax` / `DeltaE` and `AngleMin` / `AngleMax` / `DeltaAngle` as scalars.
- Capture/exclude overlap, capture-order drift, misspelled declarations, undeclared→`None` vs declared-empty→`[]` — all pinned; mutations bit.
- `test_camera_on_a_plugin_host_is_plugin_backed`'s new comment ("a Point Grey declares one capture stream") is accurate: the shared ROSTER types `UC_TestCam` as `Point Grey Camera` (`tests/test_namespace.py:82`), so it exercises the declared path and `test_an_undeclared_devicetype_keeps_the_one_image_guess` covers the default.
- Scan-folder-creation invariant: not touched. Public-repo hygiene: fixture and tests clean (192.168.x test IPs are the accepted form).

---
Ran from the worktree against the main checkout's envs: GEECS-Core `pytest tests -m "not integration"` → 144 passed, 1 skipped, 2 deselected (17 of them `test_device_streams.py`); GeecsBluesky `tests/test_namespace.py tests/test_native_optimization.py` → 48 passed; full GeecsBluesky `tests/` → 730 passed, 4 skipped; ruff 0.9.9 `check` + `format --check` clean on the 7 touched `.py` files; pydocstyle clean under the repo's ignore set; 7 mutations applied and reverted (6 caught by tests, 1 — the optimizer devicetype argument — not); `git status --porcelain` empty afterwards.

---

## Disposition (author), fix commit f3f36dd4

|---|---|---|
| 1 | P2 | **Fixed** — `exclude` / `excluded_variables` and their entries removed from this PR; the module docstring now says the serving-side exclusion list lands with the gateway's array support beside its consumer. PR body's judgment-call list updated. The Point Grey serving question (plugin-per-variable cost vs. operator-visible PVs) is PR 3's to decide, where the gateway code is in the diff. |
| 2 | P2 | **Fixed** — `GeecsDbDeviceTypes` docstring names `capture_streams` as the second consumer and both consequences (FROG → `SpatialImage` arm timeout; MagSpec drops `ImageInterp`); the WARNING text carries the same sentence. |
| 3 | P3 | **Fixed** — `test_camera_measurement_refire_and_preclaim_connection`'s fake roster is now a FROG-typed camera with `SpatialImage` + `frogTrace`; the fake frame source records its PV and the test asserts `["test:camera:frogtrace"]`. The reviewer's mutation (dropping the devicetype argument) now fails this test. |
| 4 | P3 | **Fixed** — CA-gateway claim removed; the docstring states the worker-today / PVA-gateway-with-arrays pair on the `scalar_policy` precedent, and `DESIGN.md` rule 3 records the admission beside the `db/alarms.py` exception. |
| 5 | P3 | **Fixed** — `GeecsDb.get_devicetype_variables(devicetype)` (type-level rows, no instance merge; tested against the fake connector) + `GEECS-Core/scripts/record_devicetype_variables.py`; the module docstring and the fixture's `_recorded` note name the command. |
| 6 | P3 | **Fixed** — `variable_types.rows_by_lower(rows)` is the one index; `scalar_attribute_variables` and `capture_variables` both use it (GEECS-Core suite 146 passed; the PVA gateway's suite, the other consumer, unchanged: 48 passed, 3 local failures are a missing `ophyd_async` in the local env and identical against the unmodified GEECS-Core). |
| 7 | P3 | **Fixed** — cross-reference in the module docstring: the assets registry describes LabVIEW's native files, not what the device pushes, and retires with #738. |

Confirmation of each fix by the same reviewer follows.

skbarber:
## Confirmation round (same reviewer, fix commit f3f36dd4)

Verified independently against `git diff 0c12352e...f3f36dd4` in the worktree; each disposition below is from the code/tests/greps, not the author's summary.

1. **Confirmed.** `exclude`, `excluded_variables` and `_resolve` are gone from `GEECS-Core/geecs_core/db/device_streams.py` (dataclass is `capture` only; table entries are bare `capture=` tuples) and from `tests/test_device_streams.py`. The module docstring now puts the serving-side list in the future tense ("arrives with the gateway's array support, beside its consumer, not here ahead of it"); the FROG rationale was also corrected to say `SpatialImage` "is an alignment view" rather than "stays empty". Residue grep for the identifier, the prose noun and the path across GEECS-Core and GeecsBluesky (`excluded_variables|\bexclude\b|exclusion|_resolve\(`) finds only the two deliberate future-tense sentences (module line 16, GEECS-Core CHANGELOG line 28); every other hit is unrelated pre-existing code (`assets/readback.py`'s `exclude=` parameter, `qs_client/presets.py`'s own `_resolve`). The Planning hunk agrees (exclude dropped, lands with PR 3).

2. **Confirmed.** `GeecsBluesky/geecs_bluesky/db_runtime.py:141-153` now names both consumers (`looks_triggerable` and `capture_streams`) and both consequences (the #934 misclassification; the FROG arming on `SpatialImage` with an arm timeout per `prepare`, a MagSpec camera dropping `ImageInterp`); the WARNING at 188-195 carries the second consequence too. GeecsBluesky CHANGELOG records it.

3. **Confirmed.** `tests/test_native_optimization.py:221-236` builds a `FROG`-typed roster with `SpatialImage` + `frogTrace`, the fake source records `args[0]`, and line 285 asserts `sources == ["test:camera:frogtrace"]`. Re-applied my mutation (`namespace.roster.types.get(diag.name, "")` → `""` in `measurements.py:209`): **2 failed, 8 passed** (both parametrizations of `test_camera_measurement_refire_and_preclaim_connection`), where before the fix it was 10 passed. File restored; `git status --porcelain` empty.

4. **Confirmed.** The "worker, PVA gateway and CA gateway all need one answer" sentence is gone; the docstring (line 39-43) now states the `scalar_policy` precedent and that "the CA gateway serves scalars only and never reads it" — the only CA-gateway mention in the module. `GEECS-Core/DESIGN.md:62-66` records the admission under rule 3 beside the `alarms.py` exception, in the same form.

5. **Confirmed.** `GeecsDb.get_devicetype_variables(devicetype)` at `geecs_db.py:598-624` uses the same join shape as the existing live-verified per-device query (`devicetype_variable dv LEFT JOIN choice c ON c.id = dv.choice_id WHERE dv.devicetype = %s`, cf. `dtv.devicetype` at line 517). `tests/test_geecs_db.py::test_get_devicetype_variables_is_type_level_only` drives it through `_patch_rows`, which fakes the connection *below* `_query`, so the real SQL and `("FROG",)` params are asserted (and `"variable v" not in query` pins "no instance merge"). `GEECS-Core/scripts/record_devicetype_variables.py` is new (dir absent on base), lint-covered (ruff exclude is only `extras/**`/third-party), runs `--help` offline, imports without touching the MySQL connector, and its `json.dumps(..., indent=1) + "\n"` **round-trips the committed 1355-line fixture byte-for-byte**, so a re-record produces no spurious formatting diff. The module docstring names the command and the fixture's `_recorded` note names the script.

6. **Confirmed.** `variable_types.rows_by_lower(rows)` (lines 118-134) is the one index; call sites are exactly `scalar_attribute_variables` (line 180) and `capture_variables` (device_streams line 128) — nothing else grew its own. Behaviour is unchanged: both still take the *first* row spelling a lowered name via `setdefault` and answer with that row's spelling. Its production consumer is exercised by the PVA gateway's `test_config.py` and `test_server.py`, which pass against the worktree's GEECS-Core (19 passed).

7. **Confirmed.** `device_streams.py:44-49` cross-references `geecs_bluesky.assets.registry` as the LabVIEW-native-files table (spatial + temporal FROG image, the stitcher's `interpDiv` TSV), states it retires with #738, and says "Extend this table, not that one."

None that survive. Two things I checked and dismissed: (a) the PVA gateway run shows 3 failures in `test_file_plugin.py`, but all three are `ModuleNotFoundError: No module named 'ophyd_async'` and fail identically with the main checkout's own GEECS-Core on the path — a missing dev extra in that env, not this diff; (b) `WHERE dv.devicetype = %s` under MySQL's case-insensitive collation would let `record_devicetype_variables.py frog` record a second fixture key beside `FROG` — harmless (the parity test lowercases keys and sets dedupe), a question not a work order.

---
Ran from the worktree at f3f36dd4 against the main checkout's envs: GEECS-Core `pytest tests -m "not integration"` → 146 passed, 1 skipped, 2 deselected; GeecsBluesky `tests/test_namespace.py tests/test_native_optimization.py` → 48 passed; full GeecsBluesky `tests/` → 730 passed, 4 skipped; GeecsPvaGateway `test_config.py test_file_plugin.py test_server.py` with the worktree's GEECS-Core → 19 passed, 3 failed (pre-existing `ophyd_async` env gap, identical without the worktree); ruff 0.9.9 `check` + `format --check` and pydocstyle (repo ignore set) clean on all 8 touched `.py` files including the new script; 1 mutation (finding 3) applied → 2 failed → reverted; `git status --porcelain` empty afterwards, HEAD unchanged.

---

**Author:** review closed — 7/7 confirmed, no new findings; the reviewer's dismissed item (b) is noted as a question, not acted on. CI on f3f36dd4: all nine check runs green. Ready for the maintainer's merge into `feature/nonscalar-pva`.

skbarber (commented):
I found one blocking data-loss issue in the new multi-stream path. The declaration and focused selection tests otherwise look sound.

GeecsBluesky/geecs_bluesky/namespace.py
```diff
@@ -558,12 +590,13 @@ def _build(
...
-            # LabVIEW-native path stays on beside it — PNG dual-write until
-            # PNG retirement (#738), the parity evidence of the rollout.
+            # Plugin-backed iff the devicetype's capture streams include a
+            # served (image) variable and the device's camera server serves
+            # the file plugin (#806).  The LabVIEW-native path stays on
+            # beside it — PNG dual-write until PNG retirement (#738), the
+            # parity evidence of the rollout.
             plugin_vars = (
-                primary_image_variable(rows)
+                capture_streams(rows, devicetype)
```
skbarber, line 599: [P1] Give each capture stream a distinct output file

This can now return both `Image` and `ImageInterp`, but `_build` constructs each downstream `PluginPathProvider` with only the device name. That provider ignores the data key and always returns `<device>/<device>.h5`. The two independent gateway writers therefore arm against the same path; each opens it with `h5py.File(..., "w", locking=False)`, so the second open truncates/races the first and both stream-resource documents point at the same URI and dataset. The advertised two stacks cannot be preserved. Please include the variable (at least for secondary streams) in the filename/URI and add a two-plugin prepare/resource test that asserts distinct paths.

skbarber:
## Disposition of the Codex review (P1: give each capture stream a distinct output file) — fix commit e46d555d

**Confirmed and fixed.** The mechanism was exactly as described: `PluginPathProvider.__call__` ignored its datakey and returned `filename=<device>`, so `_build` armed both of a device's plugins on `<device>/<device>.h5`; the gateway writer opens its path with `h5py.File(..., "w", locking=False)`, so the second stream to arm truncated the first, and both stream resources carried one URI.

**The fix — one folder per stream.** `PluginPathProvider` gains `variable=`. The primary stream keeps `<device>/<device>.h5` unchanged (every Point Grey, every existing reader). A second stream of the same device writes the sibling folder `<device>-<variable>/<device>-<variable>.h5` — the layout the LabVIEW-native files already use for a device's second output (`U_BCaveMagSpec-interpSpec/`, `U_FROG_Grenouille-Temporal/`), so `geecs_data_utils.io.scan_stack.find_stack_file` resolves it with no read-side change (it looks for `<folder>/<folder>.h5`), and the portal lists it the way it lists those folders today. The namespace passes `variable=var` for every stream after the first. The stream key stays `<name>-<variable>` (lowercased), so for a MagSpec camera: data key `uc_bcavemagspeccam1-imageinterp`, file `UC_BCaveMagSpecCam1-ImageInterp/UC_BCaveMagSpecCam1-ImageInterp.h5`.

One deliberate divergence to note: the folder takes the variable's **DB spelling** (`-ImageInterp`), where the legacy native folder for that camera's interpolated image is `-interp` (an asset-registry suffix that retires with #738). PR 3's array streams (`interpSpec`, `interpDiv`) land in `-interpSpec/` / `-interpDiv/`, which *are* the legacy names.

**Tests added (as requested):**
- `test_hdf_plugin_detector.py::test_two_capture_streams_write_two_files` — a MagSpec-shaped detector (`Image` + `ImageInterp`) through stage → prepare → shot → `collect_asset_docs`: distinct `FilePath` (`…\Scan001\UC_TestCam` vs `…\Scan001\UC_TestCam-ImageInterp`), distinct `FileName`, both folders created inside the claimed scan folder, two `stream_resource` docs with distinct URIs ending `UC_TestCam/UC_TestCam.h5` and `UC_TestCam-ImageInterp/UC_TestCam-ImageInterp.h5`, and `describe()` carrying both shapes.
- `test_plugin_path_provider_hands_out_both_paths` — the secondary-stream provider case (`stem`, Windows path, URI, distinct from the primary).
- `test_namespace.py` MagSpec test — the namespace wiring: the two `ADHDFDataLogic` path providers' stems are `["UC_MagSpecCam", "UC_MagSpecCam-ImageInterp"]`.
- GEECS-Core `test_declared_names_are_safe_folder_name_components` — every declared capture name is a legal Windows folder-name component (no `<>:"/\|?*`, no edge whitespace).

Shown to bite: reverting the provider's stem to `device` fails 3 tests (the two-stream test, the provider test, the namespace test); reverting the namespace to pass no variable fails the namespace test. Restored, green.

Counts on e46d555d: GeecsBluesky full suite **731 passed, 4 skipped**; GEECS-Core `test_device_streams.py` 23 passed (full suite reruns in CI).

Codex cannot be resumed from this side, so a fresh-context verifier given only the finding text and the fix commit confirms it below (the substitution the review process allows when the original reviewer is unreachable). Hardware note for the OWED section: with this layout the MagSpecCam1 check reads two stacks in **two folders**, `UC_BCaveMagSpecCam1/` and `UC_BCaveMagSpecCam1-ImageInterp/`.

skbarber:
## Verification (fresh-context verifier, fix commit e46d555d)

**Verdict: confirmed.** Two plugins of one device now arm on distinct `FilePath`/`FileName` and emit distinct stream-resource URIs through the stock `ADHDFDataLogic`; the primary stream's path is byte-for-byte unchanged; the new tests fail when either half of the fix is reverted.

### 1. End to end through the stock data logic
Read `ophyd_async/epics/adcore/_data_logic.py` (0.19.3): `ADHDFDataLogic.prepare_unbounded` calls `self.path_provider(datakey_name)`, then `prepare_file_paths` sets `FilePath = f"{directory_path}\\"`, `FileName = path_info.filename`, `FileTemplate = "%s%s.h5"`, and the provider is returned with `uri = f"{directory_uri}{filename}.h5"`. All three derive from `PluginPathProvider._stem`, which is `device` or `f"{device}-{variable}"`, so a second stream differs in directory, filename and URI at once. `GeecsDetector` builds one `ADHDFDataLogic(path_provider=provider, …)` per `hdf_plugins` entry (`detector.py:689-711`) — the test's `cam.hdf` / `cam.hdf_imageinterp` signals and `collect_asset_docs()` are that stock path on mock backends. Gateway side (`file_plugin.py:609-641, 798`): `_open_file` opens `os.path.join(FilePath, f"{FileName}.h5")` with `h5py.File(..., "w", locking=False)`; distinct paths, so the second arm no longer truncates the first.

### 2. Primary path unchanged; read side resolves the sibling
`variable=None` → `_stem = device` → `<device>/<device>.h5`, and `namespace._build` passes `None` only at index 0. Index 0 is the first *served* declared stream (`capture_streams` = declaration order filtered to image variables); the table lists `Image` first for `magspeccamera` and `magspecstitcher`, and Point Grey/FROG declare one stream, so the primary is the historic one for every declared type. Read side: `find_stack_file(device_dir)` is `device_dir / f"{device_dir.name}.h5"` — unchanged, resolves `<device>-ImageInterp/<device>-ImageInterp.h5`. Consumers: ScanAnalysis (`scan_directory / data_device_name` → `find_stack_file`; its `_acq_timestamp_column` keys on `device_name`, so a v2 diagnostic with `scan.device` = the sibling folder joins by the device's stamps), the portal (`device_kind` lists every subfolder and calls `find_stack_file(scan_folder / device)`), `callbacks.py` (stacks keyed by `data_key`, URI stored, nothing derives a device from the path — pre-fix it would have held two `_Stack`s on one URI), logbook (no stack reads). Nothing in the repo assumes one stack per device folder.

### 3. The tests bite (each mutation restored with `git checkout --`; imports verified to resolve to the worktree source)
- `hdf_plugin.py` stem always the device → **3 failed / 49 passed**: `test_two_capture_streams_write_two_files`, `test_plugin_path_provider_hands_out_both_paths`, `test_a_magspec_camera_captures_both_images_and_logs_the_waiting_arrays`.
- `namespace.py` `variable=None` for every stream → **1 failed / 51 passed**: the MagSpec test at `stems == ["UC_MagSpecCam", "UC_MagSpecCam-ImageInterp"]` (line 519).
- GEECS-Core table `"frogTrace"` → `"frog:Trace"` → `test_declared_names_are_safe_folder_name_components[FROG]` fails on its own (plus 4 parity tests).

### 4. What the fix touches
- **Scan-folder invariant:** `PluginPathProvider.__call__` still raises `FileNotFoundError` when `local_dir.parent` (the scan folder) is absent and uses `mkdir(exist_ok=True)` without `parents`; `GeecsScanPathProvider.__call__` builds a `PathInfo` only. The sibling folder is created inside the claimed scan folder under the same rule.
- **Gateway constraints:** `FilePath` exists (the provider creates it; the plugin's host sees it through the UNC mapping), `FileName` is never empty (a device name, or `device-variable`), `FileTemplate` is the stock `%s%s.h5`. Satisfied for both streams.
- **Windows validity:** stems are `<device>-<DB spelling>`; the DB spelling differs from the table's only by case (`rows_by_lower`), so the new GEECS-Core pin (no `<>:"/\|?*`, no edge whitespace) covers the stem. Current names: `Image`, `ImageInterp`, `interpSpec`, `interpDiv`, `frogTrace`, `image`.
- **Hygiene:** no home paths, accounts or hosts added; `\\nas\hdna2\data` in the tests pre-exists at f3f36dd4 in five test files.
- **Bookkeeping:** CHANGELOG entry sits under `0.97.0` = `pyproject.toml`; the GEECS-Core change is test-only on an already-bumped 0.9.0. Ruff `check` + `format --check` clean on the five changed Python files.

### New findings
- **P3 (observation, no action needed for this PR):** the portal holds only the folder name for a sibling stack, and `tiled_schema.device_acq_timestamp_column("UC_X-ImageInterp")` looks for `uc_x_imageinterp-acq_timestamp`, which no run has (the device's column is `uc_x-acq_timestamp`). The Images tab therefore serves `<device>-<variable>` stacks by ordinal (`read_shot(stack, shot - 1)`) rather than a timestamp join — the path every `-interpSpec` sibling folder takes today, so not new, but this fix creates the first *stack* that hits it. ScanAnalysis is unaffected (keys the column on the diagnostic's `name`). A suffix-stripping fallback in `device_acq_timestamp_column` would close it when wanted.
- **P3:** the folder-safety pin checks edge whitespace and reserved characters, not internal spaces; ophyd-async's `generate_directory_uri` does not percent-encode, so a future declared name with a space would put a raw space in the stream-resource URI. No current entry has one.

**Ran:** in the worktree, `tests/test_hdf_plugin_detector.py` + `tests/test_namespace.py` → 52 passed; `GEECS-Core/tests/test_device_streams.py` → 23 passed; three reverting mutations → 3 / 1 / 5 failures as above, all restored, `git status` clean; ruff 0.9.9 check + format clean on 5 files.

---

## Disposition (author)

| finding | disposition |
|---|---|
| Codex P1 (shared file) | **Fixed** in e46d555d, **confirmed** above. |
| Verifier P3 — portal Images tab reads a `<device>-<variable>` stack by ordinal, not by stamp join | **Waived for this PR** — pre-existing behaviour for every legacy sibling folder (`-interpSpec`, `-Temporal`); the first *stack* to take that path arrives with the deploy. Noted for PR 3 / the portal: a suffix-stripping fallback in `device_acq_timestamp_column`. |
| Verifier P3 — folder-safety pin does not reject internal spaces; `generate_directory_uri` does not percent-encode | **Waived** — no declared name has a space; the pin can grow that check the day one does. |

**Review closed.** Round 1 (fresh-context adversarial): 7 findings fixed in f3f36dd4, all confirmed. Round 2 (Codex): 1 P1 fixed in e46d555d, confirmed by a fresh-context verifier. CI: all nine check runs green on e46d555d. Ready for the maintainer's merge into `feature/nonscalar-pva`.

skbarber (commented):
Approved. The P1 filename collision is fixed: the primary stream keeps `<device>/<device>.h5`, secondary streams receive distinct `<device>-<variable>/<device>-<variable>.h5` paths, and the regression test verifies separate FilePath/FileName values and stream-resource URIs through prepare and collection. Focused verification: GEECS-Core 44 passed; GeecsBluesky detector/namespace/optimization 62 passed. All CI checks are green. No remaining findings. (GitHub does not permit a formal APPROVE review on one’s own PR, so this is the approval-equivalent verdict.)

PR #946, opened 2026-09-21
Data-Utils 0.35.0 + Core 0.10.0 + PvaGateway 0.12.0 + Bluesky 0.98.0: array variables served, padded and captured (non-scalar PVA arc, PR 3a)

skbarber:
PR 3a of the non-scalar-over-PVA arc (brief: `Planning/data_capture/02_nonscalar_over_pva.md` §4.4–§4.4c on this base). **Base is `feature/nonscalar-pva`**, the arc's integration branch. Follows #945 (the capture declaration). The brief's PR 3 is split: this is the gateway/decoder half; **3b** (next) is the Picoscope capture gate, which needs a worker-side read of `Enable.Ch<X>` — a scalar the CA gateway does not serve today — and so comes with its own DB curation.

## What

Today the PVA gateway serves image-typed variables only. The devices in scope also push **arrays**: the MagSpec cameras' and stitcher's lineouts (`interpSpec`, `interpDiv`) and the Picoscope's traces. This PR makes an array a stream like an image, end to end:

1. **Decoders** (GEECS-Data-Utils 0.35.0, `io/arrays.py`). Three wire shapes, chosen by **sniffing the payload**, never by devicetype: nested `[[x,y], ...]` pairs → `(n, 2)`; CRLF CSV → `(n,)`; the LabVIEW flattened waveform (six-field header, uint32 BE count, int16 BE samples) → **volts** `(n,)` with `x0`/`dx`/`samples`/`offset`/`gain`/`name` as attributes. Every decoder raises on a payload it cannot account for byte by byte. Pinned on payloads **captured live 2026-09-21** (`tests/data/wire/`, byte-exact under `.bin`, hook-excluded): a 285-row spectrum, a 189-row divergence lineout, both axes, and a 3000-sample Picoscope trace with the internal trigger on.
2. **The declaration grows its serving side** (GEECS-Core 0.10.0). `array_variables(rows)` is the `1darray` twin of `image_variables`. `DeviceTypeStreams` gains `exclude` — **array variables the gateway never serves**: the FROG's six spectra (never carry a value), the Picoscope's dead `ScopeTraces`/`wfm`/`wfm info` and its GUI twins, the stitcher's malformed `interpDiv`, the MagSpec axes (they are the lineouts' column 0) and the Point Grey lineouts (out of scope) — and `array_ceiling` (2048 camera, 16384 stitcher, `None` = native). `served_array_variables` = typed minus excluded is the one served-set rule gateway and worker both read. **No image PV is ever excluded**; the parity test requires every exclusion to be `1darray`-typed.
3. **The gateway serves arrays** (GeecsPvaGateway 0.12.0). `CameraSpec` → `DeviceSpec` (`image_variables` + `array_variables`, `devicetype`, `array_ceiling`); `streams.py` decodes per variable and **pads arrays along axis 0 to the ceiling with NaN** so a lineout whose row count moves with the magnet current keeps one PV shape, one descriptor and one stack shape per run — longer is dropped and counted as a shape error, never truncated; the file plugin takes the worker's decoder and writes **1-D and float64 stacks** (`ArraySizeY = 0`, the stock data logic drops zero dims); a waveform's axis rides as NTNDArray attributes; the fleet roster counts array-only hosts; and **an instance with nothing to serve idles on its identity PVs** instead of exiting (the Picoscope host crash-looped under NSSM after PR 1 for exactly this).
4. **The worker captures declared arrays** (GeecsBluesky 0.98.0). `capture_streams` restricts the declaration to images **plus served arrays**, so a MagSpec camera arms four plugins in four folders (`<device>/`, `-ImageInterp/`, `-interpSpec/`, `-interpDiv/`), the lineouts as `(N, 2048, 2)` float64. A declared name that is neither is now a WARNING and skipped (the "waiting for arrays" INFO is gone with the wait).

## Per-concern LOC

| concern | files | +/− |
|---|---|---|
| Array decoders + fixtures | `GEECS-Data-Utils/geecs_data_utils/io/arrays.py`, `io/__init__.py`, `tests/test_io_arrays.py`, `tests/data/wire/*` (5 captures + README) | +166, +14/−1, +144, +20 |
| Declaration: arrays, exclusions, ceilings | `GEECS-Core/geecs_core/db/device_streams.py`, `variable_types.py`, `tests/test_device_streams.py` | +164/−59, +16, +73/−14 |
| Gateway serves arrays | `config.py`, `streams.py`, `server.py`, `file_plugin.py`, `__main__.py`, `fleet.py`, `__init__.py` | +55/−19, +69, +55/−28, +33/−9, +9/−6, +10/−4, +2/−2 |
| Gateway tests | `tests/test_arrays.py` (new: pairs padded over PVA, waveform attributes, 1-D float stack with an over-ceiling drop, config roster), renames + fakes in `test_config/test_server/test_file_plugin/test_entrypoint/test_fleet` | +392, +59/−33 |
| Worker captures arrays | `GeecsBluesky/geecs_bluesky/namespace.py`, `tests/test_namespace.py` | +14/−11, +42/−16 |
| Docs, versions, changelogs, brief, hook exclude for the fixtures | four `CHANGELOG.md`, three `CLAUDE.md`, four `pyproject.toml`, `.gitattributes`, `.pre-commit-config.yaml`, `Planning/…` | +177/−22 |

1. **`CameraSpec` → `DeviceSpec`, `.cameras` → `.devices`, and `image_variables` no longer defaults to `["image"]`.** The brief asked for the rename ("stops being an honest name"); the default change follows (an array-only device must not grow a phantom image). Every construction site in the repo is updated; nothing outside the gateway imports it.
2. **Exclusions are array-only by construction.** The round-1 reviewer of #945 flagged that excluding a Point Grey's `processed image` would remove PVs operators watch; this PR resolves it by never letting `exclude` name an image (pinned).
3. **A waveform's axis parameters go on the PV as NTNDArray attributes, not yet into the HDF5.** The per-frame attribute set is fixed at arm time and the Picoscope is not captured until 3b; 3b decides the file-side representation with the gate.
4. **Padding applies to any array of a devicetype that declares a ceiling** (1-D or 2-D, along axis 0); devicetypes without one (the Picoscope) serve native length — a scope's record length is configured, and padding a 5e6-sample record would be the wrong lever.
5. **The fleet roster counts array-only hosts** (one more batched DB query in `camera_endpoints`), so the Picoscope server gets a row on the fleet screen. The `FleetHost.cameras` field name is left alone.
6. **An empty instance idles** rather than exiting 1 — a behaviour change to the CLI, pinned by the entrypoint tests. After review round 1 the config records the **served host** (`--host`, else the lab-facing local address), so the idle instance's identity PVs carry the `addr_list` IP the fleet probe asks for, not the machine name.
7. **The fleet roster grows from 14 to 22 hosts.** Counting array-only hosts is the DB-driven served-set rule applied to the roster, so every host whose only stream devices are arrays — the four DaqPad hosts, the hexapod's, the Hamamatsu's — now shows on the fleet screen and in `/fleet-status` as *not deployed*. That is honest, and it is also eleven rows of "not deployed" nobody asked for. If you would rather not see them, the lever is a declaration entry excluding those devicetypes' arrays (each needs a recorded fixture), not a roster special case.

Facility-literal check: the only `192.168.` literals added are host strings in tests, the pattern the existing tests use; the fixture README names devicetypes, not devices or hosts.

## Tests

- GEECS-Data-Utils: `test_io_arrays.py` **27 new**, io + scan_stack suites 60 passed.
- GEECS-Core: **155 passed, 1 skipped** (27 in `test_device_streams.py`).
- GeecsPvaGateway: **56 passed** locally (3 deselected: `ModuleNotFoundError: ophyd_async` in the local env, identical on the base — CI runs them).
- GeecsBluesky: **732 passed, 4 skipped** (full suite on the first commit; namespace + plugin-detector 53 passed on the fix).
- Shown to bite: padding that truncates instead of raising → 2 fail; exclusions ignored → 3 fail across Core and gateway; the worker decoding every variable as an image → the array PV tests fail; the waveform decoder ignoring leftover bytes → 1 fails. Restored, green.
- `scripts/check.sh` fails environment-shaped in a worktree (no Poetry envs); suites were run with the main checkout's envs against this worktree's source (`PYTHONPATH`), CI markers.

Deploy order matters: **gateway fleet first, worker second** (the worker only arms what the gateway serves). Gateway = move the share clone's pin to this arc branch (or wait for the master merge), then `:restart` every instance — all twelve now, the three PR-1 boxes included; the Picoscope host's service is `DEMAND_START` and needs a manual start. Worker = `geecs-qserver` restart. `192.168.7.168` does **not** go into `file_plugin_addr_list` until 3b (nothing to capture yet).

- **MagSpecCam1**: `undulator:uc_bcavemagspeccam1:interpspec` posts `(2048, 2)` float64, column 0 the energy axis from ~51.6 MeV at 0.25 MeV steps, NaN from row ~285 on; `interpdiv` `(2048, 2)`, column 0 −23.5..23.5; no `energyaxis`/`angleaxis` PV. A strict scan writes **four stacks** in four folders, the lineouts `(N, 2048, 2)`.
- **Stitcher**: `u_bcavemagspec:interpspec` `(16384, 2)` with ~8218 real rows; no `interpdiv` PV.
- **Picoscope host**: the instance stays up with two ICTs; `u_bcaveict:scopetrace_channel0` `(3000,)` volts within ±0.1 V, attribute `dx` = 4e-9; no `scopetracegui_*` PV; a disabled channel's PV stays at its initial single zero (the device pushes an empty value, which the gateway ignores).
- **Regression**: every Point Grey stream unchanged; no `horizontallineout`/`verticallineout`/`lineouts` PVs appear.

Review record: fresh-context adversarial review (2 P2 + 5 P3; all fixed or waived with reasons in 90de2bc2, confirmation by the same reviewer) — as comments below. Read-side debt recorded in the brief under PR 4: the portal's Images tab on a lineout stack, and 3b's file-side waveform axis.


skbarber:
## Adversarial review (fresh-context reviewer, commit 03a4ff49)

Scope: `git diff feature/nonscalar-pva...HEAD` (41 files), read against the root and package `CLAUDE.md`s, `GEECS-Core/DESIGN.md`, `GeecsPvaGateway/DEPLOYMENT.md`, the p4p and ophyd-async sources in the package venvs, and the consumers of the written layout (portal, parity tool, scan-path readers). Everything below survived my own attempt to refute it; the refuted items are listed at the end so nobody re-checks them.

### Findings

**1. P2 — The "idle instead of exit" change does not deliver the visibility it is justified by.**
`GeecsPvaGateway/geecs_pva_gateway/__main__.py:71-80` now keeps a device-less instance running so "the fleet screen sees it, and picks up the host's devices on the next restart". But `server.py:510-514` (`_instance_host`) names the identity PVs after `self._config.devices[0].host`, falling back to `socket.gethostname()` when `devices` is empty — and `PvaGatewayConfig` carries no host, `launch.bat:46` passes only `--experiment`. The fleet probe (`fleet.py:59-60`, `instance_pv_prefix(experiment, self.ip)`) asks `{exp}:pvagateway:<addr_list IP>:version`; the idle instance publishes `{exp}:pvagateway:<hostname>:version`. Scenario: a scope server is bootstrapped before its Picoscope is enabled in the DB → the service idles under the hostname prefix → `geecs-pva-gateway fleet` and the Phoebus screen show `[DOWN]` (timeout) for a live process, and `pvput {exp}:pvagateway:<ip>:restart 1` — the documented remote restart that would pick up the new device — finds no PV, so the operator must restart NSSM by hand. Before this diff the fallback was unreachable (no cameras → `return 1`); this diff makes it the production path for exactly the host it was written for, and no test pins the idle path (`tests/test_entrypoint.py` has none). Fix: record the served host on `PvaGatewayConfig` (the `--host` argument, else the local IP that matched in `from_geecs_experiment`) and use it in `_instance_host()`; add the test.

**2. P2 — Contract files and identifiers did not travel with the behaviour.**
- `GeecsPvaGateway/DEPLOYMENT.md:272` ("Every served **image** variable also gets the … `NDFileHDF5` PV set"), `:308` (`:connected` "Per image variable"), `:266` (`role=PVA image gateways`): arrays now get all three, and the client-facing name of a dotted array variable (`scopeTrace.Channel0` → `scopetrace_channel0`, pinned only in `tests/test_arrays.py:376`) is documented nowhere a client reads — `docs/geecs_gateway/image_pvs.md` and `GeecsPvaGateway/README.md:3,21,28` still say images only.
- `GeecsPvaGateway/CLAUDE.md:145` and `:153` still cite `CameraSpec.scalar_variables` and `_CameraWorker.subscription_variables` — both identifiers deleted in this diff (the file was edited elsewhere; identifier + prose sweep missed).
- `__main__.py:33` CLI description: "Serve this host's GEECS camera images as NTNDArray PVs."
- `GEECS-Core/DESIGN.md:62-64`: "read by the worker today and by the PVA gateway **once it serves arrays**" — now past tense, and the admission text ("which non-scalar variables a devicetype captures / never serves") does not mention the ceiling it now also carries (see 6).
- `fleet.py:148-155, 203-216` and `deploy/gen_fleet_status.py:129-130,198`: `FleetHost.cameras` / "N cameras in the DB" now count array-only devices (a scope is reported as a camera on the Phoebus fleet screen).
- Root `CLAUDE.md:18` package table still describes the gateway as "exposing that host's GEECS camera images".

**3. P3 — An empty scope record decodes as valid, and the empty-push case the author observed is noisy on the live path.**
`GEECS-Data-Utils/geecs_data_utils/io/arrays.py:127-146` accepts `samples == 0` (header `0.000000,…|` + count 0) and returns a `(0,)` array — verified: `decode_labview_waveform(wf([], "Ch0"))` → shape `(0,)`. The module docstring promises "a truncated or misread array must never come back looking valid"; an empty record is the one shape that looks valid and carries no data. Downstream (`server.py:456-461`) it is posted and becomes `_last_frame[var]`; an arm on that held frame posts `ArraySizeX=0` (descriptor shape `()`) and `_open_file` then fails — h5py: "All chunk dimensions must be positive" (verified in the gateway venv). If instead the Picoscope's "empty trace" (Planning note, `02_nonscalar_over_pva.md:59-60`) is an empty string, `decode_csv_values` raises and `_publish` logs a WARNING **with traceback** per push, per watched channel, for as long as no trigger is present. Neither is on the 3a capture path (Picoscope `capture=()`), but `scopeTrace.Channel*` PVs are served now. Suggest rejecting `count == 0` in the decoder and rate-limiting/downgrading the per-push decode warning in `_publish`.

**4. P3 — The nested-pairs decoder's byte-accounting claim is overstated.**
`arrays.py:78-97` uses `_PAIR.findall` between the outer brackets, so `[[1,2]],[[3,4]]`, `[[1,2] [3,4]]`, `[[1,2],,[3,4]]`, `[[1,2],[3,4]]]` and `[[[1,2]]]` all decode silently (verified). The docstring says every decoder "raises ValueError on a payload it cannot account for byte by byte" — true for the waveform and CSV, not for pairs. Bounded (LabVIEW's formatter emits well-formed text); either full-match the row grammar or soften the claim.

**5. P3 — The plugin's decoder seam drops the waveform attributes, so a captured scope stack would have no time axis.**
`file_plugin.py:435,443` types `decoder: Callable[[str], np.ndarray]` and `server.py:205` passes `self.decode(v, blob)[0]`; `x0`/`dx`/`samples` reach the PV (`streams.py`) but never the `.h5`. Not reachable in 3a, but 3b (Picoscope capture) has to widen this seam — e.g. write `dx`/`x0` as per-frame `NDAttributes` — so it is worth deciding now rather than after the first scope run writes axis-less stacks.

**6. P3 (placement question, not a work order) — `array_ceiling` in GEECS-Core has one consumer.**
`geecs_core/db/device_streams.py:100-103, 241-245`: the padding ceiling is read only by the gateway (`config.py:183`); the worker takes geometry from the plugin PVs and never imports it (`grep array_ceiling GeecsBluesky/` → docs only). `DESIGN.md:56-58` admits `device_streams` on the two-consumers argument and says "when in doubt, leave it out". Keeping it beside the exclusions avoids a second devicetype table (the hazard the module docstring itself warns about), so this may be the right call — but the DESIGN.md admission should say so, and a per-instance or DB-driven ceiling would move it.

**7. P3 — The read side is not "unchanged" for the analysis consumers.**
`GEECS-DataPortal/geecs_portal/resources.py:497-499`: `device_kind` classifies any device folder holding a stack as `"stack"` first, so after the first four-plugin MagSpec scan `UC_…-interpSpec/` and `-interpDiv/` flip from "unrenderable" (TSV folders) to stack devices and the Images tab renders each `(2048, 2)` NaN-padded float frame as a 2-pixel-wide strip (`to_display_png` is NaN/float-safe, so no error — verified). A 1-D stack (a captured CSV/scope stream) would raise in `Image.fromarray`. The CHANGELOG's "no read-side change" is true of the worker only; note it as the arc's read-side debt.

### Redundancy
No surviving duplicates. Checked: `geecs_data_utils/io/array1d.py` (TDMS/file readers, incl. `wf_increment` waveform *properties* — file-based, not wire), `geecs_data_utils/io/images.py` (`_flatten_string_to_bytes` is reused, not copied), `geecs_core/client/*` and `geecs_core/transport/*` (no non-scalar decoders), `image_analysis/analyzers/line_stitcher.py` and `magspec_manual_calib_analyzer.py` (read `-interpSpec` TSVs by filename), `geecs_bluesky/assets/registry.py` (native-file table, explicitly distinct), `geecs_bluesky/callbacks.py`, `geecs_bluesky/assets/handlers.py`, padding helpers (`np.pad` only in the grenouille algorithm). `_resolve` in `device_streams.py` removes a duplicate loop rather than adding one.

### Placement
`io/arrays.py` in GEECS-Data-Utils (below both consumers, beside `images.py`) and `streams.py` in the gateway (the only place that pads) are the right homes; the exclusion list in GEECS-Core fits the recorded `scalar_policy` precedent. The one question is 6.

### Refuted (so nobody re-checks)
p4p `NTNDArray.wrap(value, attrib=…)` and 1-D arrays — accepted (`p4p/nt/ndarray.py:147-181`); ophyd-async `get_ndarray_resource_info` drops zero dims and maps `Float64` (`_data_logic.py:82-99`, `_io.py:29`); `PluginPathProvider` creates the `<device>-<variable>` folder scanner-side with `mkdir(exist_ok=True)`, no `parents` (`hdf_plugin.py:117`) — invariant intact; the parity tool reports a TSV+stack folder as `capture_only`, not a mismatch (`diff.py:97-99, 209`); scan-path readers filter by shot regex / extension (`scan_paths.py:685-696, 881-887`); fixtures keep CRLF in the index (`git ls-files --eol`: `i/crlf attr/-text`); every exclusion is `1darray`-typed in the recorded fixture; `GeecsDb.get_experiment_device_types` exists (`geecs_db.py:567`); pre-commit (ruff, ruff-format, pydocstyle, yaml/toml/ast) passes on all changed files; version bumps and CHANGELOG headings agree in all four packages.

### What I ran
Data-Utils `tests/test_io_arrays.py tests/test_scan_stack.py`: 33 passed. Core `tests -m "not integration"`: 155 passed, 1 skipped. PvaGateway `tests` (3 ophyd-async-dependent tests deselected as instructed): 54 passed (the 6 `test_arrays.py` end-to-end tests confirmed executing, not skipped). Bluesky `tests/test_namespace.py tests/test_hdf_plugin_detector.py`: 53 passed. `pre-commit run --files <all 41 changed>`: all hooks passed. Mutation probes (each restored with `git checkout --`, tree clean afterwards): `pad_rows` truncating → 2 gateway tests fail; `served_array_variables` ignoring exclusions → 2 Core + 1 gateway tests fail (Bluesky namespace tests unaffected, correctly — the worker captures declared ∩ served); `capture_streams` dropping served arrays → 1 namespace test fails; waveform length check removed → 1 Data-Utils test fails; `_post_geometry` 1-D branch removed → 1 gateway test fails. No files under `tests/data/wire/` were touched.

---

## Disposition (author), fix commit 90de2bc2

|---|---|---|
| 1 | P2 | **Fixed** — `PvaGatewayConfig.host` records the served address (`--host`, else the lab-facing local address `detect_local_ip` resolves toward the device endpoints) and `_instance_host()` names the identity PVs after it, before any device endpoint or the machine name. Pinned three ways: the config test (`cfg.host == "192.168.6.100"`), a server test (idle config → `192.168.7.168`; a config with devices and no host keeps the endpoint fallback), and a CLI test for the idle path (exit 0, the WARNING, nothing listed). Shown to bite: removing the `config.host` branch fails the server test. |
| 2 | P2 | **Fixed** — DEPLOYMENT.md (gating / plugin / `:connected` wording; the `role=PVA image gateways` record name is kept, `fleet_status.sh` keys on it, and now says so), README.md, CLAUDE.md (the two stale identifiers and the decode step), the CLI description, DESIGN.md (past tense; the ceiling's admission), the root CLAUDE.md row; `docs/geecs_gateway/image_pvs.md` gains an "Array variables" section with the dotted-name normalization, shapes, NaN padding and the waveform attributes; `FleetHost.cameras` → `devices`, `camera_endpoints` → `stream_endpoints`, "stream hosts / stream devices" wording in the probe lines and the generated screen, which is regenerated. **Consequence to know:** the roster grows from 14 to 22 hosts — every host whose only stream devices are arrays (DaqPads, the hexapod, the Hamamatsu server) now appears as *not deployed*, which is the DB-driven served-set rule stated honestly; excluding those devicetypes is a declaration decision for the owner, not a roster fix. |
| 3 | P3 | **Fixed, half** — `decode_labview_waveform` rejects a 0-sample record (pinned, also through `decode_array_payload`). The noisy-warning half is **refuted**: `_DeviceWorker._on_frame` returns before decode on an empty value (`if not blob`), so the Picoscope's empty pushes without a trigger — an empty string, as captured 2026-09-21 — never reach `_publish`; no per-push traceback. |
| 4 | P3 | **Fixed** — `_PAIRS_PAYLOAD` full-matches the grammar (`[` rows `]`, rows `[..]` comma-separated, nothing else); the five variants listed are now `ValueError`s and are in the test's parametrization. |
| 5 | P3 | **Waived for 3a** — not reachable (the Picoscope captures nothing yet); recorded in the brief under PR 4/3b as "write `x0`/`dx` as per-frame `NDAttributes`, decide with the gate". |
| 6 | P3 | **Accepted as placement; text fixed** — DESIGN.md's admission now says the ceiling has one consumer and rides with the exclusions so a devicetype has one table, and that a per-instance or DB-driven ceiling would move it. |
| 7 | P3 | **Waived for 3a, recorded** — the GeecsBluesky CHANGELOG entry now says the portal's Images tab renders a lineout stack as a two-pixel-wide image (and would not render a 1-D stack) until the arc's record-side PR; the brief carries it under PR 4. No 1-D stack is captured by 3a. |

Confirmation of each fix by the same reviewer follows.

skbarber:
## Confirmation round (same reviewer, fix commit 90de2bc2)

Checked independently against `git diff 03a4ff49...90de2bc2` in the same worktree, tests run with the same recipe; every mutation restored, tree clean, `tests/data/wire/` untouched.

1. **Confirmed.** `PvaGatewayConfig.host` is set from `--host`, else `detect_local_ip(any_ip)` (`config.py:160-168`), and `_instance_host()` returns it first (`server.py:520-521`). End-to-end simulation of the production path (`launch.bat`, no `--host`) on an idle host: `cfg.devices == []`, `cfg.host == 192.168.7.168`, instance serves `undulator:pvagateway:192_168_7_168:version` — exactly what `FleetHost(ip="192.168.7.168").instance_pv()` asks for. Mutations bite: disabling the `config.host` branch fails `test_an_idle_instance_is_named_after_its_served_host_not_the_machine`; restoring `return 1` fails `test_main_idles_with_no_devices_instead_of_exiting`. See new finding A for the precedence it introduces on serving hosts.
2. **Confirmed, with residue.** DEPLOYMENT.md (`:198, :266, :272, :308`), README, gateway CLAUDE.md (`:145/:153` now `DeviceSpec` / `_DeviceWorker`), `__main__.py` description, Core DESIGN.md, root CLAUDE.md row, `image_pvs.md` new section (its normalization claim verified against `normalize_component`: `scopeTrace.Channel0` → `scopetrace_channel0`, runs collapse, lowercased), `fleet.py`/`gen_fleet_status.py` wording, and the regenerated `.bob` (0 "camera" mentions; 22 hosts incl. three array-only ones the old screen lacked). Residue the grep still finds: `fleet.py:138` docstring `:func:\`camera_endpoints\`` (renamed to `stream_endpoints` in this very commit — dangling cross-ref, new finding B); "image variable" in `file_plugin.py:1, :104, :134, :141` (the plugin is per stream variable), `gen_fleet_status.py:9, :34` "image devices", `docs/geecs_gateway/image_pvs.md:38` and `:103` ("Each image variable also has a subscription-state PV" — `:connected` is per stream variable now); `fleet.py:52, :136, :141, :153` "camera server"/"camera list"/"enabled camera" for a roster that now includes scope servers. The hardware noun "camera server" elsewhere (DEPLOYMENT/bootstrap/pyproject) is correct as is.
3. **Confirmed (decoder half); your refutation of the noise half is confirmed for the empty-string form.** `decode_labview_waveform` rejects `count == 0` (`arrays.py:137-140`; `decode_array_payload(_waveform([]))` raises; mutation removing the check fails `test_an_empty_waveform_record_is_rejected_not_posted`). `server.py:431-433` `if not blob or not isinstance(blob, str): return` is pre-existing and sits before both the plugin offer and the latest-wins slot, and the transport regex (`tcp_subscriber.py:109-111`, `(?P<value>.*?) nvar`) yields `""` for `nval, nvar` — so an empty-string push never reaches decode. What I cannot verify from the repo is *which* form the untriggered Picoscope emits (no fixture); if it is a 0-sample header rather than `""`, the new rejection turns every push into a decode error and `_publish` logs a WARNING with traceback per push per watched channel. Cheap insurance for 3b: capture one untriggered push into `tests/data/wire/`.
4. **Confirmed.** `_PAIRS_PAYLOAD` rejects all five variants plus six more I tried (`[[1,2],]`, `[,[1,2]]`, `[[1,2]][`, `[[]]`, `[[1,2],[]]`, `[[1,2]] junk`) and accepts the legitimate forms incl. inner whitespace, `[ [1,2],[3,4] ]` and CRLF tails; 285-row fixture 0.26 ms, a 16384-row stitcher-sized payload 15.9 ms (linear, no backtracking blow-up). Mutation (old `startswith/endswith` check) fails the 5 new parametrized cases.
5. **Confirmed.** `Planning/data_capture/02_nonscalar_over_pva.md:104-112` (PR 4) names the plugin decoder seam returning values only and the missing `x0`/`dx`, with per-frame `NDAttributes` as the representation to decide — accurate to `file_plugin.py:435,443` / `server.py:205`.
6. **Confirmed.** `GEECS-Core/DESIGN.md:62-68` states exclusions + ceiling, the one-consumer fact, the one-table rationale and the move condition; matches code (`array_ceiling` read only at `config.py:183`).
7. **Confirmed.** `GeecsBluesky/CHANGELOG.md` 0.98.0 wording is accurate (`device_kind` classifies stack-first → two-pixel strip; a 1-D stack would fail `Image.fromarray`), and the PR 4 line records it.

**A. P3 — On a serving host the identity PVs now follow the route probe, not the DB endpoint the fleet keys on.** `config.py:168` sets `served_host = detect_local_ip(any_ip)` whenever `--host` is absent (the production `launch.bat` path), and `_instance_host()` prefers it over `devices[0].host`. Simulation: DB endpoint `192.168.6.100` (= the `[pva] addr_list` entry), `local_ip_addresses` → `{192.168.6.100, 192.168.7.5}`, probe toward the alphabetically-first device endpoint answers `192.168.7.5` → the instance serves `undulator:pvagateway:192_168_7_5:version` while the fleet asks `…:192_168_6_100:version` → `[DOWN]` for a live, serving host; pre-fix it served `…:192_168_6_100:*`. Needs a dual-lab-addressed box (the usual camera-NIC + lab-NIC layout routes lab traffic out the lab NIC, so it is bounded), but it is a regression class for the hosts that *are* serving. Safer precedence, one line: `--host` → the endpoint IP of any served device (known to be in `hosts` and in the roster) → probe → hostname. Corollary, not a finding: an experiment with zero enabled devices gives `any_ip == ""` → `host=None` → hostname fallback (verified).

**B. P3 — Dangling cross-reference from the rename.** `fleet.py:138` `fleet_roster` docstring still cites `:func:\`camera_endpoints\``, deleted by this commit (now `stream_endpoints`). Plus the docstring residue listed under 2.

### What I ran
Data-Utils `tests/test_io_arrays.py tests/test_scan_stack.py`: 39 passed. PvaGateway `tests` (3 ophyd-async-dependent tests deselected): 56 passed. `pre-commit run --files` on the fix's 20 files: all hooks passed. Four mutations (empty-record check, pairs grammar, `config.host` branch, idle exit) → 1 / 5 / 1 / 1 tests fail, all restored. Grammar probe: 11 malformed rejected, 6 legitimate accepted. Three-case host simulation as above. Core and Bluesky suites not rerun (the fix touches only their docs).

---

## Disposition of the new findings (author), fix commit db9be1e4

|---|---|---|
| A | P3 | **Fixed** — the identity host is now `--host`, else **the served devices' DB endpoint** (the address `[pva] addr_list` names; every served device on a host shares it), else, with nothing served, the route probe. Pinned by `test_identity_host_is_the_served_endpoint_not_the_route_probe` (dual-NIC: probe answers `192.168.7.5`, served endpoint `192.168.6.100` → `cfg.host == 192.168.6.100`; nothing served → the probe). Shown to bite: reverting to the probe fails it. |
| B | P3 | **Fixed** — the `stream_endpoints` cross-reference; "stream host / stream device / stream variable" in `fleet.py`, `gen_fleet_status.py`, `file_plugin.py` (module, suffix, attribute helpers) and the orientation page's two remaining sentences. The hardware noun "camera server" is left where it names the box. |
| 3 (insurance) | — | **Recorded, not stored** — the untriggered Picoscope's form was observed live on 2026-09-21 (eight pushes, every trace variable an empty string), which is exactly the value the gateway drops before decoding; the fixture README says so. There is no 0-sample record on the wire to capture. |

One last confirmation of A and B by the same reviewer follows; this is the review's second and final round.

skbarber:
## Final confirmation (same reviewer, fix commit db9be1e4)

Checked against `git diff 90de2bc2...db9be1e4` (7 files) in the same worktree with the same recipe; the one mutation was restored, tree clean, `tests/data/wire/` untouched (the README there was read, not written).

**A. Confirmed.** `config.py:206-211` now sets `served_host = host or (served[0].host if served else probe_host)`, with `probe_host` computed only on the no-`--host` path. Re-ran my simulation against the new code: (i) idle host, production path — `host=192.168.7.168`, serves `undulator:pvagateway:192_168_7_168:version`, the name `FleetHost(ip=…).instance_pv()` asks for: MATCH. (ii) serving dual-NIC box, probe answering the other NIC — `host=192.168.6.100` (the DB endpoint), MATCH (was MISMATCH on 90de2bc2). (ii′) explicit `--host` — MATCH. (iii) experiment with no enabled devices — `host=None`, hostname fallback (pathological corner, unchanged, note only). (iv) one box with two DB endpoint IPs — `served[0].host` is the alphabetically first device's endpoint; the roster would list both IPs as separate hosts, which is the pre-existing one-instance-per-IP fleet model, not something this change introduces. `test_identity_host_is_the_served_endpoint_not_the_route_probe` covers (ii) and the idle case; reverting the precedence to `host or probe_host` fails exactly that test (1 failed, 5 passed), so it bites.

**B. Confirmed.** Re-grep of `GeecsPvaGateway/` (whole package), `docs/geecs_gateway/`, root `CLAUDE.md`, `GEECS-Core/`, `GeecsBluesky/geecs_bluesky/` for `camera_endpoints|CameraSpec|_CameraWorker`: no hits outside CHANGELOGs. "image variable" in `geecs_pva_gateway/` + `docs/geecs_gateway/`: the single remaining hit is `streams.py:57`, the docstring of `decode_image` — correct as written (that function *is* the image-variable decoder), not residue. The listed phrases are gone: `fleet.py:52,136,138,141,153` (incl. the cross-ref now `:func:\`stream_endpoints\``), `gen_fleet_status.py:9,34`, `file_plugin.py:1,104,134,141`, `image_pvs.md:38,103`.

**Finding 3's open question — closed.** `tests/data/wire/README.md:18-21` records the live observation (8 untriggered pushes, all length 0: an empty string, no 0-sample record on the wire), which is the form the pre-existing `server.py:431-433` guard drops before decode; the empty-record rejection in the decoder stays as defence in depth. Consistent with what I verified last round.

**Nothing broken by the fix.** PvaGateway `tests` (same 3 deselected): 57 passed. `pre-commit run --files` on the fix's 7 files: all hooks passed (the fixture README sits under the hooks' `tests/data/wire/` exclude, as intended). No other package's code changed in this commit.

### What I ran
PvaGateway suite: 57 passed, 3 deselected. Pre-commit on the 7 changed files: passed. Five-case host simulation (i, ii, ii′, iii, iv) as above. One mutation (precedence reverted) → 1 test fails, restored. Residue grep over the package, docs and neighbouring packages.

---

**Author: review closed.** Round 1: 2 P2 + 5 P3 → fixed (5) / waived with reasons (2), all confirmed. Round 2: 2 P3 → fixed, confirmed. The reviewer's corner (iii) — an experiment with zero enabled devices falls back to the machine name — is pathological (no experiment, no fleet) and left as is. Ready for the maintainer's merge into `feature/nonscalar-pva` once CI on db9be1e4 reports.

skbarber (commented):
Fresh review of db9be1e4: no new actionable findings.

I re-read the root and package guidance, reviewed the array decoders, stream declaration/selection, padding and drop behavior, gateway subscription and HDF paths, fleet identity/roster changes, and the Bluesky multi-plugin wiring. The previously recorded waveform-axis and portal rendering items remain explicitly tracked follow-up work; I did not duplicate them as findings.

Independent verification:
- GEECS-Core test_device_streams.py: 27 passed
- GEECS-Data-Utils test_io_arrays.py: 27 passed
- GeecsPvaGateway modified suites: 46 passed
- GeecsBluesky test_namespace.py: 39 passed
- GeecsBluesky test_hdf_plugin_detector.py: 14 passed

All current GitHub checks are green. Hardware verification remains owed as stated in the PR body.

PR #947, opened 2026-09-22
Preset run-level fields are the one source of truth (trigger_profile joins native_image_save); scanner relock

skbarber:
## What

Three housekeeping items left over from #944, each its own commit.

1. **`trigger_profile` gets the same one-source-of-truth rule as `native_image_save`.** `expand_preset` used `kwargs.setdefault("trigger_profile", preset.trigger_profile)`, so a copy of the key inside `plan.kwargs` silently overrode the top-level preset field — the exact shape the Codex review of #944 flagged for the switch. The two run-level fields are now `RUN_LEVEL_FIELDS` in `qs_client/presets.py`, handled by one rule: set at the preset's top level, assigned into the plan keyword when set, **refused** if a copy sits in `plan.kwargs`. The refusal message no longer names one client's widget (the round-3 review nit). The scanner's save-path comment says what the expander now does. GeecsBluesky 0.97.0 → 0.97.1.
2. **GeecsScanner `poetry.lock` relocked.** The path dependencies' pins had fallen behind (geecs-bluesky 0.94.0 → 0.97.1, geecs-schemas 0.29.0 → 0.31.0, geecs-core 0.8.2 → 0.8.3, geecs-data-utils 0.33.0 → 0.34.2). Editable installs imported the checkout's code regardless, so nothing ran wrong, but a freshly installed environment reported the stale numbers through `importlib.metadata` — a false "old version" readout during yesterday's #944 deploy. `poetry lock` only, no dependency upgraded; the `scrub-lock-paths` hook normalized the path URLs. GeecsScanner 0.13.0 → 0.13.1.

**Corpus check:** none of the 86 presets across all experiments in GEECS-Plugins-Configs carries either field inside `plan.kwargs`, so the refusal breaks nothing live. The scanner has dropped both copies on save since #944.

3. **Docs: the laser-off arm failure, as a note rather than an issue** (judgment-call addition, cheap to veto). The qserver README's existing `no frame within 8 s` bullet framed the case as "right after a camera server restart"; 26_0921 showed it is any laser-off session, because the PVA gateway's image subscription is gated and a shot fired with nothing monitoring seeds nothing. The bullet now carries the diagnosis (`<image>:connected` reads `Idle`) and the operator-verb recipe that worked (a monitor on the image, then two `mv` items), and the `/lab-status` skill's failure vocabulary points at it. The maintainer's call: laser-off scanning is for testing, so no engine change.

## LOC by concern

| Concern | Files | Lines |
|---|---|---|
| run-level field rule | `qs_client/presets.py`, `tests/qs_client/test_presets.py`, `scanner.js` (comment), both CHANGELOGs + pyprojects | ~+55 / −25 |
| relock | `GeecsScanner/poetry.lock` | 5 lines changed |
| laser-off note | `GeecsBluesky/qserver/README.md`, `.claude/skills/lab-status/SKILL.md` | ~+14 |

No facility literals added.

## Tests

`test_run_level_field_in_plan_kwargs_is_refused_not_merged` is the old switch test parametrized over both fields (a kwargs copy against a set and against an unset top-level value, both refused). Counts on the worktree: GeecsBluesky `tests/qs_client` + `test_plan_registry.py` + `test_optimizer_resolution.py` 135 passed; GeecsScanner `tests/` 148 passed, 1 skipped (its own worktree env, installed from the new lock).

## Hardware verification

Not applicable: no scan-execution path changes. The expander runs client-side and is exercised by the scanner's preflight; the relock changes no resolved dependency version.


skbarber:
## Adversarial review of #947 (`fix/preset-field-truth` @ 165e4acf; commits 45bfa6dd, ddb26570, 165e4acf) — correctness / redundancy / placement

*Fresh-context reviewer subagent; report pasted verbatim.*

Verdict: the change is correct and does what the CHANGELOG says; no High or Medium findings. Three Low findings survive, all consistency/scope rather than wrong behaviour. The docs commit has no surviving findings.

### Findings

1. **Low — `GeecsScanner/geecs_scanner/static/scanner.js:453` — the preset LOAD path still encodes the precedence rule this PR deletes from the expander.**
   `fillFormFromPreset` reads `kw.trigger_profile` in preference to `doc.trigger_profile` when a kwargs copy exists, and `tests/test_page.py::test_preset_trigger_control_overrides_hidden_kwarg_and_keeps_other_options` pins it (`assert state["loaded"] == "effective"`). That "kwargs copy is the effective value" rule was `setdefault`'s behaviour, which 45bfa6dd removes; the expander now refuses such a preset outright, so there is no "effective" value to show. Scenario: a hand-authored preset `{trigger_profile: "HTU-Normal", plan: {kwargs: {trigger_profile: "HTU-LaserOFF"}}}` — the form opens with LaserOFF selected and Start submits LaserOFF at top level (buildPreset reads the selector), i.e. the client silently picks the value the backend now calls illegitimate. Meanwhile `native_image_save` three lines down loads from `doc.native_image_save` only — the two RUN_LEVEL_FIELDS are loaded by opposite rules. Not reachable from the corpus (0 of 86 presets carry a copy — verified) and save/submit both drop the copy, hence Low. Fix: load `doc.trigger_profile` only, flip the harness assertion to `loaded == "top-level"` and rename the test; or, if the "show what used to win" migration is intentional, say so in the comment at :452 (there is none). The new comment at :607-608 ("expand_preset refuses a copy of either") argues for the former.

2. **Low — `GeecsScanner/CHANGELOG.md` 0.13.1 / scope — the relock fixes one of eight stale locks that produce the same false version readout on the fleet.**
   The relock itself is genuinely lock-only (4 version lines; every `[package.dependencies]`/`[package.extras]` block matches its pyproject; `content-hash` unchanged). But the symptom the entry describes — `importlib.metadata` reporting a stale number in a fresh venv — is exactly what `scripts/fleet_status.sh:471` reads per service, and every other lock in the repo pins the same path deps behind their packages: GEECS-MCP (geecs-bluesky 0.92.2 vs 0.97.1, core 0.8.2 vs 0.8.3, data-utils 0.32.2 vs 0.34.2, schemas 0.28.2 vs 0.31.0), GEECS-DataPortal, GeecsLogbook, ScanAnalysis, ImageAnalysis (data-utils 0.32.2, schemas 0.28.2), GeecsCAGateway (core 0.8.2, schemas 0.28.2), GeecsPvaGateway (core 0.8.2, data-utils 0.32.2), and the root lock. The portal/logbook/gateway hosts will show the same "old version" on their next venv rebuild. Not a defect in this PR and widening it is scope creep for a fix PR; but per the repo's own "pattern-bound mechanical pass" rule, name the owed relocks in the CHANGELOG entry or PR body (or file one issue) so the pattern is recorded rather than rediscovered host by host.

3. **Low — `GEECS-Schemas/geecs_schemas/preset.py:96-98` — the schema's own contract text now contradicts the expander.**
   `PlanCall`'s docstring: kwargs carry "the GEECS keyword arguments the worker adds to every scan verb: `shots_per_step` … and `trigger_profile` (normally the preset's own field)". After this PR a `trigger_profile` in kwargs is refused, so "normally" is "never". It does not ship: docgen renders only a model docstring's first paragraph (`docgen.py:361-363`) and the sentence is in the second, and the `kwargs` field description does not mention it — so this is Python-docstring-only. Fix: drop the `trigger_profile` clause (a one-line docstring edit; under the repo's rules that is a GEECS-Schemas patch bump + CHANGELOG line, or file it as a follow-up).

### Refuted / not findings (so nobody re-litigates them)
- Callers that put `trigger_profile` in kwargs: the scanner's `calibration_check`/`calibration_measure` (scanner.py:633-648) and `test_verbs.py` go through `_queue_item` → `submit_plan`, never `expand_preset`, and the calibration plans are `NON_SCAN_PLAN_NAMES` — unaffected and correct. Demo presets (demo.py:78/119/143) set the field at top level. GEECS-MCP has no preset/submit path (only the `trigger_profiles` listing kind). `prepare_optimizer_preset` never touches the field. `run_submit_preflight` and `ZmqQueueClient.submit_preset` both call `expand_preset`, so the refusal surfaces as the preflight's hard-gate refusal / `SubmitResult.message`, operator-facing — and the preflight already resolved the profile's devices from the TOP-LEVEL field (`_trigger_profile_devices`, submit_preflight.py:199), so the old `setdefault` let the plan run a profile the preflight had not liveness-checked; the PR closes that gap.
- Placement of `RUN_LEVEL_FIELDS`: presets.py is right. It describes the client-seam rule (preset field → plan keyword); the registry pops those keywords inline (registry.py:313, 319) and needs no shared tuple, and `plan_names.py`'s precedent (`ACQUISITION_MODES`) is for names BOTH sides consume. GEECS-Schemas is vocabulary and should not learn which fields become plan keywords. No duplicate constant exists anywhere (searched plan_names.py, registry.py, GEECS-Schemas). The JS `delete kwargs.trigger_profile/native_image_save` pair is the unavoidable cross-language twin.
- Tests bite: reverting the loop to `setdefault` for trigger_profile makes the parametrized case raise nothing → `pytest.raises` fails; top-level assignment and unset-stays-absent are pinned by the existing tests at test_presets.py:97 and :153. GeecsBluesky qs_client presets+preflight: 49 passed; GeecsScanner full suite: 148 passed, 1 skipped.
- CHANGELOG claims verified: 86 presets in GEECS-Plugins-Configs (main @ cbcf6be), zero kwargs copies of either field; the scanner drops both on save (scanner.js:609-610); the four version deltas match the lock diff.

### Docs commit 165e4acf (qserver/README.md laser-off bullet + lab-status SKILL.md line): no surviving findings.
Verified against code: the gateway subscribes only while a client holds the PV (`_ClientHandler.onFirstConnect/onLastDisconnect` → `_retain/_release`, server.py:113-116, 260-275); `:connected` is `geecs_core.pv_naming.connected_pv(...)` with states `("Idle", "Disconnected", "Connected")` and `Idle` = gated off (server.py:54-58, 401); the image PV's initial value is `np.zeros((1,1), uint16)` with no stamp (server.py:167) = the epoch-zero placeholder; the plugin's `Capture=1` arms instantly on a held frame (#894, file_plugin.py:650-662) else waits `ARM_TIMEOUT_S = 8.0` and fails with `no frame from <device> <variable> within 8 s` (file_plugin.py:674) — the exact string quoted; the held frame (`server.py:162`, set :435) is never cleared on release, so "seed once, later arms are instant" holds; `<cam>-hdf-capture didn't match True in 10.0s` is ophyd-async's `set_and_wait_for_value` on the `hdf` child with `DEFAULT_TIMEOUT = 10.0`; gated prepares with the box OFF before `SCAN` (gated.py:239, 254-260, 268) so it fails identically; the recipe's writes mirror `HTU-LaserOFF.yaml`'s ARMED (`Trigger.Source: Single shot`) and SINGLESHOT (`Trigger.ExecuteSingleShot: 'on'`) on `U_DG645_ShotControl`, and `trigger_source`/`trigger_executesingleshot` follow the pinned `settable_attribute` rule ("Position.Axis 1" → `position_axis_1`) with no namespace exclusion for the shot-control device; `MoveIn.value: float` (models.py:208) makes "the move panel takes numbers only" true; `qserver queue add plan` is the README's own form (:61-64). Consistent with GeecsPvaGateway/CLAUDE.md:98-124 and CHANGELOG #853/#854/#894 entries. One cosmetic variance only: the README writes `<image>:connected` where the gateway's CLAUDE.md/CHANGELOG write `<image PV>:connected`.

What I checked and how: full `git diff master...HEAD` (3 commits) in the worktree; root, GeecsBluesky and GeecsScanner CLAUDE.md; every `trigger_profile` mention in the repo (py/js/html/md/yaml); every `expand_preset`/`submit_preset` caller; the Preset schema and docgen; the scanner's submit/preflight/calibration service paths, demo manager and both JS harness tests; all nine poetry.lock files against their pyprojects; the configs repo's 86 presets and the Undulator/Thomson trigger profiles; GeecsPvaGateway server/file_plugin and GeecsBluesky strict/gated/registry for the docs commit; and ran the GeecsBluesky preset/preflight tests and the GeecsScanner suite read-only. No files were edited, committed, stashed or pushed.

---

## Dispositions

| # | Finding | Disposition |
|---|---|---|
| 1 | Scanner load path preferred the kwargs copy | **Fixed** in 098ca701 + 5d010359: `fillFormFromPreset` reads the field only; the harness test asserts the field loads, the copy is dropped on save, and clearing the field clears the selector. Confirmation by the same reviewer to follow. (098ca701 shipped with the harness's last step still on the old rule — 3 cases failed; 5d010359 corrected it. The commit gate on the scanner suite's exit code was missing from my chain and is restored.) |
| 2 | Eight other locks pin the same path deps behind their packages | **Waived for this PR, recorded here as owed**: GEECS-MCP, GEECS-DataPortal, GeecsLogbook, ScanAnalysis, ImageAnalysis, GeecsCAGateway, GeecsPvaGateway and the root lock. Same symptom (a fresh venv reports stale `importlib.metadata` versions; `fleet_status.sh` reads exactly that). Lock-only relocks, one per package, whenever each package is next touched — or one sweep PR if a host rebuild makes the readout matter first. |
| 3 | `PlanCall` docstring described the old precedence | **Fixed** in 098ca701 (GEECS-Schemas 0.31.1, docstring only; rendered reference unchanged, no-drift test green). Confirmation to follow. |
| — | Cosmetic: `<image>:connected` vs `<image PV>:connected` | **Waived** — both spellings name the same PV; the README's is the shorter of two forms already in the repo. |

skbarber:
## Review round 2 of #947 — fix commits 098ca701 and 5d010359 checked against each round-1 finding

*Same fresh-context reviewer, continued; report pasted verbatim.*

Finding 1 (scanner load path preferred the kwargs copy) — CONFIRMED fixed.
GeecsScanner/geecs_scanner/static/scanner.js:453-455 now reads `var trigger = doc.trigger_profile;` with a comment stating the rule (a kwargs copy is not a value; neither shown nor kept on save), so both run-level fields load by the same rule as `native_image_save` at :458. The harness `tests/test_page.py::test_preset_trigger_control_loads_the_field_and_drops_the_kwargs_copy` asserts `loaded == "top-level"`, and its last step clears the FIELD while the kwargs copy still says "effective" and expects the selector empty. I proved it bites rather than trusting the rename: I ran the same extracted-function harness (the test's own `_script_function` extraction) against the scanner.js from 165e4acf and against HEAD. Old code → `loaded: "effective"`, `cleared: "effective"` — both new assertions fail; HEAD → `loaded: "top-level"`, `cleared: ""`, kwargs copy absent from the saved plan — passes. So the 098ca701 regression (last step still clearing the kwargs copy, which made the "cleared" assertion vacuous against the old line) is genuinely corrected in 5d010359, not just made green. `tests/test_page.py`: 29 passed, 1 skipped on 5d010359. `buildPreset` is untouched, so the save/submit path is unchanged. Nothing new introduced.

Finding 2 (eight other stale locks) — CONFIRMED as a disposition; recording rather than fixing is the right call for this PR's scope, and I do not disagree. But see New finding 1: the fix commits re-opened the same gap inside this PR's own relock.

Finding 3 (`PlanCall` docstring) — CONFIRMED fixed.
GEECS-Schemas/geecs_schemas/preset.py:96-100 now names `shots_per_step`, `acquisition`, `shot_period` as the worker-added kwargs and says `trigger_profile`/`native_image_save` are preset fields, never kwargs, refused as a copy — consistent with `plans/registry.py:313-320`, which pops exactly trigger_profile, shots_per_step, acquisition, non_essential, shot_period, native_image_save (omitting `non_essential` from an "authored kwargs" list is right: the client derives it from `devices[].essential`, it is never authored). GEECS-Schemas bumped 0.31.0 → 0.31.1 with a CHANGELOG entry whose claim ("docstring only; the rendered reference shows a model's first paragraph and is unchanged") I verified two ways: `docs/geecs_schemas/schema_reference.md` is not in the diff, and the no-drift guard `tests/test_schema_reference.py::test_committed_reference_matches_schemas` ran and PASSED (not skipped — it carries a skipif, so I ran it with -v) on the worktree; full GEECS-Schemas suite 506 passed.

New finding 1 — Low — GeecsScanner/poetry.lock:~957 vs GEECS-Schemas/pyproject.toml:3 — the fix re-staled the pin this PR relocked to fix.
098ca701 bumps GEECS-Schemas to 0.31.1; GeecsScanner/poetry.lock — relocked in ddb26570 precisely so that "the path dependencies' pins match their packages" (GeecsScanner CHANGELOG 0.13.1, first bullet) — still pins `geecs-schemas = "0.31.0"`. By the PR's own stated rationale (a fresh scanner venv reporting a stale `importlib.metadata` number, the readout `scripts/fleet_status.sh:471` reads), the 0.13.1 entry is now one patch untrue for geecs-schemas. Fix: `cd GeecsScanner && poetry lock` (a one-line lock change, content-hash unchanged) in the same PR; alternatively soften the CHANGELOG wording — but the relock is cheaper and keeps the entry honest. The GeecsBluesky and GEECS-MCP locks (geecs-schemas 0.28.2) are already in the waived set and are not made worse.

Nothing else new: the amended GeecsScanner 0.13.1 CHANGELOG bullet describes the final code accurately; no facility literals or hygiene issues in the diff; no behaviour surface beyond the one JS line changed.

What I checked and how: `git diff 165e4acf..5d010359` in full; ran the harness proof (old vs new scanner.js) via node with the test module's own extractor; `tests/test_page.py` on HEAD; `tests/test_schema_reference.py -v` and the full GEECS-Schemas suite; compared every lock's `geecs-schemas` pin with the bumped pyproject; re-read `plans/registry.py:313-320` against the new docstring. No files edited, committed, stashed or pushed.

---

## Dispositions (round 2)

| Finding | Disposition |
|---|---|
| Round-1 #1, #3 | **Confirmed fixed** above (the harness proven to bite against the old line) |
| Round-1 #2 | **Waived, recorded** — the eight owed relocks are listed in the round-1 comment |
| New #1: the 0.31.1 bump re-staled the scanner's `geecs-schemas` pin | **Fixed** in 7fb63be2 (lock only, 2 lines; a fresh install of the worktree env now reports 0.31.1). Confirmation of that commit to follow as the closing line. |

Two rounds by the stopping rule; the relock confirmation closes the review.

skbarber:
## Round 2 closing confirmation — 7fb63be2

*Same reviewer, verbatim:*

Round 2 confirmation for #947 @ 7fb63be2 — New finding 1 (re-staled geecs-schemas pin): CONFIRMED fixed. `git diff 5d010359..7fb63be2` touches GeecsScanner/poetry.lock only, one line (`geecs-schemas` 0.31.0 → 0.31.1); content-hash unchanged (731a94da…) on both sides; every geecs-* pin now equals its pyproject (bluesky 0.97.1, core 0.8.3, data-utils 0.34.2, schemas 0.31.1, web-theme 0.6.1); no `.claude/worktrees/`, `file:///Users/` or `file:///home/` URL in the lock. Nothing new. Round 2 closed under the stopping rule; no further findings from me.

Every finding is dispositioned (three fixed and confirmed, one waived and recorded as owed relocks, one cosmetic waived). Review closed.

skbarber (commented):
Reviewed current head `7fb63be2`. No remaining actionable findings.

I rechecked the run-level-field rejection and unset/default semantics, the scanner load/save behavior for hidden kwargs copies, the schema documentation, and the relocked path-package versions. The earlier review findings are fixed at this head.

Local targeted verification: 60 passed, 1 skipped across the GEECS-Schemas preset/reference tests, GeecsBluesky preset-expansion tests, and GeecsScanner page tests. GitHub pre-commit and all unit-test jobs are green.

PR #948, opened 2026-09-22
Core 0.11.0 + PvaGateway 0.13.0 + Bluesky 0.99.0: per-instance capture gates for the Picoscope channels; waveform stacks carry their axis (non-scalar PVA arc, PR 3b)

skbarber:
PR 3b of the non-scalar-over-PVA arc (brief: `Planning/data_capture/02_nonscalar_over_pva.md`). **Base is `feature/nonscalar-pva`.** Follows #946 (arrays served, padded and captured). This closes the Picoscope half that 3a deferred: capturing its traces.

## What

A Picoscope has up to four channels but pushes a trace only on the ones its `Enable.Ch<X>` reads `on` (live 2026-09-20: A/B on, C/D empty on `U_BCaveICT`). Arming a plugin on a disabled channel would wait out the arm timeout on every prepare — the FROG trap again. So the capture set is **per instance**, decided at every prepare:

1. **The declaration gains a `gate` column** (GEECS-Core 0.11.0): capture variable → the device's on/off variable. `PicoscopeV2` now declares `scopeTrace.Channel0..3`, gated by `Enable.ChA..D`. `capture_gates(devicetype, rows)` resolves pairs to DB spellings; the parity test requires every gate to be an on/off choice variable of the devicetype and to gate a declared capture.
2. **The detector reads the gates once per stage** (GeecsBluesky 0.99.0, `GeecsDetector(plugin_gates=)`): a gated plugin whose gate does not read `on` at `stage` takes no part in that session — not armed, no data key. The set is latched for the session: the stock detector re-runs its prepare context on every trigger and the run's descriptor is emitted once, so a gate moving inside a run must not move the armed set (review round 1 found the per-prepare version aborted the run in bluesky's bundler either way). A channel enabled between runs is captured on the next. The gate is one of the detector's own scalar readbacks, so it costs no extra PV. **`plugin_backed` follows the latch** (round 3, Codex): a staged scope whose every channel read off is not plugin-backed for that run, so the gated plan takes its scalars through the sampler and the non-essential wrapper sits it out with a WARNING — the stock kickoff would have refused a streamless flyer and aborted the run, and bluesky refuses an empty `describe_collect` at `declare_stream`, so the flyer has to leave the list rather than idle in it. A direct fly prepare on it is refused with the channels named.
3. **The namespace passes only gates it can read.** The gate must be one of the device's read-only columns (DB `get='yes'`, `set='no'`, so the CA gateway serves it as a scalar readback). A gated stream whose gate is not subscribed — or is marked settable and therefore bound as a Movable child — is **not captured**, with a WARNING naming the right remedy for each case — the fail-safe, never an arm on a channel that may push nothing.
4. **A waveform stack carries its axis** (GeecsPvaGateway 0.13.0). The plugin's decoder seam now returns `(array, attributes)`, and an array variable's plugin declares `wave_x0`, `wave_dx` (seconds) and `wave_samples` after the scalars, written from the payload (NaN for the pairs and CSV shapes, which have no axis). The root attribute `waveform_attributes` names them; `scalar_attributes` stays scalars only. The suffixes are spelled once in `geecs_data_utils.io.arrays` and the shot join skips them, so the s-file exporter never diagnoses the axis as missing scalar columns (round 3, Codex). This closes the #946 review's P3 before the first scope run could write an axis-less stack.

## Per-concern LOC

| concern | files | +/− |
|---|---|---|
| Declaration: gates | `GEECS-Core/geecs_core/db/device_streams.py`, `tests/test_device_streams.py` | +64/−8, +52/−3 |
| One owner for the waveform keys | `GEECS-Data-Utils/geecs_data_utils/io/arrays.py`, `io/__init__.py`, `tests/test_io_arrays.py` (0.35.1) | +12/−9, +4, +4 |
| Detector gate + namespace wiring | `GeecsBluesky/geecs_bluesky/devices/detector.py`, `namespace.py` | +76/−22, +31/−1 |
| Worker tests | `tests/test_hdf_plugin_detector.py`, `tests/test_namespace.py` | +99, +40/−12 |
| Stack axis attributes | `GeecsPvaGateway/geecs_pva_gateway/file_plugin.py`, `server.py` | +64/−14, +2/−1 |
| Gateway tests | `tests/test_arrays.py` | +76 |
| Docs, versions, changelogs, brief | three `CHANGELOG.md`, two `CLAUDE.md`, three `pyproject.toml`, `Planning/…` | +95/−9 |

1. **Fail-safe polarity.** An unreadable gate means *not captured*, not *captured unconditionally*. The alternative would re-open the FROG trap on every Picoscope until the DB flags are set.
2. **The gate is read through the CA gateway, so it needs DB `get='yes'` on `Enable.ChA..D`.** That puts four enum columns in each ICT's strict row — harmless and informative. The other route (the PVA gateway publishing an `:enabled` state per trace PV) would put a GEECS scalar on the PVA gateway, which its doctrine forbids.
3. **Axis attributes are declared for every array variable**, NaN when the payload carries no axis, rather than only for waveforms — the attribute set is fixed at arm time, before the first payload says which shape it is.

Facility-literal check: no new lab literals outside tests.

## Tests

- GEECS-Core: **162 passed, 1 skipped** (34 in `test_device_streams.py`).
- GeecsBluesky: **739 passed, 4 skipped** (full suite on b00914e2).
- GeecsPvaGateway: **59 passed** locally (3 deselected: `ophyd_async` missing in the local env, identical on the base; CI runs them).
- GEECS-Data-Utils: shot-join / arrays / export / stack suites **87 passed**.
- Shown to bite: the detector ignoring its gates → the gate test fails; re-reading gates per prepare with context invalidation (the round-1 defect) → the gate test fails; the namespace passing an unreadable gate, or misdiagnosing a settable one → the namespace tests fail; the plugin dropping the axis attributes → the waveform-stack test fails; the Core resolver keeping a pair with an unknown gate, or warning twice → its warning test fails. Restored, green.
- `scripts/check.sh` fails environment-shaped in a worktree; suites ran with the main checkout's envs against this worktree's source, CI markers.

## Deploy — one DB step first, then the usual order

**Owner's DB step (no rows exist for these today):**

```sql
INSERT INTO expt_device_variable (expt_device_id, variablename, `get`, `set`) VALUES
  ('973','Enable.ChA','yes','no'), ('973','Enable.ChB','yes','no'),
  ('973','Enable.ChC','yes','no'), ('973','Enable.ChD','yes','no'),
  ('970','Enable.ChA','yes','no'), ('970','Enable.ChB','yes','no'),
  ('970','Enable.ChC','yes','no'), ('970','Enable.ChD','yes','no');
```

(973 = `U_BCaveICT`, 970 = `U_UndulatorExitICT`, `expt_device.id` on Undulator.) Then restart the CA gateway (its roster is startup-only). Then gateway fleet `:restart` (the Picoscope host's service is `DEMAND_START`), add `192.168.7.168` to the worker's `[pva] file_plugin_addr_list`, restart the worker. Until the flags are set the worker logs one WARNING per channel and captures nothing from the scopes — by design.

- **U_BCaveICT** (A/B on, C/D off): a strict scan writes two stacks, `U_BCaveICT/U_BCaveICT.h5` and `U_BCaveICT-scopeTrace.Channel1/…h5`, each `(N, 3000)` float64 in volts, with `…-wave_dx` = 4e-9 per frame; no plugin armed on Channel2/3 (the worker's INFO line names them as not `on`); the prepare does not wait on the arm timeout.
- **U_UndulatorExitICT**: was not running on 2026-09-20; check its enables and repeat.
- **Toggle**: disable ChB in LabVIEW, run again — one stack; re-enable — two. The strict rows carry `enable_cha..d` columns.
- **Regression**: MagSpec (four stacks) and Point Grey (one) unchanged; the FROG's `frogTrace` stack unchanged.
- **All channels off, scope first** (verifier's bench item): with every enable off and the ICT listed first, a gated `count` takes the ICT as its clock, so its `timestamp` must advance per trigger in that state — expect `shots` rows with `u_bcaveict-acq_timestamp` and no ICT stack; if the stamp does not advance the sampler fails loudly (`GeecsTriggerTimeoutError` naming the ICT), never silently.

Review record: fresh-context adversarial review (1 P1 + 1 P2 + 1 P3 → 9ab2189f; 2 P3 → 2013adb0; all confirmed), then Codex (2 P2 + 1 P3 → 26dcc0ce, verified by a fresh-context verifier), whose verification raised one P2 (`refuse_native_essentials` on the per-session flag → 50cbe8f5) and one residual (`shot_clock` on the same flag → 8a27a38c + b00914e2): both plan-level checks now ask the static `has_file_plugin`, so bind-time and step-time agree — as comments below.


skbarber:
## Adversarial review (fresh-context reviewer, commit 543287bb)

Branch `feat/picoscope-gate` vs `feature/nonscalar-pva`, reviewed in the worktree with the main checkout's package envs against the worktree's source. Three findings survive; the rest of the diff (the Core `gate` column and `capture_gates`, the namespace readability guard, the gateway's waveform-axis attributes and the `(array, attributes)` decoder seam, the `scalar_attributes` slice fix, the parity test) holds up against the scenarios I tried.

### 1. P1 — the gate is re-read on every **shot**, not every prepare, and a gate that changes inside a run aborts the run in bluesky's bundler

`GeecsBluesky/geecs_bluesky/devices/detector.py:846-856` (`_update_prepare_context`), `:867-881` (`_closed_gates`); contract text at `:601-609`, `GeecsBluesky/CHANGELOG.md` 0.99.0, `GeecsBluesky/CLAUDE.md:442-447`.

The stock `StandardDetector.trigger()` calls `self._update_prepare_context(trigger_info)` on every trigger when a context exists (ophyd_async `core/_detector.py:648-655`, "Ensure the data provider is still usable"), and the strict plan also issues `bps.prepare(det, STRICT_TRIGGER_INFO)` for every row (`plans/strict.py:323`); the gated plan prepares every step (`plans/gated.py:259`). So `_closed_gates()` runs per shot. `Enable.Ch<X>` is operator state in the LabVIEW GUI, changeable at any moment. When it changes mid-run the armed set differs, `_prepare_ctx` is dropped and **every** provider is rebuilt, but the run's descriptor is already emitted (bluesky caches `describe()` once per object per run, `bundlers.py:414`). Reproduced through a real `RunEngine` with the diff's own `_gated_scope` fixture, two shots, `prepare → trigger → create/read/save` per shot:

- **Channel B enabled between shot 1 and 2** → `RuntimeError: Receieved a stream_resource with data_key u_ict-scopetrace_channel1 that is not in the descriptor 'STREAM:' data_keys ['u_ict']` (`bundlers.py:893`). Run aborted.
- **Channel B disabled between shot 1 and 2** (A and B both on at start) → `RuntimeError: Received stream_datum for each of the data_keys {'u_ict'}, expected ... {'u_ict-scopetrace_channel1', 'u_ict'}` (`bundlers.py:940`). Run aborted, and channel B's plugin `Capture` was still `True` at the abort (nobody counts it, unstage did not run).
- In both directions the still-armed channel A also gets a **second** `stream_resource` (new uid) whose first datum is `indices {0..2}, seq_nums {2..4}` — frame 0 referenced twice and the datum width no longer matches the event — because the rebuild starts a fresh `StreamResourceDataProvider` (`last_emitted = 0`) over the same file.

Corollary: `_closed_gates` logs at INFO per closed channel per call, so a four-channel scope with two wired writes two "not captured this run" lines **per shot** (measured: 2 lines over 2 shots with one closed channel).

The CHANGELOG/docstring wording ("read at every prepare", "a channel enabled between **runs** is captured on the next") describes the intended granularity; the code's is per shot. Suggested shape: latch the closed set once per plugin session — in `stage()` (the scalar signals are staged there, so the gate read is a cache hit, and it is where `count_zeroed` is already keyed) — and have `_update_prepare_context` filter on the latched set. `stage()` already clears `_prepare_ctx`, so the `_armed_gates` invalidation at `:852-854` becomes unnecessary. Note that `tests/test_hdf_plugin_detector.py:397-424` currently **pins the per-prepare re-arm** (flip B on, prepare again in the same stage session, expect both armed) and would need to re-stage between the flips. I mutated out the invalidation lines and that test fails at `:415`, so the pin is real — it just pins the wrong granularity.

### 2. P2 — a gate that is subscribed **and** settable is dropped with the wrong diagnosis

`GeecsBluesky/geecs_bluesky/namespace.py:592-596` builds `readables` as subscribed **minus settables**; `:623-633` then treats a gate absent from `readables` as "not subscribed (DB get='yes')" and tells the operator to subscribe it. Scenario: the DB owner sets `set='yes'` on `Enable.ChA` (a scan-settable channel enable is an ordinary curation; today's rows do not exist yet and the Planning note's SQL writes `set='no'`, so this is one flag away). Reproduced with the test module's helpers: `Enable.ChA/ChB` subscribed and settable → all four channels dropped, four WARNINGs saying the gate "is not subscribed", while the device does carry a `CaSettable` child `enable_cha` whose readback is exactly the value needed (the namespace later `add_readables([child])`s it at `:683`). The guard should distinguish "not subscribed" from "subscribed but bound as a Movable child", and in the latter case either read the gate through the child or say so in the message. Bounded: requires a DB change; no data is lost, only capture is refused with a misleading remedy.

### 3. P3 — minor: an unknown *capture* name warns twice with the wrong noun; the waveform attribute keys are re-spelled

- `GEECS-Core/geecs_core/db/device_streams.py:279` resolves `[captured, gate]` with `which="gate"`, so a capture name the DB lacks warns once as "capture variable" (from `capture_variables`) and again as "declares **gate** variable 'scopeTrace.Channel0'" (verified with rows lacking `scopeTrace.Channel0`). `test_a_gate_the_db_does_not_list_drops_the_pair_with_a_warning` covers only the unknown-gate side. Resolve the two names with their own `which`, or skip the pair when the capture side is already gone.
- `GeecsPvaGateway/geecs_pva_gateway/file_plugin.py:134` `_WAVEFORM_ATTRIBUTE_KEYS = {"wave_x0": "x0", "wave_dx": "dx", "wave_samples": "samples"}` re-spells the literal keys `decode_array` emits at `GEECS-Data-Utils/geecs_data_utils/io/arrays.py:155-157`; the gateway already depends on Data-Utils, so the vocabulary has one owner available. Bounded because `test_a_waveform_stack_carries_its_time_axis_as_attributes` runs the real decoder end-to-end and would fail on a rename (the missing key would land as NaN).

### Redundancy / placement — checked, no duplication found
`GEECS-Schemas/geecs_schemas/preset.py:50-80` (`PresetDevice.save_images` / `essential` — per-run operator intent, a different axis from a per-instance hardware fact); `GeecsBluesky/geecs_bluesky/namespace.py` (`capture_streams`, `looks_triggerable`, `NATIVE_SAVE_VARIABLES`); `geecs_bluesky/assets/registry` (LabVIEW-native files, retires with #738); `geecs_core/db/scalar_policy.py` (the precedent `capture_gates` follows); grep for `Enable.Ch`, `capture_gates`, `plugin_gates`, `== "on"` across GEECS-Schemas, GeecsScanner, GeecsPvaGateway, GEECS-Data-Utils, GeecsBluesky — the on/off compare exists only in `_closed_gates`. `capture_gates` in GEECS-Core is the right home (shared by the worker; the PVA gateway needs no gate). The read side (`geecs_data_utils/io/scan_stack.py:105-128`) iterates the attribute group, so the new `wave_*` datasets come back as the CHANGELOG claims; `stack_scalar_variables` stays scalars-only, consistent with `waveform_attributes` being a separate manifest. Invariants: no scan-folder creation added (`PluginPathProvider` unchanged); `\\nas\hdna2\data` in the new test pre-exists on the base (`test_hdf_plugin_detector.py:270`); the Planning note's IPs are 192.168.x.

**Ran:** GEECS-Core `pytest tests -m "not integration"` → 162 passed, 1 skipped, 2 deselected; GeecsPvaGateway (3 ophyd_async-dependent tests deselected as instructed) → 59 passed; GeecsBluesky `test_hdf_plugin_detector.py test_namespace.py test_strict_plans.py` → 68 passed; scratch reproductions (RunEngine gate-flip ON/OFF, per-trigger `_closed_gates` count = 3 after 1 prepare + 2 triggers, settable-gate namespace) → 2 expected aborts + 2 passed; mutation check (context invalidation removed → `test_a_gated_plugin_is_armed_only_while_its_enable_reads_on` fails at :415) then `git checkout --` restored, worktree clean; nothing committed, staged, stashed or pushed; `tests/data/wire/` untouched.

---

## Disposition (author), fix commit 9ab2189f

|---|---|---|
| 1 | P1 | **Fixed** — the closed set is latched in `stage()` (right after the scalar signals are staged, so the read is a cache hit) and cleared in `unstage()`; `_update_prepare_context` filters on the latch and never re-reads, so every trigger's context re-check sees the same armed set and the stock code keeps reusing its providers. The `_armed_gates` invalidation is gone. The INFO line is once per closed channel per session. The docstring, CHANGELOG and CLAUDE.md now say "once per stage". The test pins the corrected granularity: a mid-session flip leaves the armed set and the **same provider objects** in place; re-staging picks the new state up, in both directions. Shown to bite: re-reading per prepare with context invalidation (the old code's shape) fails it. |
| 2 | P2 | **Fixed** — the namespace checks `settable_names` first and says so: "the DB marks settable (set='yes') so it is bound as a scan-settable child, not a read-only column — not captured; capture gating needs the gate subscribed read-only (get='yes', set='no')". The unsubscribed message is unchanged for the unsubscribed case. Pinned by `test_a_settable_gate_is_refused_with_the_right_diagnosis` (the settable channel dropped with that message, the read-only one captured, no "not subscribed" warning); reading the gate through the Movable child is not done — the detector is constructed before the namespace binds its settable children, so it could not see it, and the remedy is one DB flag. |
| 3 | P3 | **Fixed, both** — `capture_gates` looks the capture name up silently (the capture list's own resolution warns about it) and resolves only the gate with `which="gate"`; the test now asserts exactly one warning, about the gate, with unknown capture names present. `geecs_data_utils.io.arrays` exports `WAVEFORM_AXIS_KEYS` / `WAVEFORM_ATTRIBUTE_KEYS` (Data-Utils 0.35.1), the decoder builds its attributes from them, and the gateway derives `wave_<key>` suffixes and the key map from `WAVEFORM_AXIS_KEYS`. |

Confirmation of the three fixes by the same reviewer follows.

skbarber:
## Confirmation round (same reviewer, fix commit 9ab2189f)

Verified in the same worktree (HEAD = 9ab2189f, clean), same env recipe, against `git diff 543287bb...9ab2189f`; dispositions re-derived, not taken on trust.

**1. P1 (gate re-read per shot; mid-run flip aborts the run) — CONFIRMED FIXED.** `detector.py:811-813` latches `_closed_gate_logics` in `stage()` after the scalar signals are staged; `unstage()` clears it (`:821`); `_update_prepare_context` (`:857-860`) filters on the latch and no longer invalidates the context. My RunEngine reproduction re-run against the fix (three shots, `prepare → trigger → create/read/save` per shot, `Enable.ChB` flipped after shot 1): **B enabled mid-run** → run `exit_status=success`, exactly one `stream_resource` (`u_ict`), datum indices `0..1, 1..2, 2..3`, channel B's `Capture` still `False`, `_closed_gates` called **once** (at stage) and the INFO line **once** for the session; **B disabled mid-run** → success, two resources (both channels, armed at stage), each with contiguous indices, no INFO line, both plugins stopped at unstage; a following `stage()` latches B closed and the next `describe()` omits `u_ict-scopetrace_channel1`. One caveat, not a defect: with `bps.stage(obj)` at its default `wait=None` bluesky does **not** wait on a new-style stage status (`plan_stubs.py:1114-1122`), so `prepare` can race the stage coroutine and the `_closed_gate_logics is None` fallback reads the gates a second time (I saw 2 reads / 2 INFO lines that way). Production stages through `stage_all`, which waits on the group (`plan_stubs.py:1160-1161`; the stock `stage_wrapper`, and `gated.py:518` for the non-essential flyers), so the latch is in place before the first prepare and the fallback stays test-only as its comment says.

**2. P2 (settable gate misdiagnosed as "not subscribed") — CONFIRMED FIXED.** `namespace.py:623-634` refuses a gate in `settable_names` first with the `set='no'` remedy. Re-ran my own scratch (A/B subscribed **and** settable, C/D unsubscribed): A/B now get the settable message, C/D still the "not subscribed" one — the right diagnosis on each. Mutation check: removing the settable branch makes `test_a_settable_gate_is_refused_with_the_right_diagnosis` fail at `test_namespace.py:646` — the test bites. Restored.

**3. P3 (double warning with the wrong noun; re-spelled waveform keys) — CONFIRMED FIXED.** `capture_gates` (`device_streams.py:277-285`) looks the capture side up silently and resolves only the gate; restoring the two-name resolution makes `test_a_gate_the_db_does_not_list_drops_the_pair_with_a_warning` fail at `test_device_streams.py:127` (asserts exactly one warning) — bites. Restored. `WAVEFORM_AXIS_KEYS` / `WAVEFORM_ATTRIBUTE_KEYS` live once in `geecs_data_utils/io/arrays.py:51-57`, exported from `geecs_data_utils.io`, and `file_plugin.py:133-136` derives the `wave_*` suffixes and key map from them (`strict=True` zip); Data-Utils 0.35.1 + CHANGELOG travel with it; `test_io_arrays.py` pins the decoder's key order to the constant.

- **P3 — the reworked detector test does not distinguish the fix from the defect's core.** `test_hdf_plugin_detector.py:397-438` asserts "same provider objects, B not armed, key absent" across a mid-session flip. With the latch mutated away (gates re-read on every `_update_prepare_context`, i.e. the 543287bb behaviour minus its `_prepare_ctx = None` invalidation) that test **still passes**: the stock reuse path (`_detector.py:469-476`, keyed on `collections_per_event`) keeps the same providers regardless of how `_data_logics` is filtered, so every mid-session assertion holds trivially. My RE scratch does distinguish it (7 gate reads and 3 INFO lines under the mutation vs 1 and 1 with the fix). Sharpen the pin with one of: assert `_closed_gates` is not called after `stage()` (count calls, or `caplog` exactly one INFO line across a second `prepare`), or force a real rebuild after the flip — re-prepare with `collections_per_event=2`, which drops the stock reuse — and assert `u_ict-scopetrace_channel1` is still absent. Nothing ships wrong; the guard is just softer than its docstring claims.
- **P3 — prose drift from the fix.** `device_streams.py:269-271` (`capture_gates` docstring: "an unknown name on either side is dropped with a WARNING") and `GEECS-Core/CHANGELOG.md:14` say the same; the capture side is now dropped silently here (its warning comes from `capture_variables`). One clause each.

**Ran:** GEECS-Core `pytest tests -m "not integration"` → 162 passed, 1 skipped, 2 deselected; GeecsPvaGateway (the 3 ophyd_async tests deselected) → 59 passed; GeecsBluesky `test_hdf_plugin_detector.py test_namespace.py test_strict_plans.py` → 69 passed; GEECS-Data-Utils `test_io_arrays.py` (worktree source over the main env) → 27 passed; RE reproduction (ON/OFF flip, waited stage) → 2 passed; un-waited-stage trace and settable-gate scratch → 2 passed; three mutation checks (latch removed → repo test still passes but scratch fails 7≠1; settable branch removed → fails :646; double resolution restored → fails :127), each restored with `git checkout --`, worktree clean; nothing committed, staged, stashed or pushed; `tests/data/wire/` untouched.

---

## Disposition of the new findings (author), fix commit 2013adb0

|---|---|---|
| A | P3 | **Fixed** — the test wraps `_closed_gates` with a counter and asserts exactly one read after `stage()`, still one after the mid-session flip and a second `prepare`, and two after the next `stage()`. Shown to bite with the reviewer's own mutation (re-read per prepare, no invalidation): it now fails. |
| B | P3 | **Fixed** — the `capture_gates` docstring and the GEECS-Core CHANGELOG say an unknown gate warns and an unknown capture name drops the pair silently because `capture_variables` already warned. |
| caveat | — | **Noted, no change** — `bps.stage` without a wait can race `prepare` into the `None` fallback; production stages through `stage_all`, which waits, and the fallback's comment says it is for tests. |

One last confirmation of A and B by the same reviewer follows; this is the review's second and final round.

skbarber:
## Final confirmation (same reviewer, fix commit 2013adb0)

Verified in the same worktree (HEAD = 2013adb0, clean), same env recipe, against `git diff 9ab2189f...2013adb0` (3 files: the gated detector test, the `capture_gates` docstring, GEECS-Core's CHANGELOG 0.11.0; no runtime code changed).

**A. P3 (gated detector test did not distinguish the latch from a per-prepare re-read) — CONFIRMED FIXED.** `tests/test_hdf_plugin_detector.py:407-417` wraps `ict._closed_gates` with a counting coroutine and asserts `gate_reads == [1]` after `stage()` (`:419`), still `[1]` after the mid-session flip + second `prepare` (`:433`), and `[1, 1]` after the next `stage()` (`:441`). Re-applied my soft mutation to `detector.py` (replace the latch block with an unconditional `closed = await self._closed_gates()` — gates re-read on every `_update_prepare_context`, no invalidation): the test now **fails at `:433` with `assert [1, 1, 1] == [1]`** — the mid-session assertion, exactly the spot the P3 named. Unmutated it passes. Restored with `git checkout --`; worktree clean afterwards.

**B. P3 (prose said "WARNING on either side") — CONFIRMED FIXED.** `device_streams.py:268-274` docstring and `GEECS-Core/CHANGELOG.md:14-16` now read: unknown *gate* → WARNING; unknown *capture* name → dropped silently here, its warning is `capture_variables`'s, raised once. Matches the body (`:277-285`: capture side looked up via `by_lower` with a bare `continue`, only `[gate]` passed to `_resolve`) and the test that pins exactly one warning (`test_device_streams.py:115-128`).

**Broken by the fix:** nothing found. The commit touches one test and two prose sites; every suite on the clean tree passes at the same counts as the previous round.

**Ran:** GEECS-Core `pytest tests -m "not integration"` → 162 passed, 1 skipped, 2 deselected; GeecsPvaGateway (the 3 ophyd_async tests deselected) → 59 passed; GeecsBluesky `test_hdf_plugin_detector.py test_namespace.py test_strict_plans.py` → 69 passed; mutation A (latch removed) → `test_a_gated_plugin_is_armed_only_while_its_enable_reads_on` fails at `:433`, baseline 1 passed → restored, `git status --porcelain` empty; nothing committed, staged, stashed or pushed; `tests/data/wire/` untouched.

---

**Author: review closed.** Round 1: P1 (gates read per shot, mid-run flip aborted the run) + P2 + P3 → fixed in 9ab2189f, all confirmed with a RunEngine reproduction. Round 2: two P3s → fixed in 2013adb0, confirmed. Ready for the maintainer's merge into `feature/nonscalar-pva` once CI on 2013adb0 reports; the DB step in the body precedes deploy.

GeecsBluesky/geecs_bluesky/devices/detector.py
```diff
@@ -816,18 +849,46 @@ async def _update_prepare_context(self, trigger_info: TriggerInfo) -> None:
...
+        # already-emitted descriptor.
+        if self._closed_gate_logics is None:  # prepared without stage (tests)
+            self._closed_gate_logics = await self._closed_gates()
+        closed = self._closed_gate_logics
         saved = self._data_logics
         self._data_logics = tuple(
-            dl for dl in saved if _data_logic_supported(dl.prepare_unbounded)
+            dl
+            for dl in saved
+            if id(dl) not in closed
```
skbarber, line 891: [P2] Handle the all-gates-closed fly case

When every declared channel reads `off`, this filter leaves `streamable_data_providers` empty, but `plugin_backed` remains true because `_hdf_ios` still contains all configured plugins. The gated and non-essential plans therefore keep this detector in their flyer list; `StandardDetector.kickoff()` then raises `ValueError: Detector ... is not streamable, so cannot kickoff`. I reproduced this with `_gated_scope`, both enables `off`: `prepare(gated_trigger_info(1))` produced zero providers and `kickoff()` failed. Thus the fail-safe still aborts a run whenever a Picoscope has no enabled channels. Please make the run-level plugin classification reflect the latched active set (while preserving the device's scalars), or explicitly remove/refuse the detector before the fly lifecycle begins.

skbarber (commented):
Review found two behavior issues that should be addressed before merge. Focused changed-file tests pass, and CI is green, but neither suite covers the all-gates-closed fly lifecycle or the scalar-export diagnostics for the new waveform metadata.

GeecsPvaGateway/geecs_pva_gateway/file_plugin.py
```diff
@@ -124,9 +124,22 @@ def available() -> bool:
 #: ``<name>-hdf-image-maxcounts`` never spells the event column
 #: ``<name>-maxcounts``; the ``frame_`` suffixes never spell a variable.
 ATTRIBUTE_SUFFIXES = ("frame_acq_timestamp", "frame_recv_timestamp")
+#: An array variable's stack also carries what its payload said about
+#: itself: a waveform's time axis (``x0``, ``dx`` in seconds) and record
+#: length, written per frame after the scalars so a captured scope trace
+#: is never an axis-less array (``NaN`` for the array shapes that carry
+#: no axis — the pairs and CSV lineouts).  Read back like any other
+#: attribute (``geecs_data_utils.io.scan_stack.read_stack_attributes``).
+WAVEFORM_ATTRIBUTE_SUFFIXES = tuple(f"wave_{key}" for key in WAVEFORM_AXIS_KEYS)
```
skbarber: [P2] Keep waveform metadata out of scalar-join warnings

These datasets are deliberately excluded from `scalar_attributes`, but `read_stack_attributes()` returns every numeric dataset and `frame_columns_from_attributes()` treats every non-stamp suffix as a candidate scalar. Consequently every datum-only array stack feeds `wave_x0`, `wave_dx`, and `wave_samples` into `_warn_unnamed_frame_columns()`, producing a WARNING that three scalar columns were left out of the s-file. I reproduced the warning with a one-frame scope attribute map. This affects gated/non-essential array runs and offline re-export even though the files are valid. Please teach the join/export path to recognize these waveform metadata suffixes (or the new manifest) so expected axis metadata is not diagnosed as missing scalar data.

GEECS-Core/geecs_core/db/device_streams.py
```diff
@@ -96,11 +96,20 @@ class DeviceTypeStreams:
     array_ceiling :
         Rows the gateway pads this devicetype's variable-length arrays to
         (NaN fill; longer is dropped and counted).  ``None`` = native length.
+    gate :
+        Capture variable → the device's on/off variable that says whether
+        *this instance* pushes it (the Picoscope's ``Enable.Ch<X>`` per
+        channel).  The worker reads the gate on every prepare and arms the
```
skbarber: [P3] Document the stage-latched gate semantics

The implementation and Bluesky changelog now intentionally read gates once per `stage`, but this newly added field contract still says `every prepare`. The planning brief repeats the same stale wording. Please update both so a future caller does not reintroduce the per-prepare behavior that commit 9ab2189 fixed.

skbarber:
## Disposition of the Codex review (round 3) — fix commit 26dcc0ce

Codex's findings, verbatim, then what changed.

> **[P2] Handle the all-gates-closed fly case** — When every declared channel reads `off`, this filter leaves `streamable_data_providers` empty, but `plugin_backed` remains true because `_hdf_ios` still contains all configured plugins. The gated and non-essential plans therefore keep this detector in their flyer list; `StandardDetector.kickoff()` then raises `ValueError: Detector ... is not streamable, so cannot kickoff`. I reproduced this with `_gated_scope`, both enables `off`: `prepare(gated_trigger_info(1))` produced zero providers and `kickoff()` failed. Thus the fail-safe still aborts a run whenever a Picoscope has no enabled channels. Please make the run-level plugin classification reflect the latched active set (while preserving the device's scalars), or explicitly remove/refuse the detector before the fly lifecycle begins.

> **[P2] Keep waveform metadata out of scalar-join warnings** — These datasets are deliberately excluded from `scalar_attributes`, but `read_stack_attributes()` returns every numeric dataset and `frame_columns_from_attributes()` treats every non-stamp suffix as a candidate scalar. Consequently every datum-only array stack feeds `wave_x0`, `wave_dx`, and `wave_samples` into `_warn_unnamed_frame_columns()`, producing a WARNING that three scalar columns were left out of the s-file. I reproduced the warning with a one-frame scope attribute map. This affects gated/non-essential array runs and offline re-export even though the files are valid. Please teach the join/export path to recognize these waveform metadata suffixes (or the new manifest) so expected axis metadata is not diagnosed as missing scalar data.

> **[P3] Document the stage-latched gate semantics** — The implementation and Bluesky changelog now intentionally read gates once per `stage`, but this newly added field contract still says `every prepare`. The planning brief repeats the same stale wording. Please update both so a future caller does not reintroduce the per-prepare behavior that commit 9ab2189 fixed.

|---|---|---|
| 1 | P2 | **Fixed — the run-level classification follows the latch.** `GeecsDetector.plugin_backed` is `True` before stage (the namespace's listing, bind-time checks) and, once staged, `False` when every gated plugin read `off`, so the gated plan takes the device's scalars through the sampler and the non-essential wrapper sits it out (narrowed after `stage_all`, one WARNING per skipped flyer, still unstaged with the rest). A direct fly prepare on such a device is refused with the channels named. **Why not a quiet flyer:** I tried a no-op `kickoff`/`complete` first; bluesky refuses it one step earlier — `declare_stream(collect=True)` asserts a non-empty `describe_collect` (`bundlers.py:340`) — so the flyer must leave the list, not idle in it. Pinned by a RunEngine test (both off → not plugin-backed after stage, a fly prepare refused, a gated-style run that classifies after stage completes; one channel on → streams) and a gated-plan test (a non-essential scope with every channel off: `count` succeeds, no `u_ict_stream` descriptor, the WARNING). Both fail with Codex's exact error when the classification ignores the latch or the wrapper does not narrow. |
| 2 | P2 | **Fixed** — `shot_join.frame_columns_from_attributes` skips `WAVEFORM_ATTRIBUTE_SUFFIXES` (`wave_x0` / `wave_dx` / `wave_samples`), now spelled once in `geecs_data_utils.io.arrays` and imported by the gateway's file plugin, so the axis rides in the stack and never reaches `_warn_unnamed_frame_columns`. Pinned by `test_waveform_axis_attributes_are_not_scalar_columns` (fails with the skip removed). Data-Utils 0.35.1's entry records it. |
| 3 | P3 | **Fixed** — the `gate` field contract in `device_streams.py` says once per stage (and `get='yes'`, `set='no'`), the brief's step-2 line too. |

Counts on 26dcc0ce: GeecsBluesky **737 passed, 4 skipped** (full suite); GEECS-Data-Utils shot-join / arrays / export / stack **87 passed**; GeecsPvaGateway **59 passed**; GEECS-Core **162 passed**. Codex cannot be resumed from this side, so a fresh-context verifier given the finding text and the fix commit confirms below.

skbarber:
## Verification (fresh-context verifier, fix commit 26dcc0ce)

Worktree `.claude/worktrees/picoscope-gate`, diff `2013adb0...26dcc0ce`, tests run with the main checkout's package envs against the worktree source. Nothing committed; every mutation restored (`git status` clean).

### Finding 1 (P2, all-gates-closed fly case) — **confirmed fixed on the plan paths**, with one new P2 beside it

Reproduced independently through a RunEngine with a two-plugin `GeecsDetector` (`scopeTrace.Channel0/1` gated on `Enable.ChA/B`), both enables `off`, wired to the fake box (10 scenarios, `test_repro_f1.py`, scratchpad):

- Essential in `count(..., acquisition="gated")`, alone and beside a plugin camera: `exit_status: success`; descriptors `{shots}` / `{primary, shots}`; the `shots` rows carry `u_ict-meancounts` + `u_ict-acq_timestamp`; no `u_ict` stream resource; both `Capture` PVs stay `False`; `shot_clock` = `U_ICT` when alone. `plugin_backed` is `True` again after unstage.
- Gated `bp.scan` with `gated_per_step(shots_per_step=2)` over 2 positions: success, 4 rows, bins `[1, 2]` — the sampler's "device set changed between steps" guard does not trip (the latch is per stage; members are constant inside a run).
- Strict `count([cam, scope], 2)`: success, 2 primary events with the scope's scalars, zero `stream_resource` docs, both plugins unarmed. Control: channel A `on` in strict → exactly `{'u_ict'}` armed, B unarmed.
- Non-essential in a strict run and in a gated run: success, no `u_ict_stream` descriptor, one WARNING `non-essential u_ict: no capture stream armed (…)`.
- bluesky's refusal, as stated by the author: `bundlers.py:729-752` returns `[]` for an empty `describe_collect`, and `declare_stream` (`bundlers.py:340-343`) asserts exactly one stream → `AssertionError` at `declare_stream`, before any kickoff. Mutation 1 below reproduces that exact message.
- Other consumers: `strict.py:339/361` (zero-count / discard guards) now skip the all-off scope — harmless, its `_rewind_plugins` iterates `ctx.streamable_data_providers`, empty anyway; `optimize.py:236` and `namespace.py:436-440` evaluate unstaged → unchanged; `utils.RESERVED_DEVICE_ATTRIBUTES` already lists `plugin_backed`. Only `strict.py` prepares outside `gated.py`, with `STRICT_TRIGGER_INFO`, so the new fly-prepare refusal is unreachable from a plan (only a direct staged fly prepare hits it, as the unit test does).

**New finding [P2] — a native-saving scope with every channel off still aborts a gated run, inside the open run, with a misleading message.** `refuse_native_essentials` tests `d.native_save and not d.plugin_backed`. Bind-time (`registry.py:255`, unstaged) passes because `plugin_backed` is `True` before stage; the step's call (`gated.py:215`, staged) then raises `GeecsConfigurationError: gated acquisition: native-saving device(s) without a file plugin: U_ICT …` → `exit_status: fail` after `start` was emitted (scenario H; control with channel A on succeeds; the same device non-essential succeeds). Condition: the `PicoscopeV2` devicetype carries `Save` and `LocalSavingPath` (fixture `GEECS-Core/tests/fixtures/devicetype_variables.json`); the namespace builds `native_save = triggered and {save, localsavingpath} <= settable_names` (`namespace.py:563`), so this bites iff the DB marks both `set='yes'` on the instance — the standard GEECS save controls the legacy scanner drove, so worth checking rather than assuming. Suggested shape: the native check should ask "has no plugin at all" (`not d._hdf_ios`), not "not plugin-backed this session", or the step-time call should be dropped in favour of the bind-time one; the message ("without a file plugin") is wrong for a device with two.

Note (not a finding): a fly `prepare` on an all-off scope *without* a prior `stage` still yields zero providers silently (the latch is computed on demand inside `_update_prepare_context` after the `plugin_backed` check) — the original reviewer's repro path; unreachable from a plan since `stage_wrapper` always stages first.

### Finding 2 (P2, waveform metadata in scalar-join warnings) — **confirmed fixed**

One-frame attribute map with `frame_acq_timestamp`, `frame_recv_timestamp`, `wave_x0`, `wave_dx`, `wave_samples` → `frame_columns_from_attributes` returns columns `['u_ict-acq_timestamp', 'u_ict-frame_recv_timestamp']` only, and `tiled_export._warn_unnamed_frame_columns` emits nothing. With the skip removed (mutation 3) the pre-fix WARNING returns verbatim: `u_ict: 3 per-frame column(s) no scalar header names, left out of the s-file: u_ict-wave_dx, u_ict-wave_samples, u_ict-wave_x0`. The gateway still writes the three datasets (`test_a_waveform_stack_carries_its_time_axis_as_attributes` passes) and `file_plugin.WAVEFORM_ATTRIBUTE_SUFFIXES is geecs_data_utils.io.WAVEFORM_ATTRIBUTE_SUFFIXES` → `True`.

### Finding 3 (P3, stage-latched gate docs) — **confirmed fixed**

`device_streams.py` `gate` contract now says once per stage / never per prepare or per shot, and pins the gate as read-only (`get='yes', set='no'`); the Planning brief, `GeecsBluesky/CLAUDE.md` and both CHANGELOGs match. The remaining "every prepare" phrases in the brief and `device_streams.py:10` are about the FROG arm-timeout, a different mechanism.

### Mutations (each restored with `git checkout --`)

1. `plugin_backed` → `bool(self._hdf_ios)`: both named tests fail (`assert not True` on `plugin_backed`; and the run aborts with bluesky's `declare_stream contained collect=True but describe_collect did not return a single Dict[str, DataKey] for … u_ict_stream`).
2. Wrapper narrowing removed: `test_non_essential_scope_with_every_channel_off_sits_the_run_out` fails on the detector's own `read off at stage` refusal.
3. Shot-join skip removed: `test_waveform_axis_attributes_are_not_scalar_columns` fails with the three `wave_*` columns present.

### Ran

GeecsBluesky `test_hdf_plugin_detector.py test_gated_plans.py test_namespace.py test_strict_plans.py`: 90 passed; GEECS-Data-Utils `test_shot_join.py test_tiled_export.py`: 48 passed; GeecsPvaGateway `test_arrays.py`: 9 passed; own RunEngine repro: 10 scenarios, 9 pass, 1 fail (= the new P2 above); 3/3 mutations bite.

---

## Disposition of the new finding (author), fix commit 50cbe8f5

|---|---|---|
| new | P2 | **Fixed** — `GeecsDetector.has_file_plugin` is the static fact ("has any plugin at all"); `refuse_native_essentials` and the namespace's listing ask it, `plugin_backed` stays the per-session answer. Pinned by `test_a_native_saving_scope_with_every_channel_off_is_an_essential_member`: a scope with `native_save=True` and both channels off is an essential member of a gated `count` beside a plugin camera — success, its scalars in the `shots` rows, nothing streamed — and fails with the step-time refusal when the check asks `plugin_backed` again. The DB condition the verifier flagged holds today only half-way (`save` is `set='yes'` on both ICTs, `LocalSavingPath` has no row), so the ICTs are not native-saving now; the fix keeps it that way whichever flags a curation sets. `has_file_plugin` is in `RESERVED_DEVICE_ATTRIBUTES`. |
| note | — | **Accepted** — a fly prepare without a prior stage latches on demand; unreachable from a plan (`stage_wrapper` stages first). |

One last confirmation by the same verifier follows; this closes the review.

skbarber:
**Correction to the disposition above (author, prompted by the owner).** I wrote that the ICTs "are not native-saving now" because `LocalSavingPath` has no `expt_device_variable` row. That was the wrong table: the worker's `settable` flag comes from the merged variable rows (`GeecsDb.get_experiment_device_variables`, the `set` column of the devicetype/instance rows), and on both `U_BCaveICT` and `U_UndulatorExitICT` **`LocalSavingPath` and `Save` are settable** — the ICTs *do* save natively to a defined path, as the owner says. So the verifier's new P2 was live for the real devices, not hypothetical: before 50cbe8f5 a gated run with an ICT whose channels all read off would have been refused mid-run as "without a file plugin". The fix stands as written (`has_file_plugin` for the static question); only my aside about the DB state was wrong. Deploy consequence, unchanged from the cameras: the ICTs dual-write — LabVIEW's native trace files beside the plugin's stacks — until PNG/native retirement (#738).

skbarber:
## Confirmation (fresh-context verifier, head b00914e2)

Worktree `.claude/worktrees/picoscope-gate`, diff `26dcc0ce...b00914e2`, main-checkout envs against the worktree source. Nothing committed; both mutations restored (`git status` clean, HEAD b00914e2).

### Fix 1 — 50cbe8f5, native-saving scope with every channel off as a gated essential — **confirmed**

- `refuse_native_essentials` (`gated.py:170`) and the namespace's native-saving listing (`namespace.py:440`) ask the static `has_file_plugin` (`bool(self._hdf_ios)`); `plugin_backed` stays per-session; `has_file_plugin` is in `RESERVED_DEVICE_ATTRIBUTES` and `test_namespace.py:375`'s reserved-equals-public check passes. Unchanged between 50cbe8f5 and head (the later diff touches `shot_clock` only).
- My scenario H (`GeecsDetector(..., path_provider=…)` → `native_save=True`, two gated plugins both `off`, essential beside a plugin camera, `count(..., acquisition="gated")`): `exit_status: success` at head (was `fail` at 26dcc0ce).
- Bind-time refusal intact: K — a truly native-only camera (`native_save=True`, no plugin) is refused before any `start` doc and before any box write, naming `UC_A`; L — the same camera beside the all-off native scope: the message names `UC_A` only, never `U_ICT`.
- Mutation (refusal back to `not d.plugin_backed`): `test_a_native_saving_scope_with_every_channel_off_is_an_essential_member` and my H both fail (`2 failed`, the mid-run "without a file plugin: U_ICT"). Restored.

### Fix 2 — 8a27a38c + b00914e2, `shot_clock` chooses by `has_file_plugin` — **confirmed**

- `shot_clock` (`gated.py:137-143`) now builds both lists from `has_file_plugin`, so the bind-time (unstaged) and step-time (staged) answers cannot differ when a scope latches off; docstring says so.
- My scenario M (`count([ict, a], 3, acquisition="gated")`, scope first, every channel off): success; `start["shot_clock"] == "U_ICT"`, `start["shot_clock_column"] == "u_ict-acq_timestamp"`, and the rows carry **no** `uc_a-acq_timestamp` — `a` is a pure plugin, so the only way its stamp reaches a row is as the sampler's clock; its absence proves the batch ticked on the clock the start document names. Mirror N (`[a, ict]`): clock `UC_A`, rows carry `uc_a-acq_timestamp` (clock) and `u_ict-acq_timestamp` (member), `uc_a` datums `[0, 3)`.
- Mutation (`shot_clock` back to `plugin_backed`, both lists): `test_the_shot_clock_choice_does_not_move_when_a_scope_latches_off` fails, and my M fails on the run-level assertion — `uc_a-acq_timestamp` reappears in the rows while the start document still says `U_ICT`, the exact divergence I reported. Restored. (One note on method: my earlier staged probe via `RE(bps.stage(...))` was undone by the RunEngine at the end of that call, so the probe alone could not show the split; b00914e2's staging inside the RE loop is the right shape, and my run-level assertion carries the evidence regardless.)

### Anything new — nothing found in the two commits

Consumers of `plugin_backed` remaining (`gated.py:215/532` take_reading + wrapper, `strict.py:339/361`) want the per-session answer; `namespace.py:436` and `optimize.py:236` evaluate unstaged devices where the two properties coincide. One hardware-check item, not a code finding: with the scope listed first and every channel off it is now the run's clock (same preference as before this PR — a device with a plugin wins — and that configuration aborted outright before 26dcc0ce), so the run relies on the Picoscope's `timestamp` advancing per trigger with all channels disabled; if it does not, the sampler fails loudly (`GeecsTriggerTimeoutError` naming `U_ICT`) rather than silently — worth one shot on the bench.

### Ran

GeecsBluesky `test_hdf_plugin_detector.py test_gated_plans.py test_namespace.py test_strict_plans.py` at b00914e2: 92 passed (rc 0); own RunEngine repro `test_repro_f1.py`: 14 scenarios (A–N) all pass; mutations 2/2 bite (`2 failed` each), both restored.

---

**Disposition (author).** Both fixes **confirmed** by the same verifier: the P2 from its 26dcc0ce verification (a native-saving scope with every channel off aborted a gated run mid-run) is fixed in 50cbe8f5, and the residual it raised on that fix (`shot_clock` could pick a different clock at bind time and at step time) is fixed in 8a27a38c + b00914e2. Its bench item — a Picoscope with every channel disabled, listed first, becomes the gated run's clock, so its `timestamp` must advance per trigger in that state — is added to the hardware-verification list in the PR body; the failure mode is loud either way. The author's own mutation check on b00914e2 (`shot_clock` back on the per-session flag) fails the new test on the `shot_clock([ict, a]) == before` assertion. CI is green on b00914e2 (13 checks); full GeecsBluesky suite there: 739 passed, 4 skipped. Review closed at round 3 per the stopping rule; no undispositioned finding remains.

skbarber (commented):
Approved from code review at b00914e2; no remaining findings. The all-gates-closed detector now becomes a scalar member for gated acquisition and is skipped safely as a non-essential stream; static plugin capability remains distinct from per-session stream activity; waveform axis metadata is excluded from scalar-join diagnostics; and the stage-latched gate contract is documented consistently. Verified 210 focused tests locally; all 13 GitHub checks are green. GitHub does not permit the PR author account to submit a formal APPROVE review.

